Issue #5, opened 2023-07-31
Codegen for API Typescript Definitions


ChristopherChudzicki:
It would be useful to automatically generate Typescript definitions for our APIs:
- Auto-generated definitions would stay up-to-date as we introduce new APIs
- Auto-generated definitions should be more accurate than manually-generated definitions.

One approach would be to use a tool like [Openapi-Typescript](https://github.com/drwpow/openapi-typescript) to process an OpenAPI Spec via [drf-spectacular](https://drf-spectacular.readthedocs.io/en/latest/).

PR #24, opened 2023-08-10
Generate OpenAPI spec, Type definitions, and API Client

ChristopherChudzicki:
#5, See also https://github.com/mitodl/hq/discussions/2076

#### What's this PR do?
With this PR, we:
- commit an OpenAPI specification file (`openapi.yaml`) generated by [drf-spectacular](https://drf-spectacular.readthedocs.io/en/latest/).
- Use `openapi.yaml` to generate, via [OpenAPITools/openapi-generator](https://github.com/OpenAPITools/openapi-generator) (also committed in the repo):
    - Typescript type declarations for the API request/response objects
    - Runtime Typescript code for an API client using `axios`.
- Add CI checks to ensure the auto-generated, committed code is up-to-date.

**Note:** Currently the generated code is based on `channels_fields.urls` only. This is irrelevant and will change after #25 is merged.
 
On this branch:
1. Run `./scripts/generate_openapi.sh`. There should be no change.
2. Make an edit to `channel_fields/serializers.py`, for example:
    - change a serializer docstring
    - change the type annotation on a `SerializerMethodField`, e.g., make `get_full_name` return an `int` (which is wrong, but that's fine for now). 
3. **Check re-generating the code.** Re-run `./scripts/generate_openapi.sh` and inspect the git diff. You should see:
    - `openapi.yaml` has changed
    -  `/frontends/api/src/generated` has changed
4. Now check CI. We want to ensure CI fails if **either** `openapi.yaml` **or** `frontends/api/src/generated` is out of date.
5. (CI check):
    - `git checkout -b your-new-branch`.
    - commit just the serializer change, and push up to github.
    - The Python tests should fail because Django & `openapi.yaml` are out of sync. (*The new `openapi-generated-client-check` test will pass, since openapi.yaml is in-sync with the generated typescript.*)
        - Check the test failure message is useful.
6. (CI check, continued):
    - commit the freshly generated `openapi.yaml` spec, and push it up.
    - The Python tests should pass (Django and OpenAPI spec are in sync).
    - The new `openapi-generated-client-check` test will fail fairly quickly.
        - Check the test failure message is useful.


.prettierrc.json
ChristopherChudzicki, line 1: Storing our prettier settings in prettierrc means
1. Schema validation
2. passing fewer CLI args around

frontends/api/package.json
```diff
@@ -0,0 +1,8 @@
+{
+  "name": "@mit-open/api",
+  "packageManager": "yarn@3.2.0",
+  "sideEffects": false,
```
ChristopherChudzicki, line 4: `sideEffects: false` lets webpack tree-shake imports. E.g., if a a file exports `a` and `b`, and we only import `a`, then the code for `b` is omitted.

open_discussions/urls_spectacular.py
ChristopherChudzicki, line 1: `drf-spectacular` provides two ways to configure which urls contribute to the openapi spec:
- `SERVE_URLCONF` a setting in the Python files that determines what gets **served** via `SpectacularAPIView` (and the other spectacular views). Accepts **multiple urlconf.**
- `urlconf` CLI argument for **generating files**. Accepts **single urlconf**.

So if we want to generate a schema for a subset of our urls using more than one urlconf, having a separate `urls_spectacular.py` file was the only way I could think of.

ChristopherChudzicki, line 1: drf-spectacular has two ways to configure the subset of URLs used for the OpenAPI spec:
- `SERVE_URLCONF` to what gets served by SpectacularAPIView and friends (i.e., what you see at https://open.mit.edu/api/v0/schema/redoc). It accepts a single urlconf OR **multiple** urlconfs.
- `--urlconf` CLI option for generating files. It accepts only a **single** urlconf

Because `--urlconf` only accepts a single urlconf, I made a separate file where we can include many.

mbertrand (approved):
👍 , I made a suggestion but it isn't directly related to the changes in this PR, so if you prefer, I can make a new issue for that.

scripts/test/python_tests.sh
```diff
@@ -17,6 +17,7 @@ function run_test {
 
 run_test ./scripts/test/detect_missing_migrations.sh
 run_test ./scripts/test/no_auto_migrations.sh
+run_test ./scripts/test/openapi_spec_check.sh
```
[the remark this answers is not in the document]
ChristopherChudzicki: I entirely agree... Makes finding the relevant error message harder, too.

I added [`set -eo pipefail`](https://stackoverflow.com/a/4346420/2747370) to the script, which some of our other scripts have, too. Here's an example failure https://github.com/mitodl/mit-open/actions/runs/5868419021/job/15911210038?pr=24

mbertrand (approved):
👍

PR #29, opened 2023-08-16
OpenAPI Generation for Learning Resources + Expose documentation

### LINKED ISSUE #28: Expose new API's OpenAPI documentation
We should expose documentation (redoc and/or swagger) for `api/v1/learning_resources`.

ChristopherChudzicki:
#28 

#### What's this PR do?
This PR:
- Switches the OpenAPI code generate to generate based on `learning_resources` app
- Exposes swagger + redoc docs for the `learning_resources` app
- Makes `url` a required field on LearningResourceImage model.

1. Visit http://localhost:8063/api/v1/schema/redoc/#tag/learning_resources. The docs should look reasonable.


course_catalog/urls.py
```diff
@@ -72,15 +67,4 @@
...
-    re_path(
-        "api/v0/schema/swagger-ui/",
-        SpectacularSwaggerView.as_view(url_name="schema"),
-        name="swagger-ui",
-    ),
-    re_path(
-        "api/v0/schema/redoc/",
-        SpectacularRedocView.as_view(url_name="schema"),
-        name="redoc",
-    ),
```
ChristopherChudzicki, line 85: Which docs get served is controlled via the `SERVE_URLCONF` env variable. As far as I know, we can't serve the old catalog docs at one URL and the new learning resource docs at another, at least not with the builtin `drf-spectacular` views.

learning_resources/serializers.py
```diff
@@ -26,6 +27,7 @@ class Meta:
         fields = ["id", "name"]
 
 
+@extend_schema_field({"type": "array", "items": {"type": "string"}})
```
ChristopherChudzicki, line 30: For SerializerMethodFields, a normal type annotation on the method is enough for drf spectacular. But not for `serializers.Field` :(. See https://github.com/tfranzel/drf-spectacular/issues/886

learning_resources/models.py
```diff
@@ -127,12 +127,13 @@ class LearningResource(TimestampedModel):
     )
 
     @property
-    def audience(self):
+    def audience(self) -> str | None:
         """Returns the audience for the course"""
-        return self.platform.audience
+        if self.platform:
+            return self.platform.audience
```
ChristopherChudzicki, line 133: `platform` might be None, so I changed this.

learning_resources/views.py
```diff
@@ -25,6 +26,24 @@ class DefaultPagination(LimitOffsetPagination):
...
+    ),
+    new=extend_schema(
+        summary="List New",
+        description="Get a paginated list of newly released resources.",
+    ),
+    upcoming=extend_schema(
+        summary="List Upcoming",
+        description="Get a paginated list of upcoming resources.",
+    ),
+)
```
ChristopherChudzicki, line 46: This just makes the generated docs a little bit nicer.

We do not need this, and if it's annoying, we can get rid of it until we decide to do something with the docs.

Without it:
- documentation descriptions come from docstrings
- I'm not sure where the summaries come from (which are used for left sidebar links in http://localhost:8063/api/v1/schema/redoc)

mbertrand (approved):
👍

PR #31, opened 2023-08-16
Remove legacy authentication mechanisms

### LINKED ISSUE #6: Authentication: Remove legacy authentication support
We'll be using an external authentication service for this rather than our own implementation, so we want to remove our old custom authentication implementation. There won't be any user-facing UIs for authentication since Keycloak will handle all of that, which means we also don't need the DRF APIs either.

- Remove email authentication
- Remove Micromasters authentication
- Remove Touchstone authentication
- Remove authentication APIs
- Remove any unused third party libraries (e.g. `djoser`)

See also: #21

rhysyngsun:
#### Pre-Flight checklist

- [ ] Testing
  - [ ] Code is tested
  - [ ] Changes have been manually tested

Closes #6 

#### What's this PR do?

This PR removes the legacy social auth mechanisms. Auth is still possible via django-admin.
- Removes SAML login
- Removes MicroMasters login
- Removes email login
- Removes the social auth DRF API
- Touched a few imports in a minimal amount of places due to the removal of some unused imports in the PR causing an import cycle.

You should be able to run the app and still login via django-admin


authentication/urls.py
```diff
@@ -2,56 +2,7 @@
...
-        name="password-reset-confirm-api",
-    ),
-    re_path(
-        r"^api/v0/set_password/$",
-        CustomDjoserAPIView.as_view({"post": "set_password"}),
-        name="set-password-api",
-    ),
-    re_path(r"^api/v0/auths/$", get_social_auth_types, name="get-auth-types-api"),
-    re_path(r"^login/complete$", login_complete, name="login-complete"),
     re_path(r"^logout/$", auth_views.LogoutView.as_view(), name="logout"),
```
collinpreston, line 7: Can this be removed?

open_discussions/urls.py
```diff
@@ -80,11 +79,6 @@
     re_path(r"^signup/confirm/$", index, name="register-confirm"),
     re_path(r"^account/inactive/$", index, name="account-inactive"),
     re_path(r"^password_reset/", index, name="password-reset"),
```
collinpreston, line 81: Can this be removed?

collinpreston (approved):
Appears good to me.  I was able to login to Django Admin.  I added 2 comments inline.

authentication/urls.py
rhysyngsun: We should keep logout.

open_discussions/urls.py
rhysyngsun: Removed in the latest.

Issue #19, opened 2023-08-09
Implement relations between learning resources


mbertrand:
Implement relationships between the following types of learning resources:
Programs and Courses
Podcasts and Podcast Episodes
User/Staff Lists and any type of other learning resource

#### Designs and Mockups

Original idea was to use treebeard for this, but a possibly better alternative is to implement a mode/table for parent-child relationships using the https://dimagi.github.io/django-cte/ library. 

#### Acceptance Criteria:

- [ ] Relationships between the above resources can be efficiently CRUD'ed and queried.

PR #34, opened 2023-08-17
Add LearningPath (StaffList) model and API, refactor relationships between resources-

### LINKED ISSUE #13: Implement new learning list model(s)
Implement new model(s) for staff lists, user lists, favorites based on the new LearningResource models

#### Designs and Mockups

https://github.com/mitodl/mit-open/issues/4

#### Acceptance Criteria:

- [ ] Models and test factories for learning list models are implemented

#### Out of Scope

- serializers, views, etc

### LINKED ISSUE #16: Implement API (serializers, views) for new learning list models
Implement serializers and views for staff lists, user lists, favorites in the context of the new LearningResource models.

#### Designs and Mockups
https://docs.google.com/document/d/1OmtUMX9yJ2phYtYkGp0xrO8UntMFdXDhsiSoF5lKjlI/edit

#### Acceptance Criteria:

- [ ] Working API with serializers and views for staff lists, user lists, favorites

#### Out of Scope

- API for courses, programs, videos, podcasts

### LINKED ISSUE #32: Filter parameters in OpenAPI spec/docs
`drf-spectacular` is currently not picking up some API parameters, particularly filters. For example:

- `LearningResourcesViewSet` allows filtering by resource type https://github.com/mitodl/mit-open/blob/88060a6e3f1191ebe252542c38f3b620c6f42e29/learning_resources/views.py#L48
- The `openapi.yaml` does not pick up `resource_type` as a possible parameter.  https://github.com/mitodl/mit-open/blob/5ecfd99691b9fefc4b50683ddc4296aeb91260b4/openapi.yaml#L12

Pagination filters **are** being picked up.

Possibly useful:
- [`extend_schema`](https://drf-spectacular.readthedocs.io/en/latest/customization.html#step-2-extend-schema) for manual annotations
- I *think* drf-spectacular will pick up filters if they are implemented by django-filter, which we use in a few places.

### LINKED ISSUE #51: Add a `learningpath_items` field to LearningResource serializer
We need to know which learning paths a LearningResource is on for purposes of the 'Add to List' UI dialog.  Look at `course_catalog.serializers.ListsSerializerMixin` to see how it was done previously.  TBD: do it similarly or with a new approach.

mbertrand:
#### Pre-Flight checklist

- [x] Migrations
  - [x] Migration is backwards-compatible with current production code
- [x] Testing
  - [x] Code is tested
  - [x] Changes have been manually tested

Closes #13
Closes #16
Closes #32
Closes #51
Partially addresses #19 

#### What's this PR do?
- Refactors relationships between learning resources
- Adds LearningPath (formerly StaffList) model and a CRUD API for it.
- Adds Django Admin functionality for learning_resources

If you haven't already, run the following:

```
docker compose run --rm web python manage.py loaddata platforms departments offerers
docker compose run --rm web python manage.py backpopulate_xpro_data
```

Go to the following urls, they should return data that looks reasonable.  In particular, the programs returned should have a "courses" field containing a list of course LearningResource objects
`http://localhost/api/v1/learning_resources`
`http://localhost/api/v1/courses`
`http://localhost/api/v1/courses/<learning_resource_id>/`
`http://localhost/api/v1/programs`
`http://localhost/api/v1/programs/<learning_resource_id>/`

LearningPath CRUD
- Go to `http://localhost/api/v1/learningpaths/` as an admin user
- Submit the following JSON in the raw data field:
```json
{
    "title": "foo",
    "description": "bar",
    "topics": [1, 2],
    "published": true
}
```
- A new LearningPath resource should be created
- Go to the resources endpoint for it `http://localhost:8063/api/v1/learningpaths/<path_resource_id>/resources/`
- Submit the following JSON in the raw data field, using the id of a course LearningResource:
```json
{
    "child": <course_learning_resource_id>
}
```
That course should then show up in the above url response with a position of 1.  Repeat with another course, it should appear in the response with a position of 2

- Go to `http://localhost:8063/api/v1/learningpaths/<path_resource_id>/resources/<item_id>/` and submit the following JSON:
```json
{
    "position": 99
}
```
The position value should be updated.
- Go to `http://localhost:8063/api/v1/learningpaths/<path_resource_id>/resources/<item_id>/` and click the delete button, that item should be removed.
- Go to `http://localhost:8063/api/v1/learningpaths/<path_resource_id>/` and click the delete button, that learning path should be removed.

#### Any background context you want to provide?
Some differences between this implementation of `LearningPath` and the old `StaffList`:

- I did not include a`privacy_level` field ("public" or "private") on `LearningPath` because `LearningResource.published` (true/false) can be used to make that distinction.
- Also removed the distinction between a "list" (unsortable) and "path" (sortable) - a LearningPath will always be sortable.
- I did not implement `UserList` in this PR because after some discussion w/Ferdi it was decided that `UserList` should not have a foreign key to `LearningResource`.  It will have its own fields for title, description, etc and have its own many-to-many relationship to other learning resources.


learning_resources/models.py
```diff
@@ -233,15 +240,54 @@ class Program(TimestampedModel):
...
+class LearningPath(TimestampedModel):
+    """
+    Model for representing a publishable list of  learning resources
+    The LearningResource readable_id should probably be something like an auto-generated UUID.
+    """
+
+    learning_resource = models.OneToOneField(
+        LearningResource,
+        null=True,
+        blank=True,
```
rhysyngsun, line 264: I don't think this should be nullable or blank?

learning_resources/views.py
```diff
@@ -179,4 +205,104 @@ def get_queryset(self):
...
+            resource_type=LearningResourceType.learning_path.value,
+        )
+
+    def list(self, request, *args, **kwargs):
+        """Get all learningpaths for learningpath editors, public learningpaths for all others"""
+
+        if is_learning_path_editor(request) or is_admin_user(request):
+            queryset = self.get_queryset()
+        else:
+            queryset = self.get_queryset().filter(published=True)
```
rhysyngsun, line 238: Should this be applied in `get_queryset` instead?

learning_resources/views.py
```diff
@@ -179,4 +205,104 @@ def get_queryset(self):
...
+
+        page = self.paginate_queryset(queryset)
+        if page is not None:
+            serializer = self.get_serializer(page, many=True)
+            return self.get_paginated_response(serializer.data)
+
+        serializer = self.get_serializer(queryset, many=True)
+        return Response(serializer.data)
```
rhysyngsun, line 246: Is there a reason we do this instead of using `pagination_class`?

learning_resources/views.py
```diff
@@ -179,4 +205,104 @@ def get_queryset(self):
...
+        return super().update(request, *args, **kwargs)
+
+    def perform_destroy(self, instance):
+        instance.delete()
+        # Uncomment when search is ready
+        # learning_path = instance.parent
+        # if learning_path.items.count() > 0:
+        #     upsert_staff_list(staff_list.id)
+        # else:
+        #     deindex_staff_list(staff_list)
```
rhysyngsun, line 308: These should really be part of the serializer

learning_resources/views.py
```diff
@@ -179,4 +205,104 @@ def get_queryset(self):
...
+class LearningPathItemsViewSet(ResourceListItemsViewSet):
+    """
+    Viewset for LearningPath related resources
+    """
+
+    serializer_class = LearningResourceRelationshipSerializer
+    permission_classes = (permissions.HasLearningPathItemPermissions,)
+
+    def create(self, request, *args, **kwargs):
+        parent_id = kwargs[f"{DRF_NESTED_PARENT_LOOKUP_PREFIX}parent_id"]
```
rhysyngsun, line 285: Could you explain this convention a bit?

learning_resources/views.py
```diff
@@ -179,4 +205,104 @@ def get_queryset(self):
...
+        request.data["parent"] = parent_id
+        request.data[
+            "relation_type"
+        ] = LearningResourceRelationTypes.LEARNING_PATH_ITEMS.value
```
rhysyngsun, line 289: Might be work subclassing `LearningResourceRelationshipSerializer` to put that code there instead of here.

learning_resources/views.py
```diff
@@ -64,41 +89,40 @@ def _get_base_queryset(self, resource_type: str = None) -> QuerySet:
...
+            "children__child__runs__instructors",
+            "children__child__course",
+            "children__child__program",
+            "children__child__learning_path",
+            "children__child__department",
+            "children__child__platform",
+            "children__child__topics",
+            "children__child__image",
+            "children__child__offered_by",
+            "children__child__resource_content_tags",
```
rhysyngsun, line 119: :scream: We'll need to keep an eye on this

learning_resources/serializers.py
```diff
@@ -103,29 +134,202 @@ class LearningResourceBaseSerializer(serializers.ModelSerializer):
...
+        if value != LearningResourceType.learning_path.value:
+            raise serializers.ValidationError(
+                "Only LearningPath resources are editable"
+            )
+        return value
+
+    def create(self, validated_data):
+        """Ensure that the LearningPath is created by the requesting user; set topics"""
+        request = self.context.get("request")
+        if request and hasattr(request, "user") and isinstance(request.user, User):
```
rhysyngsun, line 225: Are we able to move this kind of validation into the view? To create a learning path we already should have an authenticated user.

rhysyngsun (changes requested):
Bounced around a bit and didn't review everything, but here's a first pass for today.

learning_resources/views.py
mbertrand: I think the serializer isn't called when a delete is performed?

learning_resources/views.py
mbertrand: It has to do with the nested serializer functionality described [here](https://chibisov.github.io/drf-extensions/docs/#nested-routes), the LearningResource id  of the parent LearningPath is contained in the request kwargs.  I will move this to a separate function with explanatory docstring.

learning_resources/serializers.py
mbertrand: Yes, user shouldn't even get this far if not authenticated so conditional can just be removed

learning_resources/views.py
mbertrand: Agreed, it's a bit scary how many prefetches are necessary to avoid n+1 errors.  Getting some "Potential unnecessary eager load detected" messages in the logs but I think those are less of a concern?

learning_resources/views.py
rhysyngsun: Ok, a couple of thoughts based on that:

- Maybe use [this method instead](https://github.com/chibisov/drf-extensions/blob/f7de5c0278371d35efc3f8ff929fcccdfb641c21/rest_framework_extensions/mixins.py#L69)?
- It probably shouldn't be a configurable setting, since it effects a public API schema and the schema for the frontend.

learning_resources/models.py
```diff
@@ -233,15 +240,52 @@ class Program(TimestampedModel):
...
+        LearningResource,
+        related_name="learning_path",
+        on_delete=models.CASCADE,
+    )
+    author = models.ForeignKey(
+        User, related_name="learning_paths", on_delete=models.PROTECT
+    )
+
+    def __str__(self):
+        return f"Staff List: {self.learning_resource.title}"
```
ChristopherChudzicki, line 271: IMO, keep `LearningPath` for the `__str__` name.

ChristopherChudzicki (commented):
I tried this out and it is working well (Aside from the blank `relation_type` that we discussed in Slack).

Related to this, some additional behavior I think we'll want:
1. Given a resource, decide what lists it belongs to. *E.g., for the 'add to list' dialog*
2. logic around updating positions on PATCH and DELETE requests. (e.g., deleting { position: 4 }` shifts 5+ down, PATCHing moves items toward old position)

_Not suggesting we do it here, but could create issues_

learning_resources/models.py
mbertrand: >Given a resource, decide what lists it belongs to. E.g., for the 'add to list' dialog

The list it belongs to should be in the url, ie posting to http://localhost:8063/api/v1/learningpaths/55/resources/ will add it to the LearningResource with id 55.


> logic around updating positions on PATCH and DELETE requests. (e.g., deleting { position: 4 }` shifts 5+ down, PATCHing moves items toward old position)
For patching, that is in the LearningResourceRelationshipSerializer.update() function.  For delete, we could add that, though it should make no difference to the remaining items order, there will just be a numerical gap in the position values.

mbertrand: https://github.com/mitodl/mit-open/issues/51

openapi.yaml
```diff
@@ -268,19 +713,95 @@ components:
...
+          type: string
+          nullable: true
+        readable_id:
+          type: string
+          maxLength: 128
         topics:
           type: array
           items:
             $ref: '#/components/schemas/LearningResourceTopic'
+          nullable: true
```
ChristopherChudzicki, line 781: Currently the OpenAPI spec and generated typescript types say that full `LearningResourceTopic` objects are:
- returned in GET requests
- expected for POST/PATCH requests

The POST/PATCH requests actually accept either fully LearningResourceTopic objects OR just their ids. See

learning_resources/serializers.py
```diff
@@ -103,29 +136,208 @@ class LearningResourceBaseSerializer(serializers.ModelSerializer):
     audience = serializers.ReadOnlyField()
     certification = serializers.ReadOnlyField()
     prices = serializers.ReadOnlyField()
+    topics = WriteableSerializerMethodField()
```
ChristopherChudzicki, line 139: Perhaps premature since this is still a PR, but I opened #54 about `topics` here.

openapi.yaml
ChristopherChudzicki: Opened #54 (perhaps prematurely) about this.

learning_resources/views.py
```diff
@@ -179,4 +200,86 @@ def get_queryset(self):
...
+    """
+    Viewset for LearningPath related resources
+    """
+
+    serializer_class = LearningPathRelationshipSerializer
+    permission_classes = (permissions.HasLearningPathItemPermissions,)
+
+    def create(self, request, *args, **kwargs):
+        request.data["parent"] = self.get_parent_id()
+        return super().create(request, *args, **kwargs)
```
ChristopherChudzicki, line 272: I was curious how `drf-spectacular` would deal with views modifying request data like this, so I looked into that a bit. Without any extra work, the OpenAPI spec thinks `parent` is required.

The only way I could find to model this with `drf-spectacular` is something along the lines of:
```python
    @extend_schema(
            request=extend_schema_serializer(
                exclude_fields=["parent"]
            )(LearningPathRelationshipSerializer)
    )
    def create(self, request, *args, **kwargs):
```

I'm not suggesting we add that, just writing it down for future. It's a little verbose and if the OpenAPI Spec thinks `requestBody.parent` is required, it's not really problematic since whoever is calling the API must have it already for the path param.

learning_resources/views.py
```diff
@@ -52,6 +65,13 @@ class LearningResourceViewSet(viewsets.ReadOnlyModelViewSet):
     serializer_class = LearningResourceSerializer
     permission_classes = (AnonymousAccessReadonlyPermission,)
     pagination_class = DefaultPagination
+    filter_backends = [DjangoFilterBackend]
+    filterset_fields = [
```
ChristopherChudzicki, line 69: 🎉 This fixes https://github.com/mitodl/mit-open/issues/32

rhysyngsun (changes requested):

learning_resources/views.py
```diff
@@ -179,4 +200,86 @@ def get_queryset(self):
...
+
+    serializer_class = LearningPathRelationshipSerializer
+    permission_classes = (permissions.HasLearningPathItemPermissions,)
+
+    def create(self, request, *args, **kwargs):
+        request.data["parent"] = self.get_parent_id()
+        return super().create(request, *args, **kwargs)
+
+    def update(self, request, *args, **kwargs):
+        request.data["parent"] = self.get_parent_id()
```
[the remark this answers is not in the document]
mbertrand: Changing to a `HiddenField` on the `LearningPathRelationshipSerializer` with default value of `LEARNING_PATH_ITEMS`

learning_resources/views.py
mbertrand: I wonder if there is a way to hide CRUD API specs altogether and just show the read-only API specs. Seems like CRUD requests should only happen via our own frontend (and probably need to, due to Django CSRF).

ChristopherChudzicki: @mbertrand I'm sure we could hide the CRUD (except R) API specs.

IMO, the CUD spec is somewhat useful internally since we generate the API client on our frontend. But I don't want to overstate that... there's still manual work: We generate the API client automatically, but in order to use it, we have to write some react hooks, and that portion is manual anyway.

mbertrand:
@rhysyngsun @ChristopherChudzicki ready for another look

learning_resources/serializers.py
```diff
@@ -103,29 +147,233 @@ class LearningResourceBaseSerializer(serializers.ModelSerializer):
...
+    course = CourseSerializer(read_only=True, allow_null=True)
+    learning_path = LearningPathSerializer(read_only=True, allow_null=True)
+    runs = LearningResourceRunSerializer(read_only=True, many=True, allow_null=True)
+    learningpath_items = serializers.SerializerMethodField()
```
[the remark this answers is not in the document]
mbertrand: How about just `learning_paths`?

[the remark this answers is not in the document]
ChristopherChudzicki: I was thinking that `learning_path` (for the "extra details") and `learning_paths` (for the parents) might be confusing. But I'm good with it if you like that better. *It's not actually confusing on our frontend, imo, since the typehints will be very different.*

ChristopherChudzicki (approved):
@rhysyngsun should still review, creating/updating learningpaths, adding/deleting/moving items is working well for me 👍

Issue #27, opened 2023-08-14
Add basic homepage


ChristopherChudzicki:
Add a basic homepage for `mit-open` using new OpenAPI-generated types and API client.

- demonstrate mocking generated API client in tests
- test data factories using new types
- Basic UI similar to what we have now at https://discussions-rc.odl.mit.edu/infinite, but with just one courses carousel for now.

rhysyngsun:
@ChristopherChudzicki can you assign yourself and move to in progress on the board when you start?

ChristopherChudzicki:
Done in #37 and #49

PR #37, opened 2023-08-18
Very minimal homepage

ChristopherChudzicki:
#27 , but not quite all of what we discuss (no cards yet).

#### What's this PR do?
This PR:
- adds `react-query` hooks for our generated API client
- adds new data factories based on the types from our generated API client
- adds some tests using the new API client & data factories
- Adds an **extremely** simple homepage. *More coming*. 

1. If you don't have any learning resources yet, make some fake ones. In a django shell, run:
    ```py
    from learning_resources.factories import *
    CourseFactory.create_batch(4)
    ProgramFactory.create_batch(3)
    ```
2. Visit http://localhost:8063 to see a list of learning resources. 

### Underwhelming Screenshot
*This is a list of fake resources titles generated as above.*
Screenshot 2023-08-18 at 4 36 43 PM


.github/workflows/ci.yml
```diff
@@ -176,7 +176,8 @@ jobs:
           generator-tag: v6.6.0
           command-args: |
             --output $GENERATOR_OUTPUT_DIR_CI \
-            --ignore-file-override $GENERATOR_IGNORE_FILE
+            --ignore-file-override $GENERATOR_IGNORE_FILE \
+            --additional-properties=useSingleRequestParameter=true
```
ChristopherChudzicki, line 180: The affect of this setting is that the generated API client puts all API parameters as an object in the first argument, rather than as position parameters. I.e.,
```ts
// now we get
apiClient. coursesList({ limit, offset })
// not
apiClient.courseList(limit, offset)
``` 

My thoughts are:
- position parameters will be annoying to deal with if we change the order
- position parameters are annoying when we want to pass 2nd param but not 1st.
- having all API params in the first argument is a little more convenient when dealing with Typescript, since OpenAPI Generator makes interfaces for the type signatures. *Although the interface names are extremely verbose, like `LearningResourcesApiLearningResourcesListRequest`*

frontends/infinite-corridor/src/test-utils/mockAxios.ts
ChristopherChudzicki, line 1: I moved this to the `api` package, which lets us use request mocking in more packages.

frontends/api/src/test-utils/urls.ts
```diff
@@ -0,0 +1,29 @@
...
+  return `?${new URLSearchParams(params).toString()}`
+}
+
+type Params<API extends BaseAPI, K extends keyof API> = Parameters<API[K]>[0]
+
+const learningResources = {
+  list: (params?: Params<LRApi, "learningResourcesList">) =>
+    `/api/v1/learning_resources/${query(params)}`,
+  details: (params: Params<LRApi, "learningResourcesRetrieve">) =>
+    `/api/v1/learning_resources/${params.id}/`
```
ChristopherChudzicki, line 26: This is going to be a little bit of a pain. I wish OpenAPI Generator gave us easy access to the URLs.

Overall, the generator is still a benefit though... before did (1) types, (2) api calls, (3) urls manually, now we just do (3) manually.

frontends/ol-util/src/factories.ts
```diff
@@ -15,14 +15,14 @@ const makePaginatedFactory =
...
-    ): PaginatedResult<T> => {
+    ) => {
       const results = times(pageSize ?? count, () => makeResult())
       return {
         results,
         count,
         next,
         previous
-      }
+      } satisfies PaginatedResult<T>
```
ChristopherChudzicki, line 25: Despite the fact that `yarn typecheck` and `yarn build` I was getting a type error in VS Code when *using* `makePaginatedFactory`:

```
The inferred type of 'makeListItemsPaginated' cannot be named without a reference to '../../../node_modules/ol-util/src/interfaces'. This is likely not portable. A type annotation is necessary.ts(2742)
```

This change has no real affect on the signature and made the issue go away.

Others have reported similar problems in https://github.com/microsoft/TypeScript/issues/42873 with no clear resolution.

jest-shared-setup.ts
```diff
@@ -20,3 +21,13 @@ configure({
...
 })
+
+afterEach(() => {
+  /**
+   * Clear all mock call counts between tests.
+   * This does NOT clear mock implementations.
+   * Mock implementations are always cleared between test files.
+   */
+  jest.clearAllMocks()
+  resetAllWhenMocks()
```
ChristopherChudzicki, line 32: This was previously in `infinite-corridor`'s `setupTests` file, but we should have it everywhere, so I moved it to the shared setup.

scripts/generate_openapi.sh
```diff
@@ -26,6 +26,7 @@ docker run --rm -v "${PWD}:/local" openapitools/openapi-generator-cli:${GENERATO
     -i /local/openapi.yaml \
     -g typescript-axios \
     -o /local/frontends/api/src/generated \
-    --ignore-file-override /local/frontends/api/.openapi-generator-ignore
+    --ignore-file-override /local/frontends/api/.openapi-generator-ignore \
+    --additional-properties=useSingleRequestParameter=true
```
ChristopherChudzicki, line 30: See https://github.com/mitodl/mit-open/pull/37/files#r1298815222

frontends/api/src/hooks/learningResources.ts
```diff
@@ -0,0 +1,31 @@
...
+const learningResources = createQueryKeys("learningResources", {
+  detail: (params: LearningResourcesApiLearningResourcesRetrieveRequest) => ({
+    queryKey: [params],
+    queryFn:  () => api.learningResourcesRetrieve(params).then(res => res.data)
+  }),
+  list: (params: LearningResourcesApiLearningResourcesListRequest) => ({
+    queryKey: [params],
+    queryFn:  () => api.learningResourcesList(params).then(res => res.data)
+  })
+})
```
ChristopherChudzicki, line 18: `@lukemorales/query-key-factory` is a (very small) library for managing query keys with react query. It's very similar to what we were doing in https://github.com/mitodl/mit-open/blob/e867e5e9fa0e2325836f7d757eb4a27814587a9e/frontends/infinite-corridor/src/api/learning-resources/urls.ts#L209 but a bit less manual / more organized.

See also:
- https://github.com/lukemorales/query-key-factory
- https://tanstack.com/query/v4/docs/react/community/lukemorales-query-key-factory

abeglova (approved):
LGTM

Issue #41, opened 2023-08-18
Adapt Learning Resource Cards for new APIs


ChristopherChudzicki:
We should be able to display Learning Resource Cards similar to the old ones but using the new APIs.

### Acceptance Criteria
- [ ] Resource card component should handle new learning resources API response

### Out of scope
- [ ] clicking on titles won't (yet) open drawer
- [ ] userlist / stafflist dialog controls

PR #49, opened 2023-08-23
Adapt resource cards for new API format

### LINKED ISSUE #32: Filter parameters in OpenAPI spec/docs

ChristopherChudzicki:
#41 

#### What's this PR do?
This PR:
1. Adapts our LearningResourceCard component for new API results.
    - *this will need to be improved as more resource types are added.*
2. Puts a carousel on the homepage API results. 
    - *It's labeled "upcoming courses", but it's actually just `learning_resources/`*; I'd like to resolve #32 before making it "upcoming courses". 

1. Load data via `./manage.py backpopulate_xpro_data` if you haven't already
2. Visit `http://localhost:8063/` *The search bar isn't really functional right now.*

#### Screenshots (if appropriate)
Screenshot 2023-08-24 at 1 40 35 PM


frontends/infinite-corridor/src/infinite-pages/HomePage.test.tsx
ChristopherChudzicki, line 1: I decided just to remove this.

This page **is** useful as a reference, but we have it in https://github.com/mitodl/open-discussions and at https://discussions-rc.odl.mit.edu/infinite. We don't need it here, too.

frontends/ol-search-ui/assets/learning-resource-card.scss
ChristopherChudzicki, line 1: The change here is that all the rules are nested under `..ol-lrc-root-old ` now.

frontends/ol-learning-resources/src/components/LearningResourceCard.tsx
```diff
@@ -0,0 +1,225 @@
+import React, { useCallback } from "react"
```
abeglova, line 1: Do we need a version of LearningResourceCard both here and in infinite-corridor/components? Also is this LearningResourceCard used anywhere? the Home.tsx uses the other one

abeglova (changes requested):

frontends/ol-learning-resources/src/utils/index.ts
```diff
@@ -0,0 +1,96 @@
...
+  if (!a.isValid()) return 1
+  if (!b.isValid()) return -1
+  // if both valid, sort descending
+  return -a.diff(b)
+}
+
+/**
+ * Find "best" running: prefer current, then nearest future, then nearest past.
+ */
+const findBestRun = (
```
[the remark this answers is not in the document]
ChristopherChudzicki: I'm not sure. I would prefer logic like this on the backend. On the other hand, for an API consumable by outside world, the sorting we do
>  prefer current, then nearest future, then nearest past.
seems a little arbitrary / specific.

abeglova (approved):
lgtm

PR #53, opened 2023-08-25
Faster yarn commands (linting, formatting)

ChristopherChudzicki:
None

#### What's this PR do?
This PR makes our yarn commands faster.

| cmd           | before | now  |
|---------------|--------|------|
| fmt-fix       | 39     | 15s  |
| lint-fix      | 25s    | 9s   |
| scss-lint-fix | 14s    | 3.5s |

1. Introduce a linting error (like an unused, unexported variable, `const five = 5`) and run `docker compose exec watch yarn lint-check`. You should get an error. (Check the exit code with `echo $?`.) 
2. Repeating for formatting and SCSS linting errors.


package.json
```diff
@@ -1,26 +1,23 @@
...
     "private": true,
     "workspaces": [
         "frontends/*"
     ],
     "scripts": {
-        "foreach-with-fallback": "yarn workspaces foreach --verbose --topological exec node ${PROJECT_CWD}/scripts/yarn-run-with-fallback.js $0 $1",
-        "start-dev": "yarn workspaces foreach --verbose --parallel --interlaced run start-dev",
+        "start-dev": "yarn workspaces foreach -pvi run start-dev",
         "build": "yarn workspaces foreach run build",
-        "clean": "run foreach-with-fallback clean global:clean",
```
ChristopherChudzicki, line 11: We weren't using `clean`, so I got rid of it.

Previously, we were running yarn commands via `run foreach-with-fallback <cmd> <fallback-cmd>`, the idea being that workspaces could define their own `cmd`, and default to a globally-defined `fallback-cmd`.

In https://github.com/mitodl/open-discussions this was useful because we were supporting differing linting/typechecking for old Flow+JS code than for new TS code.

It was a little complicated and is unnecessary now that everything is TS: all projects use the same linting, typechecking, etc commands.

package.json
```diff
@@ -1,26 +1,23 @@
 {
-    "name": "open-discussions-root",
+    "name": "root",
     "private": true,
     "workspaces": [
         "frontends/*"
     ],
     "scripts": {
-        "foreach-with-fallback": "yarn workspaces foreach --verbose --topological exec node ${PROJECT_CWD}/scripts/yarn-run-with-fallback.js $0 $1",
```
ChristopherChudzicki, line 8: Previously, we were running yarn commands via `run foreach-with-fallback <cmd> <fallback-cmd>`, the idea being that workspaces could define their own `cmd`, and default to a globally-defined `fallback-cmd`.

In https://github.com/mitodl/open-discussions this was useful because we were supporting differing linting/typechecking for old Flow+JS code than for new TS code.

It was a little complicated and is unnecessary now that everything is TS: all projects use the same linting, typechecking, etc commands.

package.json
```diff
@@ -1,26 +1,23 @@
...
-        "start-dev": "yarn workspaces foreach --verbose --parallel --interlaced run start-dev",
+        "foreach": "yarn workspaces foreach -pvi --exclude=root run",
```
ChristopherChudzicki, line 8: Previously, we were running yarn commands via `yarn-run-with-fallback.js <cmd> <fallback-cmd>`, the idea being that workspaces could define their own `cmd`, and default to a globally-defined `fallback-cmd`.

In https://github.com/mitodl/open-discussions this was useful because we were supporting differing linting/typechecking for old Flow+JS code than for new TS code.

It was a little complicated and is unnecessary now that everything is TS: all projects use the same linting, typechecking, etc commands.

This is much faster: *some* of the speedup comes from not using `yarn-run-with-fallback.js`, most of the speedup comes from running the `foreach` in parallel (`-p`), which we could have been doing any time since
- https://github.com/mitodl/open-discussions/pull/3874
was merged

abeglova (approved):
lgtm

PR #60, opened 2023-08-29
annotate /new, /upcoming as lists

### LINKED ISSUE #59: /upcoming and /new should be paginated in OpenAPI
### Steps to Reproduce
1. View the OpenAPI spec for `/api/v1/courses/upcoming/` https://github.com/mitodl/mit-open/blob/4713309a6fefc3d2fa4f0fea188ef04939fce9f1/openapi.yaml#L89

### Expected Behavior
OpenAPI spec indicates response is `PaginatedLearningResourceList` (which is what the API actually responds with).

### Actual Behavior

OpenAPI spec indicates response is `LearningResource`.

ChristopherChudzicki:
Fixes #59 

#### What's this PR do?
This PR annotates the `/new` and `/upcoming` learning resource APIs as returning lists.

1. Check that https://mit-open-rc.odl.mit.edu/api/v1/learning_resources/new/ and https://mit-open-rc.odl.mit.edu/api/v1/learning_resources/upcoming/ return paginated lists
2. Check that the OpenAPI schema indicates those APIs return lists.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,3 +1,3 @@
 cassettes/* linguist-generated=true
 factory_data/* linguist-generated=true
-flow-typed/* linguist-generated=true
+frontends/api/src/generated/* linguist-generated=true
--- a/learning_resources/views.py
+++ b/learning_resources/views.py
@@ -113,6 +113,7 @@ class LearningResourceViewSet(viewsets.ReadOnlyModelViewSet):
         """
         return self._get_base_queryset()
 
+    @extend_schema(responses=LearningResourceSerializer(many=True))
     @action(methods=["GET"], detail=False, name="New Resources")
     def new(self, request: Request) -> QuerySet:
         """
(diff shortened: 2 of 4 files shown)


.gitattributes
```diff
@@ -1,3 +1,3 @@
 cassettes/* linguist-generated=true
 factory_data/* linguist-generated=true
-flow-typed/* linguist-generated=true
+frontends/api/src/generated/* linguist-generated=true
```
ChristopherChudzicki, line 3: I didn't mark `openapi.yaml` as generated (even though it is) because, IMO, it is useful to see its diffs in PRs

learning_resources/views.py
```diff
@@ -113,6 +113,7 @@ def get_queryset(self) -> QuerySet:
         """
         return self._get_base_queryset()
 
+    @extend_schema(responses=LearningResourceSerializer(many=True))
     @action(methods=["GET"], detail=False, name="New Resources")
```
ChristopherChudzicki, line 117: From the drf-spectacular [FAQ](https://drf-spectacular.readthedocs.io/en/latest/faq.html#i-m-using-action-detail-false-but-the-response-schema-is-not-a-list):

> I'm using `@action(detail=False)` but the response schema is not a list
> -------------------------------------------------------------------------
>
> `detail=True/False` only specifies whether the action should be routed at `x/{id}/action` or `x/action`. The `detail` parameter in itself makes no statement about the action's response. Also note that the default for underspecified endpoints is a non-list response. To signal a listed response, you can use `@extend_schema(responses=XSerializer(many=True))`.

So since you could create non-list responses when `detail=False`, drf-spectacular doesn't automatically listify them.

mbertrand (approved):
👍

PR #62, opened 2023-08-30
Add pre-commit rules for all linting and replace Pylint with Ruff

blarghmatey:
N/A

This does a few things:
- Replaces usage of pylint with ruff for linting Python code
- Applies all automated Ruff fixes to the Python code
- Adds noqa comments to all existing linter violations found by Ruff
- Applies prettier and eslint fixes to all relevant files
- Updates eslint config to be compatible with prettier
- Upgrades Prettier to v3.x
- Adds pre-commit rules for running all linters and fixes automatically
- Removes issue and PR templates from the repo so that it will fall back to the ones in mitodl/.github

# How can this be tested?
This was tested by running `pre-commit run --all-files` and ensuring that the output all passes. There are two errors from the `prettier` hook that I didn't want to modify directly, instead preferring to leave it to someone who has a better understanding of what those changes would result in.

# Additional Context
This is related to the proposals outlined in https://github.com/mitodl/hq/discussions/1422


collinpreston:
It might be good to update the following README sections with instructions for linting: https://github.com/mitodl/mit-open#commits & https://github.com/mitodl/mit-open#testing-formatting--code-generation

package.json
```diff
@@ -1,83 +1,83 @@
...
-        "frontends/*"
-    ],
-    "scripts": {
-        "foreach": "yarn workspaces foreach -pvi --exclude=root run",
-        "start-dev": "yarn workspaces foreach -pvi run start-dev",
-        "build": "yarn workspaces foreach run build",
-        "lint-check": "yarn foreach global:lint-check",
-        "lint-fix": "yarn foreach global:lint-fix",
-        "fmt-check": "yarn foreach global:fmt-check",
-        "fmt-fix": "yarn foreach global:fmt-fix",
```
ChristopherChudzicki, line 14: Right `yarn fmt-check` (which we use on CI and in `./scripts/generate-openapi`) is only formatting JS/TS files.

I would suggest:
1. Change these two lines to `prettier --check .` and `prettier --write .`, respectively
    - I suspect this will also fix your CI issue... Now prettier will always run from the root, and should pick up `.prettierignore` and `.gitignore` automatically. [^1]
2. With (1), you should be able to clean out `.prettierignore`, except for `.yarn/` (which we want to commit, but we probably don't want to format).
3. Delete the `global:format-check` and `global:format-fix` commands below.

[^1]: I suspect it was working before because of some difference with running through prettier-eslint. :shrug:

ChristopherChudzicki (commented):
I haven't looked too much on the Python side yet, but on the JS/TS side I think this is a great improvement. I really like that we are running Prettier by itself now, not via eslint through `prettier-eslint`. "Plain Prettier + Plain Eslint" has better tooling and fewer issues, in my experience.

We do lose some customizability by not running prettier through prettier-eslint (e.g., our object keys / values are not vertically aigned). But IMO that's fine. The important thing is consistency and auto-fixes. *And the vertical alignment, while pretty, always made our git history a little noiser.*

I left once suggestion re Prettier on CI.

Regarding the failing prettier checks when running `pre-commit run --all-files`... Both of those are failing on invalid HTML in django templates. (Another great reason to be running prettier on html files). That should be easy to fix, but @collinpreston might be removing those as part of  #64 anyway (unsure)

ChristopherChudzicki:
> One question I have is whether we want to continue incorporating the lint/format step in the GH Actions workflows, or if we can just delegate that to https://pre-commit.ci/ which we are using in ol-infrastructure and ol-data-platform.

Am I understanding correctly that with https://pre-commit.ci/:
- if your PR has issues that can be fixed automatically (like formatting), then a commit is automatically added to the PR with those changes? 

If yes... that seems desirable, given how ci checks fail due to linting.

blarghmatey:
Yes, you are correct in that assessment. If e.g. Black or prettier changes need to be applied, pre-commit CI will automatically add a commit to the PR with the fixes, as defined in the pre-commit config for the repo.

blarghmatey:
It looks like the only remaining failures are related to problems in the Reddit code which will be removed as part of https://github.com/mitodl/mit-open/pull/64

blarghmatey:
This is passing tests now. The issue was that prettier was naively reformatting the Django templates, which led to invalid template files. I resolved that by adding a plugin that understands the template syntax and leads to correct formatting.

blarghmatey:
@ChristopherChudzicki can you take a look again? All of the tests are passing. You can ignore the failing check for pre-commit. I've got a patch in for the CI that will let us skip that specific step in that environment since prettierv3 + plugins currently won't work with how they get run there.

ChristopherChudzicki (changes requested):
#### Pre-commit linting failing
@blarghmatey The pre-commit linting is failing for me locally, `pre-commit run --all-files`:

```
/Users/cchudzicki/dev/open/scripts/repl.js
  3:1  error  'R' is not defined       no-undef
  4:1  error  'moment' is not defined  no-undef
  5:1  error  'S' is not defined       no-undef
  6:1  error  'sinon' is not defined   no-undef
  7:1  error  'assert' is not defined  no-undef
  8:1  error  'casual' is not defined  no-undef
```

That file is very outdated anyway, and `jest`'s watch mode is probably just as convenient for what it was supposed to accomplish. Let's just remove that file.

#### Prettier & pre-commit
Re
>  I've got a patch in for the CI that will let us skip that specific step in that environment since prettierv3 + plugins currently won't work with how they get run there.

I'm not too familiar with `pre-commit`, but since it's all NPM-based, could we just run it in the local section like we do for `eslint` and `stylelint`?

blarghmatey:
Thanks @ChristopherChudzicki that file is deleted now. As to your question about Prettier, the answer is that yes, you can rely on pre-commit running locally for those Django templates (as well as all other checks/fixes) but prettier will also still run for all of its other target file types as part of pre-commit CI

ChristopherChudzicki (approved):
Now it all passes 👍  😄

Issue #67, opened 2023-09-05
/learningpaths and /learning_resources?resource_type=learning_path return different results


ChristopherChudzicki:
As a logged-in user with permission to edit learning paths:
- http://localhost:8063/api/v1/learningpaths/ returns all paths (published + unpublished)
- http://localhost:8063/api/v1/learning_resource/resource_type=learning_path returns only published paths

I would expect the results to be the same.

*For context: It seems most convenient on the frontend to just use the unified API `/learning_resource` API, which is how I came across this.*

mbertrand:
Since learning paths are CRUDable, we need at least one API to return both published and unpublished resources of this type (but only to people who have permissions to edit them), which is what the `learningpaths` endpoint is aimed at.  The learning_resources endpoint could be modified to do the same but only for LearningPaths and only for authorized users, if that is desirable.

mbertrand:
PS the LearningPathViewSet also has custom create and destroy functions for CRUD functionality that the LearningResourceViewSet lacks.

PR #71, opened 2023-09-06
Filter 'published' the same for /learningpaths and /learning_resource?resource_type=learning_path

ChristopherChudzicki:
A potential fix for #67 

#### What's this PR do?
This PR refactors how the published filtering works so that `/learningpaths` and `/learning_resources?resource_type=learning_path` return the same results.

1. View http://localhost:8063/api/v1/learningpaths/ and http://localhost:8063/api/v1/learning_resources/?resource_type=learning_path as anonymous user or a user without `learning_path_editor` permission (or as admin).
    - both endpoints should **not** include unpublished learningpaths
2. Repeat (1) but now with the `learning_path_editor` permission (or as admin). 
    - both endpoints **should** include unpublished learningpaths

### DIFF
--- a/learning_resources/views.py
+++ b/learning_resources/views.py
@@ -122,6 +122,15 @@ class LearningResourceViewSet(viewsets.ReadOnlyModelViewSet):
             *([item.value for item in LearningResourceType]),
         )
         lr_query = lr_query.prefetch_related(*prefetches).distinct()
+
+        published_filter = Q(published=True)
+
+        if is_learning_path_editor(self.request) or is_admin_user(self.request):
+            published_filter = published_filter | Q(
+                resource_type=LearningResourceType.learning_path.value
+            )
+
+        lr_query = lr_query.filter(published_filter)
         return lr_query
 
     def get_queryset(self) -> QuerySet:
@@ -131,7 +140,7 @@ class LearningResourceViewSet(viewsets.ReadOnlyModelViewSet):
         Returns:
             QuerySet of LearningResource objects
         """
-        return self._get_base_queryset().filter(published=True)
+        return self._get_base_queryset()
 
     @extend_schema(responses=LearningResourceSerializer(many=True))
     @action(methods=["GET"], detail=False, name="New Resources")
@@ -186,9 +195,7 @@ class CourseViewSet(LearningResourceViewSet):
         Returns:
             QuerySet of LearningResource objects that are Courses
         """
-        return self._get_base_queryset(
-            resource_type=LearningResourceType.course.value
-        ).filter(published=True)
+        return self._get_base_queryset(resource_type=LearningResourceType.course.value)
 
 
 class ProgramViewSet(LearningResourceViewSet):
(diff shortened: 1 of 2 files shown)


learning_resources/views.py
```diff
@@ -122,6 +122,15 @@ def _get_base_queryset(self, resource_type: str = None) -> QuerySet:
             *([item.value for item in LearningResourceType]),
         )
         lr_query = lr_query.prefetch_related(*prefetches).distinct()
+
+        published_filter = Q(published=True)
+
+        if is_learning_path_editor(self.request) or is_admin_user(self.request):
```
mbertrand, line 128: I originally thought of this read-only viewset as intended just for browsing published resources, and if one needed to do any CRUD work, a different endpoint would be used (only LearningPath fits that need right now).  I'm fine with discarding that assumption, but one potential issue with this change is that editors/admins will always see both published and unpublished results here for learning paths, and there may be times when it is only appropriate to show published paths (for instance, a carousel of top 10 newest learning paths or top 10 newest resources in general).  One way to avoid this is to add `published` to the `filterset_fields` above, and filter by `?published=true` whenever necessary.  Can you add that in?

mbertrand (changes requested):

Ferdi:
@ChristopherChudzicki can you summarize what the changes are ?

learning_resources/views.py
mbertrand: PS One potential drawback to this though is that using `published=False` would allow any users to see unpublished courses, programs, videos, paths, etc - which may or may not be desirable?  So might need to somehow force `published=True` for anyone except admins (and editors for learningpaths only)

mbertrand: Maybe add an `api/v1/learning_resources/published/` endpoint to be used by carousels and such?

ChristopherChudzicki:
@Ferdi My goal with this change was to make it so that if a user has permission view a learningpath, then it is visible at `/learning_resources/:id` or `/learnpaths:id/`. Currently, `/learning_resources/` will not return unpublished learningpaths, whereas `/learningpaths/` will.

@mbertrand
> I originally thought of this read-only viewset as intended just for browsing published resources, and if one needed to do any CRUD work, a different endpoint would be used (only LearningPath fits that need right now).

I was not thinking to use the `/learning_resource/` endpoints for editing, but for populating a list of things that might be edited: It does seem easier to me when fetching details or listings to use the single API `/learning_resources/`: 

For example: the frontend caches API responses. If we sometimes request LearningPaths from `/learning_resources/:id`, and sometimes from `/learningpaths/:id`, then editing a LearningPath requires two cache invalidations instead of one. *This isn't a big deal. Two isn't much more complicated.*

@mbertrand Re
> there may be times when it is only appropriate to show published paths (for instance, a carousel of top 10 newest learning paths or top 10 newest resources in general). One way to avoid this is to add published to the filterset_fields above, and filter by ?published=true whenever necessary. Can you add that in?

Good point. If we do change this API, then rather than explicitly saying "published only", maybe `published=true` should be the default, and we could filter to explicitly allow unpublished.

ChristopherChudzicki:
@mbertrand I'm going to close this for now, but I'll leave the issue open:

For the learningpaths work, `/learningpaths/:id` has an advantage over `learning_resources/:id`: the former will 404 if the id is not a learningpath, which allows the frontend to redirect naturally if `/frontend_page/learningspaths/123` is actually a course. 

You raise a good point point about published filtering; I left a note about that in #67 
 Continuing to use `/learningpaths` for the fetching seems just as easy

PR #72, opened 2023-09-07
UserList model and API

### LINKED ISSUE #42: UserList model and API
Implement user lists 

#### Designs/Acceptance Criteria

- [ ] UserList should have just title and description, no topics
- [ ] UserList should not be a learning resource
- [ ] Userlist can be private or unlisted (not indexable by google but link can be shared)

mbertrand:
#### Pre-Flight checklist

- [x] Migrations
  - [x] Migration is backwards-compatible with current production code
- [x] Testing
  - [x] Code is tested
  - [x] Changes have been manually tested

Closes #42 

#### What's this PR do?
- Adds `UserList` model, permissions, serializers, and views
- Puts some topic serializer functionality into a separate `WriteableTopicsMixin` class to avoid code duplication

- Go to the `http://localhost:8063/api/v1/userlists/` as an anonymous user, you should get a 403 response
- Log in (as anyone), go to the same url, and now you should be able to successfully post the following to create a new user list:
  ```
  {
    "topics": [1,2],
    "title": "My 1st UserList",
    "description": "My UL description"
  }
  ```
- Go to `http://localhost:8063/api/v1/userlists/<id>/` and PATCH this response, it should update the title.
  ```
  {
    "title": "My 1st UserList, modified",
  }
  ```
- Go to `http://localhost:8063/api/v1/userlists/<id>/resources/` and post the following, it should add the resource to the list.
  ```
  {
    "child": <id of a course or program learning resource>
  }
  ```

- Refresh `http://localhost:8063/api/v1/userlists/<id>/resources/`, you should see all the standard details of the added resource under `results`, along with a `position` of 0
- Add another, it should be added with a `position` of 1
- Go to `http://localhost:8063/api/v1/userlists/<user_list resource id>/resources/<user_list_relationship id>/` and click `DELETE`, the resource should be deleted from the userlist.
- Log in as another user and go to `http://localhost:8063/api/v1/userlists/`, you should not get any user lists returned.
- Go to `http://localhost:8063/api/v1/userlists/<id>`, you should get a 404


learning_resources/serializers.py
```diff
@@ -36,6 +36,40 @@ class Meta:
         fields = ["id", "name"]
 
 
+class WriteableTopicsMixin(serializers.Serializer):
+    """Class for editable topics"""
+
+    topics = WriteableSerializerMethodField()
+
+    def validate_topics(self, topics):
+        """Validator for topics"""  # noqa: D401
```
ChristopherChudzicki, line 45: IMO: here and elsewhere, might as well just write it the way the linter suggests.

```suggestion
        """Validate specified topics exist.""
```

learning_resources/serializers.py
```diff
@@ -36,6 +36,40 @@ class Meta:
...
+            except ValueError as ve:
+                msg = "Topic ids must be integers"
+                raise ValidationError(msg) from ve
+            missing = set(topics).difference(valid_topic_ids)
+            if missing:
+                msg = f"Invalid topic ids: {missing}"
+                raise ValidationError(msg)
+        return {"topics": topics}
+
+    @extend_schema_field(LearningResourceTopicSerializer(many=True, allow_null=True))
```
ChristopherChudzicki, line 64:
```suggestion
    @extend_schema_field(LearningResourceTopicSerializer(many=True))
```

I **think** we do not want `allow_null=True` here.

- When I try PATCHing http://localhost:8063/api/v1/userlists/1/ with `{ "topics": null }`, I get a validation error. So it seems like requests should not have null topics.
- `get_topics()` always returns a list and as far as I can tell responses always have listified topics (possibly empty).

Note: If we remove `allow_null=True`, drf-spectacular's OpenAPI will drop `nullable: true`, but will _not_ make `topics` required on requests.

learning_resources/serializers.py
```diff
@@ -129,17 +163,27 @@ class Meta:
...
+    Serializer containing only parent and child ids for a learning path relationship
     """
 
     class Meta:
         model = models.LearningResourceRelationship
         fields = ("id", "parent_id", "child_id")
 
 
-class LearningResourceBaseSerializer(serializers.ModelSerializer):
+class MicroUserListRelationshipSerializer(serializers.ModelSerializer):
```
ChristopherChudzicki, line 176: Do you mind returning these as `parent/child`:

```
    parent = serializers.ReadOnlyField(source="parent_id")
    child = serializers.ReadOnlyField(source="child_id")
```

#77 Does this also `MicroLearningPathRelationshipSerializer` ... Maybe just change them both here and I can deal with the merge conflict? Since I'm out next week I expect this PR will be merged before #77 

**Rationale:** That way responses from MicroUserListRelationshipSerializer & UserListRelationshipSerializer are more similar.

learning_resources/serializers.py
```diff
@@ -129,17 +163,27 @@ class Meta:
...
+    """
+    Serializer containing only  parent and child ids for a user list relationship
```
ChristopherChudzicki, line 178: Extra space here.

learning_resources/serializers.py
```diff
@@ -438,3 +468,90 @@ class Meta:
...
+class UserListSerializer(serializers.ModelSerializer, WriteableTopicsMixin):
+    """
+    Simplified serializer for UserList model.
+    """
+
+    item_count = serializers.SerializerMethodField()
+
+    def get_item_count(self, instance) -> int:
+        """Return the number of items in the list"""
+        return getattr(instance, "item_count", None) or instance.resources.count()
```
ChristopherChudzicki, line 482: Question: When would the first condition be used vs the second?

I would have thought it would also use `instance.resources.count()` since the model doesn't have a. `item_count` field/attribute.

learning_resources/serializers.py
```diff
@@ -438,3 +468,90 @@ class Meta:
...
+            topics_data = validated_data.pop("topics", [])
+            with transaction.atomic():
+                userlist = super().create(validated_data)
+                userlist.topics.set(
+                    models.LearningResourceTopic.objects.filter(id__in=topics_data)
+                )
+            return userlist
+        return None
+
+    def update(self, instance, validated_data):
```
ChristopherChudzicki, line 498: **Question:** Does the docstring here mean
1. "This method will check the list is authored by requesting user"
2. or does it mean "Be sure the list is authored by the requesting user before anything calls this method".
3. or maybe something else.

I don't see (1) happening... That's taken care of by the permission class, yes?

learning_resources/models.py
```diff
@@ -356,3 +356,36 @@ class ContentFile(TimestampedModel):
...
+    and their relative positions in the list.
+    """
+
+    parent = models.ForeignKey(
+        UserList, on_delete=models.deletion.CASCADE, related_name="children"
+    )
+    child = models.ForeignKey(
+        LearningResource, related_name="user_lists", on_delete=models.deletion.CASCADE
+    )
+    position = models.PositiveIntegerField(default=0)
```
ChristopherChudzicki, line 391: When I POST a resource a newly-created, empty userlist, it gets position 1 (not 0). Weird.

ChristopherChudzicki (changes requested):
Working well. I left a few minor comments/questions.

learning_resources/models.py
```diff
@@ -356,3 +356,36 @@ class ContentFile(TimestampedModel):
     class Meta:
         unique_together = (("key", "run"),)
         verbose_name = "contentfile"
+
+
+class UserList(TimestampedModel):
+    """
+    Similar in concept to a LearningPath: a list of learning resources.
+    However, UserLists are not considered LearningResources because they
+    should only be accessible to the user who created them.
```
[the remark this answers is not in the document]
mbertrand: Forgot about that part, will add it in

learning_resources/serializers.py
mbertrand: `item_count` is added via the view queryset:
```python
    def get_queryset(self):
        """Return a queryset for this user"""
        return (
            UserList.objects.all()
            .prefetch_related("author", "topics")
            .annotate(item_count=Count("children"))
        )
```
Copied over from `course_catalog`, I believe it may be [marginally more efficient](https://stackoverflow.com/questions/7957569/in-django-is-aggregatecount-faster-or-better-than-count-in-anyway) to run `annotate(item_count=Count("children")` than `instance.resources.count()` - but the latter is there as a backup in case the serializer is used outside of a view.

learning_resources/serializers.py
mbertrand: It's making sure that if an API PATCH/POST request sends `{"author": <some_other_author_id>}", it will be overwritten by the id of the actual requestor.

```python
validated_data["author"] = request.user
```

learning_resources/models.py
mbertrand: For both learningpaths and userlists, the following code is used, which will make the 1st addition have a position of 1:

```python
        position = (
            items.aggregate(Max("position"))["position__max"] or items.count()
        ) + 1
```

learning_resources/serializers.py
```diff
@@ -129,35 +163,68 @@ class Meta:
...
-    topics = WriteableSerializerMethodField()
     course = CourseSerializer(read_only=True, allow_null=True)
     learning_path = LearningPathSerializer(read_only=True, allow_null=True)
     runs = LearningResourceRunSerializer(read_only=True, many=True, allow_null=True)
+    image = serializers.SerializerMethodField()
     learning_path_parents = serializers.SerializerMethodField()
+    user_list_parents = serializers.SerializerMethodField()
 
-    @extend_schema_field(MicroRelationshipSerializer(many=True, allow_null=True))
+    def get_image(self, instance) -> dict:
```
ChristopherChudzicki, line 207: **Requested Change:** Could you add 
```python
@extend_schema_field(LearningResourceImageSerializer(allow_null=True))
```
here?

learning_resources/serializers.py
```diff
@@ -438,3 +481,98 @@ class Meta:
...
+            if topics_data is not None:
+                userlist.topics.set(
+                    models.LearningResourceTopic.objects.filter(id__in=topics_data)
+                )
+            return userlist
+
+    class Meta:
+        model = models.UserList
+        exclude = COMMON_IGNORED_FIELDS
+        read_only_fields = ["author", "resources"]
```
ChristopherChudzicki, line 535: **Comment:** One thing I noticed is that the `resources` property and the `/userlist/<id>/resources` endpoint don't necessarily return the same order. 

Maybe we could order it or exclude it.

ChristopherChudzicki (approved):
New privacy options are working well, and images Look good 👍 👍 

I left one requested change but I don't think it needs a re-review: Just a drf-spectacular annotation.

Also a comment about `resources`

**BTW:** I merged this into my branch #77. After adding the `drf-spectacular` annotation, the conflicts + type errors were very simple. (Mostly just renaming MicroRelationship).

Issue #36, opened 2023-08-18
Reconsider `LearningResource.program` serializer


ChristopherChudzicki:
Currently, `/api/v1/learning_resources` returns objects like:
```ts
interface LearningResource {
...lots_of_properties,
course: |
    null |
    { extra_course_numbers: string[] } // Extra course properties,
program: |
    null |
    { courses: LearningResourceProgramCourse } // array of courses
}
```

`program.courses` is *almost* an array of Learning Resources, but not quite. The `LearningResourceProgramCourse` objects differ in a few ways:

1. They're courses, but they don't have a `course` property.
2. They have some extra fields like `created_on`, `updated_at`. *No big deal.*

### Suggestion

Either make `program.courses` the same shape as the courses returned by `/api/v1learning_resources/`, OR return an array of references instead.

PR #73, opened 2023-09-08
Reworked serializers and views to normalize the respsonses

rhysyngsun:

#36 

- Updates `program.courses` under a program learning resource to be the course learning resource instead of the literal `Course` object. This course learning resource is also more strongly typed than the general learning resource type because we know the objects are only of one type.
- Updated the serializers to utilize a disjoint union model, rather than a monolithic type. This will make it easier to reason about these types in the frontend without have to do assertions on nullable fields all over the place.
- Updated the API views for courses and programs to have more specifically typed responses. Compare these two APIs now:
  - `/api/v1/learningresources/`:
    
    Screenshot 2023-09-08 at 18-34-43 MIT Open Discussions Course Catalog API

  - `/api/v1/courses/`:
    
    Screenshot 2023-09-08 at 18-31-52 MIT Open Discussions Course Catalog API

- Removed `/upcoming/` and `/new/` APIs from learning paths - I don't think these make sense for that resource type. This involved some view refactoring. I cleaned up some of the OpenAPI descriptions in the process too.
- Removed filtering by the `resource_type` field from APIs other than `/api/v1/learningresources/`, This doesn't make sense on those APIs as they're prefiltered and specifying the parameter would be undefined behavior.

- [ ] Desktop screenshots

# How can this be tested?

# Additional Context


mbertrand (commented):
A few initial observations:

[This line](https://github.com/mitodl/mit-open/pull/73/files#diff-638dcb3aea0f7898675d7c87d8c639d4c572fa2a797f346a62c772db215119d7L202) is now raising an exception `LearningResourceRelationship' object has no attribute 'topics` when trying to view programs at `http://localhost:8063/api/v1/learning_resources/?resource_type=program` or `http://localhost:8063/api/v1/programs`


-------------------

> Updates program.courses under a program learning resource to be the course learning resource instead of the literal Course object.

Not sure if I'm reading this correctly, the current code in main already returns the courses as `LearningResources` under `program.courses` instead of the literal `Course` objects:

```
            "program": {
                "courses": [
                    {
                        "id": 8,
                        "offered_by": [
                            "xPRO"
                        ],
                        "resource_content_tags": [],
                        "image": {
                            "id": 8,
                            "url": "https://xpro-app-production.s3.amazonaws.com/original_images/SysEng_Course_4_Card_New.jpg",
                            "description": null,
                            "alt": null
                        },
                        "department": null,
                        "audience": "Professional Offerings",
                        "certification": "Certificates",
                        "prices": [
                            1129.0
                        ],
                        "topics": [
                            {
                                "id": 2,
                                "name": "Engineering"
                            }
                        ],
                        "course": {
                            "extra_course_numbers": null
                        },
                        etc
```            

Comparing the json result for a course between this branch and main, the biggest difference I see is that irrelevant fields are no longer included, for example no `"program": null` or `"learning_path": null`, which is definitely an improvement.


-------------------

> Removed /upcoming/ and /new/ APIs from learning paths - I don't think these make sense for that resource type. 

Definitely agree on `/upcoming`, not so sure about `/new/` - maybe we would want to display a carousel of the newest learning paths at some point?

ChristopherChudzicki (commented):
The generated OpenAPI spec and Typescript code both look exactly how I would expect for a discriminated union 👍 👍.  

Related to @mbertrand 's comment:
> Comparing the json result for a course between this branch and main, the biggest difference I see is that irrelevant fields are no longer included, for example no "program": null or "learning_path": null, which is definitely an improvement.

This makes the responses of `/learning_resources` and `/learning_paths` line up much more nicely. In particular, now `LearningPathResource` is assignmable to `LearningResource`, which is great.

PR #99, opened 2023-09-22
Adjust OfferedBy and Platform values

mbertrand:
# What are the relevant tickets?
Closes #9 (mostly - care will need to be taken to assign correct platform and offered_by values when completing #102)

# Description (What does it do?)
- Modifies the valid `LearningResourceOfferor` and `LearningPlatform` objects
- Adds platform as a filterable api query parameter
- Changes `offered_by` to a single foreign key relation on `LearningResource`

# How can this be tested?
- Run the migrations (`docker compose up` or `docker compose run --rm web ./manage.py migrate`)
- Run `./manage.py update_platforms` and `./manage.py update_offered_by`
- Run `./manage.py backpopulate_xpro_data --delete` and `./manage.py backpopulate_podcast_data --delete`
- Run `./manage.py backpopulate_xpro_data` and `./manage.py backpopulate_podcast_data`
- All xpro courses and programs should now be offered by  `xPRO`, and a mix of `platform` values, one per resource: `emeritus`, `xpro`, `globalalumni`, `whu`, etc
- All podcasts and episodes should have platform `podcast` and various `offered_by` values: `None`, `CSAIL`, etc.
- The homepage courses carousel should display `Offered By - xPRO` for each resource.

You should be able to filter the API by platform and offered_by:

http://localhost:8063/api/v1/learning_resources/?platform=emeritus
http://localhost:8063/api/v1/learning_resources/?offered_by=xpro

PR #112, opened 2023-09-29
Remove LearningResourcePlatform.audience field, add LearningResource.is_professional field

### LINKED ISSUE #97: Change type/location of field for `professional offerings` vs `open content`
# User Story
Users should be able to filter API responses by whether or not a `LearningResource` is a "Professional Offering" or not.  Currently that is stored in `LearningResourcePlatform.audience` but that should be changed.

# Description/Context
- Remove the `LearningResourcePlatform.audience`  field
- Add a `LearningResource.is_professional` boolean field (`True` if a "Professional Offering", False otherwise)
- Update the serializers and views as necessary  (Q: what are the opensearch implications if this is a boolean? )
- Add an API filter for this field
- For now, don't worry about "Open Content", which may or may not be the opposite of `is_professional`

# Acceptance Criteria

- [ ] User can filter the API response by `is_professional` == True or False

mbertrand:
Closes #97
Based on PR #99 (which should be reviewed and merged first)

- Gets rid of the `LearningResourcePlatform.audience` field
- Adds a `LearningResource.is_professional` boolean field
- Fixes everything that broke as a result

# How can this be tested?
- Run `docker compose up` or run migrations via mgmt command
- Run `docker compose run --rm web ./manage.py backpopulate_xpro_data`
- Check that all the xpro course/program `LearningResource` objects now have `is_professional=True`
- Go to `http://localhost:8063/learning_resources/?professional=true`, you should get the xpro courses and/or programs returned
- Go to `http://localhost:8063/learning_resources/?professional=false`, you should get no courses/programs, just  podcasts/episodes returned if you previously ran `backpopulate_podcast_data`, or nothing at all otherwise.

# Additional Context
Whether a course/program is professional or not currently depends on which `LearningResourceOfferor` it is related to, so the new field could have been added to `LearningResourceOfferor` instead.  But I decided to add it to `LearningResource` so that additional criteria (or exceptions) could be accounted for in the future.


learning_resources/filters.py
```diff
@@ -53,23 +49,15 @@ def filter_offered_by(self, queryset, _, value):
...
     def filter_platform(self, queryset, _, value):
         """Platform Filter for learning resources"""
         return queryset.filter(platform__platform=value)
 
     class Meta:
         model = LearningResource
         fields = [
-            "platform__audience",
-            "platform__platform",
+            "is_professional",
```
ChristopherChudzicki, line 59: I **think** that this should just be `professional`.

Currently `?professional=true` and `?is_professional=true` both work for filtering and both show up at http://localhost:8063/api/v1/schema/redoc/#tag/learning_resources/operation/learning_resources_list

**Or:** could just use `is_professional` as the name 🤷

learning_resources/serializers.py
ChristopherChudzicki, line 1: **Question:** Would it be appropriate to list `is_professional` in `read_only_fields` for LearningResourceBaseSerializerr?

I mention it because I noticed that in the OpenAPI spec generated by drf-spectacular, `is_professional` is not listed as required. It turns out that listing it as read-only will make it required, so if that's appropriate anyway, might be good.

IMO, the fact that drf-spectacular is not recognizing it as required already is wrong/disappointing. It's discussed a b it in this issue: https://github.com/tfranzel/drf-spectacular/issues/810

> You could make the argument that for a ModelSerializer on responses (and only there), the fields are always present. I don't think we make that distinction yet. But I am also not 100% sure if that always holds true. Some more investigation would be needed.

ChristopherChudzicki (changes requested):
@mbertrand I think the `is_professional` filter label is worth changing so that it only appears in the http://localhost:8063/api/v1/schema/redoc/#tag/learning_resources/operation/learning_resources_list docs once.

If I'm wrong/misunderstood the filter, let me know and I will approve. Otherwise, looks good.

ChristopherChudzicki (approved):

PR #117, opened 2023-10-03
Normalize learning resource serializers

### LINKED ISSUE #36: Reconsider `LearningResource.program` serializer

rhysyngsun:
Closes #36 

- Refactored the serializers into some smaller packages
- Updated `LearningResourceSerializer` to be a polymorphic wrapper around the resource_type specific variants. This now allows the API schema and thus the typescript generated code to be type specific in places where we know what the type is.

**Main branch:**
Screenshot 2023-10-20 at 00-14-59 MIT Open Discussions Course Catalog API

**This Branch:**
Screenshot 2023-10-20 at 00-11-29 MIT Open Discussions Course Catalog API

- Switched `ProgramResourceSerializer.program.courses` to be course learning resources (note the is now typed to `CourseResource`, not a generic LearningResource):

**Main branch:**
Screenshot 2023-10-20 at 00-16-32 MIT Open Discussions Course Catalog API

**This Branch:**
Screenshot 2023-10-20 at 00-12-05 MIT Open Discussions Course Catalog API

- All type-specific APIs now return a specific type using that type's serializer. For example, the programs API:

**Main branch:**
Screenshot 2023-10-20 at 00-19-45 MIT Open Discussions Course Catalog API

**This Branch:**
Screenshot 2023-10-20 at 00-20-13 MIT Open Discussions Course Catalog API

- Updated `factoryboy` and the factory definitions themselves so that the test data they generate is more internally consistent. Tests were breaking otherwise.

# How can this be tested?

- Check out the schema and make sure it makes sense compared to what's on main/RC.
- Test the API in the swagger UI to verify it's returning the revised data structures.
- Functional test the application.


ChristopherChudzicki (changes requested):
Still looking through changes, but I think
```
learning_resources/serializers.py
learning_resources/serializers_test.py
```
should be deleted.

Locally I deleted these and regenerated openapi—no changes. So I don't think these files were affecting anything, really.

frontends/api/src/test-utils/factories/learningResources.ts
```diff
@@ -127,56 +143,91 @@ const learningResource: Factory<LearningResource> = (
...
+  overrides = mergeOverrides(
+    {
+      resource_type: learningResourceType(),
+    },
+    overrides,
+  )
+  switch (overrides.resource_type) {
+    case ResourceTypeEnum.Program:
+      // @ts-expect-error TODO
+      return program(overrides)
```
ChristopherChudzicki, line 161: Could you add a better note here? Something about the individual resource types (ProgramResource, CourseResource...) having a `resource_type` wider than it should be.

ChristopherChudzicki (changes requested):
This looks great to me. The new serializer structure makes sense and the openapi / typescript is much more accurate.

I do think it would be worth resolving the "`resource_type` too wide on ProgramResource, CourseResource, etc types" issue. Currently, `ProgramResource`, `LearningPathResource`, etc, aren't assignable to `LearningResource`, which is... inconvenient (e.g., suppressed errors in the frontend factories). But we don't need to resolve it here.

Noticed a few minor issues (see code comments). Also, one question and one bug:

---

#### Question
 I noticed this error
```
/src/learning_resources/serializers/learning_resources.py:236: Warning [LearningPathItemsViewSet > LearningPathRelationshipSerializer > LearningResourceSerializer]: Encountered 2 components with identical names "LearningResource" and different classes <class 'learning_resources.serializers.learning_resources.LearningResourceSerializer'> and <class 'drf_spectacular.utils.PolymorphicProxySerializer'>. This will very likely result in an incorrect schema. Try renaming one.
```
when running `./scripts/generate_openapi.sh`. Do you know what's going on with that?

#### Bug?
1. Visit http://localhost:8063/api/v1/learningpaths/ *Or the frontend form at http://localhost:8063/learningpaths/*
2. Create a learning path, either by form or by POSTing someting like `{ "title": "Hello, world!" }`
3. Check the response. `resource_type` is `""`. Should be `"learning_path"`.
4.

mbertrand:
I rebased with main (which this PR needs to resolve some conflicts) and ran all the ETL pipelines, didn't notice any problems.

Dockerfile-node
```diff
@@ -1,6 +1,10 @@
 FROM node:16.20.1
```
ChristopherChudzicki, line 1:
```suggestion
FROM node:18.18.2
```
We updated this recently in docker-compose to match Heroku.

rhysyngsun:
Closing in favor of #244 244

Issue #109, opened 2023-09-27
Rich Text Editor


ChristopherChudzicki:
# User Story

As a staff member, I want to create rich-text content to be displayed on MIT Open.

# Description/Context
We want to enable staff users to create rich-text content without assistance from engineering (example: A highlight on new courses on Generative AI).

# Acceptance Criteria

- [x] Propose and implement a model for model for storing rich-text content.
- [x] Expose the model with an API
    - limited to some usergroup *conceptually "OL staff"*
- [x] Implement simple frontend such as `/post/:id` and `/post/:id/edit` for viewing/editing the rich text content. 
    - A CKEditor component already exists on the frontend. See http://discussions-rc.odl.mit.edu/infinite/article. *The [corresponding page](https://mit-open-rc.odl.mit.edu/infinite/article) on mit-open-rc is broken at th moment, I believe due to a missing environment variable.*

# Plan/Design

- **Continue using CKEditor**: In the past we have used CKEditor for rich-text editing in Open Discussions and Studio. It's works well out-of-the-box and is extensible, so we should continue using it.
- **Data Storage**: CKEditor's natural output format is HTML. Let's store data as HTML. (Not as markdown, as in OCW Studio, or as JSON, as in Open Discussions.)
    - **Concern:** We should sanitize the HTML before storing to db, using something like [nh3](https://github.com/messense/nh3)

ChristopherChudzicki:
@ferdi For this initial work, I think it makes sense to keep the data model pretty limited:
```python
class StaffPost(TimestampedModel):
    """
    Stores rich-text content created by staff members.
    """

    html = models.TextField()
    title = models.CharField(max_length=255)
```

We could additionally store author id, although based on the use case we discussed, it seems like the API permission should be similar to LearningPaths—all members of some group ("staff") can create/edit/delete, others can only view.

Is there any additional data that you think it makes sense to store/edit for now?

<img width="1582" alt="Screenshot 2023-10-02 at 3 53 29 PM" src="https://github.com/mitodl/mit-open/assets/9010790/0786fa51-4486-44e9-b7dd-9f63678a7256">

Ferdi:
As simple as possible. we can always add stuff.

PR #120, opened 2023-10-04
add Articles api

ChristopherChudzicki:
API for #109 

This PR implements a very simple `StaffPost` model and API for creation/editing of rich-text content (e.g., staff writes a highlight of newly released courses). 

# How can this be tested?
1. As a staff user, visit http://localhost:8063/api/v1/staff_posts/
2. POST some data, e.g., `{ "html": "<p>Hello world</p>", "title": "First Post" }`. and then try viewing it at http://localhost:8063/api/v1/staff_posts/POST_ID
3. Try deleting/updating posts, and sending suspicious HTML like `<script>` tags. It should be sanitized.
    - The sanitization is overly aggressive right now, but I'd like to tweak that in a subsequent PR and allow more discussion on https://github.com/mitodl/hq/discussions/2600


Ferdi:
I'd call it simply "article"  

"staff" assumes predefined permissions , which might change in the future

articles/serializers.py
```diff
@@ -0,0 +1,28 @@
...
+
+
+@extend_schema_field(str)
+class SanitizedHtmlField(serializers.Field):
+    @staticmethod
+    def to_representation(value):
+        return value
+
+    def to_internal_value(self, data):
+        return nh3.clean(data)
```
ChristopherChudzicki, line 15: See https://github.com/mitodl/hq/discussions/2600 for `nh3` / sanitization discussion.

This is currently too agressive. By default, `nh3.clean(data)` strips classes and data-attributes, which we use for things like embedly cards.

I'd like to configure nh3 more separately from this PR: need experiment more with CKEditor, too, to see exactly what tags we need.

open_discussions/views.py
```diff
@@ -30,6 +30,7 @@ def index(request, **kwargs):  # pylint: disable=unused-argument  # noqa: ARG001
             and (is_admin_user(request) or is_staff_list_editor(request)),
             "is_learning_path_editor": user.is_authenticated
             and (is_admin_user(request) or is_learning_path_editor(request)),
+            "is_article_editor": is_admin_user(request),
```
ChristopherChudzicki, line 33: I thought about using a separate group like we do for `api/v1/learningpaths/` permissions, but I decided to stay simple until we have a more concrete use case around this.

ChristopherChudzicki:
Closing this in favor of #122

Issue #134, opened 2023-10-12
Remove Article Editor feature flag


ChristopherChudzicki:
# Description/Context
We want to remove the env-variable article editor feature flag.

- It's preventing https://mit-open-rc.odl.mit.edu/articles/1 from working on RC right now.
- We don't see any reason to keep it (the flag is ~5 years old)
- Possibly nothing should be feature-flagged in MIT Open right now, but if it were, we should use our new feature flagging platform anyway.

ChristopherChudzicki:
Completed by #135

PR #135, opened 2023-10-12
remove ARTICLE_UI feature flag; bump api version

ChristopherChudzicki:
#134 

This removes the article editor feature flag.

# How can this be tested?
1. Ensure `CKEDITOR_ENVIRONMENT_ID, CKEDITOR_SECRET_KEY, CKEDITOR_UPLOAD_URL` are all set in your env file. You can get them from Heroku.
2. Set `MITOPEN_FEATURES_DEFAULT: "False"` in `docker-compose.yml` so that feature flags are not on by default. **As far as I can tell, this has to be done in docker-compose (or overrides), not `.env`**.
3. `docker compose up` and visit http://localhost:8063/api/v1/ckeditor/. You should get a token (and not a 503).


mbertrand (approved):
👍 ,  something that might be worth doing in this PR or a separate one:

Change this line in `docker-compose.yml` so that `MITOPEN_FEATURES_DEFAULT` will be overridden by the value in your `.env` file if it is there:

```yaml
  MITOPEN_FEATURES_DEFAULT: ${MITOPEN_FEATURES_DEFAULT:-True}
```

PR #140, opened 2023-10-16
search api

abeglova:
https://github.com/mitodl/mit-open/issues/110

This pr replaces the current open search endpoint with an api that allows you to 
search based on query params

The allowed search parameters are
- text 
- offset
-  limit 
- sort - The sort paramenter. Can start with '-' to sort in descentding order." Can also refer to a nested value. For example `-runs.start_date`    
-  resource_type - Valid values are ['course', 'program', 'learning_path', 'podcast','podcast_episode']
-  professional - Valid values are: ['true', 'false']
- certification 
- offered_by - Valid values are: ['mitx', 'ocw', 'bootcamps', 'xpro', 'csail', 'professional education', 'sloan executive education', 'schwarzman college of computing', 'center for transportation & logistics']

 - topic 
 -  department 
 - level 
 - resource_content_tags 
 - aggregations - Valid values are: ['resource_type', 'audience', 'certification', 'offered_by', 'platform', 'topic', 'department', 'level', 'resource_content_tags']

You can search by multiple values for each parameter except text, limit, offset and sort

ex platform=ocw,xpro

# How can this be tested?
1) Make sure you have data in both your database and your open search index. If not, run
- docker-compose run web ./manage.py backpopulate_xpro_data
- docker-compose run web ./manage.py recreate_index --all

2) Go to http://localhost:8063/api/v1/learning_resources_search/ and verify that there are search results

3) Here is the documentation for the allowed parameters http://localhost:8063/api/v1/schema/swagger-ui/#/learning_resources_search/learning_resources_search_retrieve
Play around with them and see that you have the expected results. An example query http://localhost:8063/api/v1/learning_resources_search/?resource_type=course,program&offered_by=xpro&aggregations=resource_type

The expected behavior for aggregations is to include all filters except for the aggregation parameter. For example a search where you have "offered_by" as a search filter and ["offered_by" and "resource_type"] as the aggregation the filter will affect the aggregation results for resource type but not offered by.

A known bug is that sort works for fields that are saved as keyword in the index but not those that are saved as text. So sorting by title, for example throws an error.

### DIFF
--- a/learning_resources_search/serializers.py
+++ b/learning_resources_search/serializers.py
@@ -1,7 +1,13 @@
 """Serializers for opensearch data"""
 # pylint: disable=unused-argument,too-many-lines
 import logging
+from collections import defaultdict
 
+from django.conf import settings
+from drf_spectacular.utils import extend_schema_field
+from rest_framework import serializers
+
+from learning_resources.constants import LearningResourceType, OfferedBy, PlatformType
 from learning_resources.models import Course, Program
 from learning_resources.serializers import (
     LearningResourceSerializer,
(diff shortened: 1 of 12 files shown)


abeglova:
It occurs to me that we can improve the swagger documentation by querying for the database for the available values for audience, certification, topic, etc and adding it to the documentation

learning_resources_search/serializers.py
```diff
@@ -10,10 +16,217 @@
...
+    """
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+    def to_internal_value(self, data):
+        return ",".join(data)
+
+    def to_representation(self, data):
+        return data.split(",")
```
ChristopherChudzicki, line 66: I think we should try to leverage more of Django's serializers for two reasons:
1. Let django's own serializers handle more validation
2. better openapi generation

For example, with
```
resource_type = ListSerializer(
    required=False,
    child=ChoiceSerializer(
        choices = [e.value.lower() for e in LearningResourceType]
    )
)
```
drf-spectacular will produce a list of choices for the value automatically using OpenAPI's `enum` field, rather than relying on a `help_text`.

**Note:** Using `ListSerializer` directly (rather than `StringArrayField`) means that query parameter arrays must passed with `&`: `?foo=a,b,c` becomes `?foo=a&foo=b&foo=c`. (In OpenAPI lingo, `?foo=a&foo=b&foo=c` is `explode: true`.). If we want to keep the comma-separated version, I think we could do
```
    def to_internal_value(self, data):
        return super().to_representation(data[0].split(","))
```

learning_resources_search/serializers.py
```diff
@@ -10,10 +16,217 @@
...
+    else:
+        return []
+
+
+class LearningResourcesSearchResponseSerializer(serializers.Serializer):
+    count = serializers.SerializerMethodField()
+    results = serializers.SerializerMethodField()
+    metadata = serializers.SerializerMethodField()
+
+    def get_count(self, instance):
```
ChristopherChudzicki, line 216: I believe that `drf-spectacular` is not picking up the response format at all right now.

I think the best way to inform spectacular is to decorate the methods here.
```python
def get_count(self, instance) -> int:
    pass

@extend_schema_field(LearningResourceSerializer(many=True))
def get_results(self, instance):
    pass
```

ChristopherChudzicki (changes requested):
This is working well, except possibly the `?aggregations=department` issue below.

I did leave a request/suggestion regarding serializers which I think is worth changing.

### Explode 💥
I also want to mention an issue with the OpenAPI spec that we don't necessarily need to resolve now. For array query parameters, OpenAPI has two variants:
- `explode: true` (the default) where `foo = ["a", "b", "c"]` is passed as `?foo=a&foo=b&foo=c`
- `explode: false` where `foo = ["a", "b", "c"]` is passed as `?foo=a,b,c

"Exploding" Seems also to be Django's default, in that passing `?foo=a&foo=b&foo=c` makes the QueryDict have key `"foo": ["a", "b", "c"]`.

So, question: Do we want to explode or not?

If not, we should see if we can communicate that to drf-spectacular. 

https://swagger.io/docs/specification/serialization/

### aggregation issue?


**Issue?:** When I use `?aggregations=department`, the metadata I get is:
```json
    "metadata": {
        "aggregations": {
            "department": null
        },
        "suggest": []
    }
```
That doesn't seem right?

Ferdi:
> text

`q` to keep it consistent with the current search parameters

> sort - The sort paramenter. Can start with '-' to sort in descentding order." Can also refer to a nested value. For example -runs.start_date

Do we have a `sort` params for the rest api?  does it work the same ?  @mbertrand  ?

> offered_by - Valid values are: ['mitx', 'ocw', 'bootcamps', 'xpro', 'csail', 'professional education', 'sloan executive education', 'schwarzman college of computing', 'center for transportation & logistics']

Can we use some key, like sse or sloand-exectutive-education in stead of 'sloan executive education'.  It would be easier for the clients . I would say the same thing for topics and departments. E.g.[ this is not a friendly url ](https://discussions-rc.odl.mit.edu/learn/search?o=School%20of%20Humanities%2C%20Arts%2C%20and%20Social%20Sciences)

> certification

What are the possible values ?

> You can search by multiple values for each parameter except text, limit, offset and sort
> ex platform=ocw,xpro

I had the same question as Chris. do we explode or not ?   What do we do in the rest api ?

mbertrand:
> Do we have a sort params for the rest api? does it work the same ? 

The REST API currently allows sorting on the following fields: "id", "readable_id", "last_modified", "title"

ChristopherChudzicki:
@Ferdi 
> do we explode or not ? What do we do in the rest api ?
The REST API does not support arrays for its filter parameters as far as I know. For example,
- https://mit-open-rc.odl.mit.edu/api/v1/learning_resources/?resource_type=course,program is invalid
- https://mit-open-rc.odl.mit.edu/api/v1/learning_resources/?resource_type=course&resource_type=program filters on `program` only

Ferdi:
@ChristopherChudzicki thanks !    I suggest:

* Decide on the 'explode' format we want to support and open an issue to implement it. It should not be a blocker for this PR
* Open an issue to implement the same for REST API
* Create an epic to track the API issues.  It's important we get the API right this time

abeglova:
@Ferdi So right now the api supports both exploded and not exploded parameters so both foo=a,b,c and foo=a&foo=b work correctly. Do we no longer care about supporting  foo=a,b,c ?

abeglova:
Right now the possible value for certification is just "Certification" (or null) - I think there is a plan to make it boolean similar to the professional flag but that hasn't been implemented yet

abeglova:
I'm going to restrict sort to "id", "readable_id",  "last_modified" and "start_date" for now so we don't get the error for sorting on text field. I'm going to keep the - format for sorting in descending order

abeglova:
But also we would have this issue with other fields for example multi-word departments. For example on current OCW the url for filtering for Earth Atmospheric and Planetary Sciences is https://ocw.mit.edu/search/?d=Earth%2C%20Atmospheric%2C%20and%20Planetary%20Sciences.

abeglova:
One solution is to ignore whitespace  in the filters so that we can search for department=earthatmosphericplanetarysciences which open search supports with a custom analyzer https://opensearch.org/docs/latest/analyzers/search-analyzers/

Ferdi:
> Do we have a sort params for the rest api? does it work the same ?
> The REST API currently allows sorting on the following fields: "id", "readable_id", "last_modified", "title"

Is there a good reason to sort by id  or readable_id ? @mbertrand 

> So right now the api supports both exploded and not exploded parameters so both foo=a,b,c and foo=a&foo=b work correctly.

I think that's great !

> Right now the possible value for certification is just "Certification" (or null) - I think there is a plan to make it boolean similar to the professional flag but that hasn't been implemented yet

This is not ideal. If this field is there to tell the client whether a learning resource offers a certificate or not, it should be true/false.   
We have talked about expanding this to include things like "type of certificate" etc.  but we can reflect than in the api when we do the implementation 

> I'm going to keep the - format for sorting in descending order

Ok.  @mbertrand maybe we consider doing the same in the REST api ?

> For see - it's currently saved as"sloan executive education" in the index and shows up as "sloan executive education" in the results. I assume we want it to still be "sloan executive education" in the results so we will need to implement a offered_by_aliases field in the serializer to make 'see' be a valid search value

How is the `resource_type` param is implemented ?   

esource_type - Valid values are ['course', 'program', 'learning_path', 'podcast','podcast_episode']
We should have 'learning path' in the result but param is learning_path

abeglova:
resource_type is set for all LearningResource objects. If we implement learning paths as LearningResource objects they will be added to the options automatically. But probably learning paths are different enough from learning resources that we should have a seperate data structure and search endpoint for them, like we are doing with content files.

abeglova:
Ok - I can change the param to "ordering"

abeglova:
@Ferdi so never mind - learning_path is one of the options already (although  learning paths are not in the open search index yet). I got the documentation before rebasing and it wasn't in my local instance

abeglova:
NM - I was confused by what Ferdi was saying - sound like we need learning_path to display with the space...

ChristopherChudzicki:
@Ferdi Re
> > `text`

> `q` to keep it consistent with the current search parameters

I think "q" is a fine name if that's what we want, but my preference would be not to have a bunch of single letter query parameters. Seems confusing.

"q" isn't really what the current search uses, it's what the OCW url bar displays. The current search uses a POST request with a big elasticsearch query as JSON body. IMO, "what OCW puts in the URL bar" and "api parameter names" are separate questions.

Ferdi:
> Ok - I can change the param to "ordering"

I suggest `sortby`  (imo 'sort' is more appropriate than 'order')


> I think "q" is a fine name if that's what we want, but my preference would be not to have a bunch of single letter query parameters. Seems confusing.

It's the one , main search param (q=query) and I think it's a popular option because google uses it.

> https://mit-open-rc.odl.mit.edu/api/v1/learning_resources/?ordering=id ascending (id)

ok, but with search we'd have `?ordering=-id`

Which is a different format. Ideally they'd be the same.

abeglova:
@Ferdi @Chris This is ready for re-review the updating documentation is attached.
[MIT Open Discussions Course Catalog API.pdf](https://github.com/mitodl/mit-open/files/13047475/MIT.Open.Discussions.Course.Catalog.API.pdf)

abeglova:
Changes:
1) Change the 'text' and 'sort' commands to 'sortby' and 'q'
2) Limit sortby to "id", "readable_id", "last_modified",    "runs.start_date" and (the reverse with '-' prefix)
3) Updated the serializers so that more documentation is automatically generated by swagger
4) Fix department filter

PRs for future 
1) https://github.com/mitodl/mit-open/issues/96 
2) https://github.com/mitodl/mit-open/issues/148, possibly , maybe just using + in urls is ok
3) https://github.com/mitodl/mit-open/issues/96 - the impact on the api would be that certification would be a boolean similar to professional

ChristopherChudzicki (approved):
Changes look good 👍 

Looks like the department issue I was having before was because I only had xpro courses, and they didn't have departments for some reason.

PR #128, opened 2023-10-10
Article image uploading

ChristopherChudzicki:
# What are the relevant tickets?
#123 

# Description (What does it do?)
This PR re-enables image uploading in Ckeditor, and configures our server-side HTML sanitization. Additionally, I updated Ckeditor to v40.0.0 which seems to fix #115.

# Screenshots (if appropriate):
Screenshot 2023-10-10 at 4 41 57 PM
Screenshot 2023-10-10 at 4 42 03 PM

# How can this be tested?
1. Get `CKEDITOR_ENVIRONMENT_ID`, `CKEDITOR_SECRET_KEY`, and `CKEDITOR_UPLOAD_URL` environment variables from Open's RC heroku instance. *We use a Ckeditor-hosted image upload service, with which these env vars are associated.*
2. Visit http://localhost:8063/api/v1/articles/. If you don't have any articles yet, create a few via the JSON form. *There's no UI for creation, just editing.*) Note their ids.
3. Visit http://localhost:8063/articles/ARTICLE_ID_HERE. Edit your article and try uploading an image or two. Including:
    - Setting an alt attribute *Click on the image after you've added it*
    - try both alignments (center, side)
4. Save your article. It should look the same.

### Notes
The mediaEmbed plugin (the other button in the "+" toolbar, for embedding videos and card previews of other sites) does not fully work yet. That'll be addressed in #124

PR #145, opened 2023-10-18
Articles: disable "save" while image uploading

ChristopherChudzicki:
#128 

This PR adds the CKEditor [PendingActions](https://ckeditor.com/docs/ckeditor5/latest/api/module_core_pendingactions-PendingActions.html) plugin. This plugin emits events when there are "pending actions" like an incomplete file upload.

Note the "Save" button is disabled—the image is currently uploading.

Screenshot 2023-10-18 at 7 16 27 PM

# How can this be tested?
As a logged-in user,
1. Create some articles at http://localhost:8063/api/v1/articles/ *There's no creation UI; just editing and viewing.*
2. Visit http://localhost:8063/articles/ARTICLE_ID
3. *Optional*: In your network tab, enable throttling. "fast-3g" or "slow-3g" are both good. This isn't necessary, but it makes the effect easier to see.
4. Upload an image. You can drag an image into the editor, or you can:
    - click the editor
    - click the floating "+" icon that shows up on the left hand side
5. **While the image is uploading:**
    - you will see a very thin blue progress bar near the top of the image
    - the "save" button should be disabled **This is the new behavior in this PR.**
    - When the image finishes uploading, you'll see a green check mark in its upper-right corner (briefly). 

# Additional Context
I'm not sure this is the optimal UX for this. I could see users being confused ("why did the 'Save' button flash gray when I added a new image?"). I'm open to suggestions, but we can always improve the UX later.

I did consider an approach where users can _always_ click "Save", but the button waits to make the actual API clal until the editor has no pending actions. After looking into this a bit, it seemed more complicated to implement, so unless we definitively want that behavior, I don't think it's worth the effort.

### DIFF
--- a/frontends/mit-open/src/pages/articles/ArticlesEditPage.tsx
+++ b/frontends/mit-open/src/pages/articles/ArticlesEditPage.tsx
@@ -47,6 +47,7 @@ const ArticleForm = ({
   onSaved,
 }: ArticleFormProps) => {
   const [editoryReady, setEditorReady] = useToggle(false)
+  const [busy, setBusy] = useToggle(false)
   const editArticle = useArticlePartialUpdate()
   const article = useArticleDetail(id)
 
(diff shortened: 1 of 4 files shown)


abeglova (approved):
lgtm

PR #159, opened 2023-10-24
Create favorites list for new users

### LINKED ISSUE #80: Allow users to "favorite" learning resources - backend
# User Story
As a user, I want to bookmark some learning resources as "favorites"

# Description/Context
This is a port of existing functionality in open-discussions, where any authenticated user can add/remove courses, programs, videos, podcasts, and public lists as favorites.

# Acceptance Criteria
- [ ] Via an API, each user can list/unlist any learning resource as a favorite, and view their list of favorites. 

# Plan/Design
There are several possible ways to implement this:
a) as a `FavoriteItem` model with foreign keys to `LearningResource` and `User` (basically identical to the previous open-discussions approach)
b) as a special type of `UserList` - a "Favorites" `UserList`, of which there can be only one per user and which cannot be deleted.

### LINKED ISSUE #111: Integrate pluggy into mit-open for separation of concerns
# Description/Context
Integrate pluggy (or some other alternative) to automatically trigger actions/workflows across different apps

# Plan/Design
https://github.com/mitodl/hq/discussions/1638

mbertrand:
Closes #80 
Closes #111 

- Integrates pluggy (#111)
- Uses pluggy to create a UserList with title "Favorites" for every new user (#80)
- Adds a mgmt command to add a "Favorites" list to every active user who doesn't already have one. (#80)

# How can this be tested?
- Enable keycloak as described in [this PR](https://github.com/mitodl/open-discussions/pull/4230) 
- Go to `http://od.odl.local:8063/login/ol-oidc/` and register a new user
- After you click on the verification email and are logged in, you should have a favorites list (go to http://od.odl.local:8063/api/v1/userlists/ to see it).
- Log out (http://od.odl.local:8063/logout/) then log back in.  You should still have only 1 favorites userlist.
- Run `./manage.py backpopulate_favorites_lists`, after that all your existing users (including admin user) should have a favorites list.  You can log into django admin to see them all.
- Run the above command again, there should still be only one per user.

### DIFF
--- a/authentication/pipeline/user.py
+++ b/authentication/pipeline/user.py
@@ -2,6 +2,8 @@
 
 from social_core.exceptions import AuthException
 
+from authentication.hooks import get_plugin_manager
+
 
 def forbid_hijack(
     strategy,
@@ -19,3 +21,13 @@ def forbid_hijack(
         msg = "You are hijacking another user, don't try to login again"
         raise AuthException(msg)
     return {}
+
+
+def user_created_actions(**kwargs):
+    """
+    Trigger plugins when a user is created
+    """
+    if kwargs.get("is_new"):
+        pm = get_plugin_manager()
+        hook = pm.hook
+        hook.user_created(user=kwargs["user"])
--- a/learning_resources/management/commands/backpopulate_favorites_lists.py
+++ b/learning_resources/management/commands/backpopulate_favorites_lists.py
@@ -1,0 +1,46 @@
+"""Management command to create user Favorites lists"""
+from django.contrib.auth.models import User
+from django.core.management import BaseCommand
+
+from learning_resources.constants import FAVORITES_TITLE
+from learning_resources.models import UserList
+from open_discussions.utils import now_in_utc
+
+
+class Command(BaseCommand):
+    """Create a Favorites userlist for every active user"""
+
+    help = "Create a Favorites userlist for every active user"  # noqa: A003
+
+    def add_arguments(self, parser):
+        parser.add_argument(
+            "--delete",
+            dest="delete",
+            action="store_true",
+            help="Delete all existing Favorites user lists",
+        )
+        super().add_arguments(parser)
+
+    def handle(self, *args, **options):  # noqa: ARG002
+        """Create a Favorites userlist for every active user"""
+        if options["delete"]:
+            self.stdout.write("Deleting all existing Favorites userlists")
+            UserList.objects.filter(title=FAVORITES_TITLE).delete()
+        else:
+            self.stdout.write("Creating Favorites lists for each user")
+            start = now_in_utc()
+            user_ids = UserList.objects.filter(title=FAVORITES_TITLE).values_list(
+                "author_id", flat=True
+            )
+            for user in User.objects.exclude(id__in=user_ids).exclude(is_active=False):
+                UserList.objects.get_or_create(
+                    author=user,
+                    title=FAVORITES_TITLE,
+                    defaults={"description": "My Favorites"},
+                )
+            total_seconds = (now_in_utc() - start).total_seconds()
+            self.stdout.write(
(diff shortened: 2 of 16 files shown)


authentication/pipeline/user.py
```diff
@@ -17,3 +24,15 @@ def forbid_hijack(
...
+def create_new_user_favorites_list(**kwargs):
+    """
+    Create a favorites UserList for new users.
+    """
+    if kwargs.get("is_new"):
+        pm = get_plugin_manager()
+        hook = pm.hook
+        hook.create_favorites_list(
+            user=User.objects.get(username=kwargs["details"]["username"])
+        )
```
rhysyngsun, line 38: So I think the way you want to do this a bit differently because it's unlikely there's going to be multiple ways to create a favorites list or multiple things you do in order to create it. More likely we have a discrete event (in this case a user has been created) and we want to allow arbitrary code to hook into that.

So we'd want the plugin be more abstract and also to define a spec for it:

```python
hookimpl = pluggy.HookimplMarker(APP_NAME)
hookspec = pluggy.HookspecMarker(APP_NAME)


# class is used here just to namespace the spec
class AuthenticationHooks:
    @hookspec
    def user_created(self, user: User): pass
```

Elsewhere:

```python
class FavoritesListPlugin:
    @hookimpl
    def user_created(self, user: User): 
        UserList.objects.get_or_create(
            author=user, title=FAVORITES_TITLE, defaults={"description": "My Favorites"}
        )
```

Then you'd update your pipeline step here to be a more generally named one and over time we can have other kinds of code hook into this same point.

Also, `user` should be available as a kwarg to this function so you shouldn't need to refetch.

mbertrand: Thanks, do the app/file locations make sense?  For instance, should the `hooks.py` file containing `hookimpl`, `def get_plugin_manager`, `class AuthenticationHooks` etc be located where it is in the `authentication` app, with `APP_NAME=authentication`, or in the main project app instead (still `open_discussions`, should probably be renamed at some point to `mitopen`)?

Also:
```
class FavoritesListPlugin:
    @hookimpl
    etc
```

is located in `learning_resources/plugins.py` and contains an import for hookimpl from the `authentication` app (which seems to somewhat defeat the purpose of keeping the apps as separate as possible).  

```
from authentication.hooks import hookimpl
```

Is that okay? Or should I redefine it again instead?

```
hookimpl = HookimplMarker("authentication")
```

rhysyngsun: The import should be fine

rhysyngsun: Actually, what do you think about having a convention where we import and assign `hookimpl` to the `AppConfig` and then access it like this:

```python
from django.apps import apps
hookimpl = apps.get_app_config("authentication").hookimpl
```

mbertrand: will give that a shot, thanks

mbertrand:
@rhysyngsun ready for another look

learning_resources/management/commands/backpopulate_favorites_lists.py
```diff
@@ -0,0 +1,44 @@
...
+            UserList.objects.filter(title=FAVORITES_TITLE).delete()
+        else:
+            self.stdout.write("Creating Favorites lists for each user")
+            start = now_in_utc()
+            user_ids = UserList.objects.filter(title=FAVORITES_TITLE).values_list(
+                "author_id", flat=True
+            )
+            plugin = AuthenticationHooksPlugin()
+            for user in User.objects.exclude(id__in=user_ids).exclude(is_active=False):
+                plugin.create_favorites_list(user)
```
rhysyngsun, line 38: This code is broken not because the imported class doesn't exist, and the function is `user_created` - I don't think it makes sense to call the plugin anymore and it should either just do the `get_or_create` directly - we'll only use this command once anyway.

rhysyngsun (approved):
LGTM

PR #166, opened 2023-10-25
add id filter to learning_resources_search

abeglova:
https://github.com/mitodl/mit-open/issues/164

This pr adds an id filter to the /learning_resources_search endpoint. It also updates the indexing ids for objects in open search to the learning resource object id 

# How to test
Got to http://localhost:8063/api/v1/learning_resources_search/ and verify that it loads
For ids that exist in your database try searching with the id param for example
http://localhost:8063/api/v1/learning_resources_search/?id=5147,5140
or 
http://localhost:8063/api/v1/learning_resources_search/?id=5147

This stacks with other filters too

### DIFF
--- a/learning_resources_search/serializers.py
+++ b/learning_resources_search/serializers.py
@@ -8,13 +8,9 @@ from drf_spectacular.utils import extend_schema_field
 from rest_framework import serializers
 
 from learning_resources.constants import LearningResourceType, OfferedBy, PlatformType
-from learning_resources.models import Course, Program
+from learning_resources.models import LearningResource
 from learning_resources.serializers import (
     LearningResourceSerializer,
-)
-from learning_resources_search.api import (
-    gen_course_id,
-    gen_program_id,
 )
 from learning_resources_search.constants import AGGREGATIONS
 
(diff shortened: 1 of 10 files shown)


learning_resources_search/serializers.py
```diff
@@ -116,6 +112,7 @@ class LearningResourcesSearchRequestSerializer(serializers.Serializer):
         required=False,
         child=serializers.ChoiceField(choices=AGGREGATIONS),
     )
+    id = StringArrayField(required=False, child=serializers.CharField())  # noqa: A003
```
abeglova, line 115: I'm not sure how concerned we should be about the warning that id is a python builtin in this instance. I can rename the key to 'ids' or 'learning_resource_id' and add some logic to map that to "id" in the search object

learning_resources_search/serializers_test.py
```diff
@@ -88,9 +86,7 @@ def test_serialize_bulk_courses_for_deletion():
         serializers.serialize_bulk_courses_for_deletion([course.learning_resource_id])
     ) == [
         {
-            "_id": api.gen_course_id(
-                course.learning_resource.platform, course.learning_resource.readable_id
-            ),
+            "_id": course.learning_resource.id,
```
mbertrand, line 89: One minor concern with using the database id is that if we mass delete resources then reimport them, the ids will be different, which could potentially mess up any saved query urls that include them, as opposed to a combo of platform and readable_id which should always be the same.  But that is true of querying id via REST API too anyway.

mbertrand (approved):
LGTM

PR #173, opened 2023-10-31
Refactor course numbers

### LINKED ISSUE #138: Refactor course number fields for `Course` model, serializer, search index
# Description/Context
Initially, the `LearningResource.readable_id` field was thought to be a suitable place to store the primary course number for all courses.  However, this did not take into consideration the fact that [multiple OCW courses with the same primary number would exist](https://github.com/mitodl/mit-open/issues/132).  Therefore, the `readable_id` values for OCW [had to be modified](https://github.com/mitodl/mit-open/pull/133), but the primary course number still needs to be stored somewhere for search purposes.

In addition, filtering by course number and department necessitates that the serializers and indexed documents be structured in such a way to facilitate that.

# Plan/Design

### Option A: 
Replace the `Course.extra_course_numbers` ArrayField with a `Course.course_numbers` JSONField.  The format of the data stored here could look something like this:

```json
[
       {
            "value": "8.370",  
            "listing_type": "canonical",
            "department": "Philosophy"
        }, {
            "value": "2.111",
            "listing_type": "cross-listed",
            "department": "Chemistry"
        }, {
            "value": "MITx.LASER+R01",
            "listing_type": "cross-listed",
            "department": null
        }, {
            "value": "18.435",
            "listing_type": "cross-listed",
            "department": "Social Science"
        }
]
```        
The department info above is a bit redundant because OCW course numbers begin with the department id, and the "departments" field in the API serializer will provide both the department id and name.  So it could potentially be dropped.

### Option B:

Leave the current `extra_course_numbers` field as is, and add a `primary_course_number` field.  This seems a bit simpler and perhaps more intuitive to API users if we include the two fields in the serializer as is.  The department if any could still be deduced from the course numbers.

## Serializer changes

The API `CourseSerializer` will need to be changed to reflect the model changes above.

## Search index changes

The search index will require a `department_course_numbers` field as [described here](https://github.com/mitodl/open-discussions/pull/3457) in order to search and sort by all course numbers with or without a department filter applied.  This field currently looks like this:

```json
                    "department_course_numbers": [
                        {
                            "coursenum": "21H.983J",
                            "department": "History",
                            "primary": true,
                            "sort_coursenum": "21H.983J"
                        },
                        {
                            "coursenum": "21H.109J,WGS.310J,WGS.303J",
                            "department": "History",
                            "primary": false,
                            "sort_coursenum": "21H.109J,WGS.310J,WGS.303J"
                        }
                    ],
```

This is not a field we want to include in the API serializer, nor in the response data sent back in search results, yet we need to include it in the data sent to populate the opensearch index.  

There are a couple of ways this could be done:
- Subclass the LearningResourceSerializer (`OSLearningResourceSerializer`) to add the above field when appropriate (for resources of type `course`), and use this serializer to populate the search index.
- Provide a single root function that:
  - serializes the resource using the same serializer as the APIs
  - generates an object that represents the extra properties to merge into the original.
  - does a [deepmerge](https://deepmerge.readthedocs.io/en/latest/) of those two objects

Finally, to remove this field from the search results, [an exclude pattern](https://www.elastic.co/guide/en/elasticsearch/reference/7.0/search-request-source-filtering.html#search-request-source-filtering) for the `_source` field can be added:
```json
{
    "_source": {
        "includes": [ "*" ],
        "excludes": [ "*.course.department_course_numbers" ]
    },
    "query" : {
        "term" : { "department" : "Physics" }
    }
}
```

mbertrand:
Closes #138

- Adds a `course_numbers` `JSONField` to the `Course` model, to be populated in the format:
  ```
  {
    "value": "12.01",
    "listing_type": <"Primary" or "Cross-linked">,
    "department: {
        "department_id": "12",
        "name": "Earth, Atmospheric and Planetary Sciences"
    }
  }
  ```
- Removes the `Course.extra_course_numbers` field.
- Adds migrations to update the schema and convert existing courses to have populated values for `course_numbers`
- Updates the serializers and ETL pipelines as necessary to support the above.
- adds additional fields that are needed for indexing, in this case, modifying `course.course_numbers` to include ``sorted_coursenum` and `primary`.
- Modified the search api to exclude from search results any additional fields defined above, via `SOURCE_EXCLUDED_FIELDS` constant (which will need to be updated as necessary).

# How can this be tested?
- Run `docker compose up`.  The migrations should run, and afterward, if you have any existing courses, the API should return results with each course resource having a populated `{"course": {"course_numbers"}}` section.
- Run `./manage.py recreate_index --all`
- All the ETL pipelines should still work without errors.
- Import an OCW course with extra course numbers if you don't already have any:
  ```
  ./manage.py backpopulate_ocw_data --course-name 4-401-environmental-technologies-in-buildings-fall-2018
  ```
- Search for that course by both the primary and extra course number, it should be returned in both searches:
  - http://localhost:8063/api/v1/learning_resources_search/?q=4.401
  - http://localhost:8063/api/v1/learning_resources_search/?q=4.464
- You should not see "sorted_coursenum" or "primary" fields in the `course.course_numbers` response
- The primary value for this course should be "4.401" and not "4.401+fall_2018" like the `readable_id`
-  Modify [this code](https://github.com/mitodl/mit-open/pull/173/files#diff-fc36e9091d5012053128f4d7ec939c6a644f38629fb4afa9d1f47f7a7a8f770cR342-R344) so that `SOURCE_EXCLUDED_FIELDS=[]`
- Try the above queries again, results should be the same but now `sorted_coursenum` and `primary` fields should be shown.


learning_resources_search/serializers.py
```diff
@@ -311,3 +330,25 @@ def serialize_for_deletion(opensearch_object_id):
...
+        "course.course_numbers.primary",
+    ]
+
+    def to_representation(self, instance):
+        """
+        Override to_representation to modify/add fields needed only for indexing
+        """
+        data = super().to_representation(instance)
+        if instance.resource_type == LearningResourceType.course.name:
+            add_extra_course_number_fields(data["course"]["course_numbers"])
```
mbertrand, line 353: The document structure in opensearch will look like this, (with `primary` and `sorted_coursenum` excluded from hit results output):

```json
            "course": {
                "course_numbers": [
                    {
                        "value": "4.401",
                        "department": {
                            "name": "Architecture",
                            "department_id": "4"
                        },
                        "listing_type": "primary",
                        "primary": true,
                        "sort_coursenum": "04.401"
                    },
                    {
                        "value": "4.464",
                        "department": {
                            "name": "Architecture",
                            "department_id": "4"
                        },
                        "listing_type": "cross-listed",
                        "primary": false,
                        "sort_coursenum": "04.464"
                    }
                ]
            },
```
@abeglova  is this a suitable replacement for `department_course_numbers`?

abeglova (approved):

PR #174, opened 2023-11-01
content file search

abeglova:
closes https://github.com/mitodl/mit-open/issues/142

This pr loads content file data into the open search index and creates a new import for content file search at 
http://localhost:8063/api/v1/content_file_search

The available parameters are
- aggregations array[string] | Available values : topic, file_content_type
- file_content_type: Currently this is always null in the index because only ocw files have this field. In the future it will contain the file content type such as "Assignment" or "Lecture notes"
- id array[string]
- limit integer
- offset integer
- q string - the query string
- sortby string | Available values : id, -id, resource_readable_id, -resource_readable_id
-  topic array[string]

Since we can now have data from the same source with different platforms, I also updated the update_index command to have an etl_source flag instead of a platform flag.

# How can this be tested?
- Recreate the index `docker-compose run web ./manage.py recreate_index --all`
- Verify that management commands that import file data still work, for example `docker-compose run web ./manage.py
backpopulate_xpro_files`
- Verify that the update index command works `docker-compose run web ./manage.py update_index --all`. Also try using the course_etl_source flag. For example `docker-compose run web ./manage.py update_index --content_files --course_etl_source xpro`
- Go to  http://localhost:8063/api/v1/content_file_search/ and play around with the search and check that you get expected results
- Go to  http://localhost:8063/api/v1/learning_resources_search/ and play around with the search and check that you get expected results


learning_resources/serializers.py
```diff
@@ -479,11 +478,11 @@ class ContentFileSerializer(serializers.ModelSerializer):
     resource_readable_num = serializers.CharField(
         source="run.learning_resource.resource_num"
     )
-    resource_type = serializers.SerializerMethodField()
+    file_content_type = serializers.SerializerMethodField()
```
abeglova, line 481: This refers to the "Resource Type" in the sidebar of the ocw search
https://ocw.mit.edu/search/?type=resourcefile

Examples are "Lecture Notes" , "Assignments. To avoid confusion with the resource type form Learning Resources ("course, "program") i renamed this to file_content_type. But maybe that's not great since we also have content_type and file_type fields. Any other name suggestions?

learning_resources_search/serializers.py
```diff
@@ -311,3 +386,26 @@ def serialize_for_deletion(opensearch_object_id):
...
+def serialize_content_file_for_bulk(content_file_obj):
+    """
+    Serialize a content file for bulk API request
+
+    Args:
+        content_file_obj (ContentFile): A content file for a course
+    """
+    return {
+        "_id": gen_content_file_id(content_file_obj.key),
+        **serialize_content_file_for_update(content_file_obj),
```
mbertrand, line 400: Some `content` values for ContentFiles are so large that they might [cause errors when trying to index](https://github.com/mitodl/open-discussions/issues/3881) - might want to truncate them as done in https://github.com/mitodl/open-discussions/pull/3917.  

Maybe we shouldn't show the entire `content` field in the API output either to avoid problems when it is excessively large? Or exclude it from the output altogether?

learning_resources_search/api.py
```diff
@@ -1,23 +1,41 @@
...
+    """
+    Generate the OpenSearch document id for a ContentFile
+
+    Args:
+        key (str): The key of a ContentFile object
+
+    Returns:
+        str: The OpenSearch document id for this object
+    """
+    safe_key = urlsafe_b64encode(key.encode("utf-8")).decode("utf-8").rstrip("=")
```
mbertrand, line 35: Noticed some unique constraint errors when running `backpopulate_mitxonline_files` (had also run `backpopulate_xpro_files` previously but wasn't paying attention during that one).  Not sure why, didn't investigate further yet, but could be that different courses have contentfiles with the same key.  Maybe can use the `ContentFile.id` value instead?

mbertrand (commented):
Some initial comments, will keep reviewing tomorrow.

learning_resources_search/serializers.py
mbertrand: PS I dont have any contentfiles over 10 MB locally yet, the biggest is 4.5 MB from OCW course https://live-qa.ocw.mit.edu/courses/hst-508-quantitative-genomics-fall-2005/

http://localhost:8063/api/v1/content_file_search/?q=%22CDNA-Pos%20%20Description%20%20%20Allele1%22 will return it if imported.

abeglova: So that actually looks fine in the ui since the default api view does not wrap lines

abeglova: Screenshot 2023-11-03 at 9 31 05 AM

mbertrand (changes requested):
One thing I noticed is that content files are still showing up after I remove the parent course:

- Import OCW course:
  ``` ./manage.py backpopulate_ocw_data --course-name hst-508-quantitative-genomics-fall-2005
  ```
- Wait a minute
- Get 1 result at both these urls, as expected:
  - http://localhost:8063/api/v1/content_file_search/?q=%22CDNA-Pos%20%20Description%20%20%20Allele1%22
  - http://localhost:8063/api/v1/learning_resources_search/?q=%22CDNA-Pos%20%20Description%20%20%20Allele1%22
- Remove the course
  ```./manage.py backpopulate_ocw_data --course-name hst-508-quantitative-genomics-fall-2005
  ```
- Wait a minute.
- Hit the two urls above again.  Expect no results for both, but the content file search url still returns 1 result.  When a course is deindexed, all the content files for it should be deindexed too.

learning_resources/serializers.py
abeglova: resource_category sounds great

abeglova: Or actually content_category is better

mbertrand (approved):
Works great! 👍

PR #152, opened 2023-10-19
Update node and remove docker-compose.override.yml

ChristopherChudzicki:
# What are the relevant tickets?
closes #136 
closes #141 

# Description (What does it do?)
This PR removes docker-compose.override.yml and merges it into `docker-compose.yml` and docker-compose-notebook.yml.

# Screenshots (if appropriate):

- [ ] Desktop screenshots
- [ ] Mobile width screenshots

# How can this be tested?
1. docker compose up; things should work as normal.
2. Try the notebook: *I wanted to double check this was still working after changes. Actually, it wasn't working before my changes, but now it is.*
    - `cp app.ipynb.example app.ipynb`
    -  `docker compose -f ./docker-compose-notebook.yml up` and open browser to url indicated in logs
    - open app.ipynb and run the top cell
    - run some python

# Additional Context
Based on https://github.com/mitodl/micromasters/pull/3644, I think the docker-compose.override.yml was originally intended to help share settings between local dev docker and travis docker. We don't use travis docker anymore, so it's unnecessary and confusing.

Issue #136, opened 2023-10-13
Do not commit `docker-compose.overrides.yml`


ChristopherChudzicki:
# Description/Context
We should stop committing `docker-compose.override.yml`:
- it's confusing. We set some env variables in `docker-compose.yml`, then we override them in `docker-compose.override.yml`. (example: `DEBUG`)
- it's less flexible. Occasionally we may need to override things but not want that change committed. And example is in OCW Studio where ARM mac users need to override the concourse container, but we don't want that override committed. _In Studio we **do** commit a template override file for ARM users, but not the actual override file._

# Plan/Design
Merge override into `docker-compose.yml`.

Additionally, we may want to set more env variables in an overridable way. See https://github.com/mitodl/mit-open/pull/135#pullrequestreview-1676747840

PR #176, opened 2023-11-01
fix watch container default NODE_ENV

ChristopherChudzicki:
- This fixes a bug introduced in #152 for #136 

Sets `development` as the default value of `NODE_ENV` in `watch` container.

# How can this be tested?
Remove `NODE_ENV` from your `.env` file if it there.

1. Run `docker compose up` on main branch. Visit http://localhost:8063. Nothing will load.
2. Run `docker compose up` on this branch. Visit http://localhost:8063. Frontend should load.

# Additional Context
In #152 when I merged the docker-compose.overrides.yml into docker-compose.yml, I made an error in not including NODE_ENV in the `watch` container. My best guess is that we missed it because me/reviewer had `NODE_ENV` set in their env file.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -94,6 +94,7 @@ services:
       - "8062:8062"
     environment:
       DOCKER_HOST: ${DOCKER_HOST:-missing}
+      NODE_ENV: ${NODE_ENV:-development}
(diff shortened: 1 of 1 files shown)


jonkafton (approved):
Tested locally and confirm fixed.

Issue #161, opened 2023-10-24
Collect UI components in Design Package


ChristopherChudzicki:
# Description/Context
I'd like to propose adding a new local package to our frontend structure. The new package, named something like `ol-design` would contain building block UI components ranging from `Button` to `Carousel` and `Drawer`.

We are already using a component library—`@mui/material`—that contains a variety of customizable components (Button, Link, Dialog, TextInput, CheckBox, ...). In `ol-design`, we would define our own wrapper around Button, Dialog, etc, and re-export them, then use these variants exclusively within the app (rather than importing from `@mui/material`). Creating our own wrappers has a few advantages:
1. `@mui/material` is **too** customizable, leading to inconsistent styling/behavior throughout pages. By defining our own wrappers, we can expose the customizability we want, but enforce consistency in other options.
2. Higher-level variants: Our app won't implement all the use-cases that `@mui/material` components support, so we can build higher level, easier-to-use variants. (E.g., simpler to user dropdowns or dialogs)
3. If we decide to swap `@mui/material` for another component library, we have fewer places in our own code that needs to change.

Additionally, `ol-design` could be the place where we put reusable, content-specific items like the `LearningResourceCard` and a `LearningResourceCarousel`.

PR #177, opened 2023-11-01
Collect ui components to ol-design package

ChristopherChudzicki:
#161 

See #161. This PR moves many presentation components to a new `ol-design` package and removes the `ol-forms` package. I removed some components that were completely unused or that can be replaced with builtins. 

Also, no package directly imports from `@mui/material`. Instead, `ol-design` re-exports (some of) `@mui/material. The rational for this was:

> 1. `@mui/material` is **too** customizable, leading to inconsistent styling/behavior throughout pages. By defining our own wrappers, we can expose the customizability we want, but enforce consistency in other options.
> 2. If we decide to swap `@mui/material` for another component library, we have fewer places in our own code that needs to change.

The first goal ended up not really being accomplished in this PR: I decided to directly re-export many components (e.g., Button; TextField, many others) directly from MUI. Thus, their interfaces aren't really constrained. Re-exporting had the advantage of preserving MUI's nice docstrings and also means that ref forwarding works naturally without needing to wrap components in `forwardRef`. 

# How can this be tested?
Besides CI tests passing, I would check:
1. Homepage looks same. *There is a slight difference in the chips below searchbar. But I think this difference is fine for now since we don't have final designs.*
2. Check http://localhost:8063/learningpaths/ *Need to be logged in as admin to create new paths.*
    - creating / editing paths should work
    - Check the "..." menu for editing paths. The "..." menu is new.
    -  Check the Home button new 404 page (e.g., http://localhost:8063/learningpaths/fake_id) which uses new ButtonLink component


frontends/ol-design/src/components/FormDialog.tsx
```diff
@@ -121,7 +128,6 @@ const FormDialog: React.FC<FormDialogProps> = ({
     const props: DialogProps["PaperProps"] = {
       component: "form",
       ...className,
-      // @ts-expect-error There seems to be an error with MUI's type defs:
```
ChristopherChudzicki, line 124: mui upgrade made this error go away

frontends/ol-design/src/components/SimpleMenu.tsx
ChristopherChudzicki, line 1: This is one of a few brand-new components. Most other changes are just moves.

frontends/ol-forms/src/components/ValidationError.test.tsx
ChristopherChudzicki, line 1: This was completely unused. We've been using MUI's `<Alert />` more recently.

frontends/ol-design/src/components/SimpleMenu.tsx
ChristopherChudzicki, line 1: This is one of a small number of brand new components, along with ChipLink and ButtonLink

frontends/ol-util/src/hooks/index.ts
```diff
@@ -1,3 +1,2 @@
 export { default as useSearchParams } from "./useSearchParams"
-export { default as useId } from "./useId"
```
ChristopherChudzicki, line 2: `useId` is builtin to react 18; we had back-ported it, but now that we're on React 18, I removed it.

mbertrand (approved):
Looks good to me and frontend is still functonal.  Can you try rebasing against main before merging?  I rebased locally to bring in some database schema changes and had to resolve a conflict with yarn.lock, which github is not warning about here.

Issue #180, opened 2023-11-02
flaky python test


ChristopherChudzicki:
## Expected Behavior
Tests pass on CI.

## Current Behavior
Sometimes they don't. For example:
- https://github.com/mitodl/mit-open/actions/runs/6735574072/job/18309042705
- https://github.com/mitodl/mit-open/actions/runs/6724993035/job/18278452841
- https://github.com/mitodl/mit-open/actions/runs/6724687952/job/18277481023
are all flaky failures of `test_learning_resource_filter_offered_by`.

## Steps to Reproduce
1.  change `test_learning_resource_filter_offered_by` to
    ```python
    @pytest.mark.parametrize('execution_number', range(100))
    def test_learning_resource_filter_offered_by(mock_courses, execution_number):
    ```
2. Run `pytest -s --no-cov learning_resources/filters_test.py -k test_learning_resource_filter_offered_by`
3. Observe a ~25% failure rate.

ChristopherChudzicki:
fixed in https://github.com/mitodl/mit-open/pull/173/files#diff-4d1ca0172a3b2e3be122db95147c48221f1027a510850fdcc6018a89d894fd9bR125-L128

PR #181, opened 2023-11-02
Fix a flaky python test

ChristopherChudzicki:
#180 

This PR changes LearningResourceOfferorFactory so that its `name` and `code` are consistent. See code comment for more.

# How can this be tested?
1.  change `test_learning_resource_filter_offered_by` to
    ```python
    @pytest.mark.parametrize('execution_number', range(100))
    def test_learning_resource_filter_offered_by(mock_courses, execution_number):
    ```
2. Run `pytest -s --no-cov learning_resources/filters_test.py -k test_learning_resource_filter_offered_by`. All tests should pass. *On `main`, there's a ~25% failure rate.*

### DIFF
--- a/learning_resources/factories.py
+++ b/learning_resources/factories.py
@@ -123,7 +123,7 @@ class LearningResourceDepartmentFactory(DjangoModelFactory):
 class LearningResourceOfferorFactory(DjangoModelFactory):
     """Factory for LearningResourceOfferor"""
 
-    name = FuzzyChoice([offeror.value for offeror in constants.OfferedBy])
+    name = factory.LazyAttribute(lambda obj: constants.OfferedBy[obj.code].value)
     code = FuzzyChoice([offeror.name for offeror in constants.OfferedBy])
     professional = Faker("boolean")
 
--- a/learning_resources/etl/prolearn_test.py
+++ b/learning_resources/etl/prolearn_test.py
@@ -27,15 +27,11 @@ pytestmark = pytest.mark.django_db
 @pytest.fixture(autouse=True)
 def _mock_offerors_platforms():
     """Make sure necessary platforms and offerors exist"""
-    LearningResourceOfferorFactory.create(name="CSAIL", code="csail", professional=True)
+    LearningResourceOfferorFactory.create(code="csail", professional=True)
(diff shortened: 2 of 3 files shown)


learning_resources/factories.py
```diff
@@ -123,7 +123,7 @@ class Meta:
 class LearningResourceOfferorFactory(DjangoModelFactory):
     """Factory for LearningResourceOfferor"""
 
-    name = FuzzyChoice([offeror.value for offeror in constants.OfferedBy])
+    name = factory.LazyAttribute(lambda obj: constants.OfferedBy[obj.code].value)
     code = FuzzyChoice([offeror.name for offeror in constants.OfferedBy])
     professional = Faker("boolean")
```
ChristopherChudzicki, line 128: @mbertrand below is set
```
 django_get_or_create = ("name",)
```
which I think is because `LearningResourceOfferor.name` is unique. Should `code` also be unique, and should we change `django_get_or_create` to use `code` ?

ChristopherChudzicki: I'm pretty certain the actual cause of the errors was that `filters_test.mock_courses` essentially ran
```
LearningResourceOfferor.create(code="ocw") # used a random name
LearningResourceOfferor.create(code="mitx") # used a random name
```
and if the name happened to already exist in the test database, then "code" would be ignored, because 
> When django_get_or_create is used, be aware that any new values passed to the Factory are not used to update an existing model.
>
> [factoryboy docs](https://factoryboy.readthedocs.io/en/stable/orms.html#factory.django.DjangoOptions.django_get_or_create)

mbertrand: @ChristopherChudzicki there are some changes related to this in [PR 173](https://github.com/mitodl/mit-open/pull/173/files#diff-4d1ca0172a3b2e3be122db95147c48221f1027a510850fdcc6018a89d894fd9bR125-L128)

ChristopherChudzicki: @mbertrand Ah, cool. Happy to leave it to that for fixing.

What do you think about `code` being unique?

mbertrand: Sounds good to me

Issue #103, opened 2023-09-27
Refactor field channels to work with new LearningPath model


mbertrand:
# Description/Context
The `channels_fields` app needs to be refactored to work with the `learning_resources.LearningPath` model instead of `course_catalog.UserList`

# Plan/Design
Start by replacing all references to `course_catalog.UserList` with `learning_resources.LearningPath` in the `field_ and fix whatever breaks from there.

ChristopherChudzicki:
Closed by #196

PR #196, opened 2023-11-07
Use new LearningPath model for Fields

ChristopherChudzicki:
#103 

This PR updates the old FieldChannel model we used in Infinite to use new LearningPath model. There are some additional changes to the serializers that I've noted in the code comments.

# How can this be tested?
1. Review the OpenAPI docs at http://localhost:8063/api/v1/schema/redoc/#tag/fields
    - In particular, note that subfields, lists, and featured_list are typed differently (and properly) and the request/response sections for create/update.
2. Create some learning paths, either via the ui at http://localhost:8063/learningpaths/ or at http://localhost:8063/api/v1learningpaths/. *Must be logged in as an admin.*
3. **Create Fields:** Create some fields at http://localhost:8063/api/v1/fields/
   - You can see some associated docs at http://localhost:8063/api/v1/schema/redoc/#tag/fields/operation/fields_create
   - Check that using non-learning paths for `featured_list` and `lists` gives reasonable errors.
4. **View the field details and Update:** http://localhost:8063/api/v1/fields/<field_name>. In particular, try:
    - assigning re-ordering the lists within a field
    - assigning and re-ordering subfields within a field 

# Additional Context

The associated frontend code from Infinite Corridor has not been updated.


.devcontainer/web/devcontainer.json
ChristopherChudzicki, line 1: In VS Code, this file triggers a popup like the one below. (Can also be re-opened in container via a command.)  Then the IDE can pick up a correct python environment.

Screenshot 2023-11-09 at 7 36 47 AM

I'm sure this could be set up better.  (Example: The terminal windows in this container aren't aware of our other containers, so things like `./manage.py shell` don't work.). But I've found it useful and I lean toward committing it.

If we don't want to commit it, I can just keep a local copy.

channels_fields/serializers.py
```diff
@@ -102,14 +102,14 @@ class FieldChannelSerializer(ChannelAppearanceMixin, serializers.ModelSerializer
     """Serializer for FieldChannel"""
 
     lists = serializers.SerializerMethodField()
-    featured_list = UserListSerializer(many=False, read_only=True)
+    featured_list = LearningResourceSerializer(many=False, read_only=True)
```
ChristopherChudzicki, line 105: For the field responses, we return entire LearningResource objects.

@mbertrand What do you think of returning just the IDs?
- **Pros:** No N+1 queries to optimize, simpler implementation, easier to get docs accurate. Also makes some caching issues on frontend easier.
- **Cons:** Less detail immediately available in the API response. *IMO, this isn't a big deal. If you're building a UI around `featured_list`, you probably want the list items, too. And the `/learningpaths/<id>` and `/learningpaths/<id>/resources` calls can be made in parallel.

If this change seems reasonable, I could make it in this PR or in a followup.

mbertrand: I think it would be okay. The channels_fields API exists primarily to deliver data to/from the front end anyway, so whatever makes things easier there is fine by me.  

Not particularly important, but I wonder if the app name `channel_fields` should be changed to just `channels` now, since we have gotten rid of the reddit-backed discussion channels.

mbertrand: (or changed to some other name entirely)

ChristopherChudzicki: > Not particularly important, but I wonder if the app name channel_fields should be changed to just channels now, since we have gotten rid of the reddit-backed discussion channels.
>
> (or changed to some other name entirely)

How about just "fields"?

I was thinking we should also (1) move the BaseChannel model to the `channels_fields` app, and (2) delete the `discussions` app. How about a followup PR to make all three of those changes?

channels_fields/serializers.py
```diff
@@ -280,23 +257,16 @@ def create(self, validated_data):
             self.upsert_subfields(field_channel, validated_data)
             return field_channel
 
+    def to_representation(self, data):
+        return FieldChannelSerializer(context=self.context).to_representation(data)
```
ChristopherChudzicki, line 261: My goal here was to make sure that the create/update responses were the same same as the GET responses. This is accurately reflected in our OpenAPI schema with some minor manual decoration.

channels_fields/serializers.py
```diff
@@ -35,6 +34,14 @@ def to_internal_value(self, data):
         return data
 
 
+class LearningPathPreviewSerializer(serializers.ModelSerializer):
+    """Serializer for a minimal preview of Learning Paths"""
+
+    class Meta:
+        model = LearningResource
+        fields = ("title", "url", "id")
```
ChristopherChudzicki, line 42: In our past work on field channels, attached fully serialized learning resource objects to the field, but only ever used   `title`. In practice, the attached resources are Learning Paths, which need an extra API call anyway to get the path items.

This mini serializer helps avoid N+1 queries *and* provides the information we were using before. Since it's a subset of the full learning resource, we could always replace it with the full resources later as a non-breaking change to the API. *Though I'm not sure we will want the field APIs to be public right away anyway. They seem a bit experimental.*

channels_fields/serializers.py
mbertrand: 👍  on a follow-up PR and moving the base channels model to the same app.  "fields" seems maybe a little too short/generic for an app name, makes me think of serializer/database fields, maybe Ferdi can weigh in on that.

channels_fields/serializers.py
```diff
@@ -139,17 +149,39 @@ class Meta:
...
+        allow_null=True,
+        allow_empty=True,
+        required=False,
+        queryset=LearningResource.objects.filter(
+            published=True,
+            resource_type=LearningResourceType.learning_path.name,
+        ),
+        help_text="Learng paths in this field.",
+    )
+    subfields = serializers.SlugRelatedField(
```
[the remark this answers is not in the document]
mbertrand: The initial idea was that we'd have a "Physics" field, and then under that, "Quantum Physics" as a subfield for example.  But I think that fell out of favor.

.devcontainer/web/devcontainer.json
```diff
@@ -0,0 +1,33 @@
+// This file describes a development container, e.g., for use in an IDE.
```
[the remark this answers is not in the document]
ChristopherChudzicki: Good question. Based on the number of open issues I've found relating to personal overrides, I don't think there's a great way to address that with dev containers right now:
- https://github.com/microsoft/vscode-remote-release/issues/3279
- https://github.com/microsoft/vscode-remote-release/issues/3218
- https://github.com/microsoft/vscode-remote-release/issues/7650
- https://github.com/devcontainers/spec/issues/305

I think the main thing people would want to override is extensions, as you mention. For people using VS Code, the next best thing to git-ignored personal extensions for dev containers is: 
- there is a UserSetting for `Dev Containers: Default Extensions`. I've moved the Github Copilot extension there.
    - This isn't perfect, because it applies to all dev containers, and that's not appropriate for all extensions. But it is for copilot, so, cool.
- You can install extensions in the container after it's built, but then they're rather more temporary. 
 
I would mildly prefer to commit this file, in case it's useful to someone else, though I'm also happy to have it git-ignored.

.devcontainer/web/devcontainer.json
```diff
@@ -0,0 +1,33 @@
...
+  "name": "Existing Dockerfile",
+  "build": {
+    // Sets the run context to one level up instead of the .devcontainer folder.
+    "context": "../..",
+    // Update the 'dockerFile' property if you aren't using the standard 'Dockerfile' filename.
+    "dockerfile": "../../Dockerfile"
+  },
+  "customizations": {
+    "vscode": {
+      "extensions": ["ms-python.python", "GitHub.copilot"]
```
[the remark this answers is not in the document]
ChristopherChudzicki: I think that the copilot status icon would just have an line through it to indicate it is disabled, and you might get a brief "toast" notification in one of the corners.

But I decided to remove the copilot extension here. Copilot seems like a good thing to put in UserSetting for `Dev Containers: Default Extensions`.

.devcontainer/web/devcontainer.json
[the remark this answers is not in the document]
mbertrand: I don't think it will hurt, seems like worst that might happen is someone adding extra extensions/config stuff that might not be applicable to everyone and will probably just be ignored with maybe some error messages.

channels_fields/serializers.py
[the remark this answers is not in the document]
mbertrand: It is okay to keep as is for now, subfields can be removed later if desired.

.devcontainer/web/devcontainer.json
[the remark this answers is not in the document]
mbertrand: Though on second thought if someone makes many personal customizations to it, they may have to deal with conflicts and rebases to this file quite frequently.   But I'm okay with adding it for now, it can be removed later if it does cause problems.

mbertrand (approved):
👍

PR #198, opened 2023-11-08
191 back end logout flow

collinpreston:
https://github.com/mitodl/mit-open/issues/191

This PR solves the issue described in this RFC: https://docs.google.com/document/d/17tJ-C2EwWoSpJWZKjuhMVgsqGtyPH0IN9KakXvSKU0M/edit#heading=h.96o2gm3cfv2 

This also enables the future implementation of the flow here: https://docs.google.com/document/d/17tJ-C2EwWoSpJWZKjuhMVgsqGtyPH0IN9KakXvSKU0M/edit#heading=h.qr1x27po61aq

# How can this be tested?
These instructions assume that you have a local instance of Keycloak and MIT Open running and connected via an Open ID client in Keycloak.

Configure the MIT Open client in Keycloak:

1. The client can match the existing `ol-open-discussions-local` client found in the olapps realm in QA.
2. This client must have `Backchannel logout URL` set to `http://open.ol.local:8063/backend-logout/`

1. Login to MIT Open via http://open.ol.local:8063/login/ol-oidc
3. You should now be logged into Keycloak (verify by visiting http://localhost:8080/admin/master/console/)
4. You should now be logged into MIT Open (verify by visiting http://open.ol.local:8063/admin)
5. In a separate browser or incognito window, perform step 1, 2, and 3 again.
6. In one of the browser windows, navigate to http://open.ol.local:8063/logout.
7. You should now be logged out of Keycloak in only one of the browsers (verify by visiting http://localhost:8080/admin/master/console/)
8. Perform step 6 in the browser not originally used to complete step 6.
9. You should now be logged out of MIT Open both browsers (verify by visiting http://open.ol.local:8063/admin)


README.md
```diff
@@ -35,6 +35,48 @@ The following settings must be configured before running the app:
...
+
+- `ACCESS_TOKEN_URL`
+
+  Provider endpoint where client exchanges the authorization code for tokens.
+
+- `USERINFO_URL`
+
+  Provder endpoint where client sends requests for identity claims.
+
+- `FEATURE_KEYCLOAK_ENABLED`
```
mbertrand, line 68: This doesn't seem to be used anymore.  Keycloak authentication is enabled if I leave this out altogether or set it to `False`.  Remove it here and from `app.json`?

open_discussions/settings.py
```diff
@@ -123,11 +123,15 @@
...
+# https://django-user-sessions.readthedocs.io/en/stable/installation.html#system-check-framework
+SILENCED_SYSTEM_CHECKS = ["admin.E410"]
+
 MIDDLEWARE = (
     "django.middleware.security.SecurityMiddleware",
-    "django.contrib.sessions.middleware.SessionMiddleware",
+    "user_sessions.middleware.SessionMiddleware",
     "django.middleware.common.CommonMiddleware",
-    "django.middleware.csrf.CsrfViewMiddleware",
+    # "django.middleware.csrf.CsrfViewMiddleware",
```
mbertrand, line 134: Uncomment this line

pyproject.toml
```diff
@@ -98,6 +98,7 @@ xbundle = "^0.3.1"
 social-auth-core = {extras = ["openidconnect"], version = "^4.4.2"}
 nh3 = "^0.2.14"
 retry2 = "^0.9.5"
+django-user-sessions = "^2.0.0"
```
mbertrand, line 101: Was this primarily for debugging purposes?  I experimented with removing this and using the usual `MIDDLEWARE` and `SESSION_ENGINE` settings, everything still seemed to work correctly.  I'm a little paranoid about changing those values if not necessary.  If it is primarily for debugging, maybe the `MIDDLEWARE` and `SESSION_ENGINE` values could be updated conditionally depending on whether or not a `KEYCLOAK_DEBUG` setting = True?

mbertrand (changes requested):

pyproject.toml
mbertrand: nvm, it works for 1 browser but logs me out of both if I'm signed into two different ones

mbertrand: Actually something still seems off after reverting back to the usual PR:

Flow A): Log into browser 1.  Log into browser 2.  Log out of browser 2.  Correctly logged out of keycloak and mitopen in browser 2 while still logged into both in browser 1.

Flow B): Log into browser 1.  Log into browser 2.  Log out of browser 1.  Incorrectly logged out of keycloak in browser 2 while still logged into mitopen; still logged into browser 1 keycloak url while logged out of mitopen.

collinpreston: This is being used to end the Django user's session after receiving a POST request from Keycloak.  Django does not provide a way of ending a user's session without that user making a request as the Django session is normally tied to the request.

mbertrand (commented):
Something seems off while testing this w/ the same user logging in/out from 2 different browsers, against both the RC instance of keycloak and a local instance.  Maybe this is just a consequence of testing against a local docker instance of mitopen, I'm not sure:

1. Log in as user1 on browser A.  Keycloak admin shows one session for user1
2. Log in as user1 on browser B.  Keycloak admin shows two sessions for user1
3. Log out from browser A.  Keycloak admin shows that the session created from browser B is removed, the session created from browser A still exists
4. Log out from browser B.  Keycloak admin is still showing the session created from browser A still exists.
5. Log back in from browser A.  Immediately redirected back to mitopen without the keycloak login screen, likely because the session still exists on keycloak - _this does not seem right.  Would expect to have to log in to keycloak again_,
6. Log back in from browser B.  Need to login on keycloak again, as expected.

I'm okay with approving this and seeing how things go on RC if you think it's a side effect of running mitopen locally in a docker container and not fully communicating with local/RC keycloak as it would in a live setting.

PR #200, opened 2023-11-09
only run black via pre-commit

ChristopherChudzicki:
None

This PR removes `black` from our python dependencies and github actions ci. Instead, we will only run black via `pre-commit` and will run it on CI via the pre-commit CI action.

This solves an issue you can see in #171:
- `black` got updated in poetry.lock
- but NOT in pre-commit.yaml

The two versions have slightly different preferences, so we have two CI checks, exactly one of which will pass.

# How can this be tested?
I guess check that pre-commit CI is running black. You can see it is here: https://results.pre-commit.ci/run/github/672068771/1699568416.AGsRY5f1QhOyeYOB-Cscog


blarghmatey:
This might be a good opportunity to switch to using the Ruff formatter through pre-commit? https://docs.astral.sh/ruff/formatter/

It's just a matter of adding `  - id: ruff-format` to the `hooks` list under the ruff pre-commit block. Likely just prior to the `ruff` id so that formatting is applied prior to linting.

PR #201, opened 2023-11-09
new content file search filters

abeglova:
Part of https://github.com/mitodl/mit-open/issues/199

This pr adds
run_id
resource_id
platform
offered_by

As search filters for the content file search

# How to test
Recreate the index with
docker-compose run web ./manage.py recreate_index --all

Go to 
http://localhost:8063/api/v1/learning_resources_search/ 
and try out the new filters

http://localhost:8063/api/v1/content_file_search/?run_id=1017,1018
http://localhost:8063/api/v1/content_file_search/?resource_id=5383
http://localhost:8063/api/v1/content_file_search/?platform=xpro
http://localhost:8063/api/v1/content_file_search/?offered_by=ocw

### DIFF
--- a/learning_resources_search/serializers.py
+++ b/learning_resources_search/serializers.py
@@ -123,10 +123,9 @@ class StringArrayField(serializers.ListField):
         super().__init__(*args, **kwargs)
 
     def to_internal_value(self, data):
-        return ",".join(data)
+        normalized = ",".join(data).split(",")
 
-    def to_representation(self, data):
-        return data.split(",")
+        return super().to_internal_value(normalized)
 
 
 CONTENT_FILE_SORTBY_OPTIONS = [
@@ -148,14 +147,25 @@ LEARNING_RESOURCE_AGGREGATIONS = [
     "professional",
 ]
 
-CONTENT_FILE_AGGREGATIONS = ["topic", "content_category"]
+CONTENT_FILE_AGGREGATIONS = ["topic", "content_category", "platform", "offered_by"]
 
 
 class SearchRequestSerializer(serializers.Serializer):
     q = serializers.CharField(required=False, help_text="The search text")
     offset = serializers.IntegerField(required=False)
     limit = serializers.IntegerField(required=False)
-    id = StringArrayField(required=False, child=serializers.CharField())  # noqa: A003
+    id = StringArrayField(  # noqa: A003
+        required=False, child=serializers.IntegerField()
+    )
+    offered_by = StringArrayField(
+        required=False,
+        child=serializers.ChoiceField(choices=[e.value.lower() for e in OfferedBy]),
+    )
+    platform = StringArrayField(
+        required=False,
+        child=serializers.ChoiceField(choices=[e.value.lower() for e in PlatformType]),
+    )
+    topic = StringArrayField(required=False, child=serializers.CharField())
 
 
 class LearningResourcesSearchRequestSerializer(SearchRequestSerializer):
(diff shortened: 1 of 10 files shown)


abeglova:
I added filters for just the database ids for now since the readable ids are not very readable for example
run_readable_id=course-v1:xPRO+LASERxE4+R11
and 
resource_readable_id =course-v1:xPRO+LASERxE4

That involves a lot of hard to remember dashes, columns, dashes and x's

Probably we should do a seperate PR to standardize all the readable ids across the different sources to something that's actually readable and easy to remember

learning_resources_search/serializers.py
```diff
@@ -258,12 +258,13 @@ class ContentFileSearchRequestSerializer(SearchRequestSerializer):
...
     )
     content_category = StringArrayField(required=False, child=serializers.CharField())
-    topic = StringArrayField(required=False, child=serializers.CharField())
     aggregations = StringArrayField(
         required=False,
         child=serializers.ChoiceField(choices=CONTENT_FILE_AGGREGATIONS),
     )
     resource_type = serializers.ReadOnlyField(default=[CONTENT_FILE_TYPE])
+    run_id = StringArrayField(required=False, child=serializers.CharField())
+    resource_id = StringArrayField(required=False, child=serializers.CharField())
```
ChristopherChudzicki, line 267: Would
```suggestion
    run_id = StringArrayField(required=False, child=serializers.IntegerField())
    resource_id = StringArrayField(required=False, child=serializers.IntegerField())
```
make sense here?

openapi.yaml
```diff
@@ -197,16 +201,98 @@ paths:
...
+            - susskind
+            - global alumni
+            - simplilearn
+            - emeritus
+            - podcast
+            type: string
+            description: |-
+              * `edx` - edx
+              * `ocw` - ocw
+              * `open learning library` - open learning library
```
ChristopherChudzicki, line 261: Do we want this to be `"oll"`, etc? If yes, is that in-scope for this PR?

learning_resources_search/serializers.py
```diff
@@ -158,14 +158,41 @@ def to_representation(self, data):
...
+    )
+    topic = StringArrayField(required=False, child=serializers.CharField())
+
+    def validate_offered_by(self, data):
+        if data:
+            for offered_by_value in data.split(","):
+                if offered_by_value.lower() not in [e.value.lower() for e in OfferedBy]:
+                    msg = f"{offered_by_value} is not a valid option"
+                    raise serializers.ValidationError(msg)
+        return data
```
ChristopherChudzicki, line 185: If we modify StringListField a little, I believe we get all this validation for free. Currently we do not because `StringListField.to_internal_value` does not call its parent method, but with something like
```python
class StringArrayField(serializers.ListField):
    """
    Character separated ListField.
    """

    def __init__(self, *args, **kwargs):
        super().__init__(*args, **kwargs)

    def to_internal_value(self, data):
        """
        Turn ["a", "b,c"] into ["a", "b", "c"]
        """
        normalized = ",".join(data).split(",")
        return super().to_internal_value(normalized)
```
the child serializer should handle the validation automatically.

ChristopherChudzicki (changes requested):
Behaving as described. I think the StringArrayField change is worth making, but if you disagree, np.

openapi.yaml
abeglova: Do we want it to be OLL in the results returned or just the filter? In any case that's probably a separate pr

ChristopherChudzicki (approved):

PR #211, opened 2023-11-13
Remove double formatting checks, switch to Ruff for python formatting

ChristopherChudzicki:
None

We had some formatters that were running on both github actions (`ci.yaml`) and via pre-commit-ci. Additionally, this switches to using `ruff`'s formatter instead of black:

- `ruff` formatter, not black. 
- Only run formatters in one place.
    - ruff and prettier run via pre-commit
    - scss-lint and eslint still run via ci.yaml since making them 

This solves an issue you can see in #171: a formatter (`black`) got updated in poetry.lock, but not in pre-commit.yaml. Since we were running the formatter in two places, with different versions, we couldn't make both happy.

# How can this be tested?
1. Add some formatting errors... maybe `print(1      +      1)` for python and `console.log(1         +       1)` for TS.
2. Run `pre-commit run --all` and check they get fixed.


pyproject.toml
```diff
@@ -1,27 +1,3 @@
...
-  | \.tox
-  | \.venv
-  | build
-  | dist
-  | node_modules
-)/
-'''
-
-# settings to make isort compatible with how black formats imports
-[tool.isort]
```
ChristopherChudzicki, line 17: We were not running `isort`. Ruff does sort imports. Those settings are under `[tool.ruff.lint.isort]` if we want them.

blarghmatey (approved):

PR #216, opened 2023-11-14
remove sql file logging

ChristopherChudzicki:
A followup to #196, removing accidental debugging code

This PR removes the database logger I added in #196 for debugging.

# How can this be tested?
App should work. Sql queries should not be logged to `sql.log`.


mbertrand (approved):
👍

Issue #215, opened 2023-11-14
Move BaseChannel to channels_fields and remove discussions app 


ChristopherChudzicki:
# Description/Context
Now that old Fields have been updated to use new learning resource models, we should move BaseChannel to the fields app and remove the old discussions app.

ChristopherChudzicki:
Closed by #217

Issue #48, opened 2023-08-23
Remove Reddit functionality


collinpreston:
Remove all reddit related functionality as it is no longer needed in the MIT Open project which was forked from the MIT Open Discussions project.

collinpreston:
Closing in favor of: https://github.com/mitodl/mit-open/issues/2

collinpreston:
Closing in favor of: https://github.com/mitodl/mit-open/issues/2

PR #217, opened 2023-11-14
Clean up / consolidate channels code (for fields)

ChristopherChudzicki:
#215 

This PR:
1. Renames `channels_fields` app to `channels`. This app includes our FieldChannel model, and is a home for future channel variants.
2. Moves BaseChannel to the `channels` app from `discussions`.
3. Removes the old `discussions` app. The old discussions app was very bare-bones after we removed reddit. #48 

# How can this be tested?
There should be not noticeable change in behavior: Tests should pass, and you should still be able to create/edit/view channels.

Assuming you have some learningpaths populated:

1. **Create Fields:** Create some fields at http://localhost:8063/api/v1/fields/
   - You can see some associated docs at http://localhost:8063/api/v1/schema/redoc/#tag/fields/operation/fields_create
   - Check that using non-learning paths for `featured_list` and `lists` gives reasonable errors.
2. **View the field details and Update:** http://localhost:8063/api/v1/fields/<field_name>. In particular, try:
    - assigning re-ordering the lists within a field
    - assigning and re-ordering subfields within a field 

## Notes
You will lose old FieldChannel models from the channels_fields app. I think that's OK for the current stage of this project.


channels/migrations/0001_initial.py
```diff
@@ -110,7 +110,7 @@ class Migration(migrations.Migration):
                     "field",
                     models.ForeignKey(
                         on_delete=django.db.models.deletion.CASCADE,
-                        to="channels_fields.FieldChannel",
```
mbertrand, line 113: Since this is essentially a new app and any pre-existing `FieldChannel` objects won't be preserved, Instead of copying all these migrations over and adjusting them, could just start with a clean slate:
- Run `./manage.py migrate channels zero` to clean out your local database tables for this app.
- Remove migrations folder altogether
- Run `./manage.py makemigrations channels`
- End up with just one migration file (`0001_initial.py`) that is up to date.

channels/apps.py
```diff
@@ -0,0 +1,15 @@
+"""apps for channels"""
+
+from django.apps import AppConfig
+
+
+class ChannelsConfig(AppConfig):
+    """Config for ChannelsConfig"""
```
mbertrand, line 7: "Config for Channels"

channels/models.py
```diff
@@ -1,15 +1,81 @@
...
 from open_discussions.models import TimestampedModel
+from profiles.utils import avatar_uri, banner_uri
 from widgets.models import WidgetList
 
+AVATAR_SMALL_MAX_DIMENSION = 22
+AVATAR_MEDIUM_MAX_DIMENSION = 90
+
+
+class BaseChannel(models.Model):
+    """Base abstract model for channels"""
```
mbertrand, line 21: This is fine with me as an abstract model, but I wonder if @ferdi might prefer that we refactor this similarly to what we did with `LearningResource` (which also used to be abstract).  So we would end up with one channels API with potentially different `channel_type` values (like `field_channel`) that could be filtered on.  That is a pretty major change though, and if the API for this app isn't meant to be public-facing like the learning_resources API , then I feel like it wouldn't be worth it.

channels/models.py
```diff
@@ -1,15 +1,81 @@
...
+    )
+    banner = ProcessedImageField(
+        null=True, blank=True, max_length=2083, upload_to=banner_uri
+    )
+    about = JSONField(blank=True, null=True)
+
+    # Miscellaneous fields
+    ga_tracking_id = models.CharField(  # noqa: DJ001
+        max_length=24, blank=True, null=True
+    )  # noqa: DJ001, RUF100
```
mbertrand, line 70: DJ001 is added twice for this field, I think if you remove the first one it would also do away with the need for `RUF100` ("unused-noqa").  Also wondering if we should get rid of it altogether by making this `default=""` instead of `null=True`) - https://docs.astral.sh/ruff/rules/django-nullable-model-string-field/

channels/models.py
```diff
@@ -1,15 +1,81 @@
...
+
+    def __str__(self):
+        """Str representation of channel"""
+        return self.title
+
+    class Meta:  # noqa: DJ012
```
mbertrand, line 76: I think you can get rid of `# noqa: DJ012` if `Meta` is placed above `def __str__` - https://docs.astral.sh/ruff/rules/django-unordered-body-content-in-model/

mbertrand (changes requested):

mbertrand (approved):
👍

PR #224, opened 2023-11-17
do not index any pages, for now

ChristopherChudzicki:
https://github.com/mitodl/mit-open/issues/223

Changes our noindex template to add `<meta name="robots" content="noindex">` in all environments, not just non-production environments.

# How can this be tested?
Check http://localhost:8063 when `MITOPEN_ENVIRONMENT=production`. The page should have a noindex meta tag.

### DIFF
--- a/open_discussions/templatetags/noindex_meta.py
+++ b/open_discussions/templatetags/noindex_meta.py
@@ -1,7 +1,6 @@
 """Adds in noindex meta tag to all pages for non-production environments."""
 
 from django import template
-from django.conf import settings
 from django.utils.html import format_html
 
 register = template.Library()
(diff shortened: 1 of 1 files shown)


open_discussions/templatetags/noindex_meta.py
```diff
@@ -12,6 +11,7 @@ def noindex_meta():
     """Add noindex for non-production environments."""
     return (
         format_html("""<meta name="robots" content="noindex">""")
-        if settings.ENVIRONMENT not in ("production", "prod")
-        else ""
+        # Restore conditional inclusion after launch
+        # if settings.ENVIRONMENT not in ("production", "prod")
+        # else ""
```
ChristopherChudzicki, line 16: I considered adding another environment variable, `DISALLOW_SEARCH_ENGINE_INDEXING` (or something) to control this. But the env check seems good. So then we'd have two checks, one for environment and one for `DISALLOW_SEARCH_ENGINE_INDEXING`. And we'd probably want to get rid of `DISALLOW_SEARCH_ENGINE_INDEXING` after launch, since it would be unused.

So... I just did this, since it seemed simpler.

ChristopherChudzicki:
@ChristopherChudzicki Make an issue to remove this after launch before merging.

mbertrand (approved):
👍

PR #244, opened 2023-11-25
Normalize learning resource serializers

### LINKED ISSUE #36: Reconsider `LearningResource.program` serializer

rhysyngsun:
Closes #36 

- Refactored the serializers into some smaller packages
- Updated `LearningResourceSerializer` to be a polymorphic wrapper around the resource_type specific variants. This now allows the API schema and thus the typescript generated code to be type specific in places where we know what the type is.

**Main branch:**
Screenshot 2023-10-20 at 00-14-59 MIT Open Discussions Course Catalog API

**This Branch:**
Screenshot 2023-10-20 at 00-11-29 MIT Open Discussions Course Catalog API

- Switched `ProgramResourceSerializer.program.courses` to be course learning resources (note the is now typed to `CourseResource`, not a generic LearningResource):

**Main branch:**
Screenshot 2023-10-20 at 00-16-32 MIT Open Discussions Course Catalog API

**This Branch:**
Screenshot 2023-10-20 at 00-12-05 MIT Open Discussions Course Catalog API

- All type-specific APIs now return a specific type using that type's serializer. For example, the programs API:

**Main branch:**
Screenshot 2023-10-20 at 00-19-45 MIT Open Discussions Course Catalog API

**This Branch:**
Screenshot 2023-10-20 at 00-20-13 MIT Open Discussions Course Catalog API

- Updated `factoryboy` and the factory definitions themselves so that the test data they generate is more internally consistent. Tests were breaking otherwise.

# How can this be tested?

- Check out the schema and make sure it makes sense compared to what's on main/RC.
- Test the API in the swagger UI to verify it's returning the revised data structures.
- Functional test the application.

### DIFF
--- a/learning_resources/constants.py
+++ b/learning_resources/constants.py
@@ -1,8 +1,7 @@
 """Constants for learning_resources"""
 
-from enum import Enum
-
 from django.db.models import TextChoices
+from named_enum import ExtendedEnum
 
 OPEN = "Open Content"
 PROFESSIONAL = "Professional Offerings"
@@ -11,7 +10,7 @@ CERTIFICATE = "Certificates"
 FAVORITES_TITLE = "Favorites"
 
 
-class AvailabilityType(Enum):
+class AvailabilityType(ExtendedEnum):
     """
     Enum for Course availability options dictated by edX API values.
     While these are the options coming in from edX that we store as is, we
@@ -27,7 +26,7 @@ class AvailabilityType(Enum):
     archived = "Archived"  # displayed as "Prior"
 
 
-class LearningResourceType(Enum):
+class LearningResourceType(ExtendedEnum):
     """Enum for LearningResource resource_type values"""
 
     course = "Course"
@@ -37,7 +36,7 @@ class LearningResourceType(Enum):
     podcast_episode = "Podcast Episode"
 
 
-class OfferedBy(Enum):
+class OfferedBy(ExtendedEnum):
     """
     Enum for our Offered By labels. They are our MIT "brands" for LearningResources
     (Courses, Bootcamps, Programs) and are independent of what platform.
@@ -57,7 +56,7 @@ class OfferedBy(Enum):
     ctl = "Center for Transportation & Logistics"
 
 
-class PlatformType(Enum):
+class PlatformType(ExtendedEnum):
     """
     Enum for platforms, this should be kept in sync
     with LearningResourcePlatform model objects
@@ -82,7 +81,7 @@ class PlatformType(Enum):
     podcast = "Podcast"
 
 
-class PrivacyLevel(Enum):
+class PrivacyLevel(ExtendedEnum):
     """
     Enum tracking privacy levels for user-created UserLists
     """
--- a/learning_resources/serializers.py
+++ b/learning_resources/serializers.py
@@ -1,10 +1,11 @@
 """Serializers for learning_resources"""
-
 import logging
+from uuid import uuid4
 
 from django.contrib.auth.models import User
 from django.db import transaction
 from django.db.models import F, Max
+from drf_spectacular.helpers import lazy_serializer
 from drf_spectacular.utils import extend_schema_field
 from rest_framework import serializers
 from rest_framework.exceptions import ValidationError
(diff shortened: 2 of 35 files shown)


learning_resources/constants.py
```diff
@@ -56,6 +64,14 @@ class OfferedBy(Enum):
...
     ctl = "Center for Transportation & Logistics"
 
+    @staticmethod
+    def values() -> list[str]:
+        return [c.value for c in OfferedBy]
+
+    @staticmethod
+    def names() -> list[str]:
+        return [c.name for c in OfferedBy]
+
```
mbertrand, line 74: What do you think of this idea:

```python
class OpenEnum(Enum):

    @classmethod
    def values(cls) -> list[str]:
        return [c.value for c in cls]

    @classmethod
    def names(cls) -> list[str]:
        return [c.name for c in cls]


class OfferedBy(OpenEnum):
....

class LearningResourceType(OpenEnum):
...

```

learning_resources/serializers.py
```diff
@@ -392,8 +423,49 @@ def update(self, instance, validated_data):
...
+            PodcastResourceSerializer,
+            PodcastEpisodeResourceSerializer,
+        )
+    }
+
+    def to_representation(self, instance):
+        """Serialize a LearningResource based on resource_type"""
+        serializer_cls = self.serializer_cls_mapping[instance.resource_type]
+
+        return serializer_cls(instance=instance).data
```
ChristopherChudzicki, line 468: A few of the learning resource serializers access `self.context` (see `get_learning_path_parents` or `get_user_list_parents`). Maybe we could change that, but for now I think we need to pass `context=self.context` here.


### Resulting bug
You can see the consequences of this if you:
1. Create a learningpath with some items
2. As a staff user, view `localhost:8063/api/v1/learningpaths/<PATH_ID>`
3. The `learning_path_parents` field is `[]`, but should be populated.

**We have a test that should have caught this**...but the test isn't doing what is is supposed to be doing. In `test_get_resource_learning_paths`, 
```python
for idx, item in items_json:
    assert item.get("id") == path_items[idx].id
    assert item.get("position") == path_items[idx].position
    assert item.get("child") == course.learning_resource.id
```
`items_json` and `path_items` are supposed to be the same-ish, but `items_json` is empty on this branch, so test still passes. (Actually it's empty on main, too, since that test doesn't authenticate the user.)

We should add an `assert len(items_json) == 3` or something to that test.

ChristopherChudzicki (changes requested):
The types look good! I'm still looking, but I did notice one issue. See below.

Really looking forward to having this merged.

ChristopherChudzicki (commented):
Context issue was only issue I noticed. Otherwise, looks great. Glad the `course.resource_type`, etc, got narrowed, too.

learning_resources/constants.py
rhysyngsun: What do you think about pulling in the [`named-enum`](https://pypi.org/project/named-enum/) library which provides this and a lot of other nice methods?

mbertrand: Looks good to me! Still seems to be actively maintained too.

scripts/openapi-configs/typescript-axios.yaml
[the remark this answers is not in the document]
rhysyngsun: Easy enough to put it in now

learning_resources/constants.py
rhysyngsun: Did this and updated the other enums in `constants.py` to use it too for good measure.

manage.py
```diff
@@ -2,10 +2,13 @@
...
 Standard manage.py command from django startproject
 """
-
+import contextlib
 import os
 import sys
 
+with contextlib.suppress(ImportError):
+    from traceback_with_variables import activate_by_import  # noqa: F401
+
```
ChristopherChudzicki, line 11: What's this for?

ChristopherChudzicki (approved):

PR #249, opened 2023-11-29
certifcation should be a boolean

abeglova:
closes https://github.com/mitodl/mit-open/issues/248

- Changes the "certification" parameter to a boolean similar to "professional"
- Fixes a bug that cause mitx learning resources to not be marked as having certificates

# How can this be tested?
http://localhost:8063/api/v1/learning_resources_search/?certification=true
and
http://localhost:8063/api/v1/learning_resources_search/?certification=false 
should work to filter by certification

Certification should be a boolean in the results


openapi.yaml
```diff
@@ -1914,8 +1914,14 @@ paths:
...
           items:
+            enum:
+            - 'true'
+            - 'false'
             type: string
-            minLength: 1
+            description: |-
+              * `true` - true
+              * `false` - false
+        description: true if the learning resource offers a certifacate
```
ChristopherChudzicki, line 1924: This OpenAPI spec is odd. I would expect the boolean parameter to have `type: boolean`.

In `LearningResourcesSearchRequestSerializer`, what do you think of either:

1.  using `certification = serializers.BooleanField(required=False)`
2. OR `certification = StringArrayField(child_serializer=serializers.BooleanField(), ...)`.

(1) seems better to me (no need to pass an array of booleans; no filter applied is the same as `certification=true,false`). But I see that `generate_filter_clauses` expects an iterable, and if we need to keep it that way, (2) seems OK, too.

ChristopherChudzicki (changes requested):
Works well 👍. One requested change for serializer.

learning_resources_search/views.py
```diff
@@ -57,9 +57,11 @@ def get(self, request):
             return Response(SearchResponseSerializer(response).data)
         else:
             errors = {}
-
-            for key, errors_dict in request_data.errors.items():
-                errors[key] = list(set(chain(*errors_dict.values())))
+            for key, errors_obj in request_data.errors.items():
```
abeglova, line 60: This change is needed because the error messages for serializers.BooleanField are a list while the errors for serializers.ListField is a dict with the list position of the element that has errors as the key.

learning_resources/models.py
```diff
@@ -187,17 +187,23 @@ def prices(self) -> str | None:
             return 0
 
     @property
-    def certification(self) -> str | None:
+    def certification(self) -> bool | None:
```
ChristopherChudzicki, line 190:
```suggestion
    def certification(self) -> bool:
```

Previously we returned None as the falsy value, now we return False.

(This will remove `nullable: true` from the OpenAPI spec.)

ChristopherChudzicki (changes requested):
Looks good! One minor requested change for the OpenAPI spec.

Issue #238, opened 2023-11-20
Update React Router to v6


ChristopherChudzicki:
# Description/Context
We use React Router for client-side navigation and are currently on v5. We should update to React Router v6:
- It's the latest major version **and** has been out for 2 years, so is now pretty stable.
- Its [nested routes](https://reactrouter.com/en/main/start/overview#nested-routes) seem useful for separating shared layout from page content
- it introduced some APIs that we are already using (namely, `useSearchParams`, which we backported)
- it has some other new APIs for preloading data which may be useful down the road, but we probably won't use now.

# Plan/Design
React Router has a [upgrade guide](https://reactrouter.com/en/main/upgrading/v5).

ChristopherChudzicki:
Completed in #250

PR #250, opened 2023-11-29
Update React Router to v6

ChristopherChudzicki:
## What are the relevant tickets?
#238 

## Description (What does it do?)
This PR updates our react router version from v5 to v6. There are some significant differences.

### Most notable changes:
- **No more `history` object.**  React router has always (and still) uses another package, `history` to implement its history stack for browser navigation. In `react-router` v5 and below, the `history` package was part of its public API. Now it's an implementation detail. As such, we can't access `history` directly.
    - Most usage of `history.push` (or `history.replace`) is replaceable by a hook, `useNavigate`. 
    - A significant exception to this is our API error handling code. See [this comment](https://github.com/mitodl/mit-open/pull/250/files#r1409864751) for details.
- Routes now defined as an array of objects (Rather than JSX) in `routes.tsx` 

## How can this be tested?
After `docker compose up`ing on this branch...

As an authenticated admin user:
1. Visit http://localhost:8063/learningpaths ... Navigate to individual paths. Things should work.
2. Visit http://localhost:8063/articles/new ... and create an article. Then edit it and save. Routing should work appropriately.
3. Test Refactored 404: Visit http://localhost:8063/articles/not_exist_id
    - You will se a development-only overlay. It's annoying. Click the X in upper right corner. (Possibly twice.)
    -  Then you'll see the 404 page.

Test the Refactered 403 handling:
1. In a private browser, view http://localhost:8063/articles/SOME_ARTICLE_ID. You'll be redirected to a login page. 
2.Logged in as a user without superuser/admin privileges, visit http://localhost:8063/articles/SOME_ARTICLE_ID. You should see a 403 page.


frontends/mit-open/src/routes.tsx
```diff
@@ -0,0 +1,63 @@
...
+import Header from "./components/Header"
+
+const routes: RouteObject[] = [
+  {
+    element: (
+      <>
+        <Header />
+        <Outlet />
+      </>
+    ),
```
ChristopherChudzicki, line 23: RouteObjects are `{ path?, element?, children?, errorElement? }`.

Our top-level route is a "layout route" with no `path`. This means that (a) it is always rendered, and (b) child routes are rendered into the `<Outlet />`.

frontends/mit-open/src/infinite-pages/SearchPage.test.tsx
ChristopherChudzicki, line 1: I removed DemoPage and SearchPage because:

1. They are not functional. They both use [course-search-utils](https://github.com/mitodl/course-search-utils), which needs some significant updates to work with the new search API.
2. ...and `course-search-utils` currently wants access to `history`, which React Router v6 doesn't like.

Changing course-search-utils to be compatible with react router v6 won't be hard and should be done when we update it to use new search API.

frontends/mit-open/src/infinite-pages/DemoPage.tsx
ChristopherChudzicki, line 1: I removed DemoPage and SearchPage because:

1. They are not functional. They both use [course-search-utils](https://github.com/mitodl/course-search-utils), which needs some significant updates to work with the new search API.
2. ...and `course-search-utils` currently wants access to `history`, which React Router v6 doesn't like.

Changing course-search-utils to be compatible with react router v6 won't be hard and should be done when we update it to use new search API.

abeglova (approved):

PR #251, opened 2023-11-29
API endpoints for platforms, offerers, departments

### LINKED ISSUE #246: Add REST endpoints for departments, platforms, offered_by
# Description/Context
Add database-backed REST API endpoints for:

- platforms
- departments
- offered_by

# Plan/Design
Standard django-rest-framework functionality

mbertrand:
Closes #246

Adds API endpoints for:
- LearningResourcePlatform
- LearningResourceOfferor
- LearningResourceDepartment

Makes the schema documentation for the above (plus topics) a bit more user friendly,

Screenshot 2023-11-29 at 3 00 54 PM

Screenshot 2023-11-29 at 3 00 58 PM

# How can this be tested?
You should be able to load the following and get expected results, logged in or not:

http://localhost:8063/api/v1/departments/
http://localhost:8063/api/v1/platforms/
http://localhost:8063/api/v1/offerers/

The schema documentation for the above and topics should be more user-friendly than the default (see screenshots above).


rhysyngsun (changes requested):
Some typos and it also needs a test for the `GET /api/v1/offerors/{id}/` endpoint

rhysyngsun (commented):
Another few things I noticed:

The API for ` LearningResourcePlatform` takes `code` as the url parameter, because it's designated at the primary key and this makes sense to me.

However, for `LearningResourceOfferor`, it takes `id` as the url parameter because that's the primary key, BUT in all our code we don't use `id`, we select and create on `code` as if it's the primary key (it's `unique`, so that's fine). I don't recall offhand how problematic changing the primary key can be, but maybe it makes sense to have that model/api behave the same?

I also noticed when digging into the code for this that there is still a place where `LearningResourcePlatform.platform` is being used: https://github.com/mitodl/mit-open/blob/618e92634fe858730dcd054e180a600f37f4524e/learning_resources/utils.py#L310-L315

mbertrand:
>The API for  LearningResourcePlatform takes code as the url parameter, because it's designated at the primary key and this makes sense to me. However, for LearningResourceOfferor, it takes id as the url parameter 

To me the `LearningResourcePlatform` seems more like the odd one out, since `LearningResourceTopic` (unique `name`), `LearningResourceDepartment` (unique `department_id`), `LearningResourceOffeor` (unique `code`), and `LearningResourceContentTag` (unique `name`) all have an `id` field as the primary key, which is only used for foreign key table relations (and the id argument for API detail endpoints where applicable).  Should all these have their id keys removed, or just add an id primary key to `LearningResourcePlatform` instead?  Seems complex enought that maybe this should be a separate issue/PR.

rhysyngsun (approved):
LGTM

PR #265, opened 2023-12-04
Add RestrictedRoute for controling route access

ChristopherChudzicki:
Closes https://github.com/mitodl/mit-open/issues/228

This PR adds a `<RestrictedRoute allow={user => boolean} />` component that can be used to restrict access to frontend routes. It behaves the same as 403 API errors: shows authenticated users the ForbiddenPage, and redirects unauthenticated users to login.

# How can this be tested?
**Prerequesite:** You should have access to a user that is NOT an admin/superuser. You can create one via the ui at http://localhost:8063/login/ol-oidc/ if you have keycloak creds from RC (I can share). Alternatively you can create a superuser then degrade its permissions via django shell:
```
./manage.py createsuperuser
# finish superuser creation
./manage.py shell
# In python
User = get_user_model()
user = User.objects.get(username="whatever_username")
user.is_admin = False
user.is_superuser = False 
user.save()
```

1. As an unauthenticated user, visit http://localhost:8063/articles/new ; you should be redirected to login
2. As an authenticated user WITHOUT permission (non-superuser, non-admin), visit http://localhost:8063/articles/new 
    - in dev mode, you'll see a black overlay from webpack. Click the X in upper right corner
    - Now you should see the ForbiddenPage.
3. As an authenticated user with permission (e.g., superuser), visit http://localhost:8063/articles/new. You should see the article creation page.


frontends/mit-open/src/routes.tsx
```diff
@@ -40,16 +42,21 @@ const routes: RouteObject[] = [
...
-        element: <ArticleDetailsPage />,
-      },
-      {
-        path: urls.ARTICLES_EDIT,
-        element: <ArticleEditingPage />,
-      },
-      {
-        path: urls.ARTICLES_CREATE,
-        element: <ArticlesCreatePage />,
+        element: <RestrictedRoute allow={isArticleEditor} />,
```
ChristopherChudzicki, line 45: The articles API currently required admin permissions on the backend. `< RestrictedRoute />` restricts the access at render instead of after an API returns. (Which in the case of the article creation page, wasn't until a user submits a form.)

frontends/mit-open/src/types/settings.d.ts
```diff
@@ -2,7 +2,7 @@
 export type User = {
   id: number | null
   is_authenticated: boolean
-  is_public_list_editor: boolean
+  is_article_editor: boolean
```
ChristopherChudzicki, line 5: `is_public_list_editor` was removed; we just forgot to remove it here.

`is_article_editor` already existed on `window.SETTINGS.user`, we just hadn't used it yet.

jonkafton:
Looks good @ChristopherChudzicki (and correct plural on the `Permissions` enum as users will have multiple!)

Issue #239, opened 2023-11-20
Refactor Sass styles to CSS-in-JS


jonkafton:
# Description/Context

Both are currently used for css styling in the React app, though predominantly Sass (with SCSS syntax). [Sass](https://sass-lang.com/) is a styling language that compiles to CSS while providing additional capabilities for dynamic values (variables and interpolation), mixins for reusability and some basic control flow. The class name selector paths are preserved in the CSS build bundle and so are bound to the class attributes on the nested HTML elements that React writes to the DOM.

This limits modularity and separation of concerns for reusability and portability as a nested component that can otherwise be abstracted away from the app for reuse will depend on styles declared in its parent where they are specific to that component nested in the context of the parent.

CSS-in-JS is currently used in the ol-util workspace with [Styled-components](https://styled-components.com/), a library that enables CSS to be written alongside React components inside native JavaScript template literals. As they live in JavaScript, any procedural functionality can be applied natively. TypeScript definitions are also provided for type safety, editor type ahead and strongly typed theming.

Class names are hashed dynamically, removing brittleness around selector path matching and producing CSS files for the build that can be split according to concern and cached across versions where unchanged. Only the styles that are in use are compiled to the build output and only the styles for components actually rendered are injected onto the page, enhancing performance.

As the styles live next to their component, they are easy to keep track of, improving maintenance and making it simple to spot any dead styles not in use. Normal React composition can be used for specialization where a component uses specific styles within its parent context. styled-components uses the [STYLIS](https://github.com/thysultan/stylis) preprocessor for more advanced selector patterns should they be useful for nested components and while migrating from Sass.

[Emotion](https://emotion.sh/) is a library for CSS-in-JS that takes inspiration from Styled-components. It is framework agnostic (Styled-components is for React) though has a ['styled' API](https://emotion.sh/docs/@emotion/styled) that returns a React component from tagged template literals, a very useful pattern from Styled-components.

We would like to use Emotion in favor of Styled-components as we use [MUI](https://mui.com/) extensively, which imports Emotion (so it is already a project dependency and included in our build output) and [provides interoperability](https://mui.com/material-ui/guides/interoperability/#emotion).

# Plan/Design

We'll want to refactor out all Saas for a single consistent styling solution, removing all .scss files and Sass build processing. Given that structural rework is ongoing, the goal here is to create a baseline PR to serve as a reference point for refactoring styles to Emotion as we migrate code.

- Styles should be colocated with their components.

[...]

PR #270, opened 2023-12-06
Remove styled components (use @emotion/styled)

ChristopherChudzicki:
#239 

This PR:
- replaces existing usage of `styled-components` with `@emotion/styled`, introduced in #245 
- moves BannerPage and TitledCarousel to `ol-design` from `ol-util`.
    - I also removed some props from BannerPage that we weren't using anywhere.
- Major version bump in nuka carousel. *This was incidental when I moved the dependency, but it didn't cause problems, so let's keep it.*

# How can this be tested
Homepage should look the same, as should all other pages. Homepage carousel should still work.


frontends/ol-design/src/components/BannerPage.tsx
ChristopherChudzicki, line 1: My focus in this PR was on completely removing `styled-components` in favor of `emotion`, so we have one fewer deprecated / old thing hanging around.

Separately, I think we may want to consider replacing this with a higher-level layout component so there is less repetition between our `pages`.

abeglova (approved):

Issue #257, opened 2023-12-01
Organize package.json JS dependencies better; distinguish devDeps


ChristopherChudzicki:
## Description/Context
We should better organize our Typescript dependencies.

Typescript/Javascript dependencies are listed in `package.json` files in the `dependencies` or `devDependencies` field (there are few other dependency fields that are unimportant for us). Generally, dependencies make it into the end-user bundle whereas dev deps are tools for development (linters, typechecker, testing tools, etc).

Currently:

- All our dependencies are listed as regular dependencies, even linters and testing tools.
- Some dependencies are listed in our root `package.json`, other dependencies are listed in the workspace package files (`frontends/*/package.json`).

The dependency vs devDependency distinction is metadata about dependencies that is useful for multiple reasons. For example, we can be more aggressive in upgrading development dependencies as long as CI tests pass. We could configure Renovate to auto-update dev dependencies even for major versions. Second, we can enforce via linters rules like "Only `*.test.ts` files are allowed to import development dependencies". This prevents development libraries from accidently making there way into our production bundle and inflating its size (which has occured in the past). 

We should additionally have clear guidelines as to where dependencies are installed—at the root or workspace level.

## Plan/Design
Separating dependencies into dev and devDeps will be easy, and we can enforce importing only regular dependencies into bundled files via [no-extraneous-dependencies](https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/no-extraneous-dependencies.md).

The part of this work I am currently unsure about is where we should specify dependencies. MIT Open's frontend code is set up as a monorepo:
```
package.json                (root)
frontends
    api
        package.json        (workspace)
    ol-design
        package.json        (workspace)
    mit-open
        package.json        (workspace)
    ⋮
```
We can specify dependencies:

    1. At the root
    2. per-workspace

Per-workspace is conceptually appealing—each workspace specifies only what it needs—but adds some complication, since if two workspaces both require react-query, we want them to use the same version of react-query. Specifying at the root is simple, but we may not want files within `frontends/api` to be able to use UI components.

PR #279, opened 2023-12-11
Distinguish deps vs devDeps + make each package.json specify own dependencies

ChristopherChudzicki:
#257 

This PR:
1. Moves development-only dependencies to `devDependencies`
2. Updates the package.json files so that each package specifies its own dependencies.  *This was a bit haphazard before*.
3. Enforces that each package.json specify its own dependencies moving forward using `[import/no-extraneous-dependencies`](https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/no-extraneous-dependencies.md#importno-extraneous-dependencies) ESLint rule. See [comment](https://github.com/mitodl/mit-open/pull/279/files#r1423257625) for more.
4. Adds `syncpack` to detect mismatched dependencies. See comment for more.

- [ ] Desktop screenshots

# How can this be tested?
1. Try importing a development dependency in a production file. (Example: `import { faker } from "@faker-js/faker/locale/en"` in HomePage.tsx. You should get a linting error.
2. Try importing a package not specified in a module's package.json; for example, `import { CKEditor } from "@ckeditor/ckeditor5-react"` in HomePage.tsx. You should get a linting error.
3. Visit http://localhost:8063; the app should work.

# Additional Context


.eslintrc.js
```diff
@@ -25,6 +26,20 @@ module.exports = {
         lib: "never",
       },
     ],
+    "import/no-extraneous-dependencies": [
```
ChristopherChudzicki, line 29: This linting rule disallows the use packages that are not included in the **nearest** ancestor package.json file.

This has two purposes:
1. Prevent accidental use of development dependencies in files meant for production. We don't want to accidentally bundle `@faker-js/faker`—it's big!
2. Ensure workspaces declare the modules they use as dependencies.

Regarding (2):

Our frontend code is a monorepo with each module specifying its own dependencies. Node module installers (yarn, npm) [hoist](https://classic.yarnpkg.com/blog/2018/02/15/nohoist/) packages to the root `node_modules/` folder in order to de-duplicate packages. Below, the `root` node_modules is thus guaranteed to exist, but the per-package node_modules may or may not exist after a `yarn install`.

```
mit-open (repo)
package.json                <-- root package.json
node_modules                <-- root node_modules
|--frontends/
    |--mit-open/
        package.json           <-- workspace package.json
        node_modules (MAYBE)   <-- workspace node_modules
    |--ol-util/
        package.json
        node_modules
    |--ol-ckeditor/
        package.json
        node_modules
```

But Node's module resolution algorithm searches all `node_modules/` folders in a files ancestor containing directories. Together with hoisting, this means that dependencies in one workspace are often available in second workspace, even if the second does not specify the dependency.

This ESLint rule prevents that by demanding dependencies appear in the **nearest** ancestor package.json.

.github/workflows/ci.yml
```diff
@@ -113,6 +113,9 @@ jobs:
       - name: Install dependencies
         run: yarn install --immutable
 
+      - name: Dependencies are consistent
+        run: yarn syncpack list-mismatches
```
ChristopherChudzicki, line 117: The downside to workspaces in a monrepo specifying their own dependencies is (a) repetition, and (b) things might get out of sync.

If workspace1 specifies, say, `axios^1.0.0` and workspace2 specifies `axios^0.27.1`, then Node can satisfy the dependencies by installing two separate versions of `axios`. That *can* be appropriate, if we really need two versions or are incrementally updating for some reasons. But in general, if two workspaces use the same dependency, we want the versions to match[^1]. `syncpack` is a tool for enforcing this. It's very configurable, so if we wanted intentional version differences, we could allow that. Our current usage does not require any configuration.

Note: When [Renovate](https://github.com/renovatebot/renovate) updates dependencies, it will update all the package.jsons together, so syncpack will still be happy. 

[^1]: At the very least we want the versions to be compatible, but having them match is easiest.

package.json
```diff
@@ -28,56 +28,42 @@
...
     "eslint-plugin-mocha": "^10.0.5",
     "eslint-plugin-react": "^7.30.1",
     "eslint-plugin-react-hooks": "^4.6.0",
     "eslint-plugin-testing-library": "^5.5.1",
-    "exec-sh": "^0.4.0",
-    "html-entities": "^2.3.3",
-    "iso-639-1": "^2.1.15",
     "jest": "^29.6.4",
     "jest-environment-jsdom": "^29.5.0",
-    "jest-extended": "^3.0.1",
```
ChristopherChudzicki, line 62: Our usage of this was very, very basic so I just removed it.

jonkafton (approved):
Looks good and much needed - seems there were quite a few missing in the package files. Ideally (in my view) dependencies used in a single workspace would install to a local `node_modules` and only shared dependencies would be version squashed and hoisted, ie. imports not specified would fail in addition to the new lint check. I looked at [nohoist](https://classic.yarnpkg.com/blog/2018/02/15/nohoist/), but that's all or nothing unfortunately.

Do we have any tool to scan for superfluous unused dependencies in package.json? [depcheck](https://www.npmjs.com/package/depcheck) would be useful here.

PR #282, opened 2023-12-12
Add filtering for topics, tags, level

### LINKED ISSUE #281: Additional filters for learning_resources API endpoint
Add filters for the following fields to the learning_resources API endpoint:
- topic
- resource_content tag
- level

mbertrand:
Closes #281

- Adds `/api/v1/learning_resources` filters for level, topic and resource_content_tags.
- Modifies `LearningResourceRun.level` to be an ArrayField

Screenshot 2023-12-13 at 11 50 48 AM
Screenshot 2023-12-13 at 11 50 41 AM
Screenshot 2023-12-13 at 11 50 32 AM

# How can this be tested?

The migration should complete successfully with existing learning resource data present.

The following urls should work and return relevant results (assuming you've imported some matching courses from ocw, mostly).

http://localhost:8063/api/v1/learning_resources/?resource_content_tags=exams
http://localhost:8063/api/v1/learning_resources/?topic=social+science
http://localhost:8063/api/v1/learning_resources/?level=undergraduate

Run the following, they should complete without errors (you'll need the approproate .env values set first):

```
./manage.py backpopulate_mit_edx_data
./manage.py backpopulate_ocw_data --course-name 8-591j-systems-biology-fall-2014
```

# Additional Context
I avoided specifying the choices for resource_content_tags and topics because doing so required querying the database on application load, which basically breaks pytest, [requiring hacks to get around](https://github.com/pytest-dev/pytest-django/issues/396#issuecomment-393444718). Instead, I also tried adding them to the `LearningResourceViewSet` via `OpenApiParameter` specs instead ([in this commit](https://github.com/mitodl/mit-open/pull/282/commits/f08eb9d59185f7279cac6cb915b7e16316dec141#diff-12658d780aa8728f3af111a6a3fb4e1c1fe224e462c74f8bc62555b42f258ef9R86-R158)), but that caused the ci tests to fail because the database was empty and the api specs didn't match what was generated locally with data. So I ultimately went back to my original approach of just noting that the topics and resource_content_tags api endpoints can be used to get the list of available values, instead of listing all of them.  (The resource_content_tags endpoint is being added in a separate PR).

As for levels, I decided to use an Enum.


learning_resources/models.py
```diff
@@ -233,7 +233,9 @@ class LearningResourceRun(TimestampedModel):
     image = models.ForeignKey(
         LearningResourceImage, null=True, blank=True, on_delete=models.deletion.SET_NULL
     )
-    level = models.CharField(max_length=128, null=True, blank=True)  # noqa: DJ001
+    level = ArrayField(
```
mbertrand, line 236: I thought about changing this field name to the plural `levels` but ultimately decided to keep it as is.  I can be convinced otherwise though.

abeglova (approved):

PR #285, opened 2023-12-13
documentation

abeglova:
address https://github.com/mitodl/mit-open/issues/268 except for issues that will be fixed by
- https://github.com/mitodl/mit-open/issues/278 
- https://github.com/mitodl/mit-open/issues/280
- https://github.com/mitodl/mit-open/issues/283
and
- https://github.com/mitodl/mit-open/issues/284

I think saying that I will fix all the issues with the documentation was a bit too optimistic. This pr improves many of the issues with the documentation but there will be more improvements in subsequent prs

1) Standardizes titles in the documentation for all api endpoints
2) Adds an endpoint for viewing content tag options
3) Adds options and help text for all learning_resource_search and content_file_search parameters (except content_category, I just removed it from the serializer for now since that is always null)
4) Adds help text to the documentation of most /learning_resources fields. The fields affected by the issues linked above are not documented yet, since we will need to rework the views and serializers to standerdize the param names and will add documentation once that happens. We would need to rework the views and serializers to make it possible to  documentation as well so holding off for now

# How can this be tested?
Go to http://localhost:8063/api/v1/schema/redoc/
Verify that the changes in the list above have been made


mbertrand (changes requested):
Looks great, just a couple suggestions

mbertrand (approved):
👍

PR #287, opened 2023-12-13
Populate content tags in contentfile serializer

### LINKED ISSUE #278: content_category is not populated for content files
This is the "resource type" in the sidebar for OCW search
https://ocw.mit.edu/search/?r=Readings&r=Tools&type=resourcefile. To avoid confusion with the resource_type field it is renamed content_category. Currently it's always null

Currently the value is always null
https://mit-open-rc.odl.mit.edu/api/v1/content_file_search/?aggregations=content_category

### LINKED ISSUE #291: Standardize content tags across `LearningResource` and `ContentFile` models
# Description/Context

The `LearningResource` model has a `resource_content_tag` field which is a Many-to-Many field to `LearningResourceContentTag`

The `ContentFile` model has a `learning_resource_types` field which is an array of strings, most of which match up to names of `LearningResourceContentTag` names.

These seem to be basically the same thing and we should standardize how they are treated.

# Plan/Design
- Use a standard field name for both: `content_tags`?
- Change the `ContentFile` field to be a Many-to-Many field to `LearningResourceContentTag`
- Update ETL functions to set `ContentFile.content_tags` appropriately.
- Update serializers and search index schema if necessary.
- Add documentation to the API schema (`learning_resources`, `learning_resources_search`, `content_files_search`, etc) about using the `/api/v1/content_tags` endpoint to get a list of valid values

mbertrand:
Closes #278 
Closes #291

- Renames `LearningResource.resource_content_tags` to `LearningResource.content_tags`
- Renames `ContentFile.learning_resource_types` to `ContentFile.content_tags` and makes it a many-to-many field to `LearningResourceContentTag`
- Modifies the serializers for the above models so that `content_tags` is represented as `course_feature` for `LearningResource` and `content_feature_type` for `ContentFile`
- Makes filter parameters work for these fields.

# How can this be tested?
- New migrations should complete
- Run `./manage.py recreate_index --all`
- Import some OCW courses if you don't already have some
- Go to the following urls (you can swap out "Exams" with some other value that you know your contentfiles have).  You should get some relevant results back.
  - http://localhost:8063/api/v1/learning_resources/?course_feature=Exams
  -  http://localhost:8063/api/v1/learning_resources_search/?course_feature=Exams
  - http://localhost:8063/api/v1/content_file_search/?content_feature_type=Exams
 

# Additional Context
In open-discussions this used to be a `SerializeMethodField` that would determine the value based on whether or not the parent OCW course was created through OCW Studio (use the `learning_resource_types` field) or not (ie from Plone, so use a mapping based on the `section` field).  This no longer seems necessary, I checked on open-discussions production and all published content files have a null `section` field, so it should just come from `learning_resource_types` .  The `section` field (and others related to Plone) should be removed; made a new issue for that: https://github.com/mitodl/mit-open/issues/288 and will address it in a separate PR.

I decided to name the related many-to-many model field `content_tags` in `ContentFile` and `LearningResource` since they both represent a many-to-many relationship to `LearningResourceContentTag`, but in the serializers they are represented as `ContentFile.content_feature_type` and `LearningResource.course_feature` instead (and filtered by those names in the api).


abeglova (approved):
lgtm

abeglova (approved):
lgtm

PR #290, opened 2023-12-14
Refactor to restructure front end components

jonkafton:

[RFC: Front End Component Structure #3159](https://github.com/mitodl/hq/discussions/3159)

# Description

Applies the [RFC: Front End Component Structure](https://docs.google.com/document/d/1TF4ac63IrE8heysdWP4jo1ElfvIosr4HnOASIKmfpnw) guidelines, restructuring the front end workspaces and components.

- `mit-open` routable pages (home, learning paths, article) take single export directory structure.
- Infinite pages and dependencies self contained as they are candidate for deprecation.
- `tsconfig` path mappings added to alias long relative import path.
- Leaf components moved to `mit-open/components` directory.
- `mit-open/services` directory created for `api` client and library initializers.
- `ol-design` repurposed to `ol-components` for shared leaf components.
- New `page-components` directory. Larger components meeting page component criteria are moved there.
- `ol-search-ui` deprecated and merged out to `mit-open/page-components`, `ol-components` and `ol-utilities`. Where appropriate, conflicting exports are suffixed `*Legacy`.
- `ol-learning-resources` also deprecated into `mit-open/page-components`, `ol-components` and `ol-utilities`.
- Sass files colocated to their component directories (these to be replaced with TypeScript).
- `ol-common` workspace created for items common across the app and `ol-components`. Currently just shared interfaces from `ol-search-ui` are here. Open question - should all _reused_ type and interface definitions go here or should we colocate where practical?
- `ol-utils` renamed `ol-utilities`. A couple of new formatters implemented here (`formatDate`, `decodeAndStripTags`) to avoid importing libraries such as `moment` and `html-entities` in root workspaces. Utility code from `ol-search-ui` and `ol-learning-resources` moved to files named respectively for now (utilities will benefit from refactor).
- Unit test factories from `ol-search-ui` and `ol-learning-resources` moved to `ol-utilities/test-utils`. These live in many places - there is a case for creating a dedicated `ol-unit-testing` workspace.

# How can this be tested?

- Unit tests run.
- Manual sanity check that pages render at all routes.

## Checklist:
- [x] Write up concluded RFC in markdown for repo docs, pending PR review.


ChristopherChudzicki (changes requested):
This looks good 👍 

I made a few minor notes that I think are worth changing.

Additional suggestions:
1. **Use `@/` for root import** Rather than having aliases to `components`, `pages`, etc, what do you think of just using `@/` as the standard root import?
    - demo commit https://github.com/mitodl/mit-open/compare/jk/restructure-reference...jk/restructure-reference-cc?expand=1
    - random comment: I picked this up when using Vue; their default templates alias `@/` to root. 
2.  We can use [`import/no-restricted-paths`](https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/no-restricted-paths.md#importno-restricted-paths) to enforce the layer importing rules; something like

    ```
        "import/no-restricted-paths": [
      "error",
      {
        zones: [
          {
            target: "**/components/**",
            from: "**/{pages,page-components,services}/**",
          },
        ],
      },
    ],
    ```
    I think this needs us to specify the tsconfig path for [`eslint-import-resolver-typescript`](https://github.com/import-js/eslint-import-resolver-typescript?tab=readme-ov-file#configuration) in order for it to work with the TS alias (`@` or otherwise).

Could always do (2) separately...merging a big pr like this sooner rather than later is always nice.

frontends/mit-open/src/common/routes.tsx
```diff
@@ -1,22 +1,18 @@
...
-import FieldPage from "./infinite-pages/field-details/FieldPage"
-
-import Header from "./components/Header"
-import { Permissions } from "./util/permissions"
+import HomePage from "pages/HomePage/HomePage"
+import RestrictedRoute from "components/RestrictedRoute/RestrictedRoute"
+import LearningPathListingPage from "pages/LearningPathListingPage/LearningPathListingPage"
+import LearningPathDetailsPage from "pages/LearningPathDetailsPage/LearningPathDetailsPage"
+import ArticleDetailsPage from "pages/ArticleDetailsPage/ArticleDetailsPage"
+import { ArticleCreatePage, ArticleEditPage } from "pages/ArticleUpsertPages"
```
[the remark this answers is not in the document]
jonkafton: We have the option of grouping pages that are similar and providing an index.ts facade for named import/exports. In this case both `ArticleCreatePage` and `ArticleEditPage` import and wrap inside `ArticleUpsertPage` so made sense to self contain.  Will check whether they are imported anywhere and otherwise delete.

jonkafton:
> 1. **Use `@/` for root import** Rather than having aliases to `components`, `pages`, etc, what do you think of just using `@/` as the standard root import?

Yes - I was avoiding `@` as it implies an npm namespace, though `@/` provides the distinction.

frontends/mit-open/src/common/routes.tsx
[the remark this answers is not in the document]
ChristopherChudzicki: I think I just misread the imports + missed the index file.

frontends/mit-open/src/common/routes.tsx
ChristopherChudzicki, line 1: This should not be in `common/` since it imports from `pages/`. (In contrast, I think `urls.ts` should stay in `common/`).

`routes.tsx` is more akin to `App.tsx`. App.tsx doesn't directly import any pages, but `routes.tsx` does. (Alternatively, we could define the routes directly in `App.tsx, which is rather small currently. IMO it's nice to keep them separate, though.)

This is a react router v6 thing, btw. In react router v5 the routes would have been defined as children of a `<Router>` component, so they would have more naturally gone in App.tsx in v5.

jonkafton:
> 2\. We can use [`import/no-restricted-paths`](https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/no-restricted-paths.md#importno-restricted-paths) to enforce the layer importing rules; something like
>
> ...
> 
> Could always do (2) separately...merging a big pr like this sooner rather than later is always nice.

Agree let's tackle later to get the merge in. The issue I'm seeing is that once installing [eslint-import-resolver-typescript](https://github.com/import-js/eslint-import-resolver-typescript?tab=readme-ov-file#configuration) ESLint is picking up the workspaces for the `import/no-extraneous-dependencies` rules and erroring on e.g. ```error  'ol-utilities' should be listed in the project's dependencies```. We'd likely need `.eslintrc` config for each workspace. [eslint-import-resolver-lerna](https://github.com/Dreamscapes/eslint-import-resolver-lerna) may solve the issue for us . Example branch here: https://github.com/mitodl/mit-open/tree/jk/eslint-restricted-import-paths

frontends/ol-components/src/components/SearchFilter/SearchFilter.tsx
```diff
@@ -1,3 +1,7 @@
+/*
+ * Not imported anywhere outside SearchFilterDrawer, also not imported - deprecate?
+ */
+
```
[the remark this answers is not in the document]
jonkafton: Removed

frontends/mit-open/src/common/routes.tsx
jonkafton: Moved to app root as top level code.

docs/architecture/front-end-component-structure.md
```diff
@@ -0,0 +1,179 @@
...
+
+- `ol-utilities` - shared utility code.
+
+- `ol-common` - For code common to `mit-open`, `ol-page-components` and `ol-components`. Used for TypeScript definitions and anything that has visibility of the app or is not a utility or service.
+
+## Large shared components
+
+Some components are large, contain a lot of functionality or can be considered a “mini app” for embedding within root applications. It may not necessarily be only their size that qualifies them for different treatment - perhaps they have special build chains, are not built on React, or have other requirements. The example in the project is the CKEditor, a rich text editor that has its own workspace, ol-ckeditor, in which we wrap and configure it for use in our application.
+
+These will always have their own package.json (as a defining factor), so are publishable independently to a package repository and should be, for reuse outside the project. To contain these and prevent our top level directories from sprawling, let’s introduce an ol-packages workspace that constitutes our npm packages library. Each of its child directories may contain a workspace of its own.
```
ChristopherChudzicki, line 94: All the workspaces have their own `package.json`, so all are packages in this sense. Not all are publishable, though.

>  let’s introduce an ol-packages work...

IMO, it's a bit awkward for this file to talk about things we should do. If it's committed, I would expect it to be things we have done.

docs/architecture/front-end-component-structure.md
```diff
@@ -0,0 +1,179 @@
...
+
+- `ol-components` can only import from other `ol-components` or `ol-utilities`.
+
+- `ol-services` can only import from `ol-utilities`.
+
+- `ol-utilities` cannot import outside its workspace.
+
+- Directories that contain React components should contain a component file with the same name as the directory that exports a single component.
+
+- Files can only import this component, for example, Article.tsx can import CommentBoard.tsx, but not CommentsList.tsx. We should enforce with lint rules:
```
ChristopherChudzicki, line 167: Having looked a bit at ESLint rules for imports, I am not too optimistic that we can actually enforce this. Most rules are based on glob patterns, which aren't flexible enough to enforce that X=Y in `import X from "../whatever/X/Y`.

So, may just have to be unenforced convention.

ChristopherChudzicki (approved):

docs/architecture/front-end-component-structure.md
jonkafton: Agree, removed.

PR #313, opened 2023-12-18
Remove put method from DRF views

### LINKED ISSUE #292: Remove extra api calls for update
Articles, user lists, learning paths and other write endpoints have both an update (PUT) and a partial update (PATCH) api call.  The PUT api call updates all the fields while the PATCH updates just the fields in the payload.

We should get rid of the PUT api calls and call the PATCH api calls just "update" in the redoc documentation

mbertrand:
Closes #292 

Removed the "put" http method from writable API viewsets and schema documentation

# How can this be tested?
Tests should still pass.
The schema documentation for writable api endpoints (channels, articles, learning paths, user lists) should still show PATCH, POST, and DELETE options but no PUT option.

### DIFF
--- a/articles/views.py
+++ b/articles/views.py
@@ -24,7 +24,7 @@ class DefaultPagination(LimitOffsetPagination):
     retrieve=extend_schema(summary="Retrieve"),
     create=extend_schema(summary="Create"),
     destroy=extend_schema(summary="Destroy"),
-    partial_update=extend_schema(summary="Partial Update"),
+    partial_update=extend_schema(summary="Update"),
 )
 class ArticleViewSet(viewsets.ModelViewSet):
     """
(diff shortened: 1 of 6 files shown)


ChristopherChudzicki (approved):
👍

articles/views.py
```diff
@@ -22,7 +23,6 @@ class DefaultPagination(LimitOffsetPagination):
     list=extend_schema(summary="List"),
     retrieve=extend_schema(summary="Retrieve"),
     create=extend_schema(summary="Create"),
-    update=extend_schema(summary="Update"),
     destroy=extend_schema(summary="Destroy"),
     partial_update=extend_schema(summary="Partial Update"),
```
abeglova, line 27: Can you update "Partial Update" to just "Update" here and everywhere else?

abeglova (changes requested):

abeglova (approved):

PR #321, opened 2023-12-19
Update calculation of `LearningResource.resource_num` property

### LINKED ISSUE #320: Incorrect `resource_num` value for OCW course resources
The `resource_num` value returned by `LearningResource.resource_num` property is wrong for OCW courses

## Expected Behavior
For OCW course with a readable_id of `6.001+spring_2005` the `readable_num` output should be `6.001`

## Current Behavior
The `readable_num` output is`spring_2005`

## Steps to Reproduce
Go to http://localhost:8063/api/v1/content_file_search/?platform=ocw and check the values of `resource_readable_num` fields.

## Possible Solution
Adjust the function to conditionally calculate the value based on etl source or platform.

mbertrand:
Fixes #320 

- Gets rid of this property from the `LearnbingResource` model and moves it to the `ContentFileSerializer` since that is the only place it was being used.
- Renames it to `course_number` and makes it an array of `Course.course_numbers.value` strings

# How can this be tested?

Run `./manage.py recreate_index --all`
Run `./manage.py backpopulate_ocw_data --overwrite  --course-name 11-014j-american-urban-history-ii-fall-2011` and maybe import some other courses if you don't have anymore, OCW and otherwise.

Go to the following urls, you should see `course_number` in results, as an array containing at least one value and possibly more.
http://localhost:8063/api/v1/contentfiles/
http://localhost:8063/api/v1/content_file_search/
http://localhost:8063/api/v1/content_file_search/?platform=ocw&q=21H.232J

The last one should return contentfiles with a `course_number` value of `[ "11.014J", "21H.232J"]`

# Additional Context
Not sure if it would make sense to add another filter on this (for both opensearch and db-backed api's), for now it is queryable in opensearch via the `q` parameter.

### DIFF
--- a/learning_resources/views.py
+++ b/learning_resources/views.py
@@ -492,7 +492,12 @@ class ContentFileViewSet(viewsets.ReadOnlyModelViewSet):
 
     serializer_class = ContentFileSerializer
     permission_classes = (AnonymousAccessReadonlyPermission,)
-    queryset = ContentFile.objects.filter(published=True).order_by("-updated_on")
+    queryset = (
+        ContentFile.objects.select_related("run")
+        .prefetch_related("content_tags")
+        .filter(published=True)
+        .order_by("-updated_on")
+    )
     pagination_class = DefaultPagination
     filter_backends = [DjangoFilterBackend]
     filterset_class = ContentFileFilter
--- a/frontends/api/src/generated/api.ts
+++ b/frontends/api/src/generated/api.ts
@@ -220,7 +220,7 @@ export interface ContentFile {
    * @type {string}
    * @memberof ContentFile
    */
-  resource_readable_num: string
+  course_number: string
   /**
    *
    * @type {string}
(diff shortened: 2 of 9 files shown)


learning_resources/views.py
```diff
@@ -489,7 +489,12 @@ class ContentFileViewSet(viewsets.ReadOnlyModelViewSet):
 
     serializer_class = ContentFileSerializer
     permission_classes = (AnonymousAccessReadonlyPermission,)
-    queryset = ContentFile.objects.filter(published=True).order_by("-updated_on")
+    queryset = (
+        ContentFile.objects.select_related("run")
+        .prefetch_related("content_tags")
+        .filter(published=True)
+        .order_by("-updated_on")
+    )
```
mbertrand, line 497: Gets rid of some n+1 queries I noticed during testing

abeglova (approved):

PR #325, opened 2023-12-19
Add public learning paths to the index

### LINKED ISSUE #106: Add LearningPath objects to the search index.
# Description/Context
Add LearningPath objects to the search index.

# Plan/Design
Design should be very similar to the current implementation in `course_catalog` for public StaffLists, but use the new `learning_resources` models and serializers instead.

mbertrand:
Closes #106 

- Creates, updates, or removes published learning path documents from the search index when appropriate

# How can this be tested?
- Create some learning paths if you don't have any already, with at least 1 public and 1 private, at this url (need to be logged in as an admin): http://localhost:8063/learningpaths/
- If you already have some, just run `./manage.py recreate_index --learning_paths`
- Check http://localhost:8063/learning_resources_search/?resource_type=learning_path, you should see published learning paths there only
- Edit a public learning path and make it private instead.
- Wait a minute, then go back to the search url above.  The learning path you made private should no longer show up.
- Edit a private path, making it public.
- Wait a minute, then go back to the search url above.  The learning path you made public should show up.


abeglova (approved):
Works great! One thing i noticed is that the learning path specific part of the learning resource serializer is

>             "learning_path": {
>                 "id": 1,
>                 "item_count": 0,
>                 "author": 1
>             },

I don't think the user id is referenced anywhere else. Should "author " be replaced with either the name or the username of the user? Probably a seperate pr

mbertrand:
Made a new issue for that: https://github.com/mitodl/mit-open/issues/328

PR #319, opened 2023-12-19
Add a user menu with login, logout to UI

ChristopherChudzicki:
# What are the relevant tickets?
Closes #241 

# Description (What does it do?)
This PR adds a user menu in top-right corner with log in / log out links as appropriate. A few other links are added, too.

**Authenticated user / Anonymous user** *Note, "Learning Paths" only available to those with editing permissions".
    
Screenshot 2023-12-19 at 9 57 14 AM / Screenshot 2023-12-19 at 9 57 23 AM

Additionally, I made some updates to the reusable SimpleMenu component.

# How can this be tested?

**Prerequisites:** Connect your local instance to keycloak; either local keycloak or on RC. Ask me for creds if you need them.

Assuming you are initially unauthenticated:

1. **Unauthenticated:** Visit http://localhost:8063/ ; in top right, click user menu and log in. You should be redirected to keycloak, after which you're redirected back to MIT Open.
2. **Authenticated:** Now click usermenu
    - if your user has `is_learning_path_editor` permission, they'll see a "Learning Paths" link.
    - all authenticated users see "Log out" link.
3. **Authenticated:** Try logging out. It should work.
    - Note: Does not currently redirect to MIT Open

**Some accessibility stuff:** This is most easily tested as a superuser, since you'll have multiple links int he dropdown usermenu.
- Check that you can tab to the usermenu trigger and open it with keyboard
- Once it's open, check that navigation with keyboard arrows works.
    - The indidividual menu items are navigable with keyboard arrows, NOT via tab. See https://www.w3.org/WAI/ARIA/apg/patterns/menubar/
- The whole menu item should be clickable as a link
    Screenshot 2023-12-19 at 10 12 02 AM

PR #330, opened 2023-12-21
preserve current location when redirecting to login

ChristopherChudzicki:
Small followup to #319 

In #319, when a user logged in, they would be redirected to the homepage. Instead, this PR preserves the current page.

# How can this be tested?

**Prerequisites:** Connect your local instance to keycloak; either local keycloak or on RC. Ask me for creds if you need them.

Assuming you are initially unauthenticated:

1. As an unauthenticated user, visit http://localhost:8063/learningpaths?pup=woof *The search parameters don't do anything, but let's check that they get preserved.*
2. In top right, click log in, and log in via keycloak.
3. After logging in, you should be redirected to the page you were on (http://localhost:8063/learningpaths?pup=woof)

### DIFF
--- a/frontends/mit-open/src/common/urls.test.ts
+++ b/frontends/mit-open/src/common/urls.test.ts
@@ -1,0 +1,19 @@
+import { login } from "./urls"
+
+test("login encodes the next parameter appropriately", () => {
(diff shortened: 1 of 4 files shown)


collinpreston (approved):
Works well, nice!

PR #331, opened 2023-12-22
enforce import hierarchy by linting; finish removing old code

### LINKED ISSUE #179: Enforce import conventions with linting rules
# Description/Context
It would be useful to be able to enforce import conventions with linting rules. In particular, the hierarchy rules outlined in https://github.com/mitodl/mit-open/blob/main/docs/architecture/front-end-component-structure.md#module-boundary-and-importexport-rules

# Plan/Design
The [`no-restricted-import`](https://eslint.org/docs/latest/rules/no-restricted-imports) rule is designed to lint imports, but is a little restricted in terms of things like "allow certain imports in directory X, but not other directories". See (https://github.com/eslint/eslint/discussions/17047).

There may be alternative rules we could use. Or, it's possible ESLint's [new configuration format ](https://eslint.org/blog/2022/08/new-config-system-part-2/#from---rulesdir-to-runtime-plugins) is flexible enough to overcome the aforementioned issue with `no-restricted-imports`.

ChristopherChudzicki:
Closes #179 

This PR enforces the import hierarchy described in https://github.com/mitodl/mit-open/blob/main/docs/architecture/front-end-component-structure.md via linting rules.

Additionally, this PR removes a lot of unused code. The removed code is generally
- scss files and components related to things we previous removed: (field pages in #290, old search page in #250) #router
- API service for old, non-functional course catalog API

We will likely bring some of this back at some point, but for now there's no need to keep it and simplifies some updates (like switching to `@emotion` based styling). We have it readily available in the git history and in the old repo at https://github.com/mitodl/open-discussions.

# How can this be tested?
1. Open any file in `frontends/mit-open/src/components/` directory and try to import something from `services`, `page-components`, or `pages. You should get an error with a meaningful error message.
    - Try both the typescript alias `@/pages/X/X` version of the import and the relative version, e.g., `../../pages/X/X`
2. Try similar things in other directories, as described in https://github.com/mitodl/mit-open/blob/main/docs/architecture/front-end-component-structure.md


jonkafton (commented):
Nice clean up and great to see the test utilities extracted to their own workspace.

I wasn't able to trigger the `import/no-restricted-paths` error on tsconfig compilerOptions.paths mappings, e.g. in src/components/**, `import Header from "../../page-components/Header/Header"` produces the lint error, but `import Header from "@/page-components/Header/Header"` does not (both VSCode marking and with `yarn lint-check`).

The `eslint-import-resolver-typescript` plugin appears to be configured correctly, so not sure what the reason is. Worth noting, if we add a `"import/no-unresolved": "error"` rule, all the `@/..` imports fail linting, so likely relates to the import resolver plugin itself.

Checked also on CI with https://github.com/mitodl/mit-open/compare/cc/lint-and-pruning...jk/ts-mapping-eslint-test, see the lint job on https://github.com/mitodl/mit-open/actions/runs/7399126037/job/20129819420

ChristopherChudzicki:
> I wasn't able to trigger the import/no-restricted-paths error on tsconfig compilerOptions.paths mappings, e.g. in src/components/**, import Header from "../../page-components/Header/Header" produces the lint error, but import Header from "@/page-components/Header/Header" does not (both VSCode marking and with yarn lint-check).

Good catch.

I believe this is because:
-  The `.eslintrc.js` file references the tsconfigs from root (`frontends/*/tsconfig.json`)
- `yarn lint-check` runs eslint with the workspace directories as `pwd`

My IDE (VSCode) was picking the `@/` import errors when `mit-open` opened from root via `cd ~/dev/mit-open && code .`

I think there are two possible resolutions:
1. Change `yarn lint-check` to run from project root
2. Change `.eslintrc.js` to reference tsconfigs via `./tsconfig.json`

I just updated this PR to implement (1). See changes in https://github.com/mitodl/mit-open/pull/331/commits/d5118c7b87d76555954fbc6ee15ab1ac53e4055f

I remain a little confused why this wasn't working in your IDE. If you `cmd+shift+p` and search for "settings.json`, do you have `eslint.workingDirectories` set to anything unusual? (I believe no value / the default should work... )

jonkafton (approved):

PR #359, opened 2024-01-03
Fix api urls

abeglova:
closes https://github.com/mitodl/mit-open/issues/280
closes https://github.com/mitodl/mit-open/issues/283

This pr standardizes the names of path variables in the documentation.
Additionally it fixes the issue that nested routes such as api/v1/learning_resources/[learning_resource]/items/[id] required the LearningResourceRelationship id that was not obtainable from the api. Now api/v1/learning_resources/[learning_resource]/items/ returns information about both the nested resources and the relationship

# How can this be tested?
1) Test that you can create and edit learning paths at http://localhost:8063/learningpaths/
and add items from the front page carousel to a learning path
2) Go to http://localhost:8063/api/v1/schema/swagger-ui/ and check that the names of the path variables in the urls are standardized  to `learning_resource_id` or `userlist_id` instead of the original names like `run__learning_resource` and `parent_id`. The path variables should also have a description explaining what they are
3) Test out the nested api calls under api/v1/learning_resources and api/v1/learningpaths and api/v1/userlists. They should work
4) Additionally the results for api/v1/learning_resources/[learning_resource]/items/ should show both the resources and the relationship information.


pyproject.toml
```diff
@@ -76,6 +75,7 @@ nh3 = "^0.2.14"
 retry2 = "^0.9.5"
 pluggy = "^1.3.0"
 named-enum = "^1.4.0"
+drf-nested-routers = "^0.93.5"
```
ChristopherChudzicki, line 78: Just curious, what motivated switching to this rather than `drf-extensions`?

ChristopherChudzicki (approved):
Tested out learningpaths and userlists. All looks good.

One question out of curiosity

pyproject.toml
abeglova: I switched out the library because drf-extentions didn't support having parameter names that are different from the model parameters without a bunch of custom code

Which makes sense, because unless you are using drf-spectacular to auto generate documentation, the parameter names in the urls are not visible to the user in any way

PR #363, opened 2024-01-04
error on extra keys for the search api calls

abeglova:
closes https://github.com/mitodl/mit-open/issues/286

If you misspell or use an incorrect parameter in the the search endpoints, the parameter is ignored. This is annoying because it's not really visible that the parameter is incorrect since you get results back. Show an error instead

# How can this be tested?
http://localhost:8063/api/v1/content_file_search/?train=choochoo
and
http://localhost:8063/api/v1/learning_resources_search/?train=choochoo

should error.

Requests to those endpoints with valid parameters should work normally


learning_resources_search/views.py
```diff
@@ -54,8 +54,12 @@ class LearningResourcesSearchView(ESView):
...
         request_data = LearningResourcesSearchRequestSerializer(data=request.GET)
         if request_data.is_valid():
-            response = execute_learn_search(request_data.data)
-            return Response(SearchResponseSerializer(response).data)
+            extra_keys = set(request.GET.keys()) - set(request_data.fields.keys())
+            if extra_keys:
+                return Response(f"invalid parameters: {list(extra_keys)}", status=400)
+            else:
+                response = execute_learn_search(request_data.data)
+                return Response(SearchResponseSerializer(response).data)
```
ChristopherChudzicki, line 62: **Suggestion:** Since this is validation, should it go in the serializer? https://stackoverflow.com/a/36205600/2747370

ChristopherChudzicki (approved):
Works well 👍 

I thought DRF behaved this way for normal ModelViewSet endpoints, but I guess it does not.

One suggestion—up to you.

PR #364, opened 2024-01-05
fix learning resource default in documentating for search

abeglova:
closes https://github.com/mitodl/mit-open/issues/362

This pr removes the default value for learning_resource from the learning_resources_search serializer since it's displayed oddly in the documentation

# How can this be tested?
Go to http://localhost:8063/api/v1/learning_resources_search
The search should work the same as previously when there is no resource_type parameter
The search should also work the same as previously when resource_type is specified
e.i http://localhost:8063/api/v1/learning_resources_search/?resource_type=podcast,podcast_episode

### DIFF
--- a/learning_resources_search/api.py
+++ b/learning_resources_search/api.py
@@ -453,6 +453,9 @@ def execute_learn_search(search_params):
         dict: The opensearch response dict
     """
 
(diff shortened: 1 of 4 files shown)


ChristopherChudzicki (approved):
👍

PR #385, opened 2024-01-09
Remove author from LearningPath serializer

### LINKED ISSUE #328: "author" field in LearningPath serializer: is it necessary?
# Description/Context
The `LearningPath` serializer currently includes an `author` field whose value is equal to the id of the user who created it.  

Is this field going to be used for anything? If so, would `username` or `name` be better?  If not, it should be excluded from the serializer data output.

```
        "learning_path": {
            "id": 1,
            "item_count": 0,
            "author": 1
        },
```

mbertrand:
Closes #328

Removes "author" field from LearningPathSerializer

### How can this be tested?
Unit tests should pass.
Create a learning path if you don't already have any (log in as an admin and go to http://localhost:8063/learningpaths/)
Check the serializer output at http://localhost:8063/api/v1/learningpaths.  You should see something like this, with no `author` field present:

```
            "learning_path": {
                "id": 1,
                "item_count": 2
            },
```

### DIFF
--- a/learning_resources/serializers.py
+++ b/learning_resources/serializers.py
@@ -207,7 +207,7 @@ class LearningPathSerializer(serializers.ModelSerializer, ResourceListMixin):
 
     class Meta:
         model = models.LearningPath
-        exclude = ("learning_resource", *COMMON_IGNORED_FIELDS)
+        exclude = ("learning_resource", "author", *COMMON_IGNORED_FIELDS)
 
 
 class PodcastEpisodeSerializer(serializers.ModelSerializer):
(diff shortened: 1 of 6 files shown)


learning_resources/serializers.py
ChristopherChudzicki, line 1: I think we can `author` from
```python
            models.LearningPath.objects.create(
                learning_resource=path_resource, author=request.user
            )
```
in `LearningPathResourceSerializer.create`, too, yes?

mbertrand: Only if we want to stop tracking which user created the LearningPath altogether, in which case the field should be removed from the model as well, not just the serializer.

ChristopherChudzicki (approved):
👍

ChristopherChudzicki (approved):
👍

PR #390, opened 2024-01-11
Replace Sass styles with Emotion's CSS-in-JS

jonkafton:

Completes https://github.com/mitodl/mit-open/issues/239.

This removes all .scss Sass files and style loaders (style-loader, postcss, sass-loader) used during build to generate the css bundle, removing the `sass` dependency from the project.

The Sass styles have been migrated to [Emotion](https://emotion.sh/docs/introduction) Css-inJS component styling.

### How can this be tested?

All pages and components should render without change to style or layout.


frontends/ol-ckeditor/src/webpack-utils.js
```diff
@@ -3,6 +3,13 @@ const {
   CKEditorTranslationsPlugin,
 } = require("@ckeditor/ckeditor5-dev-translations")
 
+/* TODO
+  Is this config used, ie. do we ever prebuild ol-ckeditor?
+  Currently this is built from source from the mit-open entrypoint.
```
ChristopherChudzicki, line 8: We do use this: https://github.com/mitodl/mit-open/blob/3b55e0e8057a7a5916a975133680a3a9f8d3330a/frontends/mit-open/webpack.config.js#L152

We do not pre-build CKEditor—this isn't a config, it's utils for including CKEditor into a webpack build.

frontends/mit-open/src/page-components/LearningResourceCardTemplate/LearningResourceCardTemplate.tsx
```diff
@@ -103,23 +95,162 @@ const LRCImage: React.FC<LRCImageProps> = ({
...
+const TypeRow = styled.div`
+  display: flex;
+  flex-direction: row;
+  justify-content: space-between;
+  align-items: center;
+  min-height: 1.5em; // ensure consistent height even if no certificate
+`
+
+const EllipsisTitle = styled(Dotdotdot)`
+  font-weight: bold !default;
```
ChristopherChudzicki, line 222: The `!default` bit here is invalid css; that's scss syntax.

frontends/mit-open/src/page-components/LearningResourceCardTemplate/LearningResourceCardTemplate.tsx
```diff
@@ -103,23 +95,162 @@ const LRCImage: React.FC<LRCImageProps> = ({
...
+
+  .ol-lrc-chip.MuiChip-root {
+    height: 2.5 * ${SMALL_FONT_SIZE};
+    font-size: ${SMALL_FONT_SIZE};
+
+    .MuiSvgIcon-root {
+      height: 1.25 * ${SMALL_FONT_SIZE};
+      width: 1.25 * ${SMALL_FONT_SIZE};
+    }
+  }
```
ChristopherChudzicki, line 159: The multiplication here is invalid syntax. (This was compile-time scss syntax.) (I guess it should be `${ 2.5 * SMALL_FONT_SIZE_EM } em` now.

ChristopherChudzicki (changes requested):
### Syntax issues and linting
I noticed a few invalid syntax issues. I noted a few of them, but there are more. IMO, we should set up `stylelint` to lint emotion styles, probably before we merge this. 

I took a stab at setting it up, just to see how many violations we have right now:
```yaml
extends: stylelint-config-standard-scss
rules:
  selector-class-pattern:
    - (^ck.*)|((^Mui[A-Z-])|^([a-z][a-z0-9]*)(-[a-z0-9]+)*$) # CKEditor, MUI or kebab-case
    - message: "Expected class selector to be kebab-case"
overrides:
  - files: ["**/*.tsx", "**/*.ts"]
    customSyntax: postcss-styled-syntax
```
along with changing `yarn scss-lint` command to lint css,scss,ts,tsx (instead of just scss). There aren't too many, except in the `ol-ckeditor` ckeditor-provided stylesheet.


### Global / Vendor Stylesheets
**Suggestion:** Rather than moving postcss, etc, from our webpack config, I think we should keep the ability to include raw CSS stylesheets.
- Our own styles should be entirely in emotion
- But we may want to include vendor stylesheets, like `frontends/ol-ckeditor/assets/vendor/ckeditor_content_styles.scss`

Rather than moving `ckeditor_content_styles.scss` to runtime css-in-js, I think we should keep it as regular css. (It doesn't need to be scss.)
- easier to update
- not evaluated at runtime

What do you think?

frontends/mit-open/src/page-components/LearningResourceCardTemplate/LearningResourceCardTemplate.tsx
```diff
@@ -103,23 +95,162 @@ const LRCImage: React.FC<LRCImageProps> = ({
...
+        margin-top: 0;
+      }
+
+      &:last-child {
+        margin-bottom: 0;
+      }
+    }
+  }
+
+  .ol-lrc-chip.MuiChip-root {
```
ChristopherChudzicki, line 151: maybe this should just be a `SmallChip` component or something?

frontends/ol-components/src/components/TitledCarousel/TitledCarousel.tsx
```diff
@@ -48,13 +48,52 @@ const ButtonsContainer = styled.div`
...
+
+const StyledCarousel = styled(Carousel)({
+  paddingTop: theme.custom.shadowOverflowTop,
+  paddingBottom: theme.custom.shadowOverflowBottom,
+
+  /*
+    We want the carousel cards to be:
+      1. spaced
+      2. cast a shadow when hovered
+      3. left-aligned (left edge of left-most card aligned with rest of page content)
```
ChristopherChudzicki, line 61: very minor nit on what might be my writing 😛 

very minor nit on what might be my writing 😛 

```suggestion
    We want the carousel cards to:
      1. be spaced
      2. cast a shadow when hovered
      3. be left-aligned (left edge of left-most card aligned with rest of page content)
```

frontends/mit-open/src/common/style.ts
ChristopherChudzicki, line 1: Can you remove bootstrap from `mit-open`'s package.json, too?

(It was largely unused. It had been used in a few places, but I think they're all gone with recent removals of unused code.)

jonkafton:
> ### Global / Vendor Stylesheets
> **Suggestion:** Rather than moving postcss, etc, from our webpack config, I think we should keep the ability to include raw CSS stylesheets.
> 
> * Our own styles should be entirely in emotion
> * But we may want to include vendor stylesheets, like `frontends/ol-ckeditor/assets/vendor/ckeditor_content_styles.scss`
> 
> Rather than moving `ckeditor_content_styles.scss` to runtime css-in-js, I think we should keep it as regular css. (It doesn't need to be scss.)
> 
> * easier to update
> * not evaluated at runtime
> 
> What do you think?

How we treat the workspace dependencies could do with some attention in terms of whether they are independent modules to be built in isolation, or whether they are part of the app, and must then be built from source from the app root. 

We manage dependencies for each workspace in isolation (they have a package.json), though we're breaking this if the root build config needs to know about dependencies.

The problem of building from source is that we must then pollute our root build dependencies and config with items specific to a package, so are not fully benefiting from the workspace module isolation.

I agree that `ol-ckeditor` has separate build requirements and that it should be able to contain .scss files as with any other NPM dependency, though don't think we should encourage including raw CSS by supporting it in the main build setup.

For that to work we'd need to build `ol-ckeditor` first and import as a normal dependency with the JS dist entrypoint and link the CSS in the page HTML.

I've experimented with putting back raw CSS and separating the Webpack config on a branch, https://github.com/mitodl/mit-open/tree/jk/239-refactor-out-sass-ckeditor (temporarily adding an adjacent `ol-ckeditor-2`). Though stopped short of finishing off as was taking some time and need to see if we have buy-in.

Some issues with that:

- We need to add the `ol-ckeditor` CSS output to the base HTML page. We'd want to version the bundle filenames instead of hashing them and might then benefit from publishing the package to a private NPM repo (GitHub Packages?) when CI detects a change to `ol-ckeditor` files.

- `ol-ckeditor` imports from `ol-components` and `ol-utilities`. We would want to remove any dependencies from the project so that `ol-ckeditor` is a leaf dependency.

We do import Webpack config from `ol-ckeditor`, so perhaps the compromise is to formalize our approach for doing that for any future packages, though could lead to a brittle build process.

jonkafton:
> ### Syntax issues and linting
> I noticed a few invalid syntax issues. I noted a few of them, but there are more. IMO, we should set up `stylelint` to lint emotion styles, probably before we merge this.
> 
> I took a stab at setting it up, just to see how many violations we have right now:
> 
> ```yaml
> extends: stylelint-config-standard-scss
> rules:
>   selector-class-pattern:
>     - (^ck.*)|((^Mui[A-Z-])|^([a-z][a-z0-9]*)(-[a-z0-9]+)*$) # CKEditor, MUI or kebab-case
>     - message: "Expected class selector to be kebab-case"
> overrides:
>   - files: ["**/*.tsx", "**/*.ts"]
>     customSyntax: postcss-styled-syntax
> ```
> 
> along with changing `yarn scss-lint` command to lint css,scss,ts,tsx (instead of just scss). There aren't too many, except in the `ol-ckeditor` ckeditor-provided stylesheet.

Some notes/observations:

Emotion provides an [ESLint plugin](https://emotion.sh/docs/@emotion/eslint-plugin), though it is not linting the CSS itself, only providing some rules around imports and string vs. object syntax, https://github.com/emotion-js/emotion/tree/main/packages/eslint-plugin/docs/rules. Currently we don't have any strong preference that these rules would enforce.

The `postcss-styled-syntax` package for Stylelint is able to parse CSS-in-JS in .tsx files, but does not catch issues in template literals such as the `!default` hangover from Sass. Emotion [is using Stylis v4](https://emotion.sh/docs/emotion-11#stylis-v4) to process styles, though there are no vendor Stylelint plugins for Stylis.

Emotion [best practices](https://emotion.sh/docs/best-practices#use-typescript-and-object-styles) prefer object styles over template literals, so any maths in the CSS can be handled by JavaScript, though we have the issue that unitless values are interpreted as pixels, so for rems we need to do e.g. ``{ fontSize: `${1.25 * SMALL_FONT_SIZE}rem` }`` in place of the much nicer `{ fontSize: 1.25 * SMALL_FONT_SIZE }`. In any case with object styles remove the reliance on linting as they are type checked.

[postcss-styled-syntax](https://www.npmjs.com/package/postcss-styled-syntax) is not able to correctly parse function interpolations in template literal expressions in conjunction with the `stylelint-config-standard-scss` rules, so errors on e.g. `color: ${({ theme }) => theme.custom.colorBlue5};`. These rules are now targeting `.scss` files specifically.

The `selector-class-pattern` Stylelint rule gave errors on the `ol-ckeditor` styles only once ported to .tsx. I updated the regex to allow `__` in class names - not clear why it was passing before.

frontends/mit-open/src/page-components/LearningResourceCardTemplate/LearningResourceCardTemplate.tsx
jonkafton: Removed.

frontends/mit-open/src/page-components/LearningResourceCardTemplate/LearningResourceCardTemplate.tsx
jonkafton: Replaced with styled component.

frontends/mit-open/src/page-components/LearningResourceCardTemplate/LearningResourceCardTemplate.tsx
jonkafton: Yes, that was not going to work - fixed. Unfortunately unitless values in Emotion object styles default to pixels. Generally, fixed pixel border and margin styles are more likely to preserve a good layout if the user increases font scaling as where we increase container width and height relative to default font size, we introduce extra space and reduce the available space for text, also not accounting for wrapping.

.stylelintrc.yaml
```diff
@@ -1,5 +1,13 @@
-extends: stylelint-config-standard-scss
```
ChristopherChudzicki, line 1: > The postcss-styled-syntax package for Stylelint is able to parse CSS-in-JS in .tsx files, but does not catch issues in template literals such as the !default hangover from Sass. Emotion [is using Stylis v4](https://emotion.sh/docs/emotion-11#stylis-v4) to process styles, though there are no vendor Stylelint plugins for Stylis.

I think this depends on the ruleset enabled.

> [postcss-styled-syntax](https://www.npmjs.com/package/postcss-styled-syntax) is not able to correctly parse function interpolations in template literal expressions in conjunction with the stylelint-config-standard-scss rules, so errors on e.g. color: ${({ theme }) => theme.custom.colorBlue5};. These rules are now targeting .scss files specifically.

Good point. Looking at comments in the `postcss-styled-syntax` repo, I think the SCSS ruleset is just a poor fit for CSS-in-JS. But the `stylelint-config-standard` ruleset is, I think, a good fit. 

Using 
```yaml
extends:
  - stylelint-config-standard
```
stylelint seems to be catching emotion errors much better. Adding a `min-height: 55pxx;` and `display: flex !default;` to test,
```
➤ YN0000: [mit-open]:
➤ YN0000: [mit-open]: src/page-components/Header/Header.tsx
➤ YN0000: [mit-open]:  10:17  ✖  Unexpected unknown unit "pxx"             unit-no-unknown
➤ YN0000: [mit-open]:  11:17  ✖  Unexpected unknown annotation "!default"  annotation-no-unknown
➤ YN0000: [mit-open]:
➤ YN0000: [mit-open]: 2 problems (2 errors, 0 warnings)
➤ YN0000: [mit-open]:
➤ YN0000: [mit-open]: Process exited (exit code 2), completed in 3s 261ms
```

Which seems great!

**Suggestions:**
- Let's use `stylelint-config-standard` as the base
- We can add a `.stylelintignore` at root, and maybe ignore `/vendor/**/*.css` (or maybe `.vendor.css?`), and move the ckeditor file to there. I agree...Not sure why it wasn't erroring before. (Was it css? Maybe we only linted scss before.)
    -  might also want to ignore `build/` code. I had a build/ directory lying around from when I did a prod build a some point, and ended up with 1000 css errors...

> Emotion [best practices](https://emotion.sh/docs/best-practices#use-typescript-and-object-styles) prefer object styles over template literals...

I dunno, I find their reasoning unconvincing. They say you don't get intellisense or typechecking. But it seems to me that, when it comes to CSS, `stylelint` is much better than typechecking. And I do get intellisense...

Screenshot 2024-01-18 at 2 57 12 PM

ChristopherChudzicki: Re intellisense... it turns out I have https://marketplace.visualstudio.com/items?itemName=styled-components.vscode-styled-components installed, which is what's giving the highlighting and intellisense.

jonkafton: Agree with all - nice to catch that `!default`; pushed https://github.com/mitodl/mit-open/pull/390/commits/45dd27a1d

frontends/ol-ckeditor/src/components/CkeditorDisplay.tsx
```diff
@@ -68,12 +68,15 @@ const CkeditorDisplay: React.FC<{
...
   )
   return (
-    <div
-      className={classNames("ck-content", className)}
-      dangerouslySetInnerHTML={{
-        __html: html,
-      }}
-    />
+    <>
+      <Styles />
```
ChristopherChudzicki, line 72: The
```
  .ck.ck-balloon-panel.ck-powered-by-balloon,
  .ck.ck-balloon-panel_visible.ck-powered-by-balloon {
    display: none !important;
  }
```
styles should be loaded with the editors, not with the read-only display.

For loading the above CSS (which is currently in Styles.tsx), I think either
- a component that wraps CkeditorArticle and CkeditorMarkdown 
- or a `.css` file like the vendor one
    - honestly I think this would be OK. We don't need anything dynamic, it's already scoped to only affect ckeditor stuff, and that way it's loaded exactly once, as opposed to once per editor. (Although, probalby would only have one editor on screen at a time.)

ChristopherChudzicki (changes requested):
👍Everything looks good except the one issue mentioned in `frontends/ol-ckeditor/src/components/CkeditorDisplay.tsx`.

Linting works great now!

ChristopherChudzicki (approved):
👍

Issue #395, opened 2024-01-16
Move OpenAPI Schema to facilitate versioning


ChristopherChudzicki:
### Description/Context
Currently, the OpenAPI schema we generate with `drf-spetacular` is generated to project root. We plan to have separate specs for `v0/`, `v1/` and `v2/` APIs. To avoid cluttering project root, the specs should be generated to a subdirectory. 

### Plan/Design

- Change the generator output to a (new) `openapi/` directory
- Update CI files, etc, that rely on the spec.
- Possibly move drf-spectacular configuration code there, too.

PR #397, opened 2024-01-16
move openapi spec to subdir

ChristopherChudzicki:
#395 

This PR adds a new (small) django app `openapi` with our drf-spectacular settings, url conf, etc, , and changes the spec generation to output to this directory.

Part of the motivation for outputting the spec to a subdirectory is to avoid cluttering the project when we introduce multiple spec versions (e.g., spec_v1 for v1 apis, spec_v2 for v2 apis, etc). 

### How can this be tested?
1. make an edit to `openapi/openapi_spec.yaml` (e.g., a comment or change a word).
2. Run `docker comose exec web ./scripts/test/openapi_spec_check.sh`; it should exit with status code 1
3. Delete the generated frontend client; `rm -rf frontends/api/src/generated` 
3. run `./scripts/generate_openapi.sh` to regenerate the spec
4. run `git diff`—it should be empty, you've regenerated exactly what is committed
5. Run `docker compose exec web ./scripts/test/openapi_spec_check.sh`; it should exit with status code 0

Also, visit:http://localhost:8063/api/v1/schema/redoc and http://localhost:8063/api/v1/schema/swagger-ui ... the doc sites should still work.


rhysyngsun (commented):
Could we namespace the urls and yaml with the version, for example:
```
openapi/v1/openapi_spec.yaml
openapi/v1/urls.py
openapi/v1/urls_spectacular.py
```
It'll make it easier that scripts can walk the directories and the modules are a little cleaner as we add versions.

collinpreston (approved):
Tested and looks good to me.  Thank you for the help troubleshooting!

ChristopherChudzicki:
@rhysyngsun 👍 Re putting the generated spec and urls in v1/, etc, directories.

I also did a quick experiment integrating DRF versioning via NamespaceVersioning with DRF Spectacular... it seems to me that this might be a good way to avoid having multiple `urls_spectacular` files and just use a single 'main' urlconf for the app and for the docs. https://github.com/mitodl/mit-open/commit/7883e09a1855150f88988a634ca5fcc7ac54ce4c#r137298243

rhysyngsun (approved):
LGTM

Issue #396, opened 2024-01-16
Remove legacy djnago apps


ChristopherChudzicki:
### Description/Context
We should remove legacy django apps. They are clutter and get in the way (particularly `course_catalog`, the old version of our `learning_resources` app. 

### Plan/Design
Some apps to remove:
- [ ] `course_catalog`
- [ ] `interactions`
- [ ] `channels` (old reddit channels)
- [ ] `mail` (there's a more modern implementation in our `ol-django` utilities repo.)
- [ ] `profiles`

`channel_fields` and `widgets` should be kept—we've talked about using them more in the future.

PR #398, opened 2024-01-17
Remove unused django apps

ChristopherChudzicki:
#396 

This PR removes unused django apps:
- course_catalog
- interactions
- mail

### How can this be tested?
Tests should pass; new v1 APIs should still work.

### Additional Context
I left `profiles` in place for now. Do we want to get rid of that, too?

Note: The `channels` app is not legacy. It's our non-Reddit "field channels"


abeglova (approved):

mbertrand:
`interactions` was originally added to [track/display popular resources](https://github.com/mitodl/open-discussions/pull/2583) - if this functionality is still needed (question for Peter or Ferdi?), might want to refactor it to work with `learning_resources` models instead of `course_catalog` .  If not, delete away!

PR #409, opened 2024-01-23
Improve contentfile api query performance

### LINKED ISSUE #386: contentfiles/ endpoint is extraordinarily slow
### Description/Context
1. Visit https://mitopen-rc.odl.mit.edu/api/v1/contentfiles/ or `curl https://mitopen-rc.odl.mit.edu/api/v1/contentfiles/` for better timing
2. The API takes about 20 seconds to send back 10kb data

### Plan/Design (RFC)

The initial PR for fixing this issue made the following changes:
- Delete contentfiles for unpublished courses/runs (lowering the total count)
- Add a database index on the `created_on` fields of models inheriting from `TimestampedModel`, and order the contentfiles by `-created_on`

These changes helped a bit, but not enough (from ~20 seconds to ~10 seconds with default limit/offset values on RC with ~700K contentfiles).  Also, when the offset value is increased significantly (like `offset=500000`), the response time gets far worse.

The following additional changes may improve things further:

### Sort by `-id` instead of `-created_on`.    
In retrospect the indexing of the `created_on` field was probably redundant and could perhaps be removed.   Sorting by `id` instead of `created_on` should retrieve the results in the same order. In preliminary testing, queries that sorted by `-id` were much faster than queries sorted by `-created_on`, even though both are now indexed on RC.  

### Remove the `unpublished=True` filter.  

Since unpublished contentfiles are now being deleted, this is also redundant.  Only published contentfiles should be in the database.

### CursorPagination : The good the bad, and the ugly
We could switch to cursor pagination for the `/api/v1/contentfiles` endpoint, but continue using default limit/offset pagination for the nested `/api/v1/<resource_id>/contentfiles` endpoint.

Good: According to the documentation, cursor pagination is ideal for large querysets:  "With extremely large datasets pagination using offset-based pagination styles may become inefficient or unusable. Cursor based pagination schemes instead have fixed-time properties, and do not slow down as the dataset size increases."

Bad: Ordering should always be on an indexed field with unique (or almost unique), non-null, non-float values.  This should be okay because right now it is only ordered by `-created_on` and the proposed sorting by `-id` also meets this requirement.

Ugly:  It is not possible to have arbitrary offset values in urls to grab a certain range of results from anywhere within the total set of results, for instance `offset=1200&limit=10`.   There are only `next` and `previous` urls provided for any given cursor that the user/api-bot would need to use to navigate through the results.  Furthermore, the cursor-based url parameters are not nearly as user-friendly: `?cursor=cD0xMTExMjE%3D` vs `?offset=20&limit=10`.   

Screenshot 2024-01-25 at 1 25 31 PM

However, offset and limit parameters could still be used for the contentfiles nested under the learning_resouces API endpoint.

Screenshot 2024-01-25 at 3 02 00 PM

### Option B: Remove the endpoint altogether
Contentfiles can already be accessed via the `/api/v1/learning_resources/<resource_id>/contentfiles/` endpoint, which does not have performance problems because a much smaller number of contentfiles is returned per resource - if that is good enough, then the `/api/v1/contentfiles/` endpoint could just be removed.

mbertrand:
Closes #386

- Adds a database index for the `created_on` field of the abstract `TiimeStampedModel` (leading to migrations for all models inheriting from it).
- Modifies the `indexing_api.deindex_run_content_files` function to delete unpublished contentfiles for the run after they have been deindexed, since there is no longer any use for them.  If the run ever gets published again, they will be reimported.
- Changes the default sort field for `ContentFileViewSet` from `-updated_on` to `-created_on` (more comparable to the `new` viewset for learning resources).  Though it would probably be fine to just remove this sort altogether and add a `new` viewset if needed in the future.  I can make that change if desired.
- Fixes a bug in calling the `SearchIndexPlugin` for learning resource runs that I noticed during testing, and adds a bunch of tests for that function and other related ones.

I thought about adding an index to `updated_on` as well, ultimately decided against it for now since it is not currently used for sorting (after the change in this PR).  But willing to change my mind on that.

### How can this be tested?
You can run the following queries before and after applying the migrations.  After the migrations, sorting by `-created_on` should be significantly faster than sorting by `-updated_on`

```python
import time

start = time.time();print(ContentFile.objects.filter(published=True).order_by("-created_on")[:20].values_list("id", flat=True));end = time.time();print(end - start)

<TimestampedModelQuerySet [103249, 103248, 103247, 103246, 103245, 103244, 103243, 103242, 103241, 103240, 103239, 103238, 103237, 103236, 103235, 103234, 103233, 103232, 103231, 103230]>
0.03517913818359375

start = time.time();print(ContentFile.objects.filter(published=True).order_by("-updated_on")[:20].values_list("id", flat=True));end = time.time();print(end - start)

<TimestampedModelQuerySet [103249, 24855, 103248, 24853, 24852, 24851, 24850, 24849, 24848, 24847, 24846, 24845, 24844, 24843, 24842, 24841, 103247, 24839, 103246, 103245]>
0.5206596851348877

start = time.time();print(ContentFile.objects.filter(published=True)[:20].values_list("id", flat=True));end = time.time();print(end - start)

<TimestampedModelQuerySet [77994, 77995, 77996, 77997, 77998, 89129, 1856, 1859, 89126, 89127, 89128, 89130, 89131, 89132, 89133, 90226, 90227, 90228, 90229, 90230]>
0.0017976760864257812
```


learning_resources/utils.py
```diff
@@ -358,7 +358,7 @@ def resource_run_unpublished_actions(run: LearningResourceRun):
     """
     pm = get_plugin_manager()
     hook = pm.hook
-    hook.resource_unpublished(run=run)
+    hook.resource_run_unpublished(run=run)
```
mbertrand, line 361: This was a previously unnoticed bug, added tests for all these functions

learning_resources/views.py
```diff
@@ -520,7 +520,7 @@ class ContentFileViewSet(viewsets.ReadOnlyModelViewSet):
         ContentFile.objects.select_related("run")
         .prefetch_related("content_tags")
         .filter(published=True)
-        .order_by("-updated_on")
+        .order_by("-created_on")
```
mbertrand, line 523: `created_on` seems more relevant than `updated_on` and is used for the `new` viewset for learning resources too.

ChristopherChudzicki (approved):
👍 The sorting is definitely faster. Publishing change looks good.

PR #411, opened 2024-01-24
Use DRF NamespaceVersioning to manage OpenAPI api versions

### LINKED ISSUE #402: Set up API versioning for the OpenAPI Spec
### Description/Context
`drf-spectacular`, the tool we use to generate our OpenAPI specification, supports versioning APIs via DRF's [versioning classes](https://www.django-rest-framework.org/api-guide/versioning/). We currently do not use DRF versioning classes. Using them should allow us to more easily generate separate `v0`, `v1`, `v2`, etc OpenAPI specs. and doc sites.

**Currently**:
- all APIs in the OpenAPI spec are `v1` (*Aside: This probably isn't appropriate. E.g., the `/articles` APIs should be v0*).
- API version is indicated by URL, `/v1/courses` for the new API, vs `/v0/courses/` for the old catalog API (which has since been removed).
-  we have one URL conf file (`open_discussions/urls.py`) served by Django, and a separate, manually curated, URL Conf file (`openapi/v1/urls_spectacular`) used by drf-spectacular to generate the the OpenAPI specification for our APIs
     - In this pattern, we'd need to have additional, separate URL Confs for v0 and v2 APIs, if we wanted to generate OpenAPI specifications for them. 

### Plan/Design
- Generate OpenAPI spec for v0, v1, v2 APIs via a single URL Conf file, using  DRF's [versioning classes](https://www.django-rest-framework.org/api-guide/versioning/) and `drf-spectacular`'s `--api-version` flag.
- Separate doc sites for v0, v1, etc APIs (a parameter passed to drf-spectacular's documentation views)
- change the `articles` APIs to be v0

While working on #397, I did a quick proof-of-concept [experiment](https://github.com/mitodl/mit-open/commit/7883e09a1855150f88988a634ca5fcc7ac54ce4c#r137298243) with the above.

ChristopherChudzicki:
Closes #402 

This PR introduces DRF's [NamespaceVersioning](https://www.django-rest-framework.org/api-guide/versioning/#namespaceversioning) class to version our APIs. 

With the current settings in this PR:
- `drf-spectacular` now generates documentation from our main `open_discussions.urls` url configuration; two separate versions, `openapi/specs/v0.yml` and `openapi/specs/v1.yml`, are generated.
    - previously we maintained a separate list `openapi/v1/urls_spectaular.yaml` for the schema/docs, and would have had to maintain one such list per version
- in order to appear in our generated OpenAPI specifications, API endpoints **must** have a version provided by a url [namespace](https://docs.djangoproject.com/en/5.0/topics/http/urls/#url-namespaces)
- the version must be either `"v0"` or `"v1"` (we can add more, via `settings.REST_FRAMEWORK["ALLOWED_VERSIONS"]`.
    - I made learning resource + search endpoints v1, and other things that are somewhat experimental/to-be-developed-more (ckeditor, channels) v0. There's one exception (articles) with a note.

### How can this be tested?
1. Visit the rendered docs. (Check `swagger-ui`, too, if you want.)
    - The v0 docs: http://localhost:8063/api/v0/schema/redoc/
    - The v1 docs: http://localhost:8063/api/v1/schema/redoc/
2. Manually edit the schemas so they are out of sync:
    - Edit `openapi/schemas/v0.yaml` and/or `openapi/schemas/v1.yaml` ... Just change a description or something.
    - Run `docker compose exec web ./scripts/test/openapi_spec_check.sh` it should exit with a nonzero status code.
3. Regenerate the docs:
    - Run (outside docker) `./scripts/generate_openapi.sh`
    - Check `git status`; the v0 and v1 yaml files should now be un-modified (changes from previous step overriden).
4. You could try changing an endpoint's version or adding a new endpoint. For example, change the `learning_resources_search/urls.py` urlpatterms to

    ```python
    urlpatterns = [
        re_path(r"^api/v0/", include((v1_urls, "v0"))),
        re_path(r"^api/v1/", include((v1_urls, "v1"))),
    ]
    ```

     and now the search endpoints should be available under both `api/v0` and `api/v1` and appear in both documentations.

### Additional Context

**Request Versions:** With NamespaceVersioning, the version is also available on the request object via `request.version`. This can be used (for example) to swap serializer classes based on version. I don't necessarily expect us to use this functionality, though we could. Right now the request versioning is just for schema + doc generation.


articles/urls.py
```diff
@@ -5,13 +5,16 @@
...
     views.ArticleViewSet,
     basename="articles",
 )
 
+app_name = "articles"
 urlpatterns = [
-    re_path(r"^api/v1/", include(router.urls)),
+    # TODO(Chris Chudzicki): Change this to version v0 when # noqa: FIX002
+    #  https://github.com/mitodl/mit-open/issues/269 is finished
+    re_path(r"^api/v0/", include((v0_router.urls, "v1"))),
```
ChristopherChudzicki, line 19: This API is something we're using for internal experimentation with CKEditor right now. It is not exposed on the public site. IMO, it should be a v0 API. However, a staff-only portion of the frontend is using this API via generated client, and the client is generated only from the v1 specification. This will change when client generation moves to https://github.com/mitodl/open-api-clients. Rather than changing the existing, to-be-removed client generation in MIT Open, I think it makes sense to keep this as v1 for now, and move it to v0 once https://github.com/mitodl/open-api-clients is ready.

articles/urls.py
```diff
@@ -5,13 +5,16 @@
...
-router = SimpleRouter()
-router.register(
+v0_router = SimpleRouter()
+v0_router.register(
     r"articles",
     views.ArticleViewSet,
     basename="articles",
 )
 
+app_name = "articles"
```
ChristopherChudzicki, line 15: These URLs have two nested namespaces:
- `include((v0_router.urls, "v1")))` gives namespace `v1`
- when `articles.urls` is included in another URL conf (e.g., the main url conf `open_discussions/urls.py`) the URLs will get additional namspace from `app_name = 'articles'`.

The resulting URLs have names like `articles:v1:articles-list`.

**We could omit the `app_name` portion**. If we do everything works fine. But if `app1/urls.py` and `app2/urls.py` use the same namespaces and both urlconfs are included, then Django gives warnings like
```
WARNINGS:
?: (urls.W005) URL namespace 'v0' isn't unique. You may not be able to reverse all URLs in this namespace
```
The warning is just that—a warning. Everything is OK as long as the url names are unique. But to avoid the warning, I added the `app_name` namespaces.

**Pro:**
- no warning
**Con:**
- url names are more verbose

Happy to drop the `app_name` if people don't like it.

mbertrand (changes requested):
Overall looks great, testing instructions worked as expected.  Found a bug with editing article bodies via ckeditor (wrong api version in url), had some questions/comments about a few other things.  Needs a rebase too.

articles/urls.py
mbertrand: Some of this was supposed to apply to a different endpoint, sorry (ckeditor - in v0 schema but with v1 url)

ckeditor/urls.py
```diff
@@ -1,7 +1,10 @@
...
-from django.urls import re_path
+from django.urls import include, re_path
 
 from ckeditor.views import ckeditor_view
 
-urlpatterns = [re_path(r"api/v1/ckeditor/", ckeditor_view, name="ckeditor-token")]
+v0_urls = [re_path(r"ckeditor/", ckeditor_view, name="ckeditor-token")]
+
+app_name = "ckeditor"
+urlpatterns = [re_path(r"^api/v1/", include((v0_urls, "v0")))]
```
mbertrand, line 10: ckeditor is in the v0 schema docs but has a v1 endpoint, this seems confusing.

articles/urls.py
mbertrand: articles is in v1 schema but has a v0 url - confusing

.github/workflows/ci.yml
```diff
@@ -176,7 +176,7 @@ jobs:
     # This job checks that the output of openapi-generator-typescript-axios that
     # is checked into version control is up-to-date.
     env:
-      OPENAPI_SCHEMA: ./openapi/v1/openapi_spec.yaml
+      OPENAPI_SCHEMA: ./openapi/specs/v1.yaml
```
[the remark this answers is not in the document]
ChristopherChudzicki: That's correct. The reasoning was:
- We are planning to move client generation out of this repository and into https://github.com/mitodl/open-api-clients as part of https://github.com/mitodl/mit-open/issues/269
- https://github.com/mitodl/open-api-clients will be responsible for generating v0, v1, etc clients. 
- So let's not duplicate the work here.

But, as you've seen, that has some awkward consequences. Like leaving `/articles` as a v1 API. 

I'm going to take a quick stab at generating the v0 client here, too. Probably not hard.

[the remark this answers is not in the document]
mbertrand: OK, thanks, I get it now, don't need to bother with the v0 client then since it be in the open-api-client.

mbertrand (approved):
👍

PR #415, opened 2024-01-24
Add /api/v0/users/me API

### LINKED ISSUE #414: Add a /users/me API
### Description/Context

Currently we have an API for getting your user info that is parameterized on username, which third party clients may not know (e.g. OCW). To support these we need a `/users/me` API that returns the same response as `/users/<username>` for the current user only. Anonymous access is not allowed.

### Plan/Design

- Bringing in a version of this API (with necessary tweaking) from xPro or MITx Online is probably good enough for now

rhysyngsun:

Closes #414 

This adds the typical `users/me` API we have in our more recent apps. Rationale in described in the linked issue.

### How can this be tested?

Go to `/api/v0/users/me`
- If you're anonymous you should get a 403
- If you're logged in you should see your user data

### DIFF
--- a/profiles/urls.py
+++ b/profiles/urls.py
@@ -4,9 +4,9 @@ from django.urls import include, re_path
 from rest_framework.routers import DefaultRouter
 
 from profiles.views import (
+    CurrentUserRetrieveViewSet,
     ProfileViewSet,
     UserViewSet,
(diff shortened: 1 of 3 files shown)


gumaerc (approved):
:+1:

Issue #401, opened 2024-01-18
POCs for end-to-end testing frameworks


jonkafton:
### Description/Context

We're looking at introducing an E2E testing framework, tooling for local test development and CI to run the tests.

The goal is establish which framework best suits our purposes by running POCs on popular options and assessing viability.

### Rationale

The React app currently contains unit tests. These are great for quickly offering some confidence during CI, though do not describe the product feature set and often suffer from being overly simple or testing too small a unit, ie. unlikey to catch unintentional change to behavior. A passing unit test suite does not guarantee that the application works to the degree a comprehensive E2E/acceptance test suite might.

As unit tests live alongside their components they limit non functional code change (refactoring) as to change the component requires work on the unit test. We are more interested that the application functions and less so how it functions (black box testing). We should not export otherwise private functions for the unit tests as it is then not clear which are public to the app - the testable unit entrypoints are the exports.

The other “end” of the E2E test depends on the ease of deploying server side components in isolation in CI. The tests may depend on running components in a pre-live environment where this is not practical (e.g. Cloud native services). Tests will also often depend on specific data being present to produce the expected UI and should clean out any data added during the test run (unless covered by some retention policy). An option is for the tests to spin up an ephemeral database bootstrapped with any necessary data to both produce the initial state and avoid the extra overhead of cleaning up test data. We should also consider which tests run and when. For example we may want to run a subset of read only tests to sanity check new deployments to production environments.

The terms “E2E” and “acceptance” testing are often loosely defined though we can use E2E in that we are testing the browser app within the context of its supporting backend and integration points and acceptance in that we are functionally testing the application’s requirements driven feature set. We are writing functional tests for acceptance that run end to end. In placing backend services under test we encourage cross-functional feature based development. 

### Plan/Design

Popular frameworks include [Cypress](https://www.cypress.io/), [Playwright](https://playwright.dev/) and [Testing Library](https://testing-library.com/).

Requirements:

- E2E tests must emulate a user’s behavior by interfacing the browser’s UI plane.

- E2E tests should run quickly in a headless browser in a timeframe comparable to unit tests.

- Tests should be written in a BDD style that describes the product feature specification. Project management stories and issues can be referenced in the test suite.

- Evaluate the framework for ease of element selection and traversal. Do we need to augment our code with special selector handles for the test (avoided where possible) or can we select on what we are expecting to appear (e.g. querying by text content).

[...]

PR #419, opened 2024-01-25
Initial E2E test setup with Playwright

jonkafton:

[POCs for end-to-end testing frameworks #401](https://github.com/mitodl/mit-open/issues/401)

Provides an initial test suite pointing to RC and basic test spec for the homepage.

- Adds an `e2e_testing` directory containing Playwright config and a test spec.
- Lists the directory as a Yarn workspace.
- Excludes the directory from ESlint checking with the [Testing Library plugin](https://testing-library.com/docs/ecosystem-eslint-plugin-testing-library/).

### How can this be tested?

To run the tests in headless mode:

```bash
yarn --cwd e2e_testing test
```

To open the [UI Mode](https://playwright.dev/docs/test-ui-mode) workbench:

```bash
yarn --cwd e2e_testing start
```

No impact on existing build. CI will be configured for https://github.com/mitodl/mit-open/issues/418.


.eslintrc.js
```diff
@@ -83,6 +85,12 @@ module.exports = {
         ...restrictedImports(),
       },
     },
+    {
+      files: ["e2e_testing/**/*.ts"],
+      settings: {
+        "disable/plugins": ["testing-library"],
```
ChristopherChudzicki, line 91: **Suggestion:** Rather than using an extra plugin, we could just handle this by removing the testing-library stuff from top config, and adding that in overrides:
```js
overrides: [
    {
      files: ["./frontends/**/*.test.{ts,tsx}"],
      plugins: ["testing-library"],
      extends: ["plugin:testing-library/react"],
    },
]
```

e2e_testing/playwright.config.ts
```diff
@@ -0,0 +1,83 @@
+import { defineConfig, devices } from "@playwright/test"
+import "dotenv/config"
+
+// Read from default ".env" file.
+// dotenv.config()
```
ChristopherChudzicki, line 5: Can we remove this comment since we're using `import "dotenv/config" ?

ChristopherChudzicki (changes requested):
👍 Works well, with one minor issue below.

**Issue**: I see the .env file in github PR, but when I clone, I don't get the file.
1. Can you check this behavior? I don't think I'm crazy. (e.g., rename your local branch and checkout this one from github?)
    - probably has to do with `.env` being in our gitignore.
2. Assuming what I saw was correct, maybe we should handle this with an `.env.example` that can be copied rather than a `.env` that is committed? Unsure.

ChristopherChudzicki:
@jonkafton Also... I have used playwright on and off, and for visual debugging I've always used the `--debug` flag.

Somehow I never knew about `--ui`. That's awesome!

jonkafton:
> **Issue**: I see the .env file in github PR, but when I clone, I don't get the file.

That is odd. Perhaps some cached ignore file is missing the negation. I get the file on fresh clone of the branch and also with clone of main then checkout branch.

```bash
git clone -b jk/401-evaluate-e2e-playwright git@github.com:mitodl/mit-open.git
```

Can you also check the ignore rule (exit code 1 for files not ignored):

```bash
git check-ignore e2e_testing/.env; echo $?
```

ChristopherChudzicki (approved):
👍

Issue #417, opened 2024-01-24
Add CI check to alert about breaking changes in APIs


ChristopherChudzicki:
### Description/Context
Now that other services will be using MIT Open's APIs (examples: OER csv export, OCW soon) we would like CI to alert about breaking changes in our documented APIs.

### Plan/Design
Set up a github action with [`openapi-diff`](https://github.com/OpenAPITools/openapi-diff) that fails on breaking changes.

In the near future, I imagine engineers will override this check and click "Merge anyway". But having the check in place increases visibility to breaking changes.

PR #425, opened 2024-01-29
Add github action to check for breaking OpenAPI changes.

ChristopherChudzicki:
#417 

This PR adds a github action using https://github.com/Tufin/oasdiff to check for breaking changes in our OpenAPI specifications.

### How can this be tested?
1. See https://github.com/mitodl/mit-open/pull/422 for demonstration of action
    - changelog in comment https://github.com/mitodl/mit-open/pull/422#issuecomment-1912763999
    - breaking changes cause failure https://github.com/mitodl/mit-open/actions/runs/7700014735/job/20982877114#step:7:14
2. If you want to try out `oasdiff` locally, make edits to `openapi/specs/v1.yaml` and run
    ```sh
    docker run --rm -it -v $(pwd)/openapi:/openapi:ro tufin/oasdiff \
        changelog \
        https://raw.githubusercontent.com/mitodl/mit-open/main/openapi/specs/v1.yaml \
        /openapi/specs/v1.yaml
    ```

    (or v0). The CI check uses "composed" mode where we can diff globs, but I don't think there's a way to do that with remote specs, which is most convenient for experimentation.

### DIFF
--- a/.github/workflows/openapi-diff.yml
+++ b/.github/workflows/openapi-diff.yml
@@ -1,0 +1,62 @@
+name: CI
+on: [pull_request]
+jobs:
+  openapi-diff:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout HEAD
+        uses: actions/checkout@v4
+        with:
+          ref: ${{ github.head_ref }}
+          path: head
+      - name: Checkout BASE
+        uses: actions/checkout@v4
+        with:
+          ref: ${{ github.base_ref }}
+          path: base
+      - name: Run docker manually
+        id: oasdif_changelog
+        run: | # Capture changelog as a multiline output
+          echo "changelog<<EOF" > $GITHUB_OUTPUT
+          docker run --rm \
+            --workdir ${{ github.workspace }} \
+            --volume ${{ github.workspace }}:${{ github.workspace }}:rw \
+            -e GITHUB_WORKSPACE=${{ github.workspace }} \
+            tufin/oasdiff changelog --composed \
+            'base/openapi/specs/*.yaml' \
+            'head/openapi/specs/*.yaml' \
+            >> $GITHUB_OUTPUT
+          echo "EOF" >> $GITHUB_OUTPUT
+      - name: Get summary
+        id: oasdif_summary
+        run: |
+          echo summary=$(echo "${{ steps.oasdif_changelog.outputs.changelog }}" | head -1 ) \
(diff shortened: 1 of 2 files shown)


.github/workflows/openapi-diff.yml
```diff
@@ -0,0 +1,62 @@
...
+          echo summary=$(echo "${{ steps.oasdif_changelog.outputs.changelog }}" | head -1 ) \
+          >> $GITHUB_OUTPUT
+      - name: Post changes as comment
+        uses: thollander/actions-comment-pull-request@v2
+        with:
+          message: |
+            ## OpenAPI Changes
+
+            <details>
+            <summary>Show/hide ${{ steps.oasdif_summary.outputs.summary || 'No detectable change.' }}</summary>
```
ChristopherChudzicki, line 42: Some changes aren't reported by tufin/oasdiff. For example, changing a response body property from `nullable: true` to `nullable: false` seems not to be detected. (I'm not sure if that's intentional on their part.)

.github/workflows/openapi-diff.yml
```diff
@@ -0,0 +1,62 @@
...
+      - name: Checkout BASE
+        uses: actions/checkout@v4
+        with:
+          ref: ${{ github.base_ref }}
+          path: base
+      - name: Generate oasdiff changelog
+        id: oasdif_changelog
+        run: | # Capture changelog as a multiline output
+          echo "changelog<<EOF" > $GITHUB_OUTPUT
+          docker run --rm \
```
ChristopherChudzicki, line 21: tufin/oasdiff does have a github action: https://github.com/Tufin/oasdiff-action

However, it can't run in "composed" mode, which compares globs of specs

.github/workflows/openapi-diff.yml
rhysyngsun: Changing a response body property from `nullable: true` to `nullable: false` should be safe because `int`  matches the type `int | null`. It's on the list of safe changes in their example doc as well: https://github.com/Tufin/oasdiff/blob/main/BREAKING-CHANGES-EXAMPLES.md

rhysyngsun (approved):
LGTM

.github/workflows/openapi-diff.yml
ChristopherChudzicki: It didn't surprise me that it's not a breaking change. It surprises me it's not reported at all, e.g., undo the "info" category. But... Not a big deal.

Issue #426, opened 2024-01-29
Placeholder "dashboard" page


ChristopherChudzicki:
### User Story
We want a placeholder dashboard page so that the app looks a bit less empty.

### Acceptance Criteria

- [ ] Add a /dashboard route; should appear in user menu when logged in.

PR #428, opened 2024-01-30
Add a placeholder dashboard page

ChristopherChudzicki:
#426 

Adds a placeholder dashboard page.

Not much there:
Screenshot 2024-01-30 at 10 40 56 AM

### How can this be tested?
1. Navigate to http://localhost:8063/dashboard ... 
    - If you're logged in, you should see a placeholder.
    - if you're anonymous, you'll be redirected to login

### DIFF
--- a/frontends/mit-open/src/common/urls.ts
+++ b/frontends/mit-open/src/common/urls.ts
@@ -41,3 +41,5 @@ export const login = ({
   const next = `${pathname}${encodeURIComponent(search)}`
   return `${LOGIN}?next=${next}`
 }
+
+export const DASHBOARD = "/dashboard/"
--- a/frontends/mit-open/src/pages/DashboardPage/Dashboard.test.tsx
+++ b/frontends/mit-open/src/pages/DashboardPage/Dashboard.test.tsx
@@ -3,7 +3,7 @@ import { renderTestApp, screen, waitFor } from "../../test-utils"
 describe("DashboardPage", () => {
   test("Renders title", async () => {
     renderTestApp({
-      url: `/dashboard`,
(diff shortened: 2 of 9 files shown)


frontends/mit-open/src/common/urls.ts
```diff
@@ -41,3 +41,5 @@ export const login = ({
   const next = `${pathname}${encodeURIComponent(search)}`
   return `${LOGIN}?next=${next}`
 }
+
+export const DASHBOARD = "/dashboard/"
```
jonkafton, line 45: I feel path constants should not have a trailing slash so we can concatenate predictably, though notice the backend 301s where it's missing.

frontends/mit-open/src/pages/DashboardPage/Dashboard.test.tsx
```diff
@@ -0,0 +1,16 @@
+import { renderTestApp, screen, waitFor } from "../../test-utils"
+
+describe("DashboardPage", () => {
+  test("Renders title", async () => {
+    renderTestApp({
+      url: `/dashboard`,
```
jonkafton, line 6: Do we favor double quotes over template delimiters where there is no interpolation?

We can enforce with ESLint - `quotes: ["error", "double", { avoidEscape: true }]` (`avoidEscape` is needed if we want e.g. `'abc"def'` without fighting with Prettier). The base config [defers to prettier](https://github.com/mitodl/eslint-config/blob/master/index.js#L34C5-L34C68), though it doesn't seem Prettier can fix to double quote preference over backticks.

jonkafton (approved):

PR #430, opened 2024-01-30
Upgrade tika and disable OCR via headers

### LINKED ISSUE #393: tika times out on large documents
Tika times out when parsing large documents (see https://github.com/mitodl/mit-open/issues/392 for context) . Is this something that can be improved? The observed error is on legacy discussions because mitopen-rc gets ocw data from ocw-content-live-qa which does not have this course. The relevant code is the same in mit-open and legacy discussions though.

mbertrand:
Closes #393 

- Disables OCR via the request header (OCR is enabled on RC/production but not locally, and drastically increases processing time)
- Makes the OCR header setting configurable as a setting (in case we ever want to change it)
- Upgrades tika to the same version used in rc/prod
- Upgrades the tika python library

### How can this be tested?
Try the following under two sets of settings:

Set 1: 
```
TIKA_SERVER_ENDPOINT=http://tika:9998/
TIKA_ACCESS_TOKEN=
```

Set 2:
```
TIKA_SERVER_ENDPOINT=<same as on heroku rc>
TIKA_ACCESS_TOKEN=<same as on heroku rc>
```

Run the following:

```python
from learning_resources.etl.utils import extract_text_from_url
extract_text_from_url("https://ocw.mit.edu/courses/18-725-algebraic-geometry-fall-2015/ec341c7a2524e5dba7c3e939f322613a_MIT18_725F15_notes.pdf")
```

In both cases you should get a response back within a few seconds or less.


ChristopherChudzicki (approved):
Looks good to me. I tried setting to `ocr_only` when using RC values and...yeah, it's slow.

So this PR will make us stop using OCR on RC/prod, right? Unless we specifically decide to change that by setting `TIKA_OCR_STRATEGY` ?

mbertrand:
> Looks good to me. I tried setting to `ocr_only` when using RC values and...yeah, it's slow.
> 
> So this PR will make us stop using OCR on RC/prod, right? Unless we specifically decide to change that by setting `TIKA_OCR_STRATEGY` ?

Correct.  I don't think this will lead to much info being overlooked, the vast majority of PDF's don't seem to need OCR to extract text out of them and the faster processing time should lead to fewer timeouts.

Issue #434, opened 2024-01-31
Let Poetry use a virtualenv in docker container


ChristopherChudzicki:
### Description/Context
When we [originally](https://github.com/mitodl/open-discussions/pull/4151) set up Poetry in Open Discussions, we told Poetry not to use a virtualenv. In retrospect, I think this was a mistake:
- makes installing packages awkward (must install with `-u root`)
- Poetry maintainers indicate lack of virtual env causes problems https://github.com/python-poetry/poetry/issues/6398#issuecomment-1236430479

Additionally:
- OCW Studio and MITxOnline both use a virtualenv with poetry, so changing this would bring us in-line with our other projects

### Plan/Design
Follow approach outlined in https://github.com/mitodl/ocw-studio/blob/master/Dockerfile

PR #436, opened 2024-02-01
Poetry install to virtualenv

jonkafton:

[Let Poetry use a virtualenv in docker container **#434**](https://github.com/mitodl/mit-open/issues/434)

Updates the Docker build config for Poetry’s installation to create a virtual environment for the project.

Aligns the container build with [OCW Studio](https://github.com/mitodl/ocw-studio/blob/master/Dockerfile) and [MITxOnline](https://github.com/mitodl/mitxonline/blob/main/Dockerfile).

Note that we're setting `POETRY_VIRTUALENVS_CREATE` to the default true so it reads better (it is false in other projects). The assumption is that this has no impact as true creates if one does not already exist (we create before the poetry install with `python3 -m venv $VIRTUAL_ENV`).

We get the same output for containers created with either:

```
$ poetry env list
-- empty --

$ poetry env info

Virtualenv
Python:         3.11.5
Implementation: CPython
Path:           /opt/venv
Executable:     /opt/venv/bin/python
Valid:          True

System
Platform:   linux
OS:         posix
Python:     3.11.5
Path:       /usr/local
Executable: /usr/local/bin/python3.11
```

#### Locally:

- The container for the `web` service builds successfully
```bash
docker compose build web
```

- Services start and site loads
```bash
yarn run build
docker compose up
```
Navigate to http://localhost:8063

#### RC:

- This needs a post deployment check. Confirm successful deployment to RC post release or roll back.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -21,7 +21,7 @@ RUN mkdir /var/media && chown -R mitodl:mitodl /var/media
 ## Set some poetry config
 ENV  \
   POETRY_VERSION=1.5.1 \
-  POETRY_VIRTUALENVS_CREATE=false \
+  POETRY_VIRTUALENVS_CREATE=true \
   POETRY_CACHE_DIR='/tmp/cache/poetry' \
   POETRY_HOME='/home/mitodl/.local' \
   VIRTUAL_ENV="/opt/venv"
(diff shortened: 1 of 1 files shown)


mbertrand (approved):
LGTM! 👍

PR #435, opened 2024-01-31
Release 0.2.2

odlbot:
## Chris Chudzicki
  - [ ] Add a placeholder dashboard page (#428) ([263521c8](../commit/263521c88e189178e2f480675d45fc662e4dcc2d))
  - [ ] Check for breaking openapi changes on ci (#425) ([181db9fb](../commit/181db9fbb9d6868ffd1e457d9f69668941cbad07))
  - [ ] Use DRF NamespaceVersioning to manage OpenAPI api versions (#411) ([5a9d261b](../commit/5a9d261b2f6b9836b0d935730210f3e31828e60b))

## renovate[bot]
  - [ ] Update dependency faker to v22 (#378) ([b965a711](../commit/b965a71108c6ace9a3a211923f4c8066d75f0efa))
  - [ ] Update dependency jest-fail-on-console to v3 (#380) ([ded9de18](../commit/ded9de1849c0baa7691f9cd5d31e86ce0b155efa))

## Matt Bertrand
  - [ ] Save OCW contentfiles as absolute instead of relative (#424) ([3e7b1328](../commit/3e7b1328564d0989c3db56bb229ccec3ac69d159))

## Jon Kafton
  - [ ] Initial E2E test setup with Playwright (#419) ([0fb9978d](../commit/0fb9978d39f0e28eb2c2d0842cc0703f8baa3509))

PR #437, opened 2024-02-01
fix release: rename oasdiff workflow

ChristopherChudzicki:
Related to  #435 

Renames OpenAPI diff workflow so that when it fails it does not block release.

A failing workflow indicates breaking changes in the OpenAPI spec, but that should not block release. The workflow is meant more as a warning than a hard-block. 
- The failing changes might be desirable bugfixes in the specification, without actually changing the APIs
- Especially in the near future, we may decide to release breaking changes to APIs.

### How can this be tested?
Merge it and re-trigger doof.

### DIFF
--- a/.github/workflows/openapi-diff.yml
+++ b/.github/workflows/openapi-diff.yml
@@ -1,4 +1,4 @@
-name: CI
+name: OpenAPI Diff
 on: [pull_request]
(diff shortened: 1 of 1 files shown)


rhysyngsun (approved):
LGTM

PR #442, opened 2024-02-01
Fix webhook url

mbertrand:
N/A

Fixes the OCW webhook url and tests that it is what's expected.

### How can this be tested?
Post the following to http://localhost:8063/api/v1/ocw_next_webook/, you should  not get a 404:

```
{
    "webhook_key": "<value of settings.OCW_WEBHOOK_KEY>",
    "unpublished": true, 
    "site_uid": "0bc2ffc76cb64553822955c21c91db54","version": "live"
}
```


rhysyngsun (approved):
LGTM

PR #449, opened 2024-02-02
Multiple filter options for learningresources and contenfiles API rest endpoints

### LINKED ISSUE #293: Make it possible to select multiple options in the REST apis
The open search api endpoints allow users to select multiple values for a parameter, for example offered_by=xpro,ocw

Update the REST api filters to also be able to select more than one value for a parameter

mbertrand:
Closes #293 

Allows both the learning_resources and contentfiles API endpoints to filter by multiple values for the same attribute, via two different query parameter formats, similar to the search API endpoint.  Examples:

http://localhost:8063/api/v1/learning_resources/?platform=ocw,xpro
http://localhost:8063/api/v1/learning_resources/?platform=ocw&platform=xpro
http://localhost:8063/api/v1/learning_resources/?platform=ocw

Also gets rid of some contentfiles filters that are not present in the search api (run_readable_id, resource_readable_id)

### How can this be tested?
Try variations of the above urls for different filterable fields, on both the contentfiles and learning_resources API endpoints.  They should return expected results that match the same filters for the search API endpoints.

### DIFF
--- a/frontends/api/src/generated/api.ts
+++ b/frontends/api/src/generated/api.ts
@@ -4397,41 +4397,37 @@ export const ContentfilesApiAxiosParamCreator = function (
     /**
      * Viewset for ContentFiles
      * @summary List
-     * @param {number} learning_resource_id2 id of the parent learning resource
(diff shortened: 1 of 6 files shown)


ChristopherChudzicki (approved):
Works great 👍

PR #478, opened 2024-02-07
Avoid integrity errors when loading instructors

### LINKED ISSUE #477: IntegrityError for some `LearningResourceInstructor` objects when importing courses
Sometimes an `IntegrityError` is raised when loading instructors for a course.

### Expected Behavior
Course is added, existing instructors are updated, new instructors are added, no errors.

### Current Behavior
Integrity error on LearningResourceInstructor

### Steps to Reproduce
On RC, run `./manage.py backpopulate_ocw_data --course-name=15-356-how-to-develop-breakthrough-products-and-services-spring-2004`

### Possible Solution
Modify this line:
```
        instructor, _ = LearningResourceInstructor.objects.get_or_create(
            **instructor_data
        )
```
to something like:
```
        instructor, _ = LearningResourceInstructor.objects.get_or_create(
            full_name=instructor_data.pop("full_name"), 
            defaults=instructor_data
        )
```

### Additional Details
```
django.db.utils.IntegrityError: duplicate key value violates unique constraint "learning_resources_learningresourceinstructor_full_name_key"
DETAIL:  Key (full_name)=(Prof. Eric von Hippel) already exists.
```

mbertrand:
Closes #477 

Adjusts the `learning_resources.etl.loaders.load_instructors` function to avoid IntegrityErrors.

### How can this be tested?
- Create the following instructor on your local instance if it doesn't already exist:
  ```python
  {"full_name": "Prof. Eric von Hippel", "first_name": None, "last_name": None}
  ```
- On the **main** branch, run `./manage.py backpopulate_ocw_data --overwrite --course-name=15-356-how-to-develop-breakthrough-products-and-services-spring-2004`
- An `IntegrityError` should be raised.
- Switch to this branch, restart containers, and try again.
- There should not be an `IntegrityError`, and if you refresh the instructor created in the first step, the first and last names should now be populated.

### DIFF
--- a/learning_resources/etl/loaders.py
+++ b/learning_resources/etl/loaders.py
@@ -104,7 +104,7 @@ def load_instructors(
                 "title",
                 f"{prof.get('first_name', '')} {prof.get('last_name', '')}".strip(),
             )
-        instructor, _ = LearningResourceInstructor.objects.get_or_create(
+        instructor, _ = LearningResourceInstructor.objects.update_or_create(
             full_name=prof.pop("full_name"),
             defaults={key: value for key, value in prof.items() if value},
         )
--- a/learning_resources/etl/loaders.py
+++ b/learning_resources/etl/loaders.py
@@ -98,12 +98,15 @@ def load_instructors(
 ) -> list[LearningResourceInstructor]:
     """Load the instructors for a resource into the database"""
     instructors = []
-    for instructor_data in instructors_data:
-        if "full_name" not in instructor_data:
-            instructor_data["full_name"] = instructor_data.get("title", None)
-
+    for prof in instructors_data:
+        if "full_name" not in prof:
+            prof["full_name"] = prof.pop(
+                "title",
+                f"{prof.get('first_name', '')} {prof.get('last_name', '')}".strip(),
+            )
         instructor, _ = LearningResourceInstructor.objects.get_or_create(
-            **instructor_data
+            full_name=prof.pop("full_name"),
+            defaults={key: value for key, value in prof.items() if value},
         )
         instructors.append(instructor)
 
--- a/learning_resources/etl/loaders.py
+++ b/learning_resources/etl/loaders.py
@@ -94,9 +94,9 @@ def load_departments(
 
 
 def load_instructors(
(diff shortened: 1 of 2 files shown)


shanbady:
@mbertrand code changes look good in terms of converting the 'title' attribute to full_name and removing it from the update so it doesnt cause an integrity error but this code doesn't protect against the case where the endpoint returns any arbitrary mapping that is not one of the model fields.

for example - It seems I can trigger an integrity error should the instructor data contain a 'test' field ```{"full_name": "John Doe", "test": "test" }```

We may want to consider having a predefined set of fields that get mapped (preferably pulled directly from the model)

Here is what I had in mind:
```

# first do the title -> full name conversion up here

# get the fields that the instructor model actually has
instructor_fields = [field.name for field in LearningResourceInstructor._meta.get_fields()]

# map the corresponding fields from the data sent in
for field in instructor_fields:
    if field in instructor_data:
        mapped_data[field] = instructor_data['field']

# then proceed with the update_or_create
```

mbertrand:
By the time data gets to `load_instructors`, any extra fields should be cleaned out from earlier steps in the ETL process (for example, [here](https://github.com/mitodl/mit-open/blob/7b2296761b1ba59ed418d06acc37154e63cffce6/learning_resources/utils.py#L189-L247) and [here](https://github.com/mitodl/mit-open/blob/7b2296761b1ba59ed418d06acc37154e63cffce6/learning_resources/etl/openedx.py#L226-L231)).  But it's possible that something might slip through the cracks, so I just  updated the code to account for that.

mbertrand:
PS, this line:

```
instructor_fields = [field.name for field in LearningResourceInstructor._meta.get_fields()]
```

would include fields that shouldn't be updated from the input data even if present (like `id`, `created_on`, etc) so I used this instead:
```
valid_attributes = ["first_name", "last_name"]
```

which are the secondary attributes that can be updated.   `full_name` is not included because it is the unique key on which an instructor should be created/updated, and if missing/blank is calculated from the first and last name instead.

shanbady (approved):
Ok this is looking good. I think we are good to go.

Issue #418, opened 2024-01-25
CI for E2E Testing


jonkafton:
### Description/Context

We have set up an initial [Playwright](https://playwright.dev/) E2E test suite during [**POCs for end-to-end testing frameworks #401**](https://github.com/mitodl/mit-open/issues/401).

The goal is to run this on the Github Actions CI.

There are 2 modes of E2E testing to address - pre-deployment to verify that a commit is stable and meets its feature requirements and should continue to deploy; and post-deployment to verify that a release has been successful.

We currently manually verify deployments on RC before release to Prod, though as we gain confidence in the accuracy and thoroughness of our tests we might want to release without human intervention and certainly will want to use the test results as an additional release gate or rollback trigger.

During pre-deployment we'll want to run the full test suite. This will involve standing up the built front end and any backing services in CI.

Post release, we'll want to run a subset of the tests to provide an optimal level of confidence in relation to test duration and infrastructure costs.

On production we will want to test certain critical paths in addition to basic assertions that we are on the correct new version and pages load and render. This critical paths should be determined as part of the product requirements.

#### Data Handling

Each environment will have different requirements around data handling. Data setup and teardown adds a significant code overhead to the tests, so should be avoided where possible.

An option for pre-deployment runs is to write database bootstrap migrations that include the data expected by the tests. A the database is created and destroyed during the CI run, this removes the need for any setup or teardown code in the tests.

For RC we may choose to provision the environment with the test data, run a subset of tests that don't depend on any specific data or run tests that change the data in concert with a database retention or data cleansing policy.

For our subset of tests that sanity check Production are limited to tests that expect production data and are non-destructive and/or tests that don't make assertions that depend on specific data.

Our CI strategy around which tests run and when can be adjusted to suit our purposes as the product develops, but the test setup should support each mode of testing and will include mechanisms to run only tests tagged for a specific purpose, triggers to run tests post commit or PR merge and a policy for triggering tests manually post release or perhaps automatically if we can establish a feedback loop to CI from the deployment infrastructure readiness checks. The strategy should also describe which browsers and browser versions we test, perhaps smoke testing the happy paths on a wider range of browsers post release to RC - to be decided upon with respect to run cost.

Requirements:

On code push and PR merge:

- The full test suite should run on CI.
- The suite should run against a locally running web service and database.
- The database should contain migrations to insert test data in addition to the database schema produced by Django migrations.
- An html test report should be published and served.

On manual trigger:

[...]

PR #481, opened 2024-02-08
CI and test fixtures for E2E testing

jonkafton:

[CI for E2E Testing #418](https://github.com/mitodl/mit-open/issues/418)

- Container for E2E tests.
- Docker Compose config for starting minimal services required CI run.
- Script for applying test data, plus sample fixtures.
- New GitHub Actions pipeline job to run test suite.
- Test reports are published to GitHub Pages at https://mitodl.github.io/mit-open/playwright-report/.
- README notes.

### How can this be tested?

- Services under test start correctly with Docker Compose:
```bash
docker compose -f docker-compose-e2e-tests.yml up web db nginx
```

- Test fixtures can be applied to the db service:
```bash
./e2e_testing/scripts/apply-fixtures.sh
```

- Tests run against locally running services:
```bash
yarn --cwd e2e_testing test
```

- Tests can also run from within container over Docker network:
```bash
docker compose -f docker-compose-e2e-tests.yml up e2e-tests
```

- Single Docker Compose command can also spin up services and run tests:
```bash
docker compose -f docker-compose-e2e-tests.yml up --exit-code-from e2e-tests
```

- Playwright UI Mode starts:
```bash
yarn --cwd e2e_testing start
```

- Sanity tests run against RC:
```bash
yarn --cwd e2e_testing test:rc
```

- Sanity tests run against Production:
```bash
yarn --cwd e2e_testing test:production
```

- e2e-tests job paases on CI (last run https://github.com/mitodl/mit-open/actions/runs/7832938636)

- Test report is published to GitHub Pages ([publish-pages](https://github.com/mitodl/mit-open/actions/runs/7832938636/job/21373010986) job may need to be triggered until merged to main, as runs there will overwrite): https://mitodl.github.io/mit-open/playwright-report/

- Review [README notes](https://github.com/mitodl/mit-open/tree/jk/418-e2e-tests-ci/e2e_testing#readme) for coherence.


jonkafton:
> #### ⚠️ GitGuardian has uncovered 1 secret following the scan of your pull request.

Can we mark this to be ignored in the GitGuardian dashboard? It's just the throwaway test database password [here](https://github.com/mitodl/mit-open/commit/78bd02141f11b7cd8542a56e4cdd60f9f250261b#diff-480f4ea516b0a965b94d2c2d952901b23d3e142c99055999f2b1ee723568626bR9).

e2e_testing/scripts/apply-fixtures.sh
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env bash
+
+directory="e2e_testing/fixtures"
+
+for path in "$directory"/*; do
+	if [ -f "$path" ]; then
+		file=$(basename $path)
+		echo "Applying: $file"
+		docker compose -f docker-compose-e2e-tests.yml cp $path db:/$file
+		docker compose -f docker-compose-e2e-tests.yml exec -u postgres db psql postgres postgres -f /$file
```
mbertrand, line 10: If I run locally, it fails due to some of this data already existing.

It might be better to insert this data into a separate e2e-only database that is recreated on each run, if possible, so that it doesn't alter the main database.  For example the python unit tests run on a separate `test_postgres` database.

e2e_testing/fixtures/002_carousel.sql
```diff
@@ -0,0 +1,43 @@
...
+    "readable_id": "test-100000",
+    "title": "Test 100000",
+    "description": "Description",
+    "published": true,
+    "resource_type": "program",
+    "etl_source": "micromasters",
+    "professional": false,
+    "offered_by_id": "mitx"
+  }
+]'::jsonb) AS item(data);
```
mbertrand, line 43: Instead of sql, maybe the `loaddata` command could be used instead?  For example `manage.py loaddata platforms departments offered_by` upserts data stored in JSON files from `learning_resources/fixtures` folder.

mbertrand (commented):
Just a couple initial observations, will keep taking a look.

mbertrand (commented):
When running tests via `docker compose -f docker-compose-e2e-tests.yml up e2e-tests`, they all failed with the error message "Timed out 5000ms waiting for expect(locator).toBeVisible()" - maybe due to data not completely loading in prior step?

jonkafton:
> When running tests via `docker compose -f docker-compose-e2e-tests.yml up e2e-tests`, they all failed with the error message "Timed out 5000ms waiting for expect(locator).toBeVisible()" - maybe due to data not completely loading in prior step?

Yes, that would be it. The intention was for the Docker Compose config to be used while working on the tests - the developer would be applying the tests fixtures manually. On CI the tests run from outside Docker. With that in mind the `e2e-tests` container wasn't actually very useful - I agree we should be able to run the tests with a single Docker Compose run command, though this solution may be overkill.

I pushed [a change](https://github.com/mitodl/mit-open/commit/1c22e6d1af17c50b0dbaee7058b4b3a818f108e3) to apply the fixtures from within the e2e-test container.

- Installs Docker in the test container (we need Docker in Docker to run the commands to copy the SQL and apply to the database).
- Mounts the Unix socket to communicate with the Docker daemon from inside the test container (security issue if run unintentionally on production or anywhere other than an isolated test environment).
- Adds a `apply-fixtures-dind.sh` script to run as we start the e2e-test container. This differs from the original script as instead of using `docker compose` (the compose file lives outside its context) is uses `docker` directly and queries for the container ID to exec into it.

e2e_testing/scripts/apply-fixtures.sh
jonkafton: > If I run locally, it fails due to some of this data already existing.
> 
> It might be better to insert this data into a separate e2e-only database that is recreated on each run, if possible, so that it doesn't alter the main database. For example the python unit tests run on a separate `test_postgres` database.

I've added a named volume, `pgdata-e2e-tests` to the Docker Compose file so the tests will use their own database instance.

.pre-commit-config.yaml
```diff
@@ -67,6 +67,8 @@ repos:
           - yarn.lock
           - --exclude-files
           - ".*/generated/"
+          - --exclude-files
+          - ".env.ci"
```
ChristopherChudzicki, line 71: IMO it would be preferable to use
```
DATABASE_URL=postgres://postgres:postgres@db:5432/postgres # pragma: allowlist secret
```
within `.env.ci`.

e2e_testing/fixtures/002_carousel.sql
ChristopherChudzicki: Where we can, I think we should create test data via the APIs. Once the e2e tests have authentication worked out, that should work for things like learning paths or user lists. (Tests that are actually testing the UI creation would create via UI not via APIs.) But many of our core data models have read-only APIs since the data is ingested from other platforms—e.g., courses, programs, podcasts. 

> The idea is that the tests should have as little coupling as possible to any service and so was avoiding any dependency on Django. In practical terms though I'm not sure if there's much benefit - perhaps future scenario where we want to apply data on environments where the manage.py tool isn't available.

My 2c: Anywhere we have direct access to the DB for running SQL, then we probably have access to django / `./manage.py` also. Switching away from Django doesn't seem likely, and if we ever did, converting some data loading to another format seems like a small lift compared to everything else.

So I think we should use `loaddata` **if it is more convenient**. Seems to me that the primary advantage is we can store the data as just JSON rather than as a SQL query.

**If we keep the SQL:** Then @jonkafton a slightly more convenient form of the json insertion command is
```sql
INSERT INTO learning_resources_learningresource
SELECT *
FROM jsonb_populate_recordset(null::learning_resources_learningresource, '[
  -- obj1,
  -- obj2,
  -- etc
]')
```
The first argument to `json_populate_recordset` is an object specifying the shape of the data. https://www.postgresql.org/docs/current/functions-json.html

e2e_testing/README.md
```diff
@@ -1 +1,87 @@
...
+
+```bash
+docker compose -f docker-compose-e2e-tests.yml up --exit-code-from e2e-tests
+```
+
+## Data Handling
+
+Applying and tearing down data adds significant overhead to writing tests. This can be minimized by writing SQL alongside any tests that are expecting specific data, that can be applied to a fresh ephemeral database locally and in CI that is disposed of after a test run. This also guarantees that tests are dependent only on the codebase and not any data that happens to have been set during an instances lifetime on a hosted environment. We have the added benefit that we do not need to write teardown code to leave the database in its original state.
+
+These test fixtures can be written in plain SQL and any files in the adjacent [./fixtures](./fixtures) directory will be applied to the database whilst standing up services in CI. As a suggestion, SQL inserts can be written in [JSON format](https://www.postgresql.org/docs/9.6/functions-json.html) for readability.
```
ChristopherChudzicki, line 64: nit: If we keep postgres docs, i think we should link to current version...probably even if it isn't quite the version we use.
```suggestion
These test fixtures can be written in plain SQL and any files in the adjacent [./fixtures](./fixtures) directory will be applied to the database whilst standing up services in CI. As a suggestion, SQL inserts can be written in [JSON format](https://www.postgresql.org/docs/9.6/functions-json.html) for readability.
```

e2e_testing/README.md
```diff
@@ -1 +1,87 @@
...
+
+## Testing a User's Perspective
+
+A primary benefit of E2E Testing (also Acceptance Testing in this context) is that the tests have zero or very limited visibility of the internals of the system under test. This yields a focus on the user facing feature set irrespective of the code, facilitating refactor without test rewrite and producing a test suite styled around the product requirements and a test output that describes the product's specification.
+
+The tests then should interpret the rendered UI as would a user or an assistive technology. Typically source code would be annotated with HTML attributes for the tests to select on, or tests would make use of DOM XPaths or CSS selector paths. These approaches should be avoided as they incur a reliance on the implementation detail, can be brittle and do not reflect a user's view of the system. Playwright's best practices include a recommendation to [test user-visible behavior](https://playwright.dev/docs/best-practices#test-user-visible-behavior) and an element [Locator](https://playwright.dev/docs/locators) preference to select on accessibility attributes and textual content. Where practical, this approach should be followed.
```
ChristopherChudzicki, line 17: Nit: I wouldn't say typically. (My sense is "historically" might be accurate.)

IMO:
> The tests then should interpret the rendered UI as would a user or an assistive technology. Prefer [locating](https://playwright.dev/docs/locators) elements using accessibility attributes or text content and asserting about [their user-visible behavior](https://playwright.dev/docs/best-practices#test-user-visible-behavior). Where necessary, use `data-testid`, css selectors, or XPath. These latter approaches are more fragile since they rely on implementation details.

docker-compose-e2e-tests.yml
```diff
@@ -0,0 +1,62 @@
...
+      timeout: 10s
+      retries: 5
+
+  web:
+    build:
+      context: .
+    environment:
+      PORT: 8061
+    env_file: .env.ci
+    command: ./scripts/run-django-dev.sh
```
mbertrand, line 36: I think it might be possible to use a different database, recreate it on each run, and then populate it without relying on a nested docker container. 

Modify `.env.ci:
```
DATABASE_URL=postgres://postgres:postgres@db:5432/e2e_postgres. # different db name
PGPASSWORD=postgres
```

Then adjust the docker-compose-e2e-tests.yml file, adding a health check to postgres to ensure it's accepting connections, then running commands from the web container to recreate the e2e test database:

```yaml
version: "3.7"

services:
  db:
    image: postgres:12.17
    ports:
      - 5432:5432
    environment:
      - POSTGRES_PASSWORD=postgres
    healthcheck:
      test: ["CMD", "pg_isready", "-U", "postgres"]
      interval: 5s
      timeout: 5s
      retries: 2

  nginx:
    image: nginx:1.25.3
    ports:
      - 8063:8063
    links:
      - web
    volumes:
      - ./config/nginx.conf:/etc/nginx/conf.d/web.conf
    healthcheck:
      test: ["CMD", "curl", "-f", "http://localhost:8063"]
      interval: 5s
      timeout: 5s
      retries: 5

  web:
    build:
      context: .
    environment:
      PORT: 8061
    env_file: .env.ci
    command: >
      bash -c "dropdb --if-exists -h db -U postgres e2e_postgres && createdb -h db -U postgres e2e_postgres && ./scripts/run-django-dev.sh"
    stdin_open: true
    tty: true
    ports:
      - 8061:8061
    links:
      - db
    depends_on:
      db:
        condition: service_healthy

  e2e-tests:
    build:
      context: e2e_testing
    environment:
      - CI=true
      - BASE_URL=http://nginx:8063
    depends_on:
      nginx:
        condition: service_healthy
    links:
      - nginx
```

.pre-commit-config.yaml
jonkafton: Nice

e2e_testing/fixtures/002_carousel.sql
jonkafton: > Where we can, I think we should create test data via the APIs. Once the e2e tests have authentication worked out, that should work for things like learning paths or user lists. (Tests that are actually testing the UI creation would create via UI not via APIs.) But many of our core data models have read-only APIs since the data is ingested from other platforms—e.g., courses, programs, podcasts.

I've added a note here that aims to capture this: https://github.com/mitodl/mit-open/blob/jk/418-e2e-tests-ci/e2e_testing/README.md?plain=1#L73. We can certainly use the APIs and potentially use this suite to test APIs directly. We might want to be careful mixing API and UI tests in case we ever want to port to a dedicated integration test suite.

jonkafton: > So I think we should use loaddata if it is more convenient. Seems to me that the primary advantage is we can store the data as just JSON rather than as a SQL query.

I've switched to using Django fixtures in https://github.com/mitodl/mit-open/commit/d69d8038 and it does feel much cleaner. I was thinking of a scenario where we are testing a live environment, RC, and want to apply SQL directly, though I guess if we are allowed to access the DB, we are also allowed to access the web container.

e2e_testing/scripts/apply-fixtures.sh
jonkafton: Now applying via the manage.py tool, which doesn't have this issue.

docker-compose-e2e-tests.yml
```diff
@@ -0,0 +1,67 @@
...
+    build:
+      context: e2e_testing
+    environment:
+      - CI=true
+      - BASE_URL=http://nginx:8063
+    depends_on:
+      nginx:
+        condition: service_healthy
+    links:
+      - nginx
```
mbertrand, line 67: Also add a `volumes` section here for the same reason as above, plus commands to run yarn (instead of in the Dockerfile):

```yaml
    command: bash -c "yarn install && yarn test"
    volumes:
      - ./e2e_testing:/tests
```

mbertrand (commented):
Still getting timeout errors on the 3 tests

e2e_testing/fixtures/002_carousel.json
```diff
@@ -0,0 +1,28 @@
+[
```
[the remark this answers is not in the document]
mbertrand: never mind, I see it corresponds to the test files and there are no fixtures required for the first

docker-compose-e2e-tests.yml
```diff
@@ -0,0 +1,67 @@
...
+    depends_on:
+      db:
+        condition: service_healthy
+    command: > # pragma: allowlist secret
+      bash -c "export PGPASSWORD=postgres
+      && dropdb --if-exists -h db -U postgres e2e_postgres
+      && createdb -h db -U postgres e2e_postgres
+      && python3 manage.py collectstatic --noinput --clear
+      && python3 manage.py migrate --noinput
+      && python3 manage.py loaddata e2e_testing/fixtures/*.json
```
mbertrand, line 48: This works great, but one potential worry for the future is that because data for all tests are loaded together, every test must be compatible with the data loaded for all other tests.  With python unit tests, any required data is typically created at the beginning of the test and is then auto discarded when it completes.  Not sure if there's a good way to do this for E2E testing, and that could be an enhancement for the future instead of for this PR, just figured it was worth mentioning.

docker-compose-e2e-tests.yml
```diff
@@ -0,0 +1,67 @@
...
+      && uwsgi uwsgi.ini --honour-stdin"
+    stdin_open: true
+    tty: true
+    ports:
+      - 8061:8061
+    links:
+      - db
```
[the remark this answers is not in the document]
jonkafton: I'm not sure we want this, though I'm thinking in terms that it is desirable that the container doesn't change unless intentionally built.

An issue is that the front end needs to be built locally in production mode for the static assets to correctly load from a relative path when loaded from inside another container.

> Still getting timeout errors on the 3 tests

It could be that the issue you are seeing locally with the tests failing is that you have a local front end build pointing to `http://{MITOPEN_HOSTNAME}:{WEBPACK_PORT_MITOPEN}/`, so the HTML will try to load the JS bundle from e.g. od.odl.local, which won't exist on the Docker network.

On CI and locally, I need to first build the front end with `NODE_ENV=production`. Of course the `watch` service from the main `docker-compose.yml` mounts the root directory so it will overwrite the front end build with the development version when it runs. I've added a [note](https://github.com/mitodl/mit-open/blob/47d8e8875fcd0f3e45f7fed5c6d31431b4ddb7b8/e2e_testing/README.md?plain=1#L60) for this to the README.

This is only an issue if we want to run the e2e tests from inside their container, which we don't have to. We can start the main Docker Compose services normally and run the tests from local (though the web service would need to be using the `e2e_postgres` database):

```bash
yarn --cwd e2e_testing test
```

It would be convenient for the fixture files to be available to the container when changed on local, though I've instead provided a [script](https://github.com/mitodl/mit-open/blob/jk/418-e2e-tests-ci/e2e_testing/scripts/apply-fixtures.sh) that copies them over and applies them.

docker-compose-e2e-tests.yml
jonkafton: Useful to not have to build each time - changed.

docker-compose-e2e-tests.yml
jonkafton: Yes, that's something to be aware of. There's some trade off for the convenience of using fixture files and a fresh database each run, though it's good to encourage writing tests that can also run in parallel without race conditions. If we setup and teardown data within each test it can be hard to spot conflicts when we parallelize tests to cut down the running time.

e2e_testing/README.md
```diff
@@ -1 +1,95 @@
...
+
+```bash
+docker compose -f docker-compose-e2e-tests.yml up --exit-code-from e2e-tests
+```
+
+Note that to be loaded from within the e2e-tests container, the front end must first be built for production (in development mode Webpack builds the static assets URLs to `od.odl.local`, not available on the Docker network).
+
+```bash
+NODE_ENV=production yarn build
+```
```
mbertrand, line 64: Maybe add this for docker:

"
Or build with docker (using `NODE_ENV=production` in your .env file):
```bash
docker compose run --rm watch yarn run build 
```
"

ChristopherChudzicki (commented):
Left a very minor comment and a suggestion.

Going to hold off on approving till the tests pass.

**Next Up:** I don't think it needs to be handled in this PR, but I do think there is more that should be done for the e2e tests before you move on to other stuff. One thing that would be really helpful is an example of a test using an authenticated user.

jonkafton:
> By static assets you mean the the webpack bundle, right? That loads successfully in our normal development mode.

The issue comes in to play where we run the e2e tests from inside a container. In normal development we'll likely just run the on our host, but if we want to run a single Docker Compose command to spin up everything and run the tests containerized, we need a front end build that doesn't reference our default `MITOPEN_HOSTNAME=od.odl.local` (this isn't available on the Docker network DNS).

e2e_testing/README.md
jonkafton: I've added a `build-frontend` config, similar to the `watch` container, but exits after build instead of watching for changes. I haven't quite got it to work on CI with a single build-and-run command ([numerous attempts today](https://github.com/mitodl/mit-open/actions?query=actor%3Ajonkafton)), though it exists for anyone having issues with local `yarn` commands.

```bash
docker compose -f docker-compose-e2e-tests.yml run build-frontend
```

jonkafton:
> One thing that would be really helpful is an example of a test using an authenticated user.

Agree, will write up an issue for this.

mbertrand (commented):
LGTM!  👍

PR #483, opened 2024-02-08
Load fixtures by default in dev environment

### LINKED ISSUE #482: Load fixture files by default
### Description/Context

As a developer who is setting up my local environment for the first time, I got tripped up by an issue where running ```docker compose run --rm web python manage.py backpopulate_prolearn_data ``` would not load any resources even though my .env settings were correct. 

The issue ended being due to my initial fixtures not being loaded as described [here](https://github.com/mitodl/mit-open?tab=readme-ov-file#loading-data)

It would be nice to have this baked into our initial setup script and move the fixture loading instructions to the optional section of the readme.

### Plan/Design
1. Add ```python manage.py loaddata platforms departments offered_by``` to scripts/run-django-dev.sh
2. Update the readme and move the fixture loading step to the optional section.

shanbady:

Closes #482 

the fixtures that are loaded by running ```manage.py loaddata platforms departments offered_by``` sets up some default data that is needed by other commands such as ```manage.py backpopulate_ocw_data``` and ```manage.py backpopulate_prolearn_data```. This PR adds a change that runs this by default on development and makes fixture loading an optional step.

### How can this be tested?
1. checkout the main branch.
2. make sure your environment is setup to be able to pull course data from the prolearn catalog (set PROLEARN_CATALOG_API_URL=https://prolearn.mit.edu/graphql in .env)
3. bring up your environment via:
```
docker-compose down
docker-compose up
```
4. run ```docker compose run --rm web python manage.py backpopulate_prolearn_data``` and note that it will return with ```Populated 0 resources``` due to default objects not existing in the database
5. checkout this branch (shanbady/482-load-fixture-files-by-default) and repeat steps 2-4
6. note that the backpopulate_prolearn_data command populates properly

### Checklist:
- [ ] Verified this PR via the steps listed in testing section above.


mbertrand (approved):
One minor concern is that this will run every time the docker web container is started, even though it's only necessary to run once on a fresh database.   But it finishes very quickly and doesn't get in the way.

shanbady:
> One minor concern is that this will run every time the docker web container is started, even though it's only necessary to run once on a fresh database. But it finishes very quickly and doesn't get in the way.

on that note - I wasn't sure if its something to do with how i'm locally setup or if folks have an overridden config for the postgres service locally but every time i do a docker-compose down/up the postgres data is wiped. looking at the docker-compose it doesn't seem like the postgres service is mapped to a volume or data directory for persistence.

ChristopherChudzicki:
> on that note - I wasn't sure if its something to do with how i'm locally setup or if folks have an overridden config for the postgres service locally but every time i do a docker-compose down/up the postgres data is wiped. looking at the docker-compose it doesn't seem like the postgres service is mapped to a volume or data directory for persistence.

`docker compose down` both stops and removes the containers. So in addition to losing data, you'll also need to rebuild next to you start. *Which is probably fairly fast if the dockerfile hasn't changed, and will guarantee you get any new dependencies.*.

But I prefer `docker compose stop` (or `ctrl+c`) which just stops them. Then I don't need to rebuild, and the data isn't lost.

------

All that said, @jonkafton also ran into the down => losing data issue, which we could avoid by using a named volume. There's probably not a good reason to NOT use a named volume?

PR #510, opened 2024-02-12
Avoid using get_or_create for LearningResourceImage object that has no unique constraint

### LINKED ISSUE #487: Flaky test learning_resources.etl.loaders_test. test_load_program
### Expected Behavior
CI tests should consistently pass

### Current Behavior
This test had a recent test failure on CI:

> FAILED learning_resources/etl/loaders_test.py::test_load_program[False-True-False-True] - learning_resources.models.LearningResourceImage.MultipleObjectsReturned: get() returned more than one LearningResourceImage -- it returned 2!

### Steps to Reproduce
Run  `.manage.py pytest learning_resources/etl/loaders_test.py` multiple times, all tests should pass every time, particularly the `test_load_program` test.

shanbady:
Fixes #487 

The LearningResourceImage object has no uniqueness constraint in the model. In learning_resourcs/etl/loaders.py we load_image method which attempts to populate the 'image' attribute on a LearningResource by using get_or_create to retrieve an image which will fail if we have duplicate objects (more on this caveat [here](https://www.queworx.com/django/django-get_or_create/#:~:text=Uniqueness%20Constraint,the%20object%20and%20return%20it). We can avoid this by checking for existence and getting the object manually instead of using the get_or_create. 

### How can this be tested?
1. Checkout the main branch
2. Make sure you have some data (LearningResource objects available). If not run ```manage.py backpopulate_prolearn_data```
3. get into the django shell and run the following:
```
from learning_resources.etl.loaders import load_image
lr = LearningResource.objects.first()
load_image(lr, image_data={'url': 'http://baker.com'})
load_image(lr, image_data={'url':'http://baker.com', 'alt':None, 'description':None})
load_image(lr, image_data={'url':'http://baker.com', 'alt':None})
```
4. Note that it fails with ```MultipleObjectsReturned: get() returned more than one LearningResourceImage -- it returned 2!```
5. checkout this branch (shanbady/487-flaky-test)
6. re-run step 3 and note that it succeeds without errors

Additionally - to simulate flaky test runs, apply the following patch to the test_loaders file: 
[test.patch](https://github.com/mitodl/mit-open/files/14250846/test.patch) and run the loaders test via ``` pytest learning_resources/etl/loaders_test.py -vvv -k test_load_program -x --no-cov -s --pdb``` - it should not fail at all throughout its hundreds of runs (it will fail on main branch at some point).

### Additional Context
One other route to fixing this was to introduce a uniqueness constraint via data migrations - I avoided this route because I did not have enough context around the existing state of the data and it could introduce side effects elsewhere and also lead us to clean up existing data in some way that is incompatible with existing expectations. If that is something we want to pursue it might be work creating a separate ticket for that since there are likely many non-unique image url objects we would need to consolidate.

### Checklist:
- [ ] Testing steps detailed above have been followed and run successfully
- [ ] There are no other unwanted side effects in data quality/object creation that this could potentially introduce

### DIFF
--- a/learning_resources/etl/loaders.py
+++ b/learning_resources/etl/loaders.py
@@ -123,8 +123,10 @@ def load_instructors(
 def load_image(resource: LearningResource, image_data: dict) -> LearningResourceImage:
     """Load the image for a resource into the database"""
     if image_data:
-        image, _ = LearningResourceImage.objects.get_or_create(**image_data)
-
+        if LearningResourceImage.objects.filter(**image_data).exists():
+            image = LearningResourceImage.objects.filter(**image_data).first()
+        else:
+            image = LearningResourceImage.objects.create(**image_data)
         resource.image = image
     else:
         resource.image = None
--- a/learning_resources/etl/loaders.py
+++ b/learning_resources/etl/loaders.py
@@ -123,10 +123,11 @@ def load_instructors(
 def load_image(resource: LearningResource, image_data: dict) -> LearningResourceImage:
     """Load the image for a resource into the database"""
     if image_data:
-        if LearningResourceImage.objects.filter(**image_data).exists():
-            image = LearningResourceImage.objects.filter(**image_data).first()
-        else:
-            image = LearningResourceImage.objects.create(**image_data)
+        image, _ = LearningResourceImage.objects.get_or_create(
+            url=image_data.get("url"),
+            description=image_data.get("description"),
+            alt=image_data.get("alt"),
+        )
         resource.image = image
     else:
         resource.image = None
--- a/learning_resources/etl/loaders_test.py
+++ b/learning_resources/etl/loaders_test.py
@@ -23,6 +23,7 @@ from learning_resources.etl.loaders import (
     load_content_files,
     load_course,
     load_courses,
+    load_image,
     load_instructors,
     load_offered_by,
     load_podcast,
@@ -51,6 +52,7 @@ from learning_resources.models import (
     ContentFile,
     Course,
     LearningResource,
+    LearningResourceImage,
     LearningResourceOfferor,
     LearningResourceRun,
     PodcastEpisode,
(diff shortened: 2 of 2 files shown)


learning_resources/etl/loaders.py
```diff
@@ -123,8 +123,10 @@ def load_instructors(
 def load_image(resource: LearningResource, image_data: dict) -> LearningResourceImage:
     """Load the image for a resource into the database"""
     if image_data:
-        image, _ = LearningResourceImage.objects.get_or_create(**image_data)
-
+        if LearningResourceImage.objects.filter(**image_data).exists():
+            image = LearningResourceImage.objects.filter(**image_data).first()
+        else:
+            image = LearningResourceImage.objects.create(**image_data)
```
mbertrand, line 129: A potential drawback of this is that `LearningResourceImage.objects.filter(**image_data).first()` might return the wrong match.

For example an OCW website or different websites may have an `` element in multiple places with the same url but different descriptions and/or alt text (and sometimes either of those might conceivably be null).

So if these have previously been created in the database:

```json
{'url':'http://baker.com', "description": "d1", "alt": "a1"}
{'url':'http://baker.com', "description": None, "alt": "a1"}
{'url':'http://baker.com', "description": "d1", "alt": None}
```

Then what `load_image(lr, image_data={'url':'http://baker.com'})` should do is create a new `LearningResourceImage` with  the same url and `description=None` and `alt=None`.  The PR code will return one of the other images instead.

So explicitly specifying all the possible image fields in the `get_or_create` call might be the way to go:

```python
image, _ = LearningResourceImage.objects.get_or_create(
    url=image_data.get("url"),
    description=image_data.get("description"),
    alt=image_data.get("alt"),
)
```

We could also ensure that all the ETL transform functions for the different sources populate every field for the image data even if it will always be null.  For example, in `learning_resources.etl.openedx._transform_image`:

```python
def _transform_image(image_data: dict) -> dict:
    """Return the transformed image data if a url is provided"""
    if image_data and image_data.get("src"):
        return {
            "url": image_data.get("src"),
            "description": image_data.get("description"),
            "alt": None # no alt value is ever included in the openedx api response
        }
    return None
```    

But this _might_ be overkill.

mbertrand (changes requested):

learning_resources/etl/loaders.py
shanbady: Thats a good point. just pushed the change for using get_or_create with explicit fields

mbertrand (changes requested):
Looks good! Can you also create a new unit test to ensure that the change works as intended?  I just searched for an existing `loaders_test.test_load_image` function that could be modified to do this, but looks like we never got around to creating one before (oops).

mbertrand (commented):
Great test, just one minor suggestion

mbertrand (approved):
Looks great! 👍

PR #513, opened 2024-02-15
Add SCIM client

collinpreston:
https://github.com/mitodl/hq/issues/3472

### Description

1. Adds the django-scim library, and custom extensions. (https://github.com/15five/django-scim2).
- Allows creating users via SCIM API.  This should enable Keycloak to created users in MIT-Open.
2. Adds Django Oauth toolkit.
- Allows for obtaining an Oauth access token for an admin user in order to perform SCIM requests.  This should enable Keycloak to get access and perform requests to the MIT Open SCIM API.

### How can this be tested?
Pre-req: 
1. Use this branch.
2. Perform any required migrations.
3. Create a Django admin user.

Setup:
1. Create an Oauth application that matches the configuration in the screenshot below:
Screenshot 2024-02-15 at 4 00 47 PM

2. Copy the client ID and client secret values from step 1.
3. Retrieve an access token using Oauth.  I use Postman and my settings look like:
Screenshot 2024-02-15 at 4 04 19 PM

Test creating a user:
Use the access token returned from step 3, and perform the following request:
Screenshot 2024-02-15 at 4 07 21 PM

`curl --location 'http://open.ol.local:8063/scim/v2/Users' \
--header 'Content-Type: application/json' \
--header 'Authorization: Bearer <ACCESS_TOKEN_FROM_STEP_3>' \
--data-raw '{
    "schemas": [
        "urn:ietf:params:scim:schemas:core:2.0:User"
    ],
    "userName": "your_username",
    "name": {
        "givenName": "Your-first-name",
        "familyName": "Your-last-name"
    },
    "emails": [
        {
            "primary": true,
            "value": "your-email@mit.edu"
        }
    ],
    "externalId": "some-fake-external-id"

}'`

Verify that a User and Profile record were created in MIT-Open using the information in your SCIM request.

### Additional Context
The django-scim library is pretty limited when adding it to an existing Django application.  It expects that you are able to modify the User model in order to add custom attributes that are required for the django-scim library to function.  Because of this restriction, I needed to create a custom adapter and add custom attributes to the Profile model in MIT-Open.

After this PR is approved, a SCIM configuration will need to setup in Keycloak in order for users to be created in MIT-Open.


shaidar:
Does this mean that if we wanted to propagate groups, we'd have to modify additional models?

blarghmatey (approved):
This all looks reasonable to me.

PR #514, opened 2024-02-15
Assign best date available to LearningResourceRun.start_date field

### LINKED ISSUE #512: start_date is missing from most (all?) ocw courses
https://ocw.mit.edu/search/?s=-runs.best_start_date&type=course
and 
https://live-qa.ocw.mit.edu/new-search/search/?e=resource&s=-start_date

look different - look into why that is

mbertrand:
Closes #512 

- For OCW courses, assigns a value to `LearningResourceRun.start_date` based on semester/year if present.
- For other ETL sources, if a start date value from source data is null, will substitute  the enrollment start date instead as the value.
- Modifies the old `semester_year_to_date` function a bit in terms of returned values, removing unused `ending` kwarg.

### How can this be tested?
Run `backpopulate_ocw_data --skip-contentfiles`.  It will take ~20-30 minutes but when done nearly all OCW resources should have runs with a populated start date.
Run `backpopulate` commands for other sources too (xpro, mitxonline, oll, etc).  The courses should be imported/updated and most should have runs with start dates.


learning_resources/etl/ocw.py
```diff
@@ -231,13 +232,15 @@ def transform_contentfile(
 def transform_run(course_data: dict) -> dict:
     """Convert ocw course data into a dict for a run"""
     image_src = course_data.get("image_src")
+    semester = course_data.get("term") or None
+    year = course_data.get("year") or None
```
shanbady, line 236: We can eliminate the "or None" here.

learning_resources/utils.py
```diff
@@ -89,30 +89,27 @@ def get_year_and_semester(course_run):
...
 
     Returns:
         datetime: The rough date of the course
     """
-    if semester is None or year is None:
+    if year is None:
         return None
-    if semester.lower() == "fall":
-        month_day = "12-31" if ending else "09-01"
+    if semester is None or semester in ("spring", "january_iap"):
```
shanbady, line 105: dont have enough context around the "january_iap" key but is it expected to have it as snake_cased here as opposed to [here](https://github.com/mitodl/mit-open/pull/514/files#diff-e84b4fc23d9aeac185e204b5f68f024d204b89546e680116c0f8c420d3eccb39R61)? My guess is that the referenced testcase with "January iap" used to excercise this conditional whereas now the test passes only because the default case of ```month_day = "01-01"``` is the same as the "january_iap" case

shanbady (changes requested):

mbertrand:
@shanbady ready for another look

shanbady (approved):
Looks good!

PR #535, opened 2024-02-21
Save user with is_active from SCIM request

collinpreston:
https://github.com/mitodl/mit-open/pull/513

Resolves an issue that occurs when a SCIM POST request, to the /users/ endpoint, contains the `active` attribute.  The issue is caused by not creating a User object, associated with the Profile object used for SCIM, before accessing the User model's `is_active` attribute.

### How can this be tested?
SCIM POST request to `http://open.ol.local:8063/scim/v2/Users` with the body of:
```
{
    "schemas": [
        "urn:ietf:params:scim:schemas:core:2.0:User"
    ],
    "userName": "collinpreston",
    "name": {
        "givenName": "Collin",
        "familyName": "Preston"
    },
    "emails": [
        {
            "primary": true,
            "value": "collinp@mit.edu"
        }
    ],
    "externalId": "keycloakID",
    "active": true

}
```

### Additional Context
I caught this while testing the integration with Keycloak.  Previously, while testing locally, I did not include the `active` attribute in my SCIM requests.

### DIFF
--- a/profiles/adapters.py
+++ b/profiles/adapters.py
@@ -157,7 +157,6 @@ class SCIMProfile(SCIMUser):
                 self.obj.user.username = self.obj.email
                 self.obj.user.first_name = self.obj.first_name
                 self.obj.user.last_name = self.obj.last_name
-                self.obj.user.is_active = self.obj.user.is_active
                 self.obj.user.save()
                 self.obj.name = self.display_name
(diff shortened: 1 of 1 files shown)


rhysyngsun (approved):
LGTM on basis of code review only.

PR #538, opened 2024-02-23
E2E Test and fixture for login

jonkafton:

https://github.com/mitodl/mit-open/issues/536

- Adds an E2E fixture that logs a user in via OIDC.
- Adds a test spec to log in and check authenticated and unauthenticated page access.

### How can this be tested?

- Tests should be pass locally against RC:

```bash
yarn workspace e2e_testing test:rc
```

- e2e-test CI job should run successfully (awaiting setup, see below).

### Additional Context

Putting up early for review. The linked issue assumes Keycloak is not configured for local login, but seems we do have setup for od.odl.local:8063 ([detail](https://docs.google.com/document/d/17tJ-C2EwWoSpJWZKjuhMVgsqGtyPH0IN9KakXvSKU0M/edit#heading=h.owvgv2h3odjp)).

We need to set up a test user on the CI OIDC before tests will pass on CI.

Do we have any test accounts set up? Ideally we'll use a test email address that we can set up in RC and Prod.

### DIFF
--- a/e2e_testing/.env
+++ b/e2e_testing/.env
@@ -2,3 +2,6 @@ BASE_URL=http://localhost:8063/
 # BASE_URL=http://nginx:8063/
 # BASE_URL=https://mit-open-rc.odl.mit.edu/
 # BASE_URL=https://mitopen.odl.mit.edu/
+
+TEST_USER_EMAIL=jonkafton+test1@gmail.com
+TEST_USER_PASSWORD=3edc5tgb # pragma: allowlist secret
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -220,12 +220,6 @@ jobs:
       - name: Install dependencies
         run: yarn install --immutable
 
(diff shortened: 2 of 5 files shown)


e2e_testing/.env
```diff
@@ -2,3 +2,6 @@ BASE_URL=http://localhost:8063/
 # BASE_URL=http://nginx:8063/
 # BASE_URL=https://mit-open-rc.odl.mit.edu/
 # BASE_URL=https://mitopen.odl.mit.edu/
+
+TEST_USER_EMAIL=jonkafton+test1@gmail.com
```
jonkafton, line 6: Do we have any test email addresses that we can use? or just a non existing address @mit.edu that we can set as active in the databases.

jonkafton:
Closing this one off. After a discussion with @ChristopherChudzicki we decided to hold off on E2E testing given reasonable emulation in the unit tests and until we establish guidelines for whether or when to write them.

PR #550, opened 2024-02-26
next/previous links for search api

### LINKED ISSUE #400: Add { next, previous } to search API response
### Description/Context
Our APIs that are paginated via DRF provide `next` and previous URL parameters. For example, https://mitopen.odl.mit.edu/api/v1/courses/?platform=ocw&offset=80&limit=20 returns 
```json5
{
    "count": 2496,
    "next": "https://mitopen.odl.mit.edu/api/v1/courses/?limit=20&offset=100&platform=ocw",
    "previous": "https://mitopen.odl.mit.edu/api/v1/courses/?limit=20&offset=60&platform=ocw",
    "results": [/* results */]
}
```
We should add the `{ next, previous }` urls to the search API endpoints, too. They are convenient when dealing with pagination. 

### Plan/Design
Probably can be added in search response serializers.

shanbady:
Closes #400 

This PR adds { next, previous } urls to the open-search based endpoints /api/v1/learning_resources_search and /api/v1/content_file_search

### How can this be tested?
1. checkout this branch
2. make sure you have some learning resources indexed in opensearch
3. login as a user /admin
4. visit /api/v1/learning_resources_search/ and note the next and prvious links in the response json


learning_resources_search/serializers.py
```diff
@@ -372,9 +373,33 @@ class SearchResponseMetadata(TypedDict):
...
+        if request:
+            url = request.build_absolute_uri()
+            total_record_count = self.get_count(instance)
+            offset = int(request.query_params.get("offset", 0))
+            limit = int(request.query_params.get("limit", 20))
+            if link_type == "previous":
+                offset -= limit
+            else:
+                offset += limit
+            if offset >= 0 and offset <= total_record_count:
```
mbertrand, line 391: Maybe make this `< total_record_count` instead of `<= total_record_count`?  I just happened to use a [query](http://localhost:8063/api/v1/learning_resources_search/?q=algebra) resulting in exactly 80 results, and while advancing through the next links, the last one (`http://localhost:8063/api/v1/learning_resources_search/?offset=80&q=algebra`) returned 0 results:

Screenshot 2024-02-28 at 10 12 17 AM

learning_resources_search/serializers.py
```diff
@@ -372,9 +373,33 @@ class SearchResponseMetadata(TypedDict):
...
+    previous = serializers.SerializerMethodField()
     results = serializers.SerializerMethodField()
     metadata = serializers.SerializerMethodField()
 
+    def construct_pagination_url(self, instance, request, link_type="next"):
+        if request:
+            url = request.build_absolute_uri()
+            total_record_count = self.get_count(instance)
+            offset = int(request.query_params.get("offset", 0))
+            limit = int(request.query_params.get("limit", 20))
```
mbertrand, line 386: The [default limit seems to be 10](https://opensearch.org/docs/latest/search-plugins/search-relevance/compare-search-results/#changing-the-number-of-results), that is the number of results returned in the main branch without any particular limit specified.  Maybe this could be a setting?  Looks like one already exists (`OPENSEARCH_DEFAULT_PAGE_SIZE`) but is set to 6 by default so apparently it's not used directly in the backend code, just passed to the front end.  Could use this one or maybe add a separate setting, might be worth checking with Chris about whether the default really needs to be 6 or if it could be 10 instead.

```python
limit = int(request.query_params.get("limit", settings.OPENSEARCH_DEFAULT_PAGE_SIZE))
```

If the front end needs to be a different value (or we just want to keep the option of making it different), then could add a new setting:
```python
limit = int(request.query_params.get("limit", settings.OPENSEARCH_DEFAULT_LIMIT))
```
where the default value would be `10`

Also, this limit, if differing from the opensearch default of 10, needs to be included as a request parameters in all urls as well.  The url http://localhost:8063/api/v1/learning_resources_search/ returns 10 results but the `next` link has `limit=20` so 10 results would be skipped over when advancing to that url.  Ditto for each subsequent url, unless `&limit=20` is added as a parameter.

mbertrand (changes requested):

learning_resources_search/serializers.py
ChristopherChudzicki: IMO, the frontend should just always send the pagesize it wants. So the default doesn't really matter—10 would be fine. 

> IMO, the frontend should just always send the pagesize it wants

Because:
- if we use a paginated UI (with separate buttons for page 1, page 2, page 3, etc, as in these [INITIAL](https://www.figma.com/proto/bRb19HaucO5loHGSU81nFG/MIT_Prototype?page-id=0%3A1&type=design&node-id=9-6102&viewport=4023%2C1145%2C0.25&t=6VyDzTtHVzsZ3Nyv-1&scaling=scale-down&starting-point-node-id=48%3A2437&show-proto-sidebar=1) designs, then someone might click on page 10 w/o having visited other pages, so we really need to know the page size up frontend so the frontend can determine limit / offset. Note: In this case, the `{ next, previous }` parameters aren't really helpful, unfortunately.
    - Even if we go this route and don't use `{ next, previous }` on the frontend, I think having them in the API is great for consistency with our other paginated APIs.
- If we use an infinite scroll UI, the frontend _could_ skip sending the page size, but I don't think there's really any disadvantage to sending it. And we might well have a "more compact" / "less compact" design (OCW offers this option, see https://ocw.mit.edu/search/?u=compact), in which case the frontend would have multiple page sizes. 
    - for infinite scroll, having `{ next, previous }` is definitely useful.

learning_resources_search/serializers.py
shanbady: Good catch. fixed.

learning_resources_search/serializers.py
```diff
@@ -372,9 +373,33 @@ class SearchResponseMetadata(TypedDict):
...
+                offset += limit
+            if offset >= 0 and offset <= total_record_count:
+                return replace_query_param(url, "offset", offset)
+        return None
+
+    def get_next(self, instance) -> int:
+        request = self.context.get("request")
+        return self.construct_pagination_url(instance, request, link_type="next")
+
+    def get_previous(self, instance) -> int:
```
ChristopherChudzicki, line 402:
```suggestion
    def get_previous(self, instance) -> str:
```
and similarly in `get_next`. 

Note: In addition to type hints in editors, this affects the schema that [drf-spectacular](https://github.com/tfranzel/drf-spectacular) generates

learning_resources_search/serializers.py
shanbady: Good to know we have that flexibility. I pushed a change to make the number in settings.py line up with the default opensearch page size of 10 as well as always explicitly including the 'limit' parameter in next/prev links since in the opensearch-dsl library  ['10' is kind of a magic number](https://github.com/opensearch-project/opensearch-dsl-py/blob/d570c6dbd48999ee2af6dfc8fb2c884e820c212c/opensearch_dsl/search.py#L378) at the moment.

learning_resources_search/serializers_test.py
```diff
@@ -716,8 +739,11 @@ def test_learning_resources_search_request_serializer_invalid(parameter, value):
         (response_test_raw_data_2, response_test_response_2),
     ],
 )
-def test_learning_resources_search_response_serializer(settings, raw_data, response):
+def test_learning_resources_search_response_serializer(
+    settings, raw_data, response, learning_resources_search_view
+):
```
mbertrand, line 744: Add a docstring.  Thought ruff was supposed to catch this, I might look into that later.

learning_resources_search/serializers_test.py
```diff
@@ -20,6 +25,18 @@
...
 
 
+@pytest.fixture()
+def learning_resources_search_view():
+    """Fixture with relevant properties for testing the search view"""
+    return SimpleNamespace(url=reverse("lr_search:v1:learning_resources_search"))
+
+
+def get_request_object(url):
+    request_factory = APIRequestFactory()
```
mbertrand, line 35: Add docstring

mbertrand (commented):
Works great! Just noticed that some docstrings were missing.

mbertrand (approved):
👍

PR #586, opened 2024-03-06
Shanbady/shareable program letter view

shanbady:

Closes https://github.com/mitodl/mit-open/issues/555

This PR implements a shareable program letter view for users who have earned a program certificate in micromasters.
The the flow for this is:
1. User receives program certificate in micromasters upon which it is recorded in MM database
2. MM program certificate table is pulled into data lake ultimately ends up in program certificate data mart
3. Data mart is synced directly to a ProgramCertificates table in MIT-Open
4. User visits dashboard in Open and see's link to view program letter if they have an entry in ProgramCertificates for the program dashboard they are looking at.
5. that link points to an "intercept view" which:
    - generates a "program letter" object which will have a uuid that is used in the shareable url
    - makes an api call to the micromasters cms to fetch the template content for the program letter
    - forwards the user to the program letter view which renders the letter and is shareable

### How can this be tested?
1. in your .env file, set MICROMASTERS_CMS_API_URL=https://micromasters-rc.odl.mit.edu/api/v0/wagtail/
2. get into a django shell via ```docker-compose exec web python manage.py  shell```
3. We are going to create 2 program certificates for your user. On production this would exist as a result of the data sync but we will simulate these records here. Make sure you replace the email here with whatever user you will be logging in with:
<pre><code>
</code></pre>
4. login as the user with email from above.
5. visit http://localhost:8063/program_letter/1/ and http://localhost:8063/program_letter/18/ - you should be redirected to shareable (viewable in incognito mode while logged out) program letter views. The urls should have uuids in them.


shanbady:
@mbertrand I'm closing out this branch and will be following up with a version that uses a react based template instead

PR #603, opened 2024-03-11
Avoid duplicate courses

### LINKED ISSUE #473: Prevent duplicate courses from being created if readable_id changes
Apparently it's possible for users to change the primary_course_number of an ocw course which creates a duplicate record for that course
https://github.com/mitodl/hq/issues/3499

Look into the best way to fix this. I think one option is to use the url to match to existing courses instead of the readable id (what primary_course_number is saved as) for ocw. Another is to send the old primary_course_number in the webhook request when the primary_course_number changes in studio.

mbertrand:
Closes #473 

- Adjusts the `load_course` ETL function to check for an optional data param ("unique_field") that if different from the default `readable_id`, unpublishes all other course resources with the same field value and platform.
- Adjusts the prolearn readable_id value to be more than just a number
- Changes the url used by prolearn (ignores the "course_link" field if there is no path in it).

### How can this be tested?
- If you haven't loaded prolearn courses already, do so on the main branch (set `PROLEARN_CATALOG_API_URL=https://prolearn.mit.edu/graphql` and then run `./manage.py backpopulate_prolearn_data`).
- Via django admin, check how many published courses there are from that source: http://localhost:8063/admin/learning_resources/learningresource/?etl_source=prolearn&published__exact=1&resource_type__exact=course
- If you search for "Accelerating Digital Transformation" you will get multiple results: http://localhost:8063/api/v1//learning_resources_search/?q=%22Accelerating%20Digital%20Transformation%22
- Switch to this branch and run the management command again.  
- Now the count of published courses from [this source should be lower](http://localhost:8063/admin/learning_resources/learningresource/?etl_source=prolearn&published__exact=1&resource_type__exact=course), and there should be only one search result for "[Accelerating Digital Transformation](http://localhost:8063/api/v1//learning_resources_search/?q=%22Accelerating%20Digital%20Transformation%22)".

### DIFF
--- a/learning_resources/etl/loaders.py
+++ b/learning_resources/etl/loaders.py
@@ -290,16 +290,14 @@ def load_course(
         if unique_field_name != DEFAULT_UNIQUE_KEY:
             # Some dupes may result, so we need to unpublish resources
             # with matching unique values and different readable_ids
-            dupe_resources = LearningResource.objects.filter(
+            for resource in LearningResource.objects.filter(
                 **{unique_field_name: unique_field_value},
                 platform=platform,
                 resource_type=LearningResourceType.course.name,
-            ).exclude(readable_id=resource_id)
-            if dupe_resources.count() > 0:
-                bulk_resources_unpublished_actions(
-                    [resource.id for resource in dupe_resources],
-                    LearningResourceType.course.name,
-                )
+            ).exclude(readable_id=resource_id):
+                resource.published = False
+                resource.save()
+                resource_unpublished_actions(resource)
         (
             learning_resource,
             created,
--- a/learning_resources/etl/loaders.py
+++ b/learning_resources/etl/loaders.py
@@ -12,7 +12,7 @@ from learning_resources.constants import (
     PlatformType,
 )
 from learning_resources.etl.constants import (
-    DEFAULT_UNIQUE_KEY,
+    READABLE_ID_FIELD,
     CourseLoaderConfig,
     ProgramLoaderConfig,
 )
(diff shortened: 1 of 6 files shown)


learning_resources/etl/loaders.py
```diff
@@ -309,14 +317,6 @@ def load_course(  # noqa: C901
...
             load_run(learning_resource, course_run_data)
 
-        if deduplicated_course_id and not created:
-            most_relevent_run = get_most_relevant_run(learning_resource.runs.all())
-
-            if most_relevent_run.run_id in run_ids_to_update_or_create:
-                for attr, val in resource_data.items():
-                    setattr(learning_resource, attr, val)
-                learning_resource.save()
-
```
mbertrand, line 319: This didn't seem necessary and the relevant test still passes without it.

learning_resources/etl/loaders.py
```diff
@@ -268,37 +268,45 @@ def load_course(  # noqa: C901
...
-            )
-
-            if readable_id != deduplicated_course_id:
-                duplicate_resource = LearningResource.objects.filter(
-                    platform=platform, readable_id=readable_id
-                ).first()
-                if duplicate_resource:
-                    duplicate_resource.published = False
-                    duplicate_resource.save()
-                    resource_unpublished_actions(duplicate_resource)
```
mbertrand, line 290: Seemed more complicated than necessary, so simplified it.  Relevant unit test still passes.

abeglova (approved):
👍

PR #605, opened 2024-03-11
Program Letter View

shanbady:

Closes https://github.com/mitodl/mit-open/issues/555

This PR implements a shareable program letter view for users who have earned a program certificate in micromasters.
The the flow for this is:
1. User receives program certificate in micromasters upon which it is recorded in MM database
2. MM program certificate table is pulled into data lake ultimately ends up in program certificate data mart
3. Data mart is synced directly to a ProgramCertificates table in MIT-Open
4. User visits dashboard in Open and see's link to view program letter if they have an entry in ProgramCertificates for the program dashboard they are looking at.
5. that link points to an "intercept view" which:
    - generates a "program letter" object which will have a uuid that is used in the shareable url
    - makes an api call to the micromasters cms to fetch the template content for the program letter
    - forwards the user to the program letter view which renders the letter and is shareable

### Main Changes
The changes in this PR can be grouped into 3 main parts:
1 - An api route added to create a program letter object and fetch template data from micromasters
2 - Autogenerated api schema for above route
3 - fronted react view to render the program letter

### How can this be tested?
1. in your .env file, set MICROMASTERS_CMS_API_URL=https://micromasters-rc.odl.mit.edu/api/v0/wagtail/
2. get into a django shell via ```docker-compose exec web python manage.py  shell```
3. We are going to create 2 program certificates for your user. On production this would exist as a result of the data sync but we will simulate these records here. Make sure you replace the email here with whatever user you will be logging in with:
<pre><code>
ProgramCertificate.objects.get_or_create(user_full_name='john doe',user_email=<b>'your@email.here'</b>, micromasters_program_id=2, program_title='Data, Economics, and Development Policy')
</code></pre>
4. login as the user with email from above.
5. visit http://localhost:8063/program_letter/1/ and http://localhost:8063/program_letter/18/ - you should be redirected to shareable (viewable in incognito mode while logged out) program letter views. The urls should have uuids in them.


profiles/models.py
```diff
@@ -179,4 +182,19 @@ class Meta:
         db_table = "external.programcertificate"
 
     def __str__(self):
-        return "program certificate: {self.user_full_name} - {self.program_title}"
+        return f"program certificate: {self.user_full_name} - {self.program_title}"
+
+
+class ProgramLetter(models.Model):
```
mbertrand, line 188: Add a docstring for the class

profiles/serializers.py
```diff
@@ -244,3 +252,49 @@ class Meta:
...
+    """
+    Serializer for Program Certificates
+    """
+
+    class Meta:
+        model = ProgramCertificate
+        fields = "__all__"
+
+
+class ProgramLetterTemplateFieldSerializer(serializers.Serializer):
```
mbertrand, line 267: Docstring

profiles/utils.py
```diff
@@ -340,3 +341,17 @@ def generate_svg_avatar(name, size, color, bgcolor):
             "text": xml_escape(initials.upper()),
         }
     ).replace("\n", "")
+
+
+def fetch_program_letter_template_data(letter):
```
mbertrand, line 346: Add docstring

profiles/serializers.py
```diff
@@ -244,3 +252,49 @@ class Meta:
...
+
+    id = serializers.UUIDField(read_only=True)
+
+    template_fields = serializers.SerializerMethodField()
+
+    certificate = ProgramCertificateSerializer()
+
+    @extend_schema_field(ProgramLetterTemplateFieldSerializer())
+    def get_template_fields(self, instance) -> dict:
+        """Return the number of episodes in the podcast"""
```
mbertrand, line 293: Wrong docstring.
"Get the serialized template fields"?

mbertrand (changes requested):
Overall works great!  Just noticed a couple things:

- Some docstrings are wrong or missing.  
- If I enter the wrong uuid in the letter url, I would expect to get a 404 page but instead I get an error message if it is a valid but wrong uuid, or a mostly blank letter on the frontend and a backend 'not a valid UUID' ValidationError if it isn't a valid uuid:

Screenshot 2024-03-12 at 10 29 03 AM


Screenshot 2024-03-12 at 10 22 14 AM

ChristopherChudzicki:
@mbertrand Re
> If I enter the wrong uuid in the letter url, I would expect to get a 404 page but instead I get an error message

You're actually seeing a 404 page covered up by a dev-mode error overlay. In production, you'd only get the 404 page. The error overlay has a little X in upper right corner that you can click to get rid of it.

I find the error overlay annoying / confusing for 404 and 403 errors, too. IIRC the error overlay is pretty easy to disable entirely, but I don't think I found a way to disable it just for the 404/403/401 errors that we handle with NotFound and Forbidden pages.

> or a mostly blank letter on the frontend and a backend 'not a valid UUID' ValidationError if it isn't a valid uuid:

I see that issue, too.

profiles/serializers.py
shanbady: good catch!

shanbady:
@mbertrand I resolved the blank page issue by checking for valid uuids and 404'ing server side if invalid which will trigger the 404 on the frontend.

mbertrand (approved):
Looks great! 👍

PR #608, opened 2024-03-12
Endpoint for user program certificate info and program letter links

### LINKED ISSUE #556: Create api endpoint for retrieving program letter info (or add program letter info to some other existing endpoint)
We need some way for the frontend to retrieve program letters with the shareable links. It does not seem like we currently do not have a way to get program specific data for a user. This will need to exist in some form before we tackle this - we have [this ticket](https://github.com/mitodl/mit-open/issues/323) to address that

shanbady:

Closes #556 

This Pr adds an api endpoint (/api/v0/program_certificates) to retrieve program certificates along with program letter urls for an authenticated user. The idea is that when a user goes to a program specific dashboard, we will call this endpoint to get a list of program letter urls to display.

### How can this be tested?
1. checkout this branch and generate some program certificates via 
<pre><code>
ProgramCertificate.objects.get_or_create(user_full_name='john doe',user_email=<b>'your@email.here'</b>, micromasters_program_id=2, program_title='Data, Economics, and Development Policy')
</code></pre>
2. visit endpoint http://localhost:8063/api/v0/program_certificates/ and confirm they your program certificates are listed there along with the program_letter_urls

### DIFF
--- a/profiles/serializers.py
+++ b/profiles/serializers.py
@@ -260,12 +260,29 @@ class ProgramCertificateSerializer(serializers.ModelSerializer):
     Serializer for Program Certificates
     """
 
-    program_letter_url = serializers.SerializerMethodField()
+    program_letter_generate_url = serializers.SerializerMethodField()
+    program_letter_share_url = serializers.SerializerMethodField()
 
-    def get_program_letter_url(self, instance):
-        return reverse(
+    def get_program_letter_generate_url(self, instance):
+        request = self.context.get("request")
+        letter_url = reverse(
             "profile:program-letter-intercept", args=[instance.micromasters_program_id]
         )
+        if request:
+            return request.build_absolute_uri(letter_url)
+        return letter_url
+
+    def get_program_letter_share_url(self, instance):
+        request = self.context.get("request")
+
+        user = User.objects.get(email=instance.user_email)
+        letter, created = ProgramLetter.objects.get_or_create(
+            user=user, certificate=instance
+        )
+        letter_url = letter.get_absolute_url()
+        if request:
+            return request.build_absolute_uri(letter_url)
+        return letter_url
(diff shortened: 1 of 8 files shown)


profiles/serializers.py
```diff
@@ -259,6 +260,13 @@ class ProgramCertificateSerializer(serializers.ModelSerializer):
     Serializer for Program Certificates
     """
 
+    program_letter_url = serializers.SerializerMethodField()
+
+    def get_program_letter_url(self, instance):
+        return reverse(
+            "profile:program-letter-intercept", args=[instance.micromasters_program_id]
+        )
+
```
mbertrand, line 269: This is returning urls in the format `"/program_letter/1/"`, which will only work for the certificate's logged-in user.  Wondering if maybe it should provide the shareable url instead if it exists yet (`/program_letter/<uuid>/view`), or both as separate fields (something like `create_letter_url`, `share_letter_url`) since it's possible the shareable url hasn't yet been created?  

Also, what do you think of returning absolute urls instead (`http://<domain>/program_letter/....`)?  The frontend could create a full url itself, but maybe easier to do it in the backend?

mbertrand (commented):
Looks good, but have some questions about the returned url and format.

profiles/serializers.py
shanbady: @mbertrand - made it generate absolute urls and changed the url params to return both a program_letter_generate_url and a program_letter_share_url. I also made it so the serializer itself does a get_or_create for the program letter. In the future we might consider removing the intercept view altogether but I think it still has a place for now (i'm imagning a case where we send links to view view a program letter from some marketing automation software that is only aware of program ids etc)

mbertrand (approved):
👍

PR #617, opened 2024-03-15
shanbady/separate database router and schema for program certificates

### LINKED ISSUE #3771: Next.js server memory climbs without plateau on 0.77.3, causing 15s origin timeouts
### Expected Behavior

The Next.js server's memory should stay flat under steady traffic, as it did on 0.77.2: ~520 MiB working set, unchanged for roughly two days, with zero synthetic probe failures.

### Current Behavior

Since 0.77.3 was deployed to production at ~06:10 ET on 2026-08-13, the `mit-learn-nextjs` pods climb monotonically and never plateau, and the origin intermittently stops responding within 15s.

Per-image working set (`container_memory_working_set_bytes`, namespace `mitlearn`, cluster `applications-production`):

| Image | Working set |
|---|---|
| 0.77.1 | pinned 890–916 MiB for days, spikes to 1138 |
| 0.77.2 | flat ~520 MiB for ~2 days |
| 0.77.3 | 420 MiB → 1058 MiB in 4 hours, still climbing |

Both 0.77.2 and 0.77.3 pods ran side by side during the rollout window, so this is a like-for-like comparison under the same traffic: the 0.77.2 replicas stayed flat for the whole 8 hours they were up while the 0.77.3 replicas climbed.

CPU tracks it — 0.16 cores at pod start rising to 0.6–0.79, consistent with V8 spending progressively more time in major GC.

Zero container restarts, so this is not OOMKill/crash looping. The user-visible symptom is stop-the-world GC pauses: the synthetic check against the origin hit its 15s timeout on 7 probes in 2 hours, reaching 10% of probes failing in a single hour, with `probe_http_status_code` returning 200 whenever the request completed. Time is spent in the `processing` and `transfer` phases (7.9s + 7.1s on the worst sample) while `resolve`/`connect`/`tls` stay at 4–6 ms, so it is the origin's own render/response, not network.

End users are largely shielded because Fastly absorbs it: the CDN-fronted `learn.mit.edu` probe held flat at 0.02s throughout. `next.learn.mit.edu` bypasses Fastly and hits the origin directly.

### Steps to Reproduce

1. Deploy 0.77.3 to an environment with production-like traffic.
2. Watch `container_memory_working_set_bytes` for the `nextjs-app` container over 3–4 hours.
3. Working set climbs monotonically from ~400 MiB past 1 GiB with no plateau. On 0.77.2 the same metric is flat.
4. While the heap is near full, request `/` repeatedly and observe occasional responses exceeding 15s.

### Possible Solution

Prime suspect is https://github.com/mitodl/mit-learn/pull/3727, which added `htmlToPlainText` to `standardizeMetadata`:

```diff
 export const standardizeMetadata = ({ ... }) => {
   title = `${title} | ${env("NEXT_PUBLIC_SITE_NAME")}`
+  description = htmlToPlainText(description)
```

and on the resource branch replaced a regex with the same call:

```diff
-    description = data?.description?.replace(/<\/[^>]+(>|$)/g, "") ?? ""
+    description = data?.description ?? ""
```

https://github.com/mitodl/mit-learn/blob/v0.77.3/frontends/main/src/common/metadata.ts
https://github.com/mitodl/mit-learn/blob/v0.77.3/frontends/main/src/common/htmlToPlainText.ts

Two properties make this expensive rather than merely wasteful:

- `standardizeMetadata` is reached from `generateMetadata` on essentially every route (30+ files under `frontends/main/src/app` reference it), so this moved from a per-resource regex to a jsdom parse on **every server-rendered page request**.
- `htmlToPlainText` calls `DOMPurify.sanitize(html, { RETURN_DOM_FRAGMENT: true })` through `isomorphic-dompurify`, which on the server is jsdom. Each call allocates real DOM nodes. The guard is `if (!html) return ""`, which only catches the empty string — so the homepage pays a full jsdom parse to strip tags from the constant `"Learn with MIT"`, which contains no markup at all.

That the climb is monotonic rather than a sawtooth suggests the fragments are being retained, not just churned.

Cheapest fix that would confirm it: short-circuit when there is no markup to strip, e.g. return the input unchanged when it contains no `<`, and/or keep the regex path for the non-resource case. That removes the cost from the homepage and every static-description page without changing behaviour for descriptions that do contain HTML.

### Additional Details

Worth knowing that 0.77.2 was not merely "before the regression" — it was a real improvement over 0.77.1, which sat pinned at 890–916 MiB for days (right at the heap ceiling the deployment was sized around) and produced its own intermittent probe failures, including a ~17-hour degraded period on 2026-08-10 with one window where all 5 probes failed. Whatever landed in 0.77.2 fixed a long-standing problem; 0.77.3 undid the improvement.

Infrastructure-side context, for anyone reading the pod spec: the deployment currently requests 1536Mi with a 2Gi limit and pins `--max-old-space-size` to 1312 MiB, with comments describing an earlier version of this same climb. Those comments are now partly stale, since 0.77.2 demonstrated flat memory. Raising the memory request again is a mitigation that buys hours, not a fix.

Detection of this class of regression is being improved separately on the infrastructure side; no single 5-minute availability window looked bad enough to alert, only the hourly failure rate did.

shanbady:

Closes [#3771](https://github.com/mitodl/hq/issues/3771)

The changes in this PR properly migrate the ProgramCertificate table to its own schema and also includes a database router that prevents writes from happening on tables in the external schema.

### How can this be tested?
1. checkout this branch
2. run the migrations or do a full docker-compose down+up
3. in the postgres database see that there is an external schema. in manage.py dbshell you can run the following:
```
 SELECT nspname
 FROM pg_catalog.pg_namespace;
```
4. trying to create a programcertificate in django shell manually should give you an exception:
```
ProgramCertificate.objects.get_or_create(user_full_name='john doe',user_email='test@test.com', micromasters_program_id=1, program_title='Supply Chain Management')
```

### Additional Context
Our previous deploy of the programcertificate table ended up create a table in the public schema named "external.programcertificate".

The new migrations in this PR do the following to fix that:
1. [Drop the rogue programcertificate table](https://github.com/mitodl/mit-open/blob/shanbady/separate-database-router-for-program-certificates2/profiles/migrations/0018_alter_programletter_certificate_and_more.py#L12) 
2. [recreate the programcertificate table](https://github.com/mitodl/mit-open/blob/shanbady/separate-database-router-for-program-certificates2/profiles/migrations/0018_alter_programletter_certificate_and_more.py#L13) in the external schema


profiles/models.py
```diff
@@ -179,7 +179,7 @@ class ProgramCertificate(models.Model):
 
     class Meta:
         managed = False
-        db_table = "external.programcertificate"
+        db_table = '"external"."programcertificate"'
```
ChristopherChudzicki, line 182: fun.

ChristopherChudzicki (approved):
👍

PR #618, opened 2024-03-15
Add a Search Page

### LINKED ISSUE #490: Add a search page
### Description/Context
Add a search page. It should be similar to the search pages at https://discussions-rc.odl.mit.edu/learn and https://ocw.mit.edu/search:
- Facets with counts
- text search
- sortable
- controllable pagesize (cards vs rows in OCW)
- resources & contentfiles
- show results with cards

### Plan/Design
I plan to tackle this in tandem with https://github.com/mitodl/course-search-utils/issues/65.

ChristopherChudzicki:
Closes #490 

This PR adds a search page. We don't have final designs for this, but I used https://www.figma.com/proto/bRb19HaucO5loHGSU81nFG/MIT_Prototype?page-id=0%3A1&type=design&node-id=9-6102&viewport=4023%2C1145%2C0.25&t=6VyDzTtHVzsZ3Nyv-1&scaling=scale-down&starting-point-node-id=48%3A2446&show-proto-sidebar=1 as inspiration.

**Note:** After discussing with @Ferdi, only "Topics" facet is supported for now (and "Resource Type" via the tabs).

Screenshot 2024-03-18 at 11 20 17 AM

### How can this be tested?
1. Ensure you have a working search index https://github.com/mitodl/mit-open?tab=readme-ov-file#enabling-searching-the-course-catalog-on-opensearch
2. Visit http://localhost:8063/; enter something in the searchbox and hit enter.
3. You should now be on the searchpage. Try facets in sidebar, tabs for resource_type facet, query text, and pagination.
    - results should update appropriately
    - URL should update appropriately


jonkafton:
This really does work very nicely. The only thing I noticed while testing the UI was that the search clear doesn't reset results (or clear the URL search param).

There are couple of small layout issues for mobile / smaller screen width. I appreciate we are working off designs for desktop. Do we have plans for mobile at this point?

shanbady (approved):
works like it should 👍

Ferdi:
> The only thing I noticed while testing the UI was that the search clear doesn't reset results (or clear the URL search param).

I think we should try to fix this in this PR if it's an issue. 

We can deal with perfecting mobile later.

frontends/mit-open/src/pages/SearchPage/SearchPage.tsx
```diff
@@ -0,0 +1,355 @@
...
+const FacetStyles = styled.div`
+  * {
+    color: ${({ theme }) => theme.palette.secondary.main};
+  }
+
+  input[type="checkbox"] {
+    accent-color: ${({ theme }) => theme.palette.primary.main};
+  }
+
+  .filter-section-main-title {
```
jonkafton, line 59: It would be good to find a solution that doesn't depend on classnames in dependencies. I do see the dilemma - how to write `course-search-utils` that are customizable, but still largely hold their layout and base styles without having to pass so many components in that lose code reuse.

If `course-search-utils` also used styled components, we could provide them to the css template literal for interpolation (we discussed that a bit the other day). The benefit would be that `course-search-utils` API is defined by its exports and not additionally by its classnames. Not one for this PR, but food for thought.

jonkafton (approved):

jonkafton:
> > The only thing I noticed while testing the UI was that the search clear doesn't reset results (or clear the URL search param).
> 
> I think we should try to fix this in this PR if it's an issue.

Retested after @ChristopherChudzicki last change - works perfectly!

PR #633, opened 2024-03-19
First iteration of field pages

abeglova:
https://github.com/mitodl/mit-open/issues/602

This PR moves field functionality from open discussions to the new open UI. It does not include list functionality since we probably don't need it. 

Screenshot 2024-03-18 at 11 22 36 PM
Screenshot 2024-03-18 at 11 23 03 PM

Screenshot 2024-03-18 at 11 23 16 PM

### How can this be tested?
- If you don't have a field locally already, create one at http://localhost:8063/api/v1/fields/ 
- Go to http://localhost:8063/fields/<your field>/ you should see your field's name and description
- Go to http://localhost:8063/fields/<your field>/manage - if you are not logged in, you should see a message that you don't have permission to access the page
- Login and go back to  http://localhost:8063/fields/<your field>/. Click the gear at the right hand side of the banner to go to the field management page
-  You should be able to change the title and description in the "appearance" tab
- You should be able to add new widgets 

### Additional Context
There will be follow up tickets to clean up the documentation and api call structure for the field and widget_list api calls (https://github.com/mitodl/mit-open/issues/643 and https://github.com/mitodl/mit-open/issues/644)

### DIFF
--- a/frontends/mit-open/src/pages/FieldPage/FieldPageSkeleton.tsx
+++ b/frontends/mit-open/src/pages/FieldPage/FieldPageSkeleton.tsx
@@ -1,0 +1,80 @@
+import React from "react"
+import { Link } from "react-router-dom"
+import * as routes from "../../common/urls"
+import { BannerPage, styled, Container } from "ol-components"
+import { useFieldDetail } from "api/hooks/fields"
+import FieldMenu from "@/components/FieldMenu/FieldMenu"
+import FieldAvatar from "@/components/FieldAvatar/FieldAvatar"
+
+export const FieldTitleRow = styled.div`
+  display: flex;
+  flex-direction: row;
+  align-items: center;
+
+  h1 a {
+    margin-left: 1em;
+
+    &:hover {
+      text-decoration: none;
+    }
+  }
+`
+
+export const FieldControls = styled.div`
+  position: relative;
+  flex-grow: 0.95;
+  justify-content: flex-end;
+  min-height: 38px;
+  display: flex;
+  align-items: center;
+`
+
+interface FieldSkeletonProps {
+  children: React.ReactNode
+  name: string
+}
+
+/**
+ * Common structure for field-oriented pages.
+ *
+ * Renders the field title and avatar in a banner.
+ */
+const FieldSkeletonProps: React.FC<FieldSkeletonProps> = ({
+  children,
+  name,
+}) => {
+  const field = useFieldDetail(name)
+
+  return (
+    <BannerPage
+      src={field.data?.banner ?? ""}
+      alt=""
+      omitBackground={field.isLoading}
+      bannerContent={
+        <Container>
+          <FieldTitleRow>
+            {field.data && (
+              <>
+                <FieldAvatar field={field.data} imageSize="medium" />
+                <h1>
+                  <Link to={routes.makeFieldViewPath(field.data.name)}>
+                    {field.data.title}
+                  </Link>
+                </h1>
+                <FieldControls>
+                  {field.data?.is_moderator ? (
+                    <FieldMenu fieldName={field.data.name} />
+                  ) : null}
+                </FieldControls>
+              </>
+            )}
+          </FieldTitleRow>
+        </Container>
+      }
+    >
+      {children}
+    </BannerPage>
+  )
+}
+
+export default FieldSkeletonProps
--- a/channels/serializers.py
+++ b/channels/serializers.py
@@ -58,19 +58,19 @@ class ChannelAppearanceMixin(serializers.Serializer):
             return True
         return False
 
-    def get_avatar(self, channel) -> str:
+    def get_avatar(self, channel) -> str | None:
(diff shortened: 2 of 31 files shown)


frontends/mit-open/src/page-components/FieldAvatar/FieldAvatar.tsx
```diff
@@ -0,0 +1,61 @@
...
+  return (
+    <div className={`avatar-container ${imageSize}-size`}>
+      <div className="avatar">
+        {isDefault ? (
+          <div className="avatar-initials">{initials(field.title)}</div>
+        ) : (
+          <img src={imageUrl} alt="" className={"avatar-image"} />
+        )}
+      </div>
+    </div>
```
ChristopherChudzicki, line 57: I realize that styling is not the primary goal in this PR **and** we are mostly bringing code over from infinite corridor. That said, we [switched](https://github.com/mitodl/mit-open/issues/239) to using [emotion](https://emotion.sh/docs/styled) and I think it would be good to try and use that rather than lots of explicit classnames. At least where it's reasonable straightforward.

This seems like a good candidate to me. Here's how I would do it with emotion:
```tsx
const IMG_SIZES = {
  small: "22px",
  medium: "57px",
  large: "90px",
}
const FONT_SIZES = {
  small: "15px",
  medium: "32px",
  large: "50px",
}

type AvatarStyleProps = Required<Pick<AvatarProps, "imageSize">>
const AvatarContainer = styled.div<AvatarStyleProps>`
  display: flex;
  flex-direction: row;
  align-items: center;
  width: ${({ imageSize }) => IMG_SIZES[imageSize]};
  height: ${({ imageSize }) => IMG_SIZES[imageSize]};
`
const AvatarImg = styled.img<AvatarStyleProps>`
  display: flex;
  flex-direction: row;
  align-items: center;
  justify-content: center;
  min-height: 0;
  min-width: 0;
  border-radius: 15px;
  width: ${({ imageSize }) => IMG_SIZES[imageSize]};
  height: ${({ imageSize }) => IMG_SIZES[imageSize]};
`
const AvatarInitials = styled(AvatarImg.withComponent("div"))`
  font-size: ${({ imageSize = "medium" }) => FONT_SIZES[imageSize]};
  font-weight: 600;
  color: white;
  background-color: ${({ theme }) => theme.custom.channelAvatarBg};
`

const FieldAvatar: React.FC<AvatarProps> = (props) => {
  const { field, formImageUrl, imageSize = "medium" } = props

  const imageUrl = formImageUrl || getImage(field, imageSize)

  return (
    <AvatarContainer imageSize={imageSize}>
      {!imageUrl ? (
        <AvatarInitials imageSize={imageSize}>
          {initials(field.title)}
        </AvatarInitials>
      ) : (
        <AvatarImg src={imageUrl} imageSize={imageSize} />
      )}
    </AvatarContainer>
  )
}
```
One of the goals here is to keep the avatar-related styles co-located with the avatar code.

**NOTES:**
- I made up the `large` values... :shrug:
- if you adopt this, also remove the avatar-related styles in FieldPageSkeleton

@jonkafton feel free to add anything

frontends/mit-open/src/pages/FieldPage/FieldPageSkeleton.tsx
```diff
@@ -0,0 +1,186 @@
...
+    flex-direction: row;
+    align-items: center;
+
+    img {
+      height: $avatar-size;
+      width: $avatar-size;
+    }
+
+    h1 a {
+      color: theme.$color-gray-1;
```
ChristopherChudzicki, line 28: **Request:** I think we should remove this and all other $ variables. Either replace them (like suggested below) or just delete.


All of the `$` variables are old sass variables and won't work with Emotion. I would suggest either using `white`, or you can access the theme via
```js
color: ${({ theme }) => theme.custom.colorGray1};
```

BTW: I forget if you use VS Code, but https://marketplace.visualstudio.com/items?itemName=styled-components.vscode-styled-components is a good extension to install if you do. (It says styled components, which is a library similar to Emotion. The extension works well with both.)

ChristopherChudzicki (changes requested):
This is working very well.  left some fairly minor comments and suggestions. Some of these are around styling. I realize styling isn't the main focus here, but I do think the suggestions are good general practices that can be done here without much effort.

frontends/mit-open/src/page-components/FieldAvatar/FieldAvatar.test.tsx
ChristopherChudzicki, line 1: Could you move this to `mit-open/src/components`?
- `page-components`: significant subsections of a page that might use routing and API calls (e.g., `ManageListModal`, `LearningResourceDrawer`).
- components: display / simple UI interactions

https://github.com/mitodl/mit-open/blob/main/docs/architecture/front-end-component-structure.md

ChristopherChudzicki (approved):
👍  Looks good.

One request, but no need for re-review IMO.

PR #634, opened 2024-03-19
fix flaky test

### LINKED ISSUE #609: Flaky test case: test_get_resource_learning_paths
test_get_resource_learning_paths appears to be intermittently failing. 

### Expected Behavior
The test should always succeed

### Current Behavior
It appears to intermittently fail with the following:

```
____________________ test_get_resource_learning_paths[True] ____________________

user_client = <rest_framework.test.APIClient object at 0x79ae4f40c110>
user = <User: 01HRT717NX5JBCN4NK0NX5XE9F>, is_editor = True

    @pytest.mark.parametrize("is_editor", [True, False])
    def test_get_resource_learning_paths(user_client, user, is_editor):
        """Test that the learning paths are returned for a resource"""
        update_editor_group(user, is_editor)
        course = factories.CourseFactory.create()
        path_items = sorted(
            factories.LearningPathRelationshipFactory.create_batch(
                3, child=course.learning_resource
            ),
            key=lambda item: item.position,
        )
        resp = user_client.get(
            reverse("lr:v1:courses_api-detail", args=[course.learning_resource.id])
        )
    
        expected = (
            [
                {
                    "id": path_item.id,
                    "parent": path_item.parent_id,
                    "child": course.learning_resource.id,
                }
                for path_item in path_items
            ]
            if is_editor
            else []
        )
    
>       assert resp.data.get("learning_path_parents") == expected
E       AssertionError: assert [OrderedDict(...ild', [183](https://github.com/mitodl/mit-open/actions/runs/8255551984/job/22582273447?pr=608#step:12:184)5)])] == [{'child': 18...arent': 1851}]
E         At index 0 diff: OrderedDict([('id', 1433), ('parent', [184](https://github.com/mitodl/mit-open/actions/runs/8255551984/job/22582273447?pr=608#step:12:185)4), ('child', 1835)]) != {'id': 1426, 'parent': 1836, 'child': 1835}
E         Full diff:
E           [
E         +  OrderedDict([('id', 1433), ('parent', 1844), ('child', 1835)]), OrderedDict([('id', 1439), ('parent', [185](https://github.com/mitodl/mit-open/actions/runs/8255551984/job/22582273447?pr=608#step:12:186)1), ('child', 1835)]), OrderedDict([('id', 1426), ('parent', 1836), ('child', 1835)]),
E         -  {'child': 1835,
E         -   'id': 1426,
E         -   'parent': 1836},
E         -  {'child': 1835,
E         -   'id': 1433,
E         -   'parent': 1844},
E         -  {'child': 1835,
E         -   'id': 1439,
E         -   'parent': 1851},
E           ]

learning_resources/views_learningpath_test.py:392: AssertionError
```

### Steps to Reproduce
Not able to reproduce reliably

### Additional Details
failure log here: https://github.com/mitodl/mit-open/actions/runs/8255551984/job/22582273447?pr=608#step:12:156

shanbady:

Fixes #609 

Fixes a flaky test by standardizing the order of results

### How can this be tested?
to simulate flaky test runs, apply the following patch to the learning_resources/views_learningpath_test.py file:
[test.patch](https://github.com/mitodl/mit-open/files/14653071/test.patch) and run the loaders test via  pytest learning_resources/views_learningpath_test.py -vvv -k test_get_resource_learning_paths -x --no-cov -s --pdb - it should not fail throughout its hundreds of runs (it will fail on main branch at some point).

### DIFF
--- a/learning_resources/views_learningpath_test.py
+++ b/learning_resources/views_learningpath_test.py
@@ -370,12 +370,11 @@ def test_get_resource_learning_paths(user_client, user, is_editor):
         factories.LearningPathRelationshipFactory.create_batch(
             3, child=course.learning_resource
         ),
-        key=lambda item: item.position,
+        key=lambda item: item.id,
     )
     resp = user_client.get(
         reverse("lr:v1:courses_api-detail", args=[course.learning_resource.id])
     )
-
     expected = (
         [
             {
@@ -388,5 +387,7 @@ def test_get_resource_learning_paths(user_client, user, is_editor):
         if is_editor
         else []
     )
-
-    assert resp.data.get("learning_path_parents") == expected
+    response_data = sorted(
+        resp.data.get("learning_path_parents"), key=lambda item: item["id"]
+    )
+    assert response_data == expected


learning_resources/views_learningpath_test.py
```diff
@@ -370,12 +370,11 @@ def test_get_resource_learning_paths(user_client, user, is_editor):
         factories.LearningPathRelationshipFactory.create_batch(
             3, child=course.learning_resource
         ),
-        key=lambda item: item.position,
+        key=lambda item: item.id,
```
ChristopherChudzicki, line 373: The view for LearningPathRelationships specifies
```py
    filter_backends = [OrderingFilter]
    ordering = ["position", "-child__last_modified"]
```

> If an ordering attribute is set on the view, this will be used as the default ordering.
> – [drf docs](https://www.django-rest-framework.org/api-guide/filtering/)

The "relationships" are re-orderable, and the API is intended to, by default, return the relationships ordered by position.

ChristopherChudzicki (commented):
I believe that sorting by `item.position` is intended here. If the test is failing sometimes, it's possible that's an actual bug, and not a flaky test.

learning_resources/views_learningpath_test.py
shanbady: good callout. i'm going to put this back in draft and have a deeper look

ChristopherChudzicki: My original comment was inaccurate: I did not look closely enough and misunderstood which API was being tested here.
-  My comment applies to `/api/v1/learningpaths/{LIST_ID}/items/`
- This API is testing `/api/v1/courses/{COURSE_ID}`. *Courses include a "micro" list of the items. *

So I think your original solution is fine.

ChristopherChudzicki (approved):

PR #647, opened 2024-03-21
Shanbady/endpoint to retrieve session data

### LINKED ISSUE #614: Endpoint to retrieve session data
### Description/Context

Working towards [RFC: Decouple the Frontend from the Backend Web Service #3646](https://github.com/mitodl/hq/discussions/3646).

Currently the web service injects session data into the base HTML template at [`{{ js_settings|json_script:'js-settings' }}`](https://github.com/mitodl/mit-open/blob/51587f10c0f18d171f28f09b5e7af92cd2525bad/main/templates/index.html#L19C7-L19C48).

When the web service is no longer serving the base HTML, the client app will need to fetch the authenticated user's session as it loads.

### Plan/Design

We need an endpoint, `GET /api/v1/session`, for example that returns the `user` object currently in the bootstrap with these properties:
```
{
  id: number
  first_name: string
  last_name: string
  is_authenticated: boolean
  is_learning_path_editor: boolean
  is_article_editor: boolean
}
```

shanbady:

Closes #614 

This adds fields that the frontend needs to /api/v0/user/me
In addition to that, it also replicates /api/v0/user/me in the v1 api so that open api generates clients for it. The v1 api is marked as experimental.

### How can this be tested?
1. checkout this branch
2. visit the v0 user api (/api/v0/users/me) which should now contain new fields first_name, last_name, is_article_editor, and is_learning_path_editor
3. visit the v1 api and see that it matches the v0 api  /api/v1/users/me
4. see that the new fields and new api route exist in the open api spec


shanbady:
@ChristopherChudzicki I removed the v1 endpoint

jonkafton:
Looks good 👍

jonkafton (approved):

PR #650, opened 2024-03-21
Add userlists page and refactor LearningResourceCardTemplate

gumaerc:
Closes https://github.com/mitodl/mit-open/issues/649
Part of https://github.com/mitodl/mit-open/issues/613

This PR primarily adds a page at `/userlists/` that lists the currently logged-in user's `UserList` objects. A menu item was added to `UserMenu` linking to the page. The `LearningResourceCardTemplate` page component was refactored into a `CardTemplate` page component, with `LearningResourceCardTemplate inheriting it. A new page component, `UserListCardTemplate` was also added that also inherits `CardTemplate`. Hooks were also added to handle querying user lists and user list items.

image
image

### How can this be tested?
 - Make sure you're set up for running `mit-open` locally, and then spin it up on this branch
 - Create a `UserList` by visiting http://localhost:8063/api/v1/userlists/ and putting a JSON object into the raw data input with something like:
```
{
    "topics": [],
    "title": "Test UserList",
    "description": "A UserList for testing",
    "privacy_level": "private
}
```
 - Visit the home page at http://localhost:8063/
 - Log in using the drawer in the upper right
 - Once you are returned to the home page, click the user menu again and then click "User Lists"
 - Once you are at the User Lists page, you should see "Favorites" (the default `UserList`) and the `UserList` you created

### Additional Context
It's not currently possible to create a new `UserList` on this page, or edit existing `UserList`s yet. That is part of https://github.com/mitodl/mit-open/issues/613 though, and will come in a later PR. The primary purpose of this PR is to set the stage for the rest of the work.

### DIFF
--- a/frontends/mit-open/src/page-components/LearningResourceCardTemplate/LearningResourceCardTemplate.test.tsx
+++ b/frontends/mit-open/src/page-components/LearningResourceCardTemplate/LearningResourceCardTemplate.test.tsx
@@ -30,7 +30,7 @@ describe("LearningResourceCard", () => {
     expect(coverImg.src).toBe(resourceThumbnailSrc(resource.image, imgConfig))
   })
 
-  it("does not show an image if suppressImage is true", () => {
+  it("does not show an image if and only if suppressImage is true", () => {
     const resource = factory.course()
     const imgConfig = makeImgConfig()
     const { rerender } = render(
@@ -94,7 +94,7 @@ describe("LearningResourceCard", () => {
     },
   )
 
-  it("Should show an item count iff the resource is a list", () => {
+  it("Should show an item count if the resource is a list", () => {
     const resource = factory.learningPath()
     const count = resource.learning_path?.item_count
     const imgConfig = makeImgConfig()
@@ -111,7 +111,7 @@ describe("LearningResourceCard", () => {
     expect(itemCount).toBeVisible()
   })
 
-  it("Should NOT show an item count iff the resource is NOT a list", () => {
+  it("Should NOT show an item count if the resource is NOT a list", () => {
     const resource = factory.resource({
       title: "Not a list",
       resource_type: ResourceTypeEnum.Course,
--- a/frontends/mit-open/src/page-components/LearningResourceCardTemplate/LearningResourceCardTemplate.test.tsx
+++ b/frontends/mit-open/src/page-components/LearningResourceCardTemplate/LearningResourceCardTemplate.test.tsx
@@ -41,7 +41,6 @@ describe("LearningResourceCard", () => {
       />,
     )
     const images = screen.queryAllByRole("img")
-    console.log(images.length)
     rerender(
       <LearningResourceCardTemplate
         variant="column"
@@ -50,7 +49,6 @@ describe("LearningResourceCard", () => {
         suppressImage={true}
       />,
     )
-    console.log(images.length)
     expect(screen.queryAllByRole("img").length).toBe(images.length - 1)
   })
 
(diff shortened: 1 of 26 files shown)


frontends/mit-open/src/routes.tsx
```diff
@@ -37,6 +38,10 @@ const routes: RouteObject[] = [
         path: urls.LEARNINGPATH_VIEW,
         element: <LearningPathDetailsPage />,
       },
+      {
+        path: urls.USERLIST_LISTING,
+        element: <UserListListingPage />,
+      },
```
ChristopherChudzicki, line 44: **Suggestion:** I think it makes sense to only show the userlist listing page to authenticated users. 
```tsx
      {
        element: <RestrictedRoute requires={Permissions.Authenticated} />,
        children: [
          {
            path: urls.USERLIST_LISTING,
            element: <UserListListingPage />,
          },
          {
            path: urls.DASHBOARD,                        @CAREY and remove the existing dashboard route
            element: <DashboardPage />,
          },
        ],
      },
```

The details page could be shown to everyone. Userlists can be "private" (only for that user) or "unlisted" (public, but not searchable).

ChristopherChudzicki (commented):
👍 Looks good so far.

frontends/mit-open/src/page-components/LearningResourceCardTemplate/LearningResourceCardTemplate.test.tsx
```diff
@@ -49,6 +50,7 @@ describe("LearningResourceCard", () => {
         suppressImage={true}
       />,
     )
+    console.log(images.length)
```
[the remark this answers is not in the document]
gumaerc: Thanks for the context. I had this logging in here when I was fixing an issue with the test, but that ended up being due to an error in the way I implemented `imageSlot`. Anyway, that's fixed now. Anywho, this test is unchanged from the way it is on `main`: https://github.com/mitodl/mit-open/blob/88a87ec35fa6f1980b16dd871300293bd5d5b58c/frontends/mit-open/src/page-components/LearningResourceCardTemplate/LearningResourceCardTemplate.test.tsx#L33. Were you just pointing that out in case I was expecting to see a change between the two logging locations?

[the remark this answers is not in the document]
ChristopherChudzicki: > Were you just pointing that out in case I was expecting to see a change between the two logging locations?

yes

frontends/mit-open/src/routes.tsx
gumaerc: I thought this was handled in the same way the permissions for the Learning Paths menu item was in `frontends/mit-open/src/page-components/Header/UserMenu.tsx`?

```
...
    {
      label: "User Lists",
      key: "userlists",
      allow: hasPermission(Permissions.Authenticated),
      href: urls.USERLIST_LISTING,
    },
...
```

gumaerc: Oh, I see, that file just prevents it from showing up in the menu and `routes.tsx` prevents it from being routed to entirely

gumaerc:
@ChristopherChudzicki This is ready for another look

frontends/mit-open/src/routes.tsx
ChristopherChudzicki: >  Should we do this with the Learning Path listing page as well?

I'm not sure. At some point that will be a product question.

Right now, an anonymous user visiting http://localhost:8063/learningpaths/ will see a list of all **public** learning paths (b/c that's what the API returns for anonymous users).

ChristopherChudzicki (approved):

PR #653, opened 2024-03-22
Shanbady/retrieve environment config

### LINKED ISSUE #615: Endpoint to retrieve environment config
### Description/Context

Working towards [RFC: Decouple the Frontend from the Backend Web Service #3646](https://github.com/mitodl/hq/discussions/3646).

Currently the web service injects environment config into the base HTML template at [`{{ js_settings|json_script:'js-settings' }}`](https://github.com/mitodl/mit-open/blob/51587f10c0f18d171f28f09b5e7af92cd2525bad/main/templates/index.html#L19C7-L19C48).

When the web service is no longer serving the base HTML, the client app will need to fetch the environment config as it loads.

### Plan/Design

We need an endpoint, `GET /api/v1/environment`, for example, that returns the bootstrap object with these properties:
```
{
  embedlyKey: string
  ocw_next_base_url: string
  search_page_size: number
  ckeditor_upload_url: string
  environment: string
  sentry_dsn: string
  release_version: string
}
```

The user session object is covered by https://github.com/mitodl/mit-open/issues/614.

It may be that these values must not change between release, in which case we have the alternative option of baking them to the front end HTML at build time. The base HTML will be served from CDN with a low TTL and will have the front end code version (or file hashes) included, so we can similarly include the environment values.

shanbady:

Closes #615

This change moves search_page_size, embedlyKey, ocw_next_base_url, sentry_dsn, release_version, and environment from window.SETTINGS (which currently gets in a django template via js_settings) into a webpack injected APP_SETTINGS variable that pulls these from environment variables. 

### How can this be tested?
1. checkout this branch
2. everything should function normally as expected. The settings that are loaded into the global APP_SETTINGS should be pulled in directly from your environment. 

### Additional Context
The env file that this uses pulls from the same env that we use django-side.


frontends/mit-open/webpack.config.js
```diff
@@ -84,6 +84,19 @@ const getWebpackConfig = ({ mode, analyzeBundle }) => {
           env: { NODE_ENV: JSON.stringify(mode) },
         },
       }),
+      new webpack.DefinePlugin({
+        APP_SETTINGS: {
+          embedlyKey: JSON.stringify(process.env.EMBEDLY_KEY),
+          ocw_next_base_url: JSON.stringify(process.env.OCW_BASE_URL),
```
ChristopherChudzicki, line 90: You can remove `ocw_next_base_url`. It was used in https://github.com/mitodl/open-discussions but is not used in MIT Open.

frontends/ol-utilities/src/components/embedly/util.ts
```diff
@@ -82,9 +82,9 @@ const createStylesheet = (doc: Document, css: string) => {
 }
 
 const getEmbedlyKey = (): string | null => {
-  const key = window.SETTINGS?.embedlyKey
+  const key = APP_SETTINGS.embedlyKey
   if (typeof key === "string") return key
-  console.warn("window.SETTINGS.EMBEDLY_KEY should be a string.")
+  console.warn("APP_SETTINGS.EMBEDLY_KEY should be a string.")
```
ChristopherChudzicki, line 87: It's `APP_SETTINGS. embedlyKey` now, right? But if you want to SHOUT CASE the keys, that's ok, too.

ChristopherChudzicki (changes requested):
Can you remove `ocw_next_base_url`? Otherwise looks good 👍

ChristopherChudzicki (approved):

Issue #640, opened 2024-03-20
Search Page Cleanup


ChristopherChudzicki:
### Description/Context
Minor tweaks for the search page:

- [x] Remove the "active facets" display, but keep the "Filters" label and "Clear All" button. "Clear all" should only be visible when facets are active. 
    - *Rationale for removal:* It shows redundant information. Additionally, it causes layout shift when facets get added/removed
- [x] Remove horizontal padding on tabpanels
- [x] Add "Offered by" filter
- [x] Use the same URL query parameters as the search API.
- [x] URL query parameters should work even if the corresponding facet isn't in the UI.
    - This makes demoing potential changes to the app easier, is simple, and doesn't have any real downsides. Additionally, it could be useful for creating "search lists" to be used with channels. 

### Plan/Design
URL parameter changes will be handled in course-search-utils.

steven-hatch:
I would expect A - Clear all resets sidebar filters. 

I am saying this because the "clear all" is in close proximity to filters, and not visually near the tabs. If these were somehow visually connected then clear all would be expected to reset the page. 

This is not related to your work yet, but if user expects to be reset to original search when selecting "clear all" and starts filtering again they my not realize they are still on podcasts tab. The content in the cards, maybe the styling of the card, and empty states need to remind user that they still have 'podcasts' selected (in addition to the tab being selected). Thanks!

PR #675, opened 2024-03-25
Search page cleanup

ChristopherChudzicki:
For #640 

This PR makes several changes to the search page:

1. Removes the "Active Facets" display above facet list. ([These chips](https://github.com/mitodl/mit-open/assets/9010790/d38b271a-20e1-43f7-bf6d-8383cc05648b) have been removed.)
2. Now, if a facet is active, its checkbox should always be shown.
     - Facets options are only included in the API results if their result count is nonzero. So if an active facet corresponds to zero results, it previously was not shown. This situation can occur by toggling a facet on, then entering query text. For example: https://mitopen-rc.odl.mit.edu/search/?q=quaternions&t=Visual+Arts
3. The results title count has been removed. (No more '524 results for "Physics".)
4. "Clear all facets" no longer clears the resource tab selection.
5. Adds "Offered by" facet.
6. Facets now use API query names ("department" not "d")
7. Facets now work even if they aren't explicitly included in UI.
    - e.g., `http://localhost:8063/search?platform=mitx` 

Screenshot 2024-03-25 at 3 01 27 PM

Screenshot 2024-03-25 at 2 50 42 PM

### How can this be tested?
1. Visit http://localhost:8063/search
2. Try:
    - click resource tabs
    - entering query text
    - toggling left sidebar filters on and off
    - using the "Clear all" filters button

    Things to look out for
    - URL bar should use "real" API parameter names ("department" not "d")
3. Append a filter not included in the UI to the URL; for example `?platform=mitxonline` or (`&platform=mitxonline`). The results should update.


frontends/mit-open/src/pages/SearchPage/SearchPage.tsx
```diff
@@ -222,47 +232,95 @@ const TABS: TabConfig[] = [
...
 
+/**
+ * Augment the facet buckets for `groupKey` with active values that have no
+ * results.
+ */
+const includeActiveZerosInBuckets = (
+  groupKey: string,
+  aggregations: LearningResourceSearchResponse["metadata"]["aggregations"],
+  params: LRSearchRequest,
+) => {
```
ChristopherChudzicki, line 243: @abeglova Maybe the `FacetDisplay` component should handle this? OCW does seem to add-in active facets with zero count, but I don't think it's in course-search-utils currently?

abeglova (changes requested):
This works great!

I agree with you that includeActiveZerosInBuckets should be moved to course-search-utils. 

Additionally it looks like you implemented useValidatedSearchParams in course-search-utils but are not using it in this pr - I can search for an invalid param such as http://localhost:8063/search/?topic=monkey and it adds monkey to the topics list

ChristopherChudzicki:
> Additionally it looks like you implemented useValidatedSearchParams in course-search-utils but are not using it in this pr - I can search for an invalid param such as http://localhost:8063/search/?topic=monkey and it adds monkey to the topics list

`useValidatedSearchParams` isn't actually exported from course-search-utils. But it is used in implementing `useResourceSearchParams` and `useContentFileSearchParams`. For example:

```ts
const useResourceSearchParams = ({
  searchParams,
  setSearchParams,
  facets = [],
  onFacetsChange
}: UseValidatedSearchParamsProps<ResourceSearchRequest>) => {
  return useValidatedSearchParams<ResourceSearchRequest>({
    searchParams,
    setSearchParams,
    facets,
    validators: resourceSearchValidators,
    onFacetsChange
  })
}
```

> I can search for an invalid param such as http://localhost:8063/search/?topic=monkey and it adds monkey to the topics list

The validation is based on the OpenAPI spec, by which `"monkey"` is a valid topic. (For topic, the spec just says "string"; department and platform, we have an enum; similarly https://mitopen-rc.odl.mit.edu/api/v1/learning_resources_search/?topic=monkey is a 200 response with zero hits, whereas https://mitopen-rc.odl.mit.edu/api/v1/learning_resources_search/?department=monkey is a 400.)  My preference would be to continue validating based only on the OpenAPI spec...It's not detrimental to the UI (since filter choices are populated off the API response) and keeps things simple.

frontends/mit-open/src/pages/SearchPage/SearchPage.tsx
ChristopherChudzicki: Sounds good. I'd like to do that separately from https://github.com/mitodl/course-search-utils/pull/78, but I can do it before merging this PR.

https://github.com/mitodl/course-search-utils/pull/78 is big enough already, and I'd like replace the Enzyme-based facet tests if I'm going to be adding more there.

abeglova (approved):

Ferdi:
@ChristopherChudzicki if it's not too late can we change the order of tabs to all, courses, programs, podcasts ?

ChristopherChudzicki:
> @ChristopherChudzicki if it's not too late can we change the order of tabs to all, courses, programs, podcasts ?
Screenshot 2024-03-27 at 11 25 45 AM

abeglova (approved):

Issue #629, opened 2024-03-18
CI and deployment config for decoupled front end


jonkafton:
### Description/Context

Working towards [RFC: Decouple the Frontend from the Backend Web Service #3646](https://github.com/mitodl/hq/discussions/3646).

Depends on issue [Load session from API and configure decoupled frontend for local development #616](https://github.com/mitodl/mit-open/issues/616).

The task is to set up the infrastructure and CI pipeline to deploy a static front end build to a fileserver, place a CDN, route traffic to the base html and route API traffic to the Django web service.

As this requires some Devops and URL design and makes some assumptions above the deployment architecture, we may want to convert it to an RFC for further discussion.

Some decisions needed:

- **CDN**: Ultimately we want to load the assets to CDN. Do we want to add CDN distribution in front of RC and Prod so they are like-for-like, or do we consider this overkill for RC (and potentially for Prod while we are pre-live).
- **Heroku**: Heroku is limiting for applications growing in complexity, expensive, and itself a PaaS layer running on top of AWS. For file storage, Heroku [delegates to AWS S3](https://devcenter.heroku.com/articles/s3) and for CDN they [delegate to AWS Cloudfront](https://devcenter.heroku.com/articles/using-amazon-cloudfront-cdn). There isn't necessary much value in adding CDN to Heroku via its [Edge Add-on](https://elements.heroku.com/addons/edge) when it just sends you off to [CloudFront](https://aws.amazon.com/cloudfront), particularly given that it's priced by the hour rather than [by egress](https://aws.amazon.com/cloudfront/pricing/). We can happily deploy directly to S3 and add a CloudFront distribution while Heroku continues to host the backend web service. Are we outgrowing it though? At this point we might consider [AWS Amplify](https://aws.amazon.com/amplify/), which does a similar job to Heroku, but with the benefit that we have the full suite of AWS' underlying as wider ranging requirements emerge. Are we generally invested in AWS? GCP has [Firebase](https://firebase.google.com/) as an equivalent. Are our ETL pipeline deployed to AWS and using OpenSearch?
- **URL Design**: The base HTML will need to be served from our domain roots, which will involve pointing the DNS CNAME to point to the public S3 bucket or CloudFront distribution endpoint. This is currently pointing to the web service on Heroku. We could add routing rules to redirect `/api` paths back to the web service, though it is more efficient to serve the api from its own domain with DNS that resolves directly, e.g. api.mitopen.odl.mit.edu.

### Plan/Design

[...]

Issue #616, opened 2024-03-15
Load session from API and configure decoupled frontend for local development


jonkafton:
### Description/Context

Working towards [RFC: Decouple the Frontend from the Backend Web Service #3646](https://github.com/mitodl/hq/discussions/3646).

Depends on:
- https://github.com/mitodl/mit-open/issues/614
- https://github.com/mitodl/mit-open/issues/615

This task covers building the base HTML template with Webpack and serving it for local development.

There is a reference branch here: https://github.com/mitodl/mit-open/compare/main...jk/frontend-dev-server that provides a base HTML template and adds the `html-webpack-plugin` to inject the front end assets.

The `watch` container currently watches for front end build changes and shares a root directory volume mount with the `web` service. This container can be repurposed to run Webpack Dev Server to serve the front end.

### Plan/Design

- Add the `html-webpack-plugin` and base HTML template.

- Add an NPM script to serve locally:
```
"start-dev:local": "ENVIRONMENT=local NODE_ENV=development webpack serve"
```

- We can also run the front end locally and point to hosted environments, so for RC can add:
```
"start-dev:rc": "API_BASE_URL=https://mit-open-rc.odl.mit.edu ENVIRONMENT=local NODE_ENV=development webpack serve"
```

- Fetch the user session from the new endpoint from https://github.com/mitodl/mit-open/issues/614 (remove the hardcoded `window.SETTINGS` if using the reference branch).

- Front end code to fetch the environment data from the new endpoint from https://github.com/mitodl/mit-open/issues/615 _or_ we may instead bootstrap that during build and can use the [EnvironmentPlugin](https://webpack.js.org/plugins/environment-plugin/).

- For the `watch` container:
  - Remove the `.:/src` volume mount.
  - Move the build context to `./frontends`.
  - I would suggest moving `Dockerfile-node` to `./frontends` and repurposing for Webpack Dev Server.
 
- Clean out the `collectstatic` task from the web service. This gathers static assets for serving in the `web` service, which will now be served by the `watch` service. It might be an idea to rename `web` and `watch` to `backend` and `frontend`. Note we'll want zero impact to Django development, so need to establish workflows (does everyone use `docker compose up`?).

- We should also be able to remove the wait on the `watch` service in the `./scripts/run-django-dev.sh` script for developers wanting to run in production mode locally.

- The `./scripts/run-watch-dev.sh` script can be removed.

- The `./scripts/http-ok.sh` script can be removed.

[...]

PR #678, opened 2024-03-26
Self contained front end using Webpack to build HTML and Webpack Dev Server to serve

jonkafton:

[Load session from API and configure decoupled frontend for local development #616](https://github.com/mitodl/mit-open/issues/616)

This PR isolates the front end so the backend web service is no longer responsible for serving the base HTML or any static assets, with the intention that these are to be served separately from a storage service fronted by endpoints or CDN.

The frontend package root has moved from the project root and into `./frontends/` along with the `./static` images folder.

Nginx has been configured to serve the built frontend files where there is a path match in `./frontends/mit-open/build`, falling back to the index.html, located there. API and admin routes are passed to the Django web service.

This PR **should not be merged** until the task to deploy the static site is completed, [CI and deployment config for decoupled front end #629](https://github.com/mitodl/mit-open/issues/629), as it may prevent release. At minimum we'll need to establish that it is quickly deployable - aiming to minimize impact, the Nginx configuration should be deployable to Heroku to serve the built frontend if we choose to do so while working on #629 , although the `nginx.conf` changes have not been applied to the Nginx `nginx.conf.erb` config for Heroku. 

- Moves the frontend package root into `./frontends` for a self contained front end, decluttering the project root.

- Configures Webpack Dev Server for lightweight front end development (serve and watch the front end in isolation, hot module replacement for quick change feedback).

- Configures Webpack to inject scripts and build the HTML index file (using [HTMLWebpackPlugin](https://webpack.js.org/plugins/html-webpack-plugin/))

- Configures Webpack to copy images (moved from `./static/`) from `./frontends/mit-open/public/images/` to the front end build (uses [CopyWebpackPlugin](https://webpack.js.org/plugins/copy-webpack-plugin/)).

- Configures Webpack Dev Server to proxy API requests to target provided on environment. Overwrites the Origin header to the proxy target for Django admin's CSRF check.

- Configures Nginx to serve on file matches in `./frontends/mit-open/build` falling back to index.html for the single page app. `^/(api|login|logout|admin|static/admin|static/hijack)/` route to the web service on 8061. _Are there any other routes that should be handled there?_.  The project root is no longer mounted to the `nginx` container in the Docker Compose run config, only the frontend build (and Nginx config).

- Updates the `manage.py collectstatic` path to collect images from above (do we need this? intends to preserve dj-static file serving, though `/static/` on the web service should never be hit for front end assets, though is still in use for admin panel assets).

- Provides NPM scripts to run front end dev server in isolation against RC or Prod:
  - `yarn watch` to run the frontend dev server locally, proxying API routes to our locally running Docker Compose stack at `localhost:8063`. Serves on port 8080.
  - `yarn watch:docker` for running inside the Docker `watch` container in build/watch mode, proxying to `nginx:8063`.
  - `yarn watch:rc` for running in isolation in serve/watch mode, proxying API requests to `mitopen-rc.odl.mit.edu`.
  - `yarn watch:prod` for running in isolation in serve/watch mode, proxying API requests to `mitopen.odl.mit.edu`.
  
- Repurposes the Docker Compose `watch` container to run Webpack Dev Server, using `./frontends` as build context and not mount project root. `Dockerfile-node` moved into `./frontends`. The watch container doesn't now serve on any port - it's job is to build the frontend on start so it is available to be served by Ngnix so as not to impact the application running locally with the single `docker compose up`. Also watches for changes. `run-watch-dev.sh` script removed and wait on `watch` logic removed from `run-django-dev.sh`.

- Github Actions CI jobs updated for changes to project structure.

- UI fetches the user session from the API at `GET /api/v0/users/me/` so it is no longer reliant on `window.SETTINGS`. Unit tests updated to either mock the endpoint to expect the call to it, or to load the user into the TanStack Query cache where tests depend on first component render.

#### Housekeeping:

- Upgrade yarn to 4.1.1 (in container build and across workspaces).
- Adds `private: true` to all package.json files (yarn workspaces requirement).
- Uses the lighter Alpine distribution as the base image for frontend Dockerfile.
- Upgrade webpack-dev-server v4->v5.
- Upgrade webpack-bundle-tracker v1->v3.
- Refactor down the Webpack config, removing static file path logic not needed (removes build env vars `MITOPEN_HOSTNAME` and `WEBPACK_PORT_MITOPEN`).
- Configures Webpack to clean the build folder each build (no stale build files).
- Supported architectures added to `.yarnrc.yml` to build the `@swc/core` Linux binaries for the `watch` container (fixes "bindings not found" error).
- Yarn lockfile and releases ignored from the `check-added-large-files` pre-commit check.

#### Not done:

- The Nginx config for Heroku, `nginx.conf.erb` is not updated to align with `nginx.conf`, to be covered by [CI and deployment config for decoupled front end #629](https://github.com/mitodl/mit-open/issues/629) and likely not needed if we go straight to deploying the front end on an S3 static site and CloudFront or similar.

- Reimplement Django Hijack, issue here: https://github.com/mitodl/mit-open/issues/733.

### How can this be tested?

**Existing workflow**:

- There should be no changes to existing workflows for local development - `docker compose up` should work as before and serve the application at `http://localhost:8063`.
- Login and logout should redirect to OIDC.
- All admin routes should function correctly.
- Changes to front end source are built by the watch container and updated on page refresh.

**New frontend dev server**:

Copying README notes over from [Frontend Dev Server with Local Backend](https://github.com/mitodl/mit-open/blob/fc7a4dd985984594ce794d18360a059dccedbc4f/frontends/README.md?plain=1#L61). This all needs to test correctly:

  - **Frontend Dev Server with Local Backend**. In this mode the front end is served with Wepback Dev Server, enabling Hot Module Replacement and faster feedback to changes. Dev Server proxies API requests through to a locally running backend stack. The front end is served at `http://localhost:8080` and changes are hot reloaded into the page.

```bash
docker compose up nginx web db
yarn watch
```

  - **Frontend Dev Server proxying to RC or Prod**. When working on the front end in isolation or to test changes against APIs already running in RC and Prod, the frontend dev server is configured to run against our hosted API environments without running the backend stack locally. The front end is served at `http://localhost:8080` and changes are hot reloaded into the page. Run the front end with one of:

```bash
yarn watch:rc
yarn watch:prod
```

### Additional Context

[RFC: Decouple the Frontend from the Backend Web Service #3646](https://github.com/mitodl/hq/discussions/3646)


config/nginx.conf
```diff
@@ -13,21 +13,18 @@ server {
...
     location = /.well-known/dnt-policy.txt {
         return 204;
     }
 
-    location = /favicon.ico {
-        try_files /static/images/favicon.ico /favicon.ico;
-    }
-
-    location / {
+    location ~ ^/(api|login|logout|admin|static/admin|static/hijack)/ {
```
jonkafton, line 27: Have I missed any routes for the web service here?

channels/views_test.py
```diff
@@ -174,7 +174,9 @@ def test_patch_field_channel_image(client, field_channel, attribute):
     png_file = os.path.join(  # noqa: PTH118
         os.path.dirname(__file__),  # noqa: PTH120
         "..",
-        "static",
+        "frontends",
```
jonkafton, line 177: Does this path change affect the Django app elsewhere? I'm not clear on what it's testing.

frontends/.yarnrc.yml
```diff
@@ -0,0 +1,19 @@
...
+yarnPath: .yarn/releases/yarn-4.1.1.cjs
+
+# https://github.com/vitejs/vite-plugin-react-swc/issues/74#issuecomment-1520484130
+# https://github.com/swc-project/swc/issues/5616#issuecomment-1265639797
+# This setting ensures we always install the Linux binaries when running `yarn install`. This is needed for running
+# swc natively in Docker (addresses swc-loader bindings not found error).
+supportedArchitectures:
+  os:
+    - current
+    - linux
```
ChristopherChudzicki, line 12: This is nice. After yarn installing on my host (as well as on docker up) I was able to compile files in both docker and on host. Running the tests is a lot faster on host.

frontends/Dockerfile
```diff
@@ -0,0 +1,16 @@
+FROM node:20.12-alpine
```
ChristopherChudzicki, line 1: It might be nice to use something with bash. I think people bash into containers somewhat regularly. `sh` works, too, though.

frontends/README.md
```diff
@@ -31,7 +33,60 @@ Most workspaces are shared dependencies built using typescript and tested with j
...
+
+### Frontend Dev Server proxying to RC or Prod
+
+When working on the front end in isolation or to test changes against APIs already running in RC and Prod, the frontend dev server is configured to run against our hosted API environments without running the backend stack locally.
+
+Run the front end with one of:
+
+```bash
+yarn watch:rc
+yarn watch:prod
```
ChristopherChudzicki, line 89: I am personally a bit hesitant to add these commands. I don't like the idea of someone's `localhost` hitting production APIs. Seems too easy to inadvertently change something on production. The site is largely read-only, but not entirely. **And some of the writeable data is publicly viewable.** So people could edit things on localhost and end up changing public production data.

frontends/.yarnrc.yml
```diff
@@ -0,0 +1,19 @@
+nodeLinker: node-modules
+
+yarnPath: .yarn/releases/yarn-4.1.1.cjs
```
ChristopherChudzicki, line 3: After `docker compose up`, I see `frontends/.yarn/install-state.gz` as an untracked file.

frontends/mit-open/src/components/RestrictedRoute/RestrictedRoute.tsx
```diff
@@ -41,7 +38,9 @@ const RestrictedRoute: React.FC<RestrictedRouteProps> = ({
   children,
   requires,
 }) => {
-  if (!hasPermission(requires)) {
+  const { isLoading, data: user } = useUserMe()
+
```
ChristopherChudzicki, line 42:
```suggestion
if (isLoading) return null
```
IMO better to show nothing if the userquery is loading.

Since the header loads `useUserMe`, users navigating around the app would probably not trigger the early return. But without it, someone going directly to a restricted page (typing in the URL, following an external link) would get a flash of the page before being redirected.

config/nginx.conf
ChristopherChudzicki: I'm not sure it's released yet, but there is an API http://localhost:8063/_features/. Possibly it should be api/v0. It's private insofar as it's a list of feature flags. On the other hand... the repository is public.

ChristopherChudzicki:
### Watch mode

I'm having trouble running the frontend in a watch mode. After
```
docker compose up --build
```
I hit http://localhost:8063/ and see the frontend, but it does not reload when I save changes to a file. I can see changes if I reload the page with `cmd+r`.

-------------------------------------- 
### Images
@rhysyngsun Would you be able to review the image-related portion of this? It certainly makes sense to me to not serve the images via django.
---------------------------------------
**Breaking it up:** Given issues noted above and my request for someone with more django experience to look, maybe it makes sense to move some the changes to a separate PR? E.g., I think the user API queries are most likely to cause nontrivial  merge conflicts. Do you think it would be feasible to move that to a separate PR?

ChristopherChudzicki (changes requested):
### Watch mode

I'm having trouble running the frontend in a watch mode. After
```
docker compose up --build
```
I hit http://localhost:8063/ and see the frontend, but it does not reload when I save changes to a file. I can see changes if I reload the page with `cmd+r`.

-------------------------------------- 
### Images
@rhysyngsun Would you be able to review the image-related portion of this? It certainly makes sense to me to not serve the images via django.

---------------------------------------

**Breaking it up:** Given issues noted above and my request for someone with more django experience to look, maybe it makes sense to move some the changes to a separate PR? E.g., I think the user API queries are most likely to cause nontrivial  merge conflicts. Do you think it would be feasible to move that to a separate PR?

jonkafton:
> I hit http://localhost:8063/ and see the frontend, but it does not reload when I save changes to a file. I can see changes if I reload the page with `cmd+r`.

That is by design - some notes on this added here https://github.com/mitodl/mit-open/blob/fc7a4dd985984594ce794d18360a059dccedbc4f/frontends/README.md?plain=1#L59

It's the only concession on previous behavior, as we now use Nginx on port 8063 to serve the static files - it's not able to hot reload. Hot reload is now provided by Webpack Dev Server, though the watch container is now running Webpack in watch and build mode. We can't easily have it both build to file and bundle in memory and hot reload at the same time (likely possible, though we'd need to then serve on a new port, separately from Nginx). The idea is that the watch container is used when working on the backend and just ensures the front end build is synced to source, though for front end development there's no real benefit for using the watch container as there's a significantly improved developer experience (reload speed, local memory footprint) by running directly with `yarn watch`.

frontends/README.md
jonkafton: Perhaps we shouldn't encourage using the Prod API for local development, though it's intended to verify front end changes against a released API.

Prod and RC should have access controls in place to make sure that no one is able to make changes without the necessary permissions. I don't see a huge distinction between a developer running the front end locally on hitting Prod and a developer loading the Prod site (both are hitting production APIs from localhost) though I did consider adding a prominent dev banner when running in the mode. 

I do think a frontend developer who has no reason to fire up the backend stack locally (it's cumbersome), should generally do their work against RC (in the absence of any hosted dev or PR validation environment). If we're working on some feature that needs super user that we don't have on RC, we'd need to fire up the backend locally, so should be covered.

frontends/Dockerfile
jonkafton: I normally just use `sh` to poke around containers, but let me know if you'd like me to install `bash`.

frontends/.yarnrc.yml
jonkafton: Added to the ignore file.

channels/views_test.py
rhysyngsun: This particular change is fine.

config/nginx.conf
rhysyngsun: It's not ideal to have a fixed mapping, nginx provides the capability to avoid all of this anyway:
```suggestion
    location @proxy {
```

And the `try_files` directive above becomes:

```
try_files /static$uri $uri @proxy;
```

docker-compose.yml
```diff
@@ -24,6 +22,8 @@ services:
       - 5432:5432
     environment:
       - POSTGRES_PASSWORD=postgres
+    volumes:
+      - pgdata:/var/lib/postgresql/data
```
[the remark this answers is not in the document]
rhysyngsun: This might wipe out anyone's local state, before you merge this you should make everyone very aware they should do a backup before pulling down these changes.

frontends/Dockerfile
```diff
@@ -0,0 +1,16 @@
...
+LABEL maintainer "ODL DevOps <mitx-devops@mit.edu>"
+
+COPY . /src
+
+WORKDIR /src
+
+RUN yarn set version 4.1.1
+RUN yarn install
```
rhysyngsun, line 9: Should this be `yarn install --immutable`?

frontends/README.md
rhysyngsun: I think it's better if we can have this done via other configuration like environment variables.

config/nginx.conf
jonkafton: I was hoping for a better solution. The issue is that we need to serve the index.html at unknown routes for the frontend to load at all of its paths, so that needs to be the static file fallback. We could instead list all SPA handled first path parts, or list the backend's. This seemed the lesser of evils, though perhaps there's another option?

jonkafton: I've added `/_features` for the time being.

docker-compose.yml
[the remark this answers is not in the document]
jonkafton: I've removed it, probably best. We can revisit in a separate PR if any else is hitting the issue.

frontends/README.md
jonkafton: These are just setting API_BASE_URL on the environment. I've removed the package script for proxying to Prod just in case.

frontends/Dockerfile
jonkafton: Updated.

config/nginx.conf
rhysyngsun: Ok, that's fair, let's leave this for now then. I guess at the very least the failure modes for the backend urls are more conspicuous.

docker-compose.yml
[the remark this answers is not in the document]
ChristopherChudzicki: Echoing Nathan: In principle, I do think this is worth having, since it would allow you to preserve data between `down`s but still give you the flexibility to destroy the volume via `down -v` if you want to.

ChristopherChudzicki:
re `docker compose up` and watching the frontend 

> It's the only concession on previous behavior, as we now use Nginx on port 8063 to serve the static files - it's not able to hot reload. 

After talking with @rhysyngsun, I do think we need to preserve the existing behavior: `docker compose up` should get you a working stack with watchable frontend. Some of the engineers don't have—and can't easily run—Node on their host machines. 

If necessary, running the frontend on a separate port seems like it could be OK. I don't really have enough (hardly any) nginx experience to tell whether that is necessary.

jonkafton:
> > It's the only concession on previous behavior, as we now use Nginx on port 8063 to serve the static files - it's not able to hot reload.
> 
> After talking with @rhysyngsun, I do think we need to preserve the existing behavior: `docker compose up` should get you a working stack with watchable frontend. Some of the engineers don't have—and can't easily run—Node on their host machines.
> 
> If necessary, running the frontend on a separate port seems like it could be OK. I don't really have enough (hardly any) nginx experience to tell whether that is necessary.

I didn't envisage anyone working on the front end would be interested in running Dev Server inside Docker. It's quite a bit slower to pick up changes (perhaps related to [watchpack](https://www.npmjs.com/package/watchpack) compatibility with the Docker filesystem volume mounts). I found hot reload to take 10-15s in Docker and 2-3s running directly, though also heavily dependent on Docker resource allocation.

We're now using Nginx as our static fileserver and could instead point it to the `watch` container inside the Docker network to get the front end on 8063. The issue is that any backend development is dependent on also always running the `watch` container (running Webpack Dev Server), which is slower to start and serve.

I've left 8063 as-is and the `watch` container still listens and rebuilds any changes, though a few things where needed to get HMR working on port 8062:

- The `watch` container starts Webpack Dev Server in addition to only watching/building with Webpack and exposes 8062 in the Docker Compose config.
- Sets Webpack Dev Server to also write to file (for the built static files to still be available to Nginx on 8063).
- Removes the Hot Module Replacement fallback to live reload / page refresh (`hot: "only"`).
- Adds a fix in App.tsx for an issue in Webpack 5 breaking HMR ([info](https://github.com/webpack-contrib/webpack-hot-middleware/issues/390)). Applies only if `ENVIRONMENT` is `"docker"`.
- Uses the [EnvironmentPlugin](https://webpack.js.org/plugins/environment-plugin/) to provide the env var in app on `process.env`.
- Set the Dev Server HMR socket to listen on 0.0.0.0 (Docker network interfaces).

Definitely a good improvement, the only concession with this PR now being that for frontend changes 8063 needs manual refresh, while the new 8062 hot reloads.

Changes here: https://github.com/mitodl/mit-open/pull/678/commits/a0c147546

ChristopherChudzicki (changes requested):
Thanks for restoring the `docker compose up` watch behavior. I think keeping that workflow is important. 

I did notice a few issues:

1. **API routes don't seem to have their normal UI.**:  For example, visit http://localhost:8063/api/v1/courses/ in your browser and compare to https://mitopen-rc.odl.mit.edu/api/v1/courses/ [^1]
2. **http://localhost:8063/admin** Doesn't work. (But `/admin/` does).
3. **Webpack HMR error:** `You are calling ReactDOMClient.createRoot() on a container that has already been passed to createRoot() before. Instead, call root.render() on the existing root instead if you want to update it.`
     - I was able to resolve this by using https://www.npmjs.com/package/@pmmmwh/react-refresh-webpack-plugin. The docs say it's experimental, but a cursory glance at github history makes me think it's fairly stable, and it's NPM usage is...large.  
4. **`yarn watch` env vars** Running `docker compose up nginx; cd frontends; yarn watch` isn't picking up environment variables. If we want running the dev server outside of docker to be an option, it should pick up environment variables.

**Additionally:** In my opinion, there are too frontend-related ports:
- 8062 for HMR with docker
- 8080 for HMR without docker
- 8063 for docker but no HMR 

If we can't clearly put everything on 8063, then IMO it would be better to have 8063 as backend-only and 8062 as frontend only, and not use 8080 at all. 

[^1]: If both links work for you, it might be an issue on my end. I'm testing this on a bus without the best wifi. I don't think that should matter since I am pretty sure all the assets are local. And I do have _some_ wifi.

ChristopherChudzicki (approved):
👍 This is working well for me now. All issues from https://github.com/mitodl/mit-open/pull/678#pullrequestreview-1999251394 look resolved 👍 

> So we now have:
> - 8062 - Dev Server running locally or inside the and providing HMR.
> - 8063 - Docker but no HMR.

HMR actually seems to be working on 8063 for me. So that's nice.


### ⚠️ One requested change 
That I do not think warrants a re-review. The line count was a red flag... `+29,754 −1,320 `. Could you remove the root-level `yarn.lock`?

### Merging
I tagged this "blocked" for now. Is it? What needs to be done before merging?

jonkafton:
> ### ⚠️ One requested change
> That I do not think warrants a re-review. The line count was a red flag... `+29,754 −1,320 `. Could you remove the root-level `yarn.lock`?

Removed. I suspect it snuck its way back in during merge.

> ### Merging
> I tagged this "blocked" for now. Is it? What needs to be done before merging?

I've updated the Nginx config for Heroku to align with our local container config in this commit: https://github.com/mitodl/mit-open/pull/678/commits/5aa641ca4.

We'll need to do a hands on merge into main to monitor RC as I'm not sure whether we can test the Nginx config in advance. We could always deploy to a new Heroku environment first, but may not be worthwhile as we can quickly roll back RC while we try the release.

The other aspect hard to test before attempting the release is that Nginx is now expecting to find the built front end on Heroku in /app/frontends/mit-open/build to serve the static content. I've added a step to upload the the build as an artifact during the javascript-tests job (we may want a dedicated job, though we have it here and can avoid building twice). This should be fine as the `release-candidate` and `production`  workflows trigger from the `ci` workflow, though there may be an issue sharing between workflows if they are treated separately and possibly the filepath it arrives on in Heroku. The existing Nginx has a root at `/app`, though this isn't a project directory. It may be a Heroku dyno default path - we'll need a trial run.

PR #682, opened 2024-03-27
Add Posthog integration to backend

### LINKED ISSUE #589: Feature flag support (Posthog)
### Description/Context
We need support for feature flags in Open, at a minimum we want:

- Standard feature flags
- A/B testing

### Plan/Design

Currently we don't have anything for Posthog setup in this project, it only has our traditional environment variable driven flags.

- Add django implementation of Posthog
  - Incorporate learnings from https://github.com/mitodl/mitxonline/issues/2116
- Add Posthog to frontend
  - Determine if any of the learnings from the above issue could also be applied to the frontend or if we need to consider an alternative approach.

jkachel:

Partially closes #589 

Adds support for querying feature flags via Posthog, with fallback to the standard local environment. Posthog support is toggleable and queries to Posthog are cached locally in the database in case their API becomes unavailable.

To configure:
- Set the Posthog specific `.env` file settings:
   - `POSTHOG_PROJECT_API_KEY` - the API key for the project the app should look at
   - `POSTHOG_HOST` - should be `https://app.posthog.com` generally but can be changed if needed
   - `POSTHOG_ENABLED` - boolean
- Run `./manage.py createcachetable` to create the table for the durable cache

### How can this be tested?

Automated tests should pass.

The code exposes an internal meta API at `/_/features` that provides read operations on feature flags. Listing feature flags will additionally load the cache for the default case. (Posthog identifies users via a ID for the instance and an ID for the user; the default is to use the app's hostname for the instance ID, and the environment and user ID or None for the user ID, so this is what you'll get here.) The features API is intended as a sanity-check so user-specific flag settings aren't there (but we could add that).


main/features.py
```diff
@@ -1,34 +1,180 @@
...
+            person. I recommend this be user.id for logged-in users to allow for
+            more readable user flags as well as more clear troubleshooting. For
+            anonymous users, a persistent ID will help with troubleshooting and tracking
+            efforts.
 
     Returns:
-        bool: True if the feature flag is enabled
-    """  # noqa: D401
-    return settings.FEATURES.get(name, default or settings.MITOPEN_FEATURES_DEFAULT)
+        bool: True if the feature flag is enablede
```
shanbady, line 143: spelling * enablede

pyproject.toml
```diff
@@ -78,7 +78,7 @@ drf-nested-routers = "^0.93.5"
 django-scim2 = "^0.19.1"
 django-oauth-toolkit = "^2.3.0"
 youtube-transcript-api = "^0.6.2"
-ruff = "0.3.4"
```
shanbady, line 81: was removing ruff 0.3.4 intentional?

shanbady (changes requested):
looks ok. left some comments about minor stuff

pyproject.toml
jkachel: think this was a rebase error... I put it back but looking at it I'm not sure why it's here and in dev dependencies (with a different version?) but that's sort of out of scope for this PR.

ChristopherChudzicki: I believe it's in there for IDE support. pre-commit is the source of truth for linting/formatting, but IDEs (to my knowledge) can't run pre-commit's ruff.

It's sometimes a bit out of date with pre-commit's, but they both get updated (pre-commit by its autoupdate, dev deps by renovate) and if they're slightly out of sync, it's not a big deal.

shanbady (approved):
looks good.

Issue #642, opened 2024-03-20
Learning Resource Drawer


ChristopherChudzicki:
### User Story
As a user looking for resources, I want more details than offered by the resource cards.

### Description/Context
We should adapt the learning resource drawer for the `/api/v1/learning_resources` resource format. 

### Acceptance Criteria
- [ ] Clicking a course card should open the learning resource drawer.
- [ ] Drawer should show info similar to drawer at https://discussions-rc.odl.mit.edu/learn/

PR #686, opened 2024-03-27
Basic learning resources drawer

ChristopherChudzicki:
#642 

This PR adds a basic learning resource drawer:
- minimal resource metadata for now
- renders playable video for video, image for everything else
- opens by clicking title of LearningResourceCard

Also:
- Moves `EmbedlyCard` to the `ol-components` package. This seems like a better place for it. This is why so many files are touched.

**Note:** Resource descriptions can contain HTML strings. Currently, I'm rendering this as a string. If we want to render as HTML (not sure we do) we can address that in a followup PR.

Screenshot 2024-03-29 at 4 22 13 PM

### How can this be tested:

1. View the "ExpandedLearningResourceDetails" story.
    - `docker compose up`
    - `docker compose exec watch yarn storybook`
    - View the last "story" on left hand sidebar, "ExpandedLearningResourceDetails". This shows roughly what the card looks lik. There's a dropdown where you can pick different resource types. Right now only "Video" is different from the others.
2. Go to `htpp://localhost:8063`. Or view the searchpage.  Any learning resource card should have a clickable title that opens the resource drawer.

Some finer points to look for:
- When the drawer content is loading, it should show a skeleton UI. 
- The drawer content should be visible even while the drawer is opening/closing. (This is a little hard to see unless you add something like `transitionDuration={2000}` to the RoutedDrawer in LearningResourceDrawer.tsx


shanbady (approved):
changes look good although I was frequently noticing a delay in the image load. dont think its a blocker but i'm wondering if there is an easy way to eliminate the brief flash before the image load

PR #692, opened 2024-03-28
Add botocore to ignored deprecation warnings, remove old python 3.7 ignore line

mbertrand:
Hopefully gets tests to pass on #349

Adds `botocore` to 3rd-party deprecation warning ignore list
Removes an old ignore line for a deprecation warning related to python 3.7

### How can this be tested?
- Tests should pass
- If you apply this change locally to #349 (after rebuilding docker containers), tests should pass there too.

PS I initially tried upgrading boto3 and moto to the latest versions on the PR 349 branch, but the deprecation warning was still present.


shanbady (approved):
looks good. appears to have also worked in merge with the renovate python upgrade branch

PR #693, opened 2024-03-28
Adds Posthog support to the frontend.

### LINKED ISSUE #589: Feature flag support (Posthog)

jkachel:

Closes #589 

Adds PostHog integration to the frontend with these settings:
- The timeout is configurable in the `.env` file
- The PostHog provider is pre-loaded with the feature flags the Django app knows about
- Enabling the provider can also be set in the `.env` file.

The PostHog provider is situated near the top of the provider stack for the app - right after `StrictMode` - which should allow for feature flag checking and event collection for all parts of the app.

### How can this be tested?

Automated tests should pass.

Test setup of the PostHog provider: If the `POSTHOG_ENABLED` flag is set to `True`, you should see the `PostHogProvider` in the Components tree in the React dev tools. You should not see it if the flag is set to False. 

Test pre-loading of flags: With the provider enabled, you should be able to see the flags that the Django app knows about pre-loaded into the provider by navigating to the provider component in React dev tools and looking through the options. The preloading will use the currently logged-in user's username as the unique user ID for retrieving flags. Doing it this way should also load the flags from cache (see PR https://github.com/mitodl/mit-open/pull/682) if there's a cached value.


frontends/mit-open/src/AppProviders.test.tsx
```diff
@@ -0,0 +1,52 @@
...
+    window.SETTINGS = {
+      ...window.SETTINGS,
+      posthog: {
+        api_key: "", // pragma: allowlist secret
+        enabled: true,
+      },
+    }
+
+    const resources = factory.resources({ count: 4 })
+    setMockResponse.get(urls.learningResources.list(), resources)
```
ChristopherChudzicki, line 25: `renderTestApp` renders the whole app, which makes some API calls, you you had to mock, which is a pain.

I would recommend using `renderWithProviders` for this:
```
renderWithProviders(<div data-testid="some-children" />)
```
then you won't need to mock the API calls.

main/views.py
```diff
@@ -31,6 +33,12 @@ def index(request, **kwargs):  # pylint: disable=unused-argument  # noqa: ARG001
             and (is_admin_user(request) or is_learning_path_editor(request)),
             "is_article_editor": is_admin_user(request),
         },
+        "posthog": {
+            "api_key": settings.POSTHOG_PROJECT_API_KEY,
+            "enabled": settings.POSTHOG_ENABLED,
+            "timeout": settings.POSTHOG_TIMEOUT_MS,
+            "bootstrap_flags": all_flags,
+        },
```
ChristopherChudzicki, line 41: Recently @shanbady did some work to remove environment settings from the django view. Instead, any env variables that the frontend needs can be injected via the webpack build: https://github.com/mitodl/mit-open/blob/87fa1ad3e93905a391ad7cbb18c09fb1ed771904/frontends/mit-open/webpack.config.js#L87-L99

This is part of some work to make the HTML static so that it can be served by a CDN.
- https://github.com/mitodl/hq/discussions/3646

ChristopherChudzicki (changes requested):
https://github.com/mitodl/mit-open/blob/87fa1ad3e93905a391ad7cbb18c09fb1ed771904/frontends/mit-open/webpack.config.js#L87-L99

main/views.py
ChristopherChudzicki: (`user` is still injected via django, Jon is doing some work to stop that and get user info via an api)

jkachel: I changed this to not depend on Django injection at all - however this does change how the flag bootstrapping works; instead of being able to pull from the durable cache present there, it'll use the settings in the `.env` file instead. The default is for it to look at items that begin with `FEATURE_` and this is configurable by setting `MITOPEN_FEATURES_PREFIX` (which will also change it in the Django app). It doesn't do any processing of the feature flag name other than stripping the prefix so it's up to us to make sure these agree between the app and the PostHog settings. (The code does normalize a Python "True" into a JavaScript true - other values are left as-is.)

ChristopherChudzicki: Could we remove the `posthog` HTML injection all together now?

frontends/mit-open/src/AppProviders.tsx
```diff
@@ -17,23 +20,42 @@ interface AppProps {
...
-          </HelmetProvider>
-          <ReactQueryDevtools
-            initialIsOpen={false}
-            toggleButtonProps={{ style: { opacity: 0.5 } }}
-          />
-        </QueryClientProvider>
-      </ThemeProvider>
+      <PostHogProvider apiKey={phSettings.api_key} options={phOptions}>
+        {interiorElements}
+      </PostHogProvider>
```
ChristopherChudzicki, line 55: Would be nice if PostHog's provider had an `enabled` prop that made it a no-op of `enabled=false`.

ChristopherChudzicki (approved):
👍 

One request, but IMO no need to re-review

PR #717, opened 2024-04-02
Shanbady/add record hash field for hightouch sync

### LINKED ISSUE #714: Add record hash field for hightouch sync
### Description/Context
The program certificate model/table needs to add a record hash field for hightouch to use as a primary key for the certificate sync

shanbady:

Closes #714

This PR adds a "record_hash" field that hightouch requires for sync

### How can this be tested?
Checkout this branch. do a docker-compose  restart all or docker-compose down/up and see that the external "program certificate" table has a record_hash field.

### Additional Context
This field was added to a table sitting in an external schema and is completely unmanaged. The PR contains several migration files because the sequence of migrations looks like so:
1. seet model too unmanaged
2. introduce new nullable non-unique record_hash field
3. populate existing models using uuid
4. make record_hash non-nullable and unique
5. unset the 'managed' flag on programcertificate model


profiles/models.py
```diff
@@ -140,6 +140,15 @@ def __str__(self):
 class ProgramCertificate(models.Model):
     """An external model that syncs with data from our data platform"""
 
+    # an external hash that hightouch uses for the sync
+    record_hash = models.CharField(
```
blarghmatey, line 144: This needs to be treated as the primary key field for the table. There isn't any semantic utility in maintaining the auto-generated `id` field for this table

shanbady: made row_hash the primary key and dropped id

blarghmatey (approved):

PR #723, opened 2024-04-03
adding initial models for user subscription

### LINKED ISSUE #709: Create PercolateQuery and PercolateQueryUser models for subscribe feature
We need two new models for saving user search subscriptions
- PercolateQuery which holds the query to save. This model also has a 'source type' field to note the purpose/source. We should save the original "search input" in addition to the opensearch query
- PercolateQueryUser which ties a user to the query

shanbady:

Closes #709

This PR introduces 2 models used to manage percolate queries:
 - PercolateQuery stores an opensearch query along with mit-open search params
 - PercolateQueryUser associates the PercolateQuery with a user


learning_resources_search/models.py
```diff
@@ -0,0 +1,47 @@
...
+    source_type = models.CharField(
+        max_length=255, choices=[(choice, choice) for choice in SOURCE_TYPES]
+    )
+
+    def __str__(self):
+        return f"Percolate query {self.id}: {self.query}"
+
+
+class PercolateQueryUser(TimestampedModel):
+    """
```
mbertrand, line 30: What do you think of adding a `ManyToMany` relationship to `User` on the `PercolatorQuery` model instead of creating this model?  I think the end result would be a similar table in the database.

```python
users = ManyToManyField(
    User, 
    related_name="percolate_queries"
). # many users possibly associated to query, many queries possibly associated to user
```

mbertrand (commented):
Looks good, have a suggestion for an alternative to the 2nd model but I might be missing some context.

learning_resources_search/models.py
shanbady: The reason behind that intermediate model is to have something that stores the reason for subscribing to the query. so for example, in future use cases we might have a "subscribe to physics" newletters in which case we would have a broad query stored "topic=physics" - some users may have saved this via site-search and subscribed to it via the button and other users are subscribed to it through some newsletter campaign.

mbertrand (approved):

PR #731, opened 2024-04-04
Configurable, Tabbed Carousels

ChristopherChudzicki:
https://github.com/mitodl/hq/issues/3940

This PR:
- Removes the existing TitledCarousel component, replacing it with simpler `Carousel`.
- Adds a `TabbedCarousel` component. See screenshot. The idea is to create a reusable tab+tabpanel component where each tabpanel has carousel, and the carousel data is pulled from an API based on the configuration.

Screenshot 2024-04-04 at 3 36 27 PM

### How can this be tested?
1. View the homepage. It should have two "Tabbed Carousels", one for upcoming courses, the other for Media. They should work. (try clicking tabs, clicking the buttons, etc).
    - NB: The "Professional" tab in Upcoming Courses might be empty unless you're using production xpro data.
3. You might like to try changing the tab configuration.
    - In frontends/mit-open/src/pages/HomePage/HomePage.tsx, try changing `UPCOMING_COURSES_CAROUSEL` or `MEDIA_CAROUSEL`.  You can change the `params` objects to change the API filters, or add a new object to get a new tab.


frontends/ol-components/src/components/Carousel/Carousel.stories.tsx
ChristopherChudzicki, line 1: If you want to view this in action:
```
docker compose exec watch yarn storybook
```
then view the "Carousel" story.

e2e_testing/tests/002_carousel.spec.ts
ChristopherChudzicki, line 1: @jonkafton I decided to remove a bit of this test so as not to block this PR.
- I'm not sure the best way to deal with the `learning_resources/upcoming` API. It pulls from data strictly in the future (with a few other restrictions). I guess we could just set the course dates to be far in the future, but that doesn't seem great.
- I had a hard time running the e2e tests locally. (I haven't succeeded yet, but am going to try a little more based on the instructions in https://github.com/mitodl/mit-open/pull/481. Do we have that written down anywhere else?)

frontends/mit-open/src/page-components/TabbedCarousel/TabbedCarousel.tsx
```diff
@@ -0,0 +1,142 @@
...
+ *  - each TabConfig generates a tab + tabpanel that pulls data from an API based
+ *    on the config
+ *  - data is lazily when the tabpanel first becomes visible
+ */
+const TabbedCarousel: React.FC<TabbedCarouselProps> = ({ config }) => {
+  const [tab, setTab] = React.useState("0")
+
+  return (
+    <TabContext value={tab}>
+      <TabList onChange={(e, newValue) => setTab(newValue)}>
```
gumaerc, line 115: Wasn't exactly sure where to put this comment as it's more of a suggestion, but here seems appropriate. Something I noticed here is that when you switch tabs, the currently rendered `Carousel` just instantly vanishes, and then the next one takes some time to load. When it does load, it just instantly appears. This all feels a little jarring. I don't know if it's been discussed in any design meetings, but it might be better if we did something here like what the Reorder functionality does in `SortableList`? While moving items, they have `disabled` set so that the whole UI gets slightly greyed out while it's talking to the server. I realize that this is in place to prevent you from sending off conflicting API requests for ordering, but it also provides a nice visual transition when loading data. Something similar here would be appropriate, in my opinion.

gumaerc (changes requested):
This mostly works fine, just a few suggestions as well as an incorrect icon I noticed:

frontends/mit-open/src/page-components/TabbedCarousel/TabbedCarousel.tsx
gumaerc: (If it wasn't clear this is just a suggestion and can be ignored)

ChristopherChudzicki: I think this is a good suggestion. There are a few ways we could handle it...
1. the panels can be pre-loaded. That's more API requests, but they're in parallel, so seems ok.
2. or we could have a skeleton ui or something for placeholder cards
3. or we could try to leave the existing panel in place until the new data has loaded, grayed out a bit like in the sortable list.

I expect we will do one of these things. But I'm inclined to leave it as is for now. I rather expect the carousel design to change a bit. (Arguably (1) and (3) aren't really design issues.)

gumaerc (approved):
👍 LGTM!

PR #740, opened 2024-04-09
channel search ui

abeglova:
closes https://github.com/mitodl/mit-open/issues/687

This pr adds a filtered search section to field pages when search_filter is set on the field

Screenshot 2024-04-09 at 9 05 15 AM

### How can this be tested?
Either make a new field using http://localhost:8063/api/v0/channels/ with search_filter set to something like `offered_by=xpro,bootcamps`

Or go to http://llocalhost:8063/admin/channels/fieldchannel/ and set search_filter from there for an existing field. There is currently a bug so you need to add some json, for example `{"ok": "ok"}` to the configuration field to be able to save it. 

Go to http://llocalhost:8063/c/[field type]/[filed name]. Verify that you see a search section with dropdowns. Verify that the search is filtered to xpro and bootcamps learning resources and you are able to add additional filters by toggling the drop downs. You should also be able to use the text search

Clear out search_filter from the admin page .

Go back to Go to http://localhost:8063/c/[field type]/[filed name]. Verify that you no longer see the search section.

### DIFF
--- a/frontends/api/src/clients.ts
+++ b/frontends/api/src/clients.ts
@@ -7,6 +7,7 @@ import {
   ArticlesApi,
   ProgramLettersApi,
   LearningResourcesSearchApi,
+  PlatformsApi,
 } from "./generated/v1/api"
 
 import { ChannelsApi, WidgetListsApi } from "./generated/v0/api"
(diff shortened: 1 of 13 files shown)


mbertrand (approved):
Looks good! 👍

PR #743, opened 2024-04-09
fix channel configuration

abeglova:
closes https://github.com/mitodl/mit-open/issues/742

There was a bug preventing channels from being saved from the admin because django counts empty json as blank and  default value for `configuration` was empty json but `configuration`  was set as blank=False

### How can this be tested?
Go to http://localhost:8063/admin/channels/fieldchannel/. You should be able to update a channel from the admin


mbertrand (approved):
👍

PR #746, opened 2024-04-10
Capture page views with more information

jkachel:

Closes mitodl/hq#3926

Adds more descriptive page view events to the Learning Resource Drawer. 

The base PostHogProvider code that was added earlier captures a basic page view event which includes a set of PostHog data and the page URL. For the learning resource drawer, we'd like to track a bit more data to make it easier to identify views for learning resources and to make them easier to aggregate. 

This adds a component that gets rendered alongside the drawer content when it's open and fires a custom `lrd_view` with the following information:
- `resourceId`: the database ID of the resource
- `readableId`: the readable ID of the resource (courses-v1:MITxT+etc, etc.)
- `platformCode`: the code for the platform that the resource belongs to
- `resourceType`: the type of resource (course, podcast, etc.) 

In addition, this fixes a small bug whereby disabling the PostHog integration in the environment would not actually do that - you had to remove the `POSTHOG_ENABLED` key (since env values are strings, so "False" was also truthy). Now it will check for `POSTHOG_ENABLED=True` specifically; any other value (or capitalization) will be regarded as JS `false`.

### How can this be tested?

Automated tests should pass.

You will need a PostHog instance or account to test this. Once you have that set up, you should be able to see `lrd_view` events come in. Clicking on the arrows next to the event name should show you the additional information captured with the event:
image

If PostHog is disabled, this should do nothing. 

### Additional Context

Next step is getting this out of PostHog and into OpenSearch


frontends/mit-open/webpack.config.js
```diff
@@ -28,7 +28,10 @@ const getFeatureFlags = () => {
 }
 
 const getPostHogSettings = () => {
-  if (process.env.POSTHOG_ENABLED && process.env.POSTHOG_PROJECT_API_KEY) {
+  if (
+    process.env.POSTHOG_ENABLED === "True" &&
```
shanbady, line 32: Would it make sense to coerce this such that any non-falsey ( !== '',0,False,false) value means True? Even though it is explicit in this PR what the value should be I could see this causing confusion in the future.

shanbady (changes requested):
Seems to work as intended. just left a suggestion about the POSTHOG_ENABLED value

frontends/mit-open/webpack.config.js
jkachel: Updated in https://github.com/mitodl/mit-open/pull/746/commits/04e60505ac34327809533a5dca425ddea003890b to do similar - it now lowers and trims and checks for "true". (This is pretty similar to how the Django app does it, except it'll raise an exception if it's not either True or False; I don't see a need to be that exacting about it.)

shanbady (commented):
LGTM

shanbady (approved):

frontends/mit-open/src/page-components/LearningResourceDrawer/LearningResourceDrawer.tsx
```diff
@@ -27,14 +28,38 @@ const PAPER_PROPS: RoutedDrawerProps["PaperProps"] = {
...
+  const posthog = usePostHog()
+
+  if (APP_SETTINGS.posthog?.enabled && resource.isSuccess && open) {
+    posthog.capture("lrd_view", {
+      resourceId: resourceId || "unknown",
+      readableId: resource.data.readable_id,
+      platformCode: resource.data.platform?.code,
+      resourceType: resource.data.resource_type,
+    })
+  }
```
ChristopherChudzicki, line 46: We should wrap this in `useEffect`:
```ts
useEffect(() => {
  if (APP_SETTINGS.posthog?.enabled && resource.isSuccess && open) { ... }
}, [id])
```
otherwise:
- **important**  `posthog.capture` will be called on every render, even if the `id` hasn't changed. 
- **theoretical** React has been saying for a long time that side-effects should always be in `useEffect`, not in the render function. I believe that's because there's a planned feature ("concurrent mode") where react may start rendering something and abort the render in order to render some other, higher priority component. The idea is that `useEffect` would only be called when the component finishes rendering

**Aside:** I would just do this as a hook, rather than a component.

If you keep it as a component, you can return `null` rather than the empty fragment.

ChristopherChudzicki (changes requested):

frontends/mit-open/src/page-components/LearningResourceDrawer/LearningResourceDrawer.tsx
ChristopherChudzicki: @jkachel It's possible that nothing right now is triggering extra renders of the drawer, but we shouldn't count on it. And with the above `useEffect`, you might even remove the `open` dependency: The drawer content probably does re-render during closing, which maybe is why the `open` check got added in first place.

jkachel: I refactored this to be a hook in https://github.com/mitodl/mit-open/pull/746/commits/5e9394480c6c837df4732a11855fa79f57134e50 .

frontends/mit-open/src/page-components/LearningResourceDrawer/LearningResourceDrawer.tsx
```diff
@@ -1,17 +1,43 @@
...
 
+const useCapturePageView = (resourceId: number) => {
+  const { data, isSuccess } = useLearningResourcesDetail(Number(resourceId))
+  const posthog = usePostHog()
+
+  useEffect(() => {
+    if (!APP_SETTINGS.posthog?.enabled) return
+    if (!isSuccess) return
+    posthog.capture("lrd_view", {
+      resourceId: data?.id || "unknown",
```
ChristopherChudzicki, line 20: I meant to say this earlier, but I would be extraordinarily surprised if `"unknown"` ever came through.

frontends/mit-open/src/page-components/LearningResourceDrawer/LearningResourceDrawer.tsx
```diff
@@ -34,7 +60,13 @@ const LearningResourceDrawer = () => {
       requiredParams={RESOURCE_DRAWER_PARAMS}
       PaperProps={PAPER_PROPS}
     >
-      {({ params }) => <DrawerContent resourceId={Number(params.resource)} />}
+      {({ params }) => {
+        return (
+          <>
+            <DrawerContent resourceId={Number(params.resource)} />
+          </>
```
ChristopherChudzicki, line 67: Any reason for the fragment?

ChristopherChudzicki (approved):

PR #752, opened 2024-04-11
adding manual migration to fix foreign key type

### LINKED ISSUE #748: Foreignkey for programletter->programcertificate fails with foreignkey type error: invalid input syntax for type integer
### Expected Behavior
Program letter objects should continue to be able to reference program certificates without this foreign key error

### Current Behavior
Since removing id as primary key from programcertificate, a silent issue was introduced where the relation program program letter to program certificate does not recognize this type change.

### Steps to Reproduce
Easiest way to see this fail is to try creating a programcertficate and program letter pointing to that cert in django shell:
```
from profiles.models import ProgramLetter, ProgramCertificate
settings.EXTERNAL_MODELS = []
cert = ProgramCertificate.objects.create(user_email='test@test.com', micromasters_program_id=18, program_title='test')
ProgramLetter.objects.create(certificate=cert, user=<some_user>)
```
### Possible Solution
Introduce a new migration that runs sql to manually change the type of profiles_programletter.certificate_id to a varchar(256) since the django migration mechanism does not automatically pick this up.

shanbady:

This contains fixes for #748 and #753

#748 - a bug that can be repro'd in steps below
#753 - migrates certain fields that we need to be nullable for program certificate data sync

Full description of bug in #748 is in the ticket.
To repro #748  locally, checkout main. in django shell run:
```
from profiles.models import ProgramLetter, ProgramCertificate
User = get_user_model()
settings.EXTERNAL_MODELS = []
cert = ProgramCertificate.objects.create(user_email='test@test.com', micromasters_program_id=18, program_title='test')
ProgramLetter.objects.create(certificate=cert, user_id=User.objects.first().id)
```
you should encounter ```invalid input syntax for type integer: "f99df281-93ab-4fc5-9be...```

### How can this be tested?
to test the bug in #748 
 - Checkout this branch. Ensure the migrations have run. re-run the script from above in django shell and it should succeed.

to verify #753:
- checkout this branch - run migrations and ensure that the nullable fields in the programcertificate model line up with the fields in the ticket #753 

### Additional Context

The oddity I saw around #741 was that the script above doesn't appear to fail on main when using django factory models. 
```
from profiles.factories import ProgramLetterFactory, ProgramCertificateFactory
User = get_user_model()
settings.EXTERNAL_MODELS = []
cert = ProgramCertificateFactory.create(user_email='test@test.com', micromasters_program_id=18, program_title='test')
ProgramLetterFactory.create(certificate=cert, user_id=User.objects.first().id)
```


quazi-h (approved):
Set up mit-open locally and tested using the Django shell commands on main as well as this branch.
I was able to reproduce the issue in main:
`DataError: invalid input syntax for type integer: "1364f723-3d97-4baf-819c-e00df8dec0f5"
LINE 1: ...('7e1c45be-d157-4cf8-8807-38f077609660'::uuid, 1, '1364f723-...`
After checking out this branch and rerunning the script, I saw no exceptions or errors being thrown:
`Out[1]: <ProgramLetter: program letter:  - test>`

When checking the ProgramCertificate, a describe on the table in postgres confirms that only the following 4 fields are "not_null", as expected: 
user_email
program_title
user_full_name
record_hash

PR #778, opened 2024-04-16
Add channel url to topic, department, and offeror serializers

mbertrand:
Closes https://github.com/mitodl/hq/issues/4011

Adds a `channel_url` field to the topic, department, and offeror serializers

### How can this be tested?
- If you haven't already, run `./manage.py backpopulate_field_channels`
- Run `manage.py recreate_index --all`
- Go to the following urls, each entity should have a populated `channel_url` field, and the urls should work.
  - http://localhost:8063/api/v1/topics/
  - http://localhost:8063/api/v1/departments/
  - http://localhost:8063/api/v1/offerors/
  - http://localhost:8063/api/v0/channels/
- Go to http://localhost:8063/learningpaths (as an admin user) and created/edit some channels, adding/updating the topics.  The functionality should still work.

### DIFF
--- a/frontends/mit-open/src/page-components/ManageListDialogs/ManageListDialogs.tsx
+++ b/frontends/mit-open/src/page-components/ManageListDialogs/ManageListDialogs.tsx
@@ -61,6 +61,7 @@ const learningPathFormSchema = Yup.object().shape({
       Yup.object().shape({
         id: Yup.number().required(),
         name: Yup.string().required(),
+        channel_url: Yup.string().nullable().required(),
       }),
     )
     .min(1, "Select between 1 and 3 subjects.")
--- a/learning_resources/serializers.py
+++ b/learning_resources/serializers.py
@@ -11,6 +11,7 @@ from drf_spectacular.utils import extend_schema_field
 from rest_framework import serializers
 from rest_framework.exceptions import ValidationError
 
+from channels.models import FieldChannel
 from learning_resources import constants, models
 from learning_resources.constants import LearningResourceType, LevelType
 from learning_resources.etl.loaders import update_index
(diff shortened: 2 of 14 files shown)


learning_resources/serializers.py
```diff
@@ -34,9 +35,16 @@ class LearningResourceTopicSerializer(serializers.ModelSerializer):
     Serializer for LearningResourceTopic model
     """
 
+    channel_url = serializers.SerializerMethodField(read_only=True)
```
abeglova, line 38: drf_spectacular is not picking up that channel_url can be null. That is why you needed the  @ts-expect-error in ManageListDialogs.tsx
 
```suggestion
    channel_url = serializers.SerializerMethodField(read_only=True, allow_null=True)
```

frontends/mit-open/src/page-components/ManageListDialogs/ManageListDialogs.tsx
```diff
@@ -61,6 +61,7 @@ const learningPathFormSchema = Yup.object().shape({
       Yup.object().shape({
         id: Yup.number().required(),
         name: Yup.string().required(),
+        channel_url: Yup.string().optional(),
```
abeglova, line 64:
```suggestion
        channel_url:  Yup.string().nullable().required()
```

frontends/mit-open/src/page-components/ManageListDialogs/ManageListDialogs.tsx
```diff
@@ -129,8 +130,10 @@ const UpsertLearningPathDialog = NiceModal.create(
     >["onSubmit"] = useCallback(
       async (values) => {
         if (resource?.id) {
+          // @ts-expect-error channel_url
           await updateList.mutateAsync({ ...values, id: resource.id })
         } else {
+          // @ts-expect-error channel_url
```
abeglova, line 136: If you make the change above  and the change in the serializer you can remove this and the @ts-expect-error  in line 133

learning_resources_search/constants.py
```diff
@@ -135,13 +136,15 @@ class FilterConfig:
...
             "name": {"type": "keyword"},
+            "channel_url": {"type": "keyword"},
         },
     },
     "offered_by": {
         "type": "nested",
         "properties": {
             "code": {"type": "keyword"},
             "name": {"type": "keyword"},
+            "channel_url": {"type": "keyword"},
```
abeglova, line 147: Currently the search api returns `channel_url=null ` for the department nested inside course_numbers add `"channel_url": {"type": "keyword"}` there too so that it returns the correct url

abeglova (changes requested):
Overall works great but found a couple of small bugs

abeglova (approved):
👍

PR #779, opened 2024-04-16
Filter channels api by channel_type

mbertrand:
- Closes https://github.com/mitodl/hq/issues/3973

- Adds a `channel_type` filter option to the channels API
- Removes some n+1 queries from the channels API view

### How can this be tested?
If you haven't already, run `./manage.py backpopulate_field_channels`

The following should work, and filter appropriately:
- http://localhost:8063/api/v0/channels/?channel_type=offeror
- http://localhost:8063/api/v0/channels/?channel_type=department
- http://localhost:8063/api/v0/channels/?channel_type=topic
- http://localhost:8063/api/v0/channels/

### DIFF
--- a/channels/serializers.py
+++ b/channels/serializers.py
@@ -137,8 +137,11 @@ class FieldChannelBaseSerializer(ChannelAppearanceMixin, serializers.ModelSerial
         """Return the field's list of LearningPaths"""
         return [
             LearningPathPreviewSerializer(field_list.field_list).data
-            for field_list in instance.lists.all()
-            .prefetch_related("field_list")
+            for field_list in FieldList.objects.filter(field_channel=instance)
+            .prefetch_related(
+                "field_list", "field_channel__lists", "field_channel__featured_list"
+            )
+            .all()
             .order_by("position")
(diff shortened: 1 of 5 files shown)


shanbady (changes requested):
The changes look good. but It doesnt seem like we do any pagination on this view. Wondering if that is something worth considering now or something to do in a future update. Another thing that occurred to me is that the name of the "channels" package could potentially clash in unexpected ways if we decided to adopt it at some point.

mbertrand:
> The changes look good. but It doesnt seem like we do any pagination on this view. Wondering if that is something worth considering now or something to do in a future update. 

The view is currently using `LargePagination` (1000 per page).  I could change it to `DefaultPagination` (10 per page) - @ChristopherChudzicki would that cause any issues on the current front end functionality?


> Another thing that occurred to me is that the name of the "channels" package could potentially clash in unexpected ways if we decided to adopt it at some point.

I think it would be best to leave as is for now, and we can rename it if it becomes a problem in the future.

shanbady (approved):

PR #782, opened 2024-04-17
create api routes for user subscribe/unsubscribe to search

### LINKED ISSUE #755: create api routes for "subcribe/unsubscribe to search"
We need an api route that the frontend can call to subscribe and unsubscribe a user to a search. The route will receive the same search paramters as the search page which will get stored as the "original_query" parameter in the PercolateQuery model. The original query will get transformed into an opensearch percolator compatable  query which is stored as the "query" in the PercolateQuery object.

shanbady:

Closes #755 

This PR introduces a "user subscription" (/api/v1/learning_resources_search_subscription) endpoint that allows for the following:
1. Subscribing a user to a search query via POST
2. Unsubscribing a user from a query via DELETE
3. Checking the subscription status for a user via GET

### How can this be tested?
How I have been testing this on the frontend is by going to the search page http://od.odl.local:8063/search/?q=testing and then opening up dev-tools and copying the json request "as fetch" and then modifying it for the learning_resources_search_subscription route. 

My fetch requests look like so. you should be able to re-use by adding a valid csrf token (run these in developer tools console):

**Subscribe user**
<pre><code>
fetch("http://od.odl.local:8063/api/v1/learning_resources_search_subscribe/", {
  "headers": {
    "accept": "application/json, text/plain, */*",
    "accept-language": "en-US,en;q=0.9",
    "Content-Type": "application/json",
    "x-csrftoken":  <b>"csrf-token"</b>
  },
  "body": JSON.stringify({
  "aggregations":[
    "resource_type",
    "topic",
    "offered_by"
  ],
  "limit":"10",
  "offset":"0",
  "q":"testing",
  "resource_type":[
    "course",
    "program",
    "podcast"
  ]
}),
  "method": "POST",
  "mode": "cors",
  "credentials": "include"
});
</code></pre>

**Unsubscribe user**
<pre>
<code>

fetch("http://od.odl.local:8063/api/v1/learning_resources_search_subscribe/", {
  "headers": {
    "accept": "application/json, text/plain, */*",
    "accept-language": "en-US,en;q=0.9",
    "Content-Type": "application/json",
    "cache-control": "no-cache",
    "pragma": "no-cache",
    "x-csrftoken": <b>"csrf-token"</b>
  },
  "referrer": "http://od.odl.local:8063/search/?q=testing",
  "referrerPolicy": "same-origin",
  "body": JSON.stringify({
  "aggregations":[
    "resource_type",
    "topic",
    "offered_by"
  ],
  "limit":"10",
  "offset":"0",
  "q":"testing",
  "resource_type":[
    "course",
    "program",
    "podcast"
  ]
}),
  "method": "DELETE",
  "mode": "cors",
  "credentials": "include"
});

</code>
</pre>

You can then check the subscribed/unsubscribed status by visiting the get request for the endpoint (http://od.odl.local:8063/api/v1/learning_resources_search_subscribe/?aggregations=resource_type&aggregations=topic&aggregations=offered_by&limit=10&offset=0&q=testing&resource_type=course&resource_type=program&resource_type=podcast)

### Additional Notes
When storing the original query we strip 'limit', 'sortby' and 'offset' which are not relevant to the query - for example 
```
{
'query': 'test',
'limit': 20,
'offset': 5
}
```
 and 
```{
'query': 'test',
'limit': 100,
'sortby': 'title',
}
```
should resolve to the same query ```{'query': 'test'}```


learning_resources_search/views.py
```diff
@@ -74,6 +80,77 @@ def get(self, request):
...
+            for key, errors_obj in request_data.errors.items():
+                if isinstance(errors_obj, list):
+                    errors[key] = errors_obj
+                else:
+                    errors[key] = list(set(chain(*errors_obj.values())))
+            return Response(errors, status=400)
+
+    @extend_schema(summary="Unsubscribe")
+    def delete(self, request):
+        request_data = LearningResourcesSearchRequestSerializer(data=request.data)
```
mbertrand, line 138: I think it may not be best API practice to send json data in a DELETE request, and in this case it's not really deleting the PercolateQuery, just removing the user from it.  

Could this be a distinct `unsubscribe` action/url instead with a url like `http://localhost:8063/api/v1/learning_resources_search_subscribe/unsubscribe`?


Will there ever be a need to get a list of all the user's search subscriptions?  In that case I think it might be better to have a more standard DRF viewset where:

GET `http://localhost:8063/api/v1/learning_resources_search_subscriptions` returns all the user's subscribed queries.  This could be used on the frontend to determine if the current query matches a subscribed query (and facilitate allowing the user to view/manage their saved queries)

POST `http://localhost:8063/api/v1/learning_resources_search_subscriptions` with JSON data upserts a PercolateQuery with the user as a member, very similar to how it currently works in this PR

POST `http://localhost:8063/api/v1/learning_resources_search_subscriptions/<id>/unsubscribe` (no json data) or `http://localhost:8063/api/v1/learning_resources_search_subscriptions/unsubscribe` (with json data) removes the user's subscription to the query specified by id (or json data)

learning_resources_search/api.py
```diff
@@ -517,9 +531,101 @@ def execute_learn_search(search_params):
...
+        search_params (dict): The opensearch query params returned from
+        LearningResourcesSearchRequestSerializer
+
+    Returns:
+        bool: Whether or not the user has subscribed to the query
+    """
+    try:
+        percolate_query = PercolateQuery.objects.get(
+            source_type=PercolateQuery.SEARCH_SUBSCRIPTION_TYPE,
+            original_query=adjust_original_query_for_percolate(search_params),
```
mbertrand, line 622: This seems sensitive to the order of the search parameters.

For instance, after subscribing to a query,
GET http://localhost:8063/api/v1/learning_resources_search_subscribe/?aggregations=resource_type&aggregations=topic&aggregations=offered_by&limit=10&offset=0&q=testing&resource_type=course&resource_type=program&resource_type=podcast

returns `{"is_subscribed": true}`

but if I reorder the parameters slightly:
GET http://localhost:8063/api/v1/learning_resources_search_subscribe/?aggregations=resource_type&aggregations=offered_by&aggregations=topic&limit=10&offset=0&q=testing&resource_type=course&resource_type=program&resource_type=podcast

I get `{"is_subscribed": false}`

mbertrand (changes requested):

learning_resources_search/api.py
ChristopherChudzicki: `PercolateQuery` stores the query as JSON, right? I have not tested this, but my guess would be that queries
```
// equivalent queries, both stored internally as { a: 1, b: "B" }
?a=1&b=B
?b=B&a=1

// NOT equivalent because `{ a: [1, 2], b: "B" } and { a: [2, 1], b: "B" }`
// are distinct JSON objects.
?a=1&a=2&b=B
?a=2&a=1&b=B
```

Maybe we could sort keys? Either via `save()` int he model, or in one of these api handlers?

learning_resources_search/api.py
```diff
@@ -449,17 +449,31 @@ def generate_aggregation_clauses(search_params, filter_clauses):
...
+    Remove keys that are irrelevent when storing original queries
+    for percolate uniqueness such as "limit" and "offset"
+    """
+    if "limit" in query:
+        del query["limit"]
+    if "offset" in query:
+        del query["offset"]
+    if "sortby" in query:
+        del query["offset"]
+    return query
```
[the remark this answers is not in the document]
shanbady: fixed

learning_resources_search/api.py
```diff
@@ -517,9 +531,101 @@ def execute_learn_search(search_params):
...
+        LearningResourcesSearchRequestSerializer
+
+    Returns:
+        dict: The opensearch response dict
+    """
+
+    percolate_query, _ = PercolateQuery.objects.get_or_create(
+        source_type=PercolateQuery.SEARCH_SUBSCRIPTION_TYPE,
+        original_query=adjust_original_query_for_percolate(search_params),
+    )
```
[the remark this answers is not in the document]
shanbady: moved this out of the save method

learning_resources_search/signals.py
```diff
@@ -5,11 +5,26 @@
...
+    hook.percolate_query_delete(percolate_query=percolate_query)
+
+
+def percolate_query_saved_actions(percolate_query: PercolateQuery):
+    """
+    Trigger plugins when a LearningResource is created or updated
+    """
+    pm = get_plugin_manager()
+    hook = pm.hook
+    hook.percolate_query_upserted(percolate_query=percolate_query)
```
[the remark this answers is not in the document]
shanbady: I was initially running into some circular imports but I think they have resolved after some refactoring. fixed.

learning_resources_search/api.py
shanbady: Good, catch. What I have now done is added a utility that recursively orders the params object prior to creating the query and added a test to make sure the ordering doesnt matter. It made sense to do this outside of save to avoid unintended consequences for example if someone in the future tried to lookup an item for existence only to get a mysterious duplicate error when saving.

learning_resources_search/views.py
shanbady: I do recall long ways back running into odd issues with client libraries not universally allowing for a body on delete methods.

 I've refactored this so that:

**GET** /api/v1/**learning_resources_user_subscription** -> lists percolate queries for user
**POST** /api/v1/learning_resources_user_subscription/**subscribe** -> subscribe user to query
**POST** /api/v1/learning_resources_user_subscription/**unsubscribe** -> unsubscribe user to query
**POST** /api/v1/learning_resources_user_subscription/**<pk>/unsubscribe** -> unsubscribe user by id

learning_resources_search/views.py
```diff
@@ -74,11 +86,97 @@ def get(self, request):
...
 
 
+class UserSearchSubscriptionViewSet(viewsets.ViewSet):
+    """
+    View for listing percolate query subscriptions for a user
+    """
+
+    permission_classes = (IsAuthenticated,)
+    serializer_class = UserPercolateQueryRequestSerializer
+    filter_backends = (DjangoFilterBackend,)
```
mbertrand, line 96: Suggestion:

```python
class UserSearchSubscriptionViewSet(viewsets.ModelViewSet):
    """
    View for listing percolate query subscriptions for a user
    """

    permission_classes = (IsAuthenticated,)
    serializer_class = PercolateQuerySerializer
    filter_backends = (DjangoFilterBackend,)
    http_method_names = ["get"]

    def get_queryset(self):
        """
        Generate a QuerySet for fetching valid PercolateQueries for this user

        Returns:
            QuerySet of PercolateQuery objects subscribed to by request user
        """
        queryset = self.request.user.percolate_queries.all()
        for backend in list(self.filter_backends):
            queryset = backend().filter_queryset(self.request, queryset, view=self)
        return queryset
```

Then you can remove the `list` and `filter_queryset` functions below.

learning_resources_search/views.py
```diff
@@ -74,11 +86,97 @@ def get(self, request):
...
+        methods=["POST"],
+        url_path="unsubscribe",
+        name="Unsubscribe user from query",
+    )
+    @extend_schema(
+        summary="Unsubscribe user to query",
+        parameters=[LearningResourcesSearchRequestSerializer()],
+        responses=PercolateQuerySerializer(),
+    )
+    def unsubscribe_user(self, request):
```
mbertrand, line 146: I think just one of the two methods (`unsubscribe_user`, sending json in body OR `unsubscribe`, sending query id in url) is probably sufficient.  I'd defer to @ChristopherChudzicki on which one is most practical to call from the front end.

learning_resources_search/views.py
shanbady: much cleaner. done

learning_resources_search/views.py
```diff
@@ -74,11 +84,97 @@ def get(self, request):
...
+    @action(
+        detail=True,
+        methods=["POST"],
+        url_path="unsubscribe",
+        name="Unsubscribe user from query by id",
+    )
+    @extend_schema(
+        summary="Unsubscribe user to query",
+        parameters=[LearningResourcesSearchRequestSerializer()],
+        responses=PercolateQuerySerializer(),
```
ChristopherChudzicki, line 138: I think you can add a `description` kwarg here... Right now all the descriptions are
```
View for listing percolate query subscriptions for a user
```

openapi/specs/v1.yaml
```diff
@@ -3258,6 +3258,66 @@ paths:
...
+          description: ''
+  /api/v1/learning_resources_user_subscription/{id}/unsubscribe/:
+    post:
+      operationId: learning_resources_user_subscription_unsubscribe_create_2
+      description: View for listing percolate query subscriptions for a user
+      parameters:
+      - in: path
+        name: id
+        schema:
+          type: string
```
ChristopherChudzicki, line 3284: Should this be `int` ?

learning_resources_search/views.py
ChristopherChudzicki: If it's just a question of `POST /../{id}/unsubscribe` vs `POST /../unsubscribe/` + body `{ id }`, it really doesn't matter—the generated API client will treat them both as `someFunction({ id })`. 

But I agree with Matt: let's just have one.

**Also:** Can you double check the OpenAPI spec? I think some request bodies might be missing.

openapi/specs/v1.yaml
shanbady: this shouldnt be there at all at. just pushed a fix that includes proper schema

learning_resources_search/views.py
shanbady: just pushed.a fix for the schemas and request bodies and removed the /unsubscribe/:id route

ChristopherChudzicki: Just talked to @shanbady. I misunderstood the alternatives (partly due to missing openapi request bodies!).

> choice was just between:
> 1. POST /unsubscribe/{id}/ (no body)
> 2. POST /unsubscribe/ with original query in the body

IMO, (1) is preferable.

shanbady:
@mbertrand @ChristopherChudzicki I've made the changes - 
the unsubscribe is now a DELETE.

In terms of the api spec - for some reason, drf spectacular is injecting an {id} param of type string (in addition to the pk of type integer). I havent been able to tease out the exact issue.
```
  /api/v1/learning_resources_user_subscription/{id}/unsubscribe/:
    delete:
      operationId: learning_resources_user_subscription_unsubscribe_destroy
      description: |-
        Unsubscribe a user from a query

        Args:
        pk (integer): The id of the query

        Returns:
        PercolateQuerySerializer: The percolate query
      summary: Unsubscribe user from query
      parameters:
      - in: path
        name: id
        schema:
          type: string
        required: true
      - in: path
        name: pk
        schema:
          type: integer
        required: true
```

learning_resources_search/api.py
```diff
@@ -517,9 +545,118 @@ def execute_learn_search(search_params):
...
+    adjusted_original_query = adjust_original_query_for_percolate(search_params)
+
+    percolate_query, _ = PercolateQuery.objects.get_or_create(
+        source_type=PercolateQuery.SEARCH_SUBSCRIPTION_TYPE,
+        original_query=adjusted_original_query,
+        query=adjust_query_for_percolator(adjusted_original_query),
+    )
+    if not percolate_query.users.filter(id=user.id).exists():
+        percolate_query.users.add(user)
+    return percolate_query
```
mbertrand, line 593: You could add the call `percolate_query_saved_actions(percolate_query)` here now and get rid of the post_save signal.   Can be in a subsequent PR too.

mbertrand (approved):
Works great!  Just a couple minor comments.

learning_resources_search/api.py
shanbady: will leave this for a subsequent PR since currently there is an admin integration which I think will be handy in the next step.

ChristopherChudzicki (approved):

learning_resources_search/views.py
shanbady: Leaving a note here on the concensus we came to after discussing. 

- We decided to go with the /{id}/unsubscribe/ route vs POST /unsubscribe/ 
- After discussing discussing with @ChristopherChudzicki and @mbertrand I leaned towards using a DELETE instead of POST for /{id}/unsubscribe/ since from an api user's perspective we are 'deleting a subscription' along with the fact that there are no post parameters for that route , it seemed more natural to use delete

PR #786, opened 2024-04-18
ETL for Open Learning events via API

mbertrand:
Closes https://github.com/mitodl/mit-open/issues/681

Refactors the Open Learning events ETL pipeline to use API calls instead of scraping HTML from the events page.

### How can this be tested?
Run `./manage.py backpopulate_news_events --delete` if you already have some news/event items.
Run `./manage.py backpopulate_news_events -p ol_events_etl`

It should complete successfully and create 1 `FeedSource` and about 3 `FeedItems` of type `events`, with populated image, audience, location, and event_type fields.  All the event dates should be today or later, no past events.

You can view them in django admin or http://localhost:8063/api/v0/news_events/


news_events/etl/ol_events.py
```diff
@@ -5,171 +5,192 @@
...
+Regular event series include Open Learning Talks, which bring together leaders
+to discuss research-based ideas, technologies, and efforts in education; and xTalks,
+which provide MIT faculty, researchers, staff, and students an opportunity to share
+their experiences developing and using digital technologies in the classroom.
+Explore upcoming events below, including webinars, workshops, and more.
+"""
 
 # One URL for now, might switch to a list of audience/type-specific URLs later
-OL_EVENTS_URLS = ["/events"]
+OL_EVENTS_API_URL = (
```
shanbady, line 31: should this url build off of OL_EVENTS_BASE_URL?

news_events/etl/ol_events.py
```diff
@@ -5,171 +5,192 @@
...
+        .get("relationships", {})
+        .get("field_media_image", {})
+        .get("data", {})
+        .get("meta", {})
+    )
+    if image_src:
+        return {
+            "url": urljoin(OL_EVENTS_BASE_URL, image_src),
+            "alt": attrs.get("alt"),
+            "description": attrs.get("title"),
```
shanbady, line 141: curious if there is a specific reason for changing title from the api to description.

news_events/etl/ol_events.py
```diff
@@ -5,171 +5,192 @@
...
-            event_page_data.find("div", class_="field--type-text-with-summary")
+        "guid": event_data["id"],
+        "url": urljoin(OL_EVENTS_BASE_URL, event_path) if event_path else None,
+        "title": attributes.get("title"),
+        "image": transform_event_image(
+            *extract_event_image(
+                event_data.get("relationships", {}).get("field_event_image", {})
+            )
         ),
+        "summary": strip_tags(attributes.get("body", {}).get("value") or ""),
```
shanbady, line 176: seems like this should be trying to get the summary from the body like so ```attributes.get("body", {}).get("summary")```

shanbady (changes requested):
looks good. just some minor notes/questions

news_events/etl/ol_events.py
shanbady: not a blocker since I'm guessing this url may never change but I was wondering if it would be better to have this sit in settings.py with all the other urls

news_events/etl/ol_events.py
mbertrand: That's what I was originally thinking of doing but the value is always "".

news_events/etl/ol_events.py
mbertrand: The front end html for the events site never seems to include a title attribute,  but I noticed it is in the API response.  About half the time it is blank, the other half it is identical to the alt tag, but it seemed worthwhile to record that info here anyway since it was available.

news_events/etl/ol_events.py
mbertrand: Changed it just in case it might be different than "" in the future, if not the same value will still be in the `content` field

news_events/etl/ol_events.py
mbertrand: The other news_events ETL urls are also stored in their respective python files.  For learning resources, we originally had the urls come from settings because RC and prod would often hit different endpoints for those (RC MITx Online vs prod MITx Online, for example) but recently we moved away from that, hitting the prod versions of those api endpoints all the time.  

If moving to settings, might be best to do all of them at once in a separate PR.

shanbady (approved):
looks good

PR #789, opened 2024-04-19
Adds ETL to pull PostHog view events into the database; adds popular resource APIs

jkachel:

Closes mitodl/hq#4008

Two main things:
- This adds a pipeline to the normal learning resources ETL pipeline stack that pulls the learning resource view events (`lrd_view`) from PostHog, and stores them in the local database.
- This also adds a `popular` endpoint to some learning resource APIs, which will pull resources in order of popularity (most hits to least). This is done in the same way that new and upcoming resources are done.

### How can this be tested?

Automated tests should pass.

To test you will need to have PostHog set up and enabled - see the README.md for this. Note that you will also need a personal API key for this to work, so if you had this set up before you will need to add this setting. It's new in this PR. You will also need to generate some views, and you'll need to pull the view data into the system. 

_Testing the pipeline manually_: You can run the `load_posthog_lrd_view_events` management command to run the pipeline manually. 
_Testing the task-based pipeline runner_: The task is called `get_learning_resource_views` and you should be able to queue it up and it should run.
_Testing the endpoints_: You should be able to append `popular` to the `learning_resources`, `courses`, and `programs` endpoints. This will return a list of the given resource sorted by the number of views each has. (I find the easiest way to test this is to load the API directly in a browser that has JSON View or something similar, or on the command line piped into `jq`.)


jkachel:
This was going to be a separate PR, but it seems this more or less came for free with the changes I've made here: https://github.com/mitodl/mit-open/pull/789/commits/acc58491686664540fb5421fcdf25fcaff98759f adds "views" as a sortable option for searching learning resources, and the view counts are present in the output, so you should be able to run a search with `sortby=views` or `-views` to search for resources and order by popularity.

You'll need to record a bunch of views (either by creating records manually or triggering the view event) for something, load those views in to the database, and queue an index update to test this. Once that's done, though, you should be able to do something like this and get results as you'd expect: `http://od.odl.local:8063/api/v1/learning_resources_search/?q=Computational%20Data%20Science%20in%20Physics%20II&resource_type=course&sortby=-views` 

(There's no UI for this so you won't be able to change this there - there doesn't seem to be any UI for sorting at all?)

Ferdi:
Are you planning for this data to make it to elasticsearch where popularity would be on of the sort_by options ?

learning_resources/views.py
```diff
@@ -227,6 +227,35 @@ def upcoming(self, request: Request) -> QuerySet:  # noqa: ARG002
...
+            new=extend_schema(
+                description=f"Get a paginated list of popular {name}.",
+                responses=cls.serializer_class(many=True),
+            ),
+        )(cls)
+        super().__init_subclass__()
+
+    @extend_schema(summary="List Popular")
+    @action(methods=["GET"], detail=False, name="Popular Resources")
+    def popular(self, request: Request) -> QuerySet:  # noqa: ARG002
```
shanbady, line 249: This appears to also be applied to /api/v1/learning_resources_search/popular which returns odd results for the views attribute at the moment. we should either resolve it so it works with learning_resources_search or remove the /popular from the opensearch endpoint.

jkachel:
> Are you planning for this data to make it to elasticsearch where popularity would be on of the sort_by options ?

It already does; the learning_resources_search pulls from ES.

README.md
```diff
@@ -187,6 +187,26 @@ To login via the Keycloak client, open http://od.odl.local:8063/login/ol-oidc in
...
+
+The keys and ID can be found in the Settings section of the project in PostHog that you're using for the app. The project key and ID are under "Project", and you can generate a personal API key under "User"->"Personal API Keys".
+
+> [!WARNING]
+> Be careful with the API keys! The project API key is **not secret** and is sent in clear text with the frontend. The personal API key **is** secret. Don't mix them up.
+
+Personal API keys only need read permission to Query. When creating a personal API key, choose "Read" under Query for Scopes. The key needs no other permissions (unless you need them for other things). Additionally, if you select either option besides "All-access" under "Organization & project access", make sure you assign the correct project/org to the API key.
+
+Once these are set (and you've restarted the app), you should see events flowing into the PostHog dashboard.
+
```
shanbady, line 209: Not sure how pressing/breaking not updating is but I did notice this banner in the dashboard
Screenshot 2024-04-22 at 11 59 26 AM

shanbady (changes requested):
tested pulling in data and seeing things populate in admin. left some minor comments

learning_resources/views.py
jkachel: This was a bit confusing because (other than the constants that got added) I didn't touch this at all. But: the URL path for this was using a `re_path` so putting anything after `learning_resources_search` worked. Changed this in https://github.com/mitodl/mit-open/pull/789/commits/e4de0aa0185d6b8ba74bc24666497a2bf249e909 to use a normal, non-regex `path` call instead, so it won't let you add path elements after the `learning_resources_search`. 

(I am assuming this is OK - the tests pass, it wasn't exposed in the openapi schema, and intrinsically it doesn't seem like this was a real necessity, but just habit to use re_path rather than path. The other route in there is still using the `re_path` but can switch that over too if it doesn't really need a regex.)

README.md
jkachel: I didn't see this get nabbed in the dependency dashboard so I went ahead and bumped it.

learning_resources/views.py
shanbady: ah ok. interesting.

shanbady (approved):
looks like there are some conflicts to resolve. other than that looks good.

PR #800, opened 2024-04-20
Buttons

ChristopherChudzicki:

- [ ] Desktop screenshots


frontends/ol-components/src/components/Button/Button.tsx
ChristopherChudzicki, line 1: I decided just to build the button variants from `styled.button` rather than customizing MUI's button. My reason was that:
- button, even with variants, seemed fairly simple
- MUI doesn't have a square or filled version of IconButton. We could probably customize it, but starting with `styled.button` seemed easy, too.

Customizing MUI's button via the `theme` probably would have been fine, too.

In contrast, I did use MUI's chip (aka "pill"). Chips initially seem pretty simple, but actually a little complicated:
- They might be clickable or not clickable, in which case they should be `button` or `div`[^1].
- ... or they might be a link
- MUI's deletable chips for their `onDelete` action when `Backspace` is pressed. This seems like good behavior.
- slots for icons on left, delete icon on right

All in all, the issues above led me to customize MUI's chip rather than building our own, and the customization was fairly simple.


[^1]: MUI actually uses a `div` by default for all chips, even when they are clickable. But it adds `role=button` AND adds keyboard interactivity.

frontends/ol-components/src/components/ThemeProvider/typography.ts
```diff
@@ -56,26 +57,21 @@ const styles = {
     fontSize: pxToRem(10),
     lineHeight: 1.26,
   },
-  p1: {
+  body1: {
```
ChristopherChudzicki, line 60: The p1, p2, p3, p4 names were a mistake. They should have been `body1`, etc.

frontends/ol-components/src/components/ThemeProvider/colors.ts
ChristopherChudzicki, line 1: The colors from https://www.figma.com/file/Eux3guSenAFVvNHGi1Y9Wm/MIT-Design-System?node-id=0%3A1&mode=dev

frontends/ol-components/src/components/ThemeProvider/buttons.ts
```diff
@@ -1,25 +1,18 @@
...
-import { pxToRem } from "./typography"
 
-const buttonComponent: NonNullable<ThemeOptions["components"]>["MuiButton"] = {
+const buttonBaseComponent: NonNullable<
+  ThemeOptions["components"]
+>["MuiButtonBase"] = {
   defaultProps: {
-    size: "large",
-    disableElevation: true,
+    disableRipple: true,
```
ChristopherChudzicki, line 7: Although I decided not to use MUI's button component (see above) I did add customization for their `ButtonBase`, specifically disabling the ripple and using browser's default `:focus-visible` behavior.

MUI's ButtonBase is used by several of their other components, e.g., Tabs, Pagination, all use ButtonBase.

frontends/ol-components/src/components/Chips/Chip.stories.tsx
```diff
@@ -0,0 +1,102 @@
...
+
+export const Variants: Story = {}
+
+export const Buttons: Story = {
+  args: {
+    onClick: fn(),
+  },
+}
+
+export const Deleteable: Story = {
```
jonkafton, line 72: Should these be removed in the Storybook UI when deleted?

frontends/ol-components/src/components/Chips/Chip.stories.tsx
```diff
@@ -0,0 +1,102 @@
...
+  {
+    variant: "outlined",
+    label: "Outlined",
+  },
+] as const
+
+const meta: Meta<typeof Chip> = {
+  title: 'smoot-design/Chip ("Pill")',
+  component: Chip,
+  render: (args) => {
```
jonkafton, line 25: The stories are missing their controls in the Storybook UI. Perhaps not much value if we're just wrapping the MUI component, but could be useful for designers if we can import/derive them from @mui/material without too much trouble.

frontends/ol-components/src/components/ThemeProvider/chips.tsx
```diff
@@ -0,0 +1,113 @@
...
+
+const chipComponent: NonNullable<ThemeOptions["components"]>["MuiChip"] = {
+  defaultProps: {
+    size: "medium",
+    deleteIcon: <ClearIcon />,
+  },
+  styleOverrides: {
+    root: {
+      borderRadius: "100vh",
+      borderWidth: "1px",
```
jonkafton, line 15: Number literals ideally for these, assuming not conflicting.

frontends/mit-open/src/pages/ErrorPage/ErrorPage.tsx
```diff
@@ -40,7 +41,9 @@ const ErrorPage = () => {
      * This should not happen, but it's better than the app crashing.
      */
     <ErrorPageTemplate title="Error">
-      <h1>Something went wrong.</h1>
+      <Typography variant="h3" component="h1">
```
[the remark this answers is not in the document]
ChristopherChudzicki: > Is this for where we want to apply an h3 theme to an h1 element

Yes

> should we be able to bend rules around presentation hierarchy?

Do you mean "is styling h3 as h1 something we should encourage"? IMO, yes: The h1 style in design docs is enormous. Easy for me to imagine pages (like this one) where we want a semantic h1 but we want it styled differently.

Also easy for me to imagine cases where we want something styled as a heading but not semantically a heading at all. (E.g., so that a screenreader doesn't announce it as a heading.)

frontends/github-pages/src/App.tsx
```diff
@@ -16,7 +16,7 @@ const PageTitle = styled.h1`
   margin-top: 6rem;
   margin-bottom: 0.5rem;
   font-size: 50px;
-  color: ${({ theme }) => theme.custom.colorBlue5};
+  color: "#03152d";
```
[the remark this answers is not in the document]
ChristopherChudzicki: This was the only usage. it's not one of the design system colors: https://www.figma.com/file/Eux3guSenAFVvNHGi1Y9Wm/MIT-Design-System?node-id=0%3A1&mode=dev Maybe I ought have done it in a separate PR, but I was a bit aggressive in cleaning up the old color variables in this PR. My thought is:

**Is the color in the [color designs](https://www.figma.com/file/Eux3guSenAFVvNHGi1Y9Wm/MIT-Design-System?node-id=0%3A1&mode=dev)**?
- Yes: then it goes in Theme
- No. Then:
    - it will be replaced anyway when that page gets proper designs
    - Replace it with something close that is in the theme for now
    - or, failing that, leave it hardcoded as raw value.  

Probably I should have just replaced this with black, the color the designs are using for generic text.

frontends/ol-components/src/components/Chips/Chip.stories.tsx
ChristopherChudzicki: No. It just logs the onDelete events:

Screenshot 2024-04-22 at 4 09 41 PM

frontends/ol-components/src/components/Button/Button.tsx
```diff
@@ -0,0 +1,224 @@
...
+  }
+  const { palette } = theme
+  return [
+    {
+      // font
+      lineHeight: 1,
+      fontFamily: theme.typography.fontFamily,
+      fontWeight: 500,
+      color: theme.palette.text.primary,
+      // display
```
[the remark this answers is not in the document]
ChristopherChudzicki: That's true. I just put them there as dividers, really. Grouping related CSS properties made it easier for me to read, at least.

frontends/ol-components/src/components/Button/Button.tsx
```diff
@@ -0,0 +1,224 @@
...
+ * 24x24px by default, but the path may be smaller. I.e., there is generally
+ * whitespace around the path.
+ *
+ * The negative margin below accounts (approximately) for this whitespace.
+ * The whitespace varies by icon, so it's not perfect.
+ */
+const IconContainer = styled.span<{ side: "start" | "end"; size: ButtonSize }>(
+  ({ size, side }) => [
+    side === "start" && {
+      marginLeft: "-4px",
```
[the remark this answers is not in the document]
ChristopherChudzicki: Is there a reason? Preferred by whom? I don't have a strong opinion, but units generally seem good to me.

frontends/ol-components/src/components/ThemeProvider/chips.tsx
ChristopherChudzicki: The first can't be a number. I'll add a comment about why it's 100vh, though.

frontends/mit-open/src/pages/DashboardPage/DashboardPage.tsx
```diff
@@ -8,7 +8,7 @@ const DashboardPage: React.FC = () => {
       <MetaTags>
         <title>Dashboard</title>
       </MetaTags>
-      <h1>Dashboard</h1>
+      <Typography variant="h1">Dashboard</Typography>
```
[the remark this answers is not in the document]
ChristopherChudzicki: We could. I don't have a strong opinion. It seems likely pages will import Typography anyway, and having a standard interface for text might be desirable? But `<Typography variant="h1" />` certainly is more verbose than `<H1 />`.

frontends/ol-components/src/components/Chips/Chip.stories.tsx
ChristopherChudzicki: I added some controls. Apparently storybook infers controls via a tool [react-docgen](https://github.com/reactjs/react-docgen) that scales the files, and doesn't work great with MUI components. I didn't look into it much, but that's why I provided specific argTypes.

Overall, my goal with the Chip and Button stories was mostly making the Docs page where you can see all the variations. Then it's easy to see how a tweak to the code changes the different possible buttons/chips/wahtever.

That said, adding simple controls is easy and definitely improvement 👍 👍

frontends/ol-components/src/components/Button/Button.tsx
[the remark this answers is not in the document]
jonkafton: I'm taking that cue from React appending "px" by default to the style attribute on components and using numbers in their docs, and from it being the preferred form in docs for CSS-in-JS libraries such as Emotion (e.g. https://emotion.sh/docs/object-styles#numbers). 

Perhaps this is more shorthand than convention.

jonkafton (approved):

PR #812, opened 2024-04-24
Updating the LearningResourceViewEvent to cascade delete

jkachel:

n/a

Changes the `on_delete` setting in the LearningResourceViewEvent model to `CASCADE`; it was set to `DO_NOTHING`, which causes errors if you try to delete the parent LearningResource.

### How can this be tested?

Deleting a LearningResource that has view events should work.

### Additional Context

I don't think we really need to store the learning resource view event data if the parent's been deleted; that data still exists in PostHog (and potentially elsewhere depending on whether or not the data platform has ingested it).


shanbady (approved):
looks good

PR #813, opened 2024-04-24
#629 Build front end during release CI

jonkafton:
For [Load session from API and configure decoupled frontend for local development #616](https://github.com/mitodl/mit-open/issues/616) we are serving the static content with Nginx on Heroku.

This commit attempted to download the frontend build artifact from the previous CI workflow, but failed to do so.

This PR builds the front end during release workflows.

This is an interim solution as we will ultimately deploy the front end separately for [CI and deployment config for decoupled front end #629](https://github.com/mitodl/mit-open/issues/629).

### DIFF
--- a/.github/workflows/production.yml
+++ b/.github/workflows/production.yml
@@ -19,11 +19,22 @@ jobs:
         with:
           ref: release
 
-      - name: Download frontend build
-        uses: actions/download-artifact@v4
+      - uses: actions/setup-node@v4
         with:
-          name: frontend-build
-          path: frontends/mit-open/build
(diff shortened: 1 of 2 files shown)


shaidar:
I wouldn't worry about the `release-candidate` workflow as we rely on Heroku to trigger the build once a commit is merged to the release-candidate branch. I paused that workflow for now. @rhysyngsun please correct me if I'm wrong

shaidar:
We decided to disable the Heroku trigger in lieu of using GitHub Actions.

shaidar (approved):

PR #828, opened 2024-04-25
Add format field to LearningResource model and ETL pipelines

mbertrand:
Closes https://github.com/mitodl/hq/issues/3972 (API/search filter is a separate issue and will be added in a separate PR)

- Adds a `format` Array[Char] field to the `LearningResource` model with 3 possible values ("Online", "In person", "Hybrid") and a default value of ["Online"]
- Adds `format` to the serializers and search index schema
- Modifies the prolearn and xpro ETL pipelines to assign the correct value(s) for the field based on the external API response data.  Other ETL pipelines do not include this field and will have the default value of `["Online"]`

### How can this be tested?
- Run `docker compose up`, the new migration should complete
- Run `./manage.py recreate_index --all` to update the search index schema
- Run `./manage.py backpopulate_prolearn_data` and `./manage.py backpopulate_xpro_data`
- Check that:
  - Programs and courses have a non-empty `format` value (`["Online"]` for most)
  - Other types of resources have an empty list for the `format` value 
  - Some of the prolearn courses have "format" values other than just ["Online"] for example:
    - http://localhost:8063/api/v1/learning_resources_search/?q=%22Developing%20a%20Leading%20Edge%20Operations%20Strategy%22
- Run `./manage.py backpopulate_mitxonline_data`.  All mitxonline courses/programs should still have a `format` value of `["Online"]`
- Run `./manage.py backpopulate_podcast_data`, all podcasts should still have a `format` value of `[]`

PS currently all xpro courses and programs only have an "Online" format.


ChristopherChudzicki (approved):
👍 This looks good to me.

What about other sources? Like OCW / edx? Should those have "online"? Seems like we have a lot of "Online" courses that aren't listed as "Online" yet, which would be problematic if we had a search filter.

mbertrand:
@ChristopherChudzicki The default value for courses/programs will be ["Online"] if nothing particular is specified in the ETL source data: https://github.com/mitodl/mit-open/pull/828/files#diff-6858e9413829359601596915fc5fc1219fd7c9c5c54d3249ad26ef1c00077aa9R251

PR #829, opened 2024-04-26
Self contained front end and fixes for building on Heroku

jonkafton:

[Load session from API and configure decoupled frontend for local development #616](https://github.com/mitodl/mit-open/issues/616)

Includes [Self contained front end using Webpack to build HTML and Webpack Dev Server to serve #678](https://github.com/mitodl/mit-open/pull/678) which failed to release due to the front end not being buildable on Heroku with the front end package and yarn workspace root having moved from the project root. The Heroku build has been tested and is successful.

The additional changes add the package.json back to the project root, though keep the frontends/package.json to retain a degree of self-contained frontend (Typescript, Jest configs and the watch container Dockerfile now live in `./frontends`).

The `watch` container Dockerfile again needs the project root as the build context. There is a wider issue of yarn workspaces having the issue for monorepo setups that by hoisting node_modules and the lockfile to the workspace root, we don't have independently deployable workspaces that can be used as Docker context.

This also upgrades Yarn to 4.1.1 as I was having issues with the nesting resulting from the intermediary workspace in `./frontends`. An `in-workspaces` package.json scripts excludes this and the root from being treated as workspaces for the previous `yarn foreach` commands.

### How can this be tested?

Detail in https://github.com/mitodl/mit-open/pull/678


jonkafton:
Labelling this wip as there's an issue running the unit tests.

frontends/package.json
ChristopherChudzicki, line 1: Any reason to have this in addition to the root level package.json? Having 2 levels of workspaces (root + frontends/workspace-name/` seems simpler than having 3 levels (root, + frontends/ + frontends/workspace-name/)

jonkafton:
This change fixes the react and react-dom versions to 18.2.0. Described in https://github.com/testing-library/react-testing-library/issues/1310, there is an issue with 18.3.0 conflicting with @testing-library that breaks a lot of tests with error:

```
Warning: `ReactDOMTestUtils.act` is deprecated in favor of `React.act`. Import `act` from `react` instead of `react-dom/test-utils`. See https://react.dev/warnings/react-dom-test-utils for more info.
```

frontends/package.json
jonkafton: The goal is for frontends to be self-contained. Unfortunately the issue of Heroku needing to find the package.json at root meant that it needed to be reintroduced there though, as we should only be needing to build the front end there temporarily, it seemed sensible to only pull what is needed to build rather than add all the front end test and lint config files etc back to the project root, so that we can more easily remove from the root once the front end is served elsewhere.

jonkafton:
> This change fixes the react and react-dom versions to 18.2.0. Described in [testing-library/react-testing-library#1310](https://github.com/testing-library/react-testing-library/issues/1310), there is an issue with 18.3.0 conflicting with @testing-library that breaks a lot of tests with error:
> 
> ```
> Warning: `ReactDOMTestUtils.act` is deprecated in favor of `React.act`. Import `act` from `react` instead of `react-dom/test-utils`. See https://react.dev/warnings/react-dom-test-utils for more info.
> ```

React released 18.3.1 last Friday to export `act` from React itself. Making that change when we upgrade should resolve the issue.

docker-compose.yml
```diff
@@ -80,30 +78,24 @@ services:
...
-      dockerfile: Dockerfile-node
-    working_dir: /src
-    command: ./scripts/run-watch-dev.sh
-    ports:
-      - "8052:8052"
-      - "8062:8062"
-      - "6006:6006"
+      dockerfile: frontends/Dockerfile
+    command: >
+      /bin/sh -c 'yarn rebuild && yarn watch:docker'
```
ChristopherChudzicki, line 90: `yarn rebuild` doesn't seem to exist, so `docker compose up` isn't working.

frontends/README.md
```diff
@@ -31,7 +33,57 @@ Most workspaces are shared dependencies built using typescript and tested with j
...
+The application is served at `http://localhost:8063`.
+
+### Local Frontend Dev Server with Local Backend
+
+The `watch` container can be slow to respond to changes to the filesystem mounted onto Docker. If [Hot Module Replacement](https://webpack.js.org/concepts/hot-module-replacement/) is slow for you, we can run Wepback Dev Server directly outside of Docker. Dev Server proxies API requests through to a locally running backend stack.
+
+Run the front end with:
+
+```bash
+yarn watch
```
ChristopherChudzicki, line 68:
```suggestion
yarn workspace frontends watch
```

ChristopherChudzicki (changes requested):
This looks good except that `docker compose up` is broken. Missing command `rebuild`. Was it renamed?

frontends/README.md
jonkafton: I've added package scripts at root for these.

ChristopherChudzicki (approved):
Linting, tests, watch mode (backend + frontend) all seem good. Env vars are working.

👍

PR #833, opened 2024-04-26
Minor updates for PostHog settings

jkachel:

Related to #831

Two things:
- Removes PostHog settings from the `js_settings` - the frontend build pulls these from the environment, so they're not needed here at all
- Updates the integration to use the `POSTHOG_PROJECT_API_KEY` to determine whether or not to be enabled
   - This also eliminates the `POSTHOG_ENABLED` setting, since it's no longer used

### How can this be tested?

Automated tests should pass, in both sides of the app. (Make sure you run this with and without the project API key set in `.env`.)

On the frontend, you should see the PostHogProvider in the Components in React Devtools when there's a project key set up, and it should go away when it's not there. 

In the Django app, you should be able to hit the meta API to list flags and it should work if there's a project key in place, and (again) should not if there isn't. The remainder of the app should work normally.


shanbady (changes requested):
There are still a few references to POSTHOG_ENABLED that need to be cleaned up.

seeing one in:
- app.json
- frontends/webpack.config.js
- main/features_test.py

shanbady (approved):
looks good

PR #844, opened 2024-04-30
Add nesting learning resource topics

jkachel:

Closes mitodl/hq#4089

This PR allows topics for learning resources to be nested, so we can have a hierarchy of topics. This involves:
- Some model changes to allow for nesting
- A management command to bulk load the topic data from OCW (or compatible source)
- Minor updates to the API to support parents

The importer expects to pull these from the OCW course site config. This is available [here](https://github.com/mitodl/ocw-studio/blob/dd823c6eec258650508b3498ff5d8657738ba9ea/static/js/resources/ocw-course-site-config.json). Use the `update_topics` management command to kick off upserting items from the file. The system expects the OCW JSON file to live in `learning_resources/data/` but you can also specify a path if you wish. 

You can technically also add topics from other sources if they're in the same format OCW stores them in. 

When storing a LearningResource record, it will queue a task to automatically backfill parent topics, so if you select a topic that's nested under another topic, the resource will have all the applicable topics. (Note that this is queued so if you change this in, say, Django Admin, you will need to reload the record to see the parent topics be added.) 

When backfilling/importing data from other data sources, any topics that are listed in the data source but that don't exist in Open will be skipped. A warning message gets emitted to the log when this happens. (This is a change from before - it would create the topics instead.)

Topics should otherwise work as expected, though.

### How can this be tested?

Automated tests should pass.

Test importing using the `update_topics` command. It should work - you will need to get an OCW course site config file (linked above).

Test adding a learning resource to a topic: 
- Test loading from the ETL pipelines - run the backfill commands or queue tasks. These should process successfully and things that have topics should have them attached accordingly. No new topics should be created.
- Manual testing - edit a learning resource in Django Admin, and add a topic or two to some. For best results, including testing search and channels, add topics that are nested - i.e., add "Engineering" to one resource, then add something like "Biomaterials" to another. The second should get "Biomaterials", plus "Biological Engineering" and "Engineering" as topics. 

Test creating a channel for a topic: Create a field channel record of type "Topic" with search filter `topic=<a topic>`, then update the search index. Loading the channel URL should show you the resources that have that topic assigned. If you use the same scenario listed above, create a channel for "Engineering" and one for "Biomaterials"; you should see two courses in the first and one in the second. 

Test searching: searching for a topic should work as expected. (Channels work using search so you should get back the same stuff in a search that you'd get back for an equivalent channel.)


Ferdi:
Should we make a copy of the ["ocw topics"](https://github.com/mitodl/ocw-studio/blob/dd823c6eec258650508b3498ff5d8657738ba9ea/static/js/resources/ocw-course-site-config.json)  locally? 

Are we going to keep the same [mapping format ](https://github.com/mitodl/mit-open/tree/main/learning_resources/data)?

jkachel:
> Should we make a copy of the ["ocw topics"](https://github.com/mitodl/ocw-studio/blob/dd823c6eec258650508b3498ff5d8657738ba9ea/static/js/resources/ocw-course-site-config.json) locally?

I added a smaller version of the site config to the repo. But this should really be kept in one spot? 

> Are we going to keep the same [mapping format ](https://github.com/mitodl/mit-open/tree/main/learning_resources/data)?

I didn't see any reason to modify it; the mappings should still match to OCW ones.

(this was mentioned on Slack too, adding here for completeness)

frontends/api/src/test-utils/factories/learningResources.ts
```diff
@@ -180,7 +181,7 @@ const _learningResourceShared = (): Partial<
     image: learningResourceImage(),
     offered_by: maybe(learningResourceOfferor) ?? null,
     platform: maybe(learningResourcePlatform) ?? null,
-    prices: [0.0],
+    prices: "",
```
abeglova, line 184: Is this change intentional?

learning_resources/utils.py
```diff
@@ -493,3 +493,74 @@ def offeror_delete_actions(offeror: LearningResourceOfferor):
...
+
+    Args:
+    - topics (dict): the topics to process
+    - parent (None or LearningResourceTopic): the parent topic (for inner loops)
+    Returns:
+    - None
+    """
+
+    for topic in topics:
+        lr_topic, _ = LearningResourceTopic.objects.filter(
```
abeglova, line 516: I think this is where the duplicate topic gets created since the get_or_create includes the parent

abeglova (changes requested):

learning_resources/utils.py
jkachel: Yes - this should have used an update_or_create rather than the get_or_create. As is the topics weren't getting saved so the channel topic detail records were also getting created incorrectly. Fixed in https://github.com/mitodl/mit-open/pull/844/commits/be4c4184ef6763891705530a8b2a23304081ef90 .

learning_resources/models.py
```diff
@@ -196,19 +222,17 @@ def audience(self) -> str | None:
...
         """Returns the prices for the learning resource"""
         if self.resource_type in [
             LearningResourceType.course.name,
             LearningResourceType.program.name,
         ]:
             return list(
-                set(
-                    flatten([(run.prices or [Decimal(0.0)]) for run in self.runs.all()])
-                )
+                set(flatten([run.prices for run in self.runs.all() if run.prices]))
```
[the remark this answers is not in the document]
jkachel: Not sure where this change came from. I pulled this back from main and re-added the updated model I changed.

frontends/api/src/test-utils/factories/learningResources.ts
jkachel: Reverted to the `[0.0]` - I changed this because the typecheck was failing for some reason (not realizing the model change that had happened inadvertently).

jkachel:
https://github.com/mitodl/mit-open/pull/844/commits/de23fa1433347d2ffef371063d96ce417b132f72 resolves some issues with saving and resolves an issue where the backpopulate_resource_channels command removes topics that aren't attached to a learning resource. Still need to do testing with existing topics.

jkachel:
The duplicate issue went a bit deeper than it seemed. There are a few things at play:
- The unique was moved off of the topic name attribute and onto the topic name _plus_ the parent - these changes allow you to have a duplicate name as long as it's not in the same place in the hierarchy
- The channel creation depends on the topic name being unique across all of the topics, which wasn't the case anymore
- There were existing topics in the system that were also represented in different places in the OCW topic hierarchy

So, this required a few more changes:
- The LearningResourceTopic now has a `full_name` which returns the full name of the topic (i.e. with any parent topic names prepended)
- The channel creation uses the topic's full_name as the name and as part of the search query
- The OpenSearch indexing stuff was updated to ensure the full_name came over, and that searching on the topic field hit the full name and not the individual topic name (so we could search for an exact topic; mainly to support channels)
- The regular direct LearningResource search now also supports nested topic searching (up to 3 levels)

The upshot of this is that the OCW topics shouldn't cause an issue with the topics list that's on RC now. 

The other question that's maybe not a code issue is that at least some of the topics on RC have representations in the OCW topic data, which we're using as the canonical source. The code changes here allow for that (and there are some use cases for having them in here even if this wasn't an issue - thinking of a pretty generic term being used as a topic in several different contexts, maybe) but maybe the topics between RC and OCW should be matched up so they're not dupes? For instance, when I was testing this, I noticed there's "Applied Mathematics", which is on RC but in OCW is under "Mathematics". We probably don't need both. Note that topics aren't going to be culled now if they aren't attached to things - we want to keep the OCW list in place.

abeglova:
After running update_topics it looks like i have duplicate topics - one with a parent and one without
Screenshot 2024-05-07 at 12 31 35 PM

learning_resources/migrations/0043_add_nestable_topics.py
```diff
@@ -0,0 +1,37 @@
...
+                on_delete=django.db.models.deletion.CASCADE,
+                to="learning_resources.learningresourcetopic",
+            ),
+        ),
+        migrations.AlterField(
+            model_name="learningresourcetopic",
+            name="name",
+            field=models.CharField(max_length=128),
+        ),
+        migrations.AddConstraint(
```
abeglova, line 29: This migration is not reversable 
abeglova@Anastasias-MacBook-Pro-2 mit-open % docker-compose run web ./manage.py migrate learning_resources 0042
[+] Creating 3/0
 ✔ Container mit-open-db-1            Running                                                                                        0.0s
 ✔ Container mit-open-redis-1         Running                                                                                        0.0s
 ✔ Container opensearch-node-mitopen  Running                                                                                        0.0s
Operations to perform:
  Target specific migration: 0042_update_ondelete_for_events, from learning_resources
Running migrations:
  Rendering model states... DONE
  Unapplying learning_resources.0043_add_nestable_topics...Traceback (most recent call last):
  File "/opt/venv/lib/python3.11/site-packages/django/db/backends/utils.py", line 87, in _execute
    return self.cursor.execute(sql)
           ^^^^^^^^^^^^^^^^^^^^^^^^
psycopg2.errors.UniqueViolation: could not create unique index "learning_resources_learningresourcetopic_name_2a9dc4dd_uniq"
DETAIL:  Key (name)=(Innovation) is duplicated.


The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/src/./manage.py", line 14, in <module>
    execute_from_command_line(sys.argv)
  File "/opt/venv/lib/python3.11/site-packages/django/core/management/__init__.py", line 442, in execute_from_command_line
    utility.execute()
  File "/opt/venv/lib/python3.11/site-packages/django/core/management/__init__.py", line 436, in execute
    self.fetch_command(subcommand).run_from_argv(self.argv)
  File "/opt/venv/lib/python3.11/site-packages/django/core/management/base.py", line 412, in run_from_argv
    self.execute(*args, **cmd_options)
  File "/opt/venv/lib/python3.11/site-packages/django/core/management/base.py", line 458, in execute
    output = self.handle(*args, **options)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/venv/lib/python3.11/site-packages/django/core/management/base.py", line 106, in wrapper
    res = handle_func(*args, **kwargs)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/venv/lib/python3.11/site-packages/django/core/management/commands/migrate.py", line 356, in handle
    post_migrate_state = executor.migrate(
                         ^^^^^^^^^^^^^^^^^
  File "/opt/venv/lib/python3.11/site-packages/django/db/migrations/executor.py", line 141, in migrate
    state = self._migrate_all_backwards(plan, full_plan, fake=fake)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/venv/lib/python3.11/site-packages/django/db/migrations/executor.py", line 219, in _migrate_all_backwards
    self.unapply_migration(states[migration], migration, fake=fake)
  File "/opt/venv/lib/python3.11/site-packages/django/db/migrations/executor.py", line 279, in unapply_migration
    state = migration.unapply(state, schema_editor)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/venv/lib/python3.11/site-packages/django/db/migrations/migration.py", line 193, in unapply
    operation.database_backwards(
  File "/opt/venv/lib/python3.11/site-packages/django/db/migrations/operations/fields.py", line 240, in database_backwards
    self.database_forwards(app_label, schema_editor, from_state, to_state)
  File "/opt/venv/lib/python3.11/site-packages/django/db/migrations/operations/fields.py", line 235, in database_forwards
    schema_editor.alter_field(from_model, from_field, to_field)
  File "/opt/venv/lib/python3.11/site-packages/django/db/backends/base/schema.py", line 831, in alter_field
    self._alter_field(
  File "/opt/venv/lib/python3.11/site-packages/django/db/backends/postgresql/schema.py", line 288, in _alter_field
    super()._alter_field(
  File "/opt/venv/lib/python3.11/site-packages/django/db/backends/base/schema.py", line 1094, in _alter_field
    self.execute(self._create_unique_sql(model, [new_field]))
  File "/opt/venv/lib/python3.11/site-packages/django/db/backends/postgresql/schema.py", line 48, in execute
    return super().execute(sql, None)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/venv/lib/python3.11/site-packages/django/db/backends/base/schema.py", line 201, in execute
    cursor.execute(sql, params)
  File "/opt/venv/lib/python3.11/site-packages/django/db/backends/utils.py", line 102, in execute
    return super().execute(sql, params)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/venv/lib/python3.11/site-packages/sentry_sdk/integrations/django/__init__.py", line 644, in execute
    result = real_execute(self, sql, params)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/venv/lib/python3.11/site-packages/django/db/backends/utils.py", line 67, in execute
    return self._execute_with_wrappers(
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/venv/lib/python3.11/site-packages/django/db/backends/utils.py", line 80, in _execute_with_wrappers
    return executor(sql, params, many, context)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/venv/lib/python3.11/site-packages/django/db/backends/utils.py", line 84, in _execute
    with self.db.wrap_database_errors:
  File "/opt/venv/lib/python3.11/site-packages/django/db/utils.py", line 91, in __exit__
    raise dj_exc_value.with_traceback(traceback) from exc_value
  File "/opt/venv/lib/python3.11/site-packages/django/db/backends/utils.py", line 87, in _execute
    return self.cursor.execute(sql)
           ^^^^^^^^^^^^^^^^^^^^^^^^
django.db.utils.IntegrityError: could not create unique index "learning_resources_learningresourcetopic_name_2a9dc4dd_uniq"
DETAIL:  Key (name)=(Innovation) is duplicated.

jkachel:
https://github.com/mitodl/mit-open/pull/844/commits/e4a809d08a4e01573ac229477c52786f24c3fceb makes a number of changes:

- Rolls back implementation a good bit (to undo changes to support duplicated topic names) - this includes moving the unique constraint back to being on just the topic name in `LearningResource`
- Moves the code to recursively add parent topics to `learning_resources/utils.py`
- Updates the loader code to call said parent topic function; removes the task and the call to it from the `LearningResource.save` function
- Moved the migration to eliminate a merge migration

Channels will be created with the topic name (not full name). Running `update_topics` will move any existing topics into the hierarchy represented by the input file (which by default comes from OCW). The parent topics will be added when the ETL pipeline runs but if you add a topic manually you'll need to call the utility function manually as well.

I noticed while testing backpopulation tasks that there were a few errors in the task for edX data and for xPRO. These seem to be issues elsewhere in the code so not addressing them here. (The edX errors I saw were relating to iterating over course numbers that didn't exist and xPRO hits a resource format that's not found.)

ChristopherChudzicki:
@jkachel I'm looking at this as I work on 
- https://github.com/mitodl/hq/issues/4186

Two things I noticed:
1. The topics json file might not be the one we want (though I see it's the one in the linked issue).. See https://github.com/mitodl/hq/issues/4089#issuecomment-2101210548
2. The topics API seems fairly slow now. Locally, it's taking about 1.5 seconds for me. I think this is because of `full_name`. When I remove it, the response time drops a lot. Maybe we could speed it up, or just remove it from the serializer until we need it. (Do we need it?)

jkachel:
> @jkachel I'm looking at this as I work on
> 
>     * [Topics Listing Page hq#4186](https://github.com/mitodl/hq/issues/4186)
> 
> 
> Two things I noticed:
> 
>     1. The topics json file might not be the one we want (though I see it's the one in the linked issue).. See [Add a topic hierarchy for MIT Open hq#4089 (comment)](https://github.com/mitodl/hq/issues/4089#issuecomment-2101210548)

Converted the loader to use the yaml doc linked above in commit https://github.com/mitodl/mit-open/pull/844/commits/d886874a96044be35ebd1aeb1e782b81beab538c . It's the same format so it's just swapping the loader. The commit also includes a cut-down version of that (just topics; it was tested with the full file too).

> 
>     2. The topics API seems fairly slow now. Locally, it's taking about 1.5 seconds for me. I think this is because of `full_name`. When I remove it, the response time drops a lot. Maybe we could speed it up, or just remove it from the serializer until we need it. (Do we need it?)

Pulled `full_name` in https://github.com/mitodl/mit-open/pull/844/commits/d886874a96044be35ebd1aeb1e782b81beab538c

abeglova (changes requested):
Works as advertised! One last minor issue related to the subtopics property

abeglova (approved):
lgtm

PR #858, opened 2024-05-02
Additional routes to the Django app

jonkafton:
Adds additional routes for Nginx to forward to the Django backend:

- `/_/features/`
- `/scim/`
- `/o/`
- `/disconnect/`
- `/podcasts/rss_feed`
- `/__debug__/`
- `/program_letter/([0-9]+)/`

This matches for front end handling of the view route:
- `/program_letter/([0-9]+)/view`

### DIFF
--- a/config/nginx.conf
+++ b/config/nginx.conf
@@ -20,6 +20,10 @@ server {
         try_files /static$uri $uri @index;
     }
 
+    location ~ ^/program_letter/([0-9]+)/view$ {
+        try_files /index.html =404;
+    }
+
     location @index {
         try_files /index.html =404;
     }
@@ -28,7 +32,7 @@ server {
         return 204;
     }
 
-    location ~ ^/(api|login|complete/ol-oidc|logout|admin|static/admin|static/rest_framework|static/hijack|_features) {
+    location ~ ^/(api|login|complete/ol-oidc|logout|admin|static/admin|static/rest_framework|static/hijack|_/features/|scim/|o/|disconnect/|podcasts/rss_feed|__debug__/|program_letter/([0-9]+)/) {
         include uwsgi_params;
         uwsgi_pass web:8061;
         uwsgi_pass_request_headers on;
--- a/config/nginx.conf
+++ b/config/nginx.conf
@@ -32,7 +32,7 @@ server {
         return 204;
     }
 
-    location ~ ^/(api|login|complete/ol-oidc|logout|admin|static/admin|static/rest_framework|static/hijack|_/features/|scim/|o/|disconnect/|podcasts/rss_feed|__debug__/|media/|program_letter/([0-9]+)/) {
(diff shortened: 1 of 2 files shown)


shanbady (changes requested):
the views listed appear to work now. I think we also need to add /media

Also - is there a need for manually pinning these paths at all? I'm finding that if we leave location as ```location ~ ^/``` it automatically falls back to the django paths. this way we dont need to maintain a list of paths manually

shanbady:
> the views listed appear to work now. I think we also need to add /media
> 
> Also - is there a need for manually pinning these paths at all? I'm finding that if we leave location as `location ~ ^/` it automatically falls back to the django paths. this way we dont need to maintain a list of paths manually

@ChristopherChudzicki wanted a second take on this. we might not need to explicitly pin anything at all to the location unless there is a specific case that we need this for

jonkafton:
> Also - is there a need for manually pinning these paths at all? I'm finding that if we leave location as `location ~ ^/` it automatically falls back to the django paths. this way we dont need to maintain a list of paths manually

The issue is that we need everything to fall back to the index.html so that the less predictable paths and 404s are handled by the front end SPA.

jonkafton:
> I think we also need to add /media

Added that one.

shanbady (approved):
This works for the routes listed. At some point we may want to consider handling this by adding a catch all url path that redirects to the frontend index file if all django paths are exhausted.

shanbady (changes requested):
sorry just noticed another /profile/

shanbady:
> > Also - is there a need for manually pinning these paths at all? I'm finding that if we leave location as `location ~ ^/` it automatically falls back to the django paths. this way we dont need to maintain a list of paths manually
> 
> The issue is that we need everything to fall back to the index.html so that the less predictable front end paths and 404s are handled by the front end SPA.

There is a way to make this happen.

we can set location to ```location ~ ^/``` and then add a catch all route that goes to the frontend in django ```re_path(r".*/", index, name="main-index"),``` after [this line](https://github.com/mitodl/mit-open/blob/main/main/urls.py#L65) . What that does is it make it so that once all the django routes are exhausted, it defaults back to the frontend.

jonkafton:
> sorry just noticed another /profile/

Does the Django up server profile images? Added just in case.

shanbady (approved):
going to approve this since it resolves the routes we currently have but if maintenance of pinning the urls gets to be problematic we should consider the default routing mechanism via django i suggested and remove the pinned paths.

jonkafton:
> There is a way to make this happen.
> 
> we can set location to `location ~ ^/` and then add a catch all route that goes to the frontend in django `re_path(r".*/", index, name="main-index"),` after [this line](https://github.com/mitodl/mit-open/blob/main/main/urls.py#L65) . What that does is it make it so that once all the django routes are exhausted, it defaults back to the frontend.

Once we deploy the front end to a CDN the Django app will have no visibility of the index file. Likely `mitopen.odl.mit.edu` will point to the static file service / CDN and something like `api.mitopen.odl.mit.edu` will point to the Django app and it will handle requests on all routes. This nginx routing bit is just to keep things working until the rest is set up, so we probably shouldn't put too much energy into it.

PR #862, opened 2024-05-02
Add certificate as a real database field to LearningResource

mbertrand:
- Closes https://github.com/mitodl/hq/issues/3970
- Will partially address https://github.com/mitodl/hq/issues/4082 (some n+1 queries were due to on-the-fly calculation of certification property).

Makes `certificate` a real db field for `LearningResource` instead of a property, and adds calculation of the appropriate value for it to ETL pipelines (fwiw, the old calculation in the property was sometimes wrong).

### How can this be tested?
- Copy .env values for `PROLEARN_*`, `MITX_ONLINE_*`, and `EDX_API_*` from heroku.
- Run the migration, either via `docker compose up` or `./manage.py migrate`
- Run the following:
   ```
   ./backpopulate_xpro_data
   ./backpopulate_mit_edx_data
   ```
- Check that all courses from the following professional offerors have `certification=True`: xpro, csail, ctl, scc, see, mitpe, bootcamps.
- Courses from offeror mitx will have values True or False depending on whether any published runs have an availability value other than "Archived"


learning_resources/etl/utils.py
```diff
@@ -666,3 +668,20 @@ def transform_format(resource_format: str) -> list[str]:
 
     """
     return [RESOURCE_FORMAT_MAPPING[resource_format]]
```
abeglova, line 670: Unrelated to your PR but while testing this, the backpopulate_xpro_data task broke because a course had resource_format "Other" can we change it to 

```suggestion
        return [RESOURCE_FORMAT_MAPPING.get(resource_format, None)]

```

here? Or i can add a seperate pr

abeglova (approved):
LGTM

learning_resources/etl/utils.py
mbertrand: Yeah, I had a conversation about this w/Peter last week, "Other" should not be an option for xpro, it's not valid.  But at the time, no course/program was using it.  As of this morning, one xpro course is using it, and I created a new issue about it: https://github.com/mitodl/hq/issues/4201

mbertrand: For now I will change it to log an error and assign a default of "Online" if there is a `KeyError` in this PR.

PR #863, opened 2024-05-02
configure api BASE_PATH

gumaerc:
Closes https://github.com/mitodl/mit-open/issues/859

This PR adds a new env setting, `MITOPEN_AXIOS_BASE_PATH`. This setting controls `BASE_PATH` in `frontends/api/clients.ts` and allows you to build the API targeting a specific host rather than a relative path.

### How can this be tested?
It's difficult to test actually making a CORS request, but we can set `BASE_PATH` to an invalid value and intentionally break the frontend

 - Set `MITOPEN_AXIOS_BASE_PATH=https://example.com/` in your `.env` file
 - Spin up `mit-open` with `docker compose up`
 - Hit http://localhost:8063/search?topic=Artificial+Intelligence&resource_type=course
 - You should see "No results found for your query." and see errors in the console
 - Remove the value from your env and restart your containers
 - Repeat the process above, and you should see search results

### Additional Context
This functionality is being added in pursuit of the future addition of a separate webpack module bundle of exported mit-open components for use in other applications such as OCW


abeglova (approved):
👍

PR #875, opened 2024-05-06
Ab sort dropdown

abeglova:
closes https://github.com/mitodl/hq/issues/4128

This pr adds a dropdown for sort to the search ui. It also extracts dropdown code into a component so we can style all dropdowns togather

Screenshot 2024-05-09 at 11 26 36 AM
Screenshot 2024-05-09 at 11 27 15 AM

Screenshot 2024-05-09 at 11 29 13 AM
Screenshot 2024-05-09 at 11 29 01 AM

### How can this be tested?
- Go to http://localhost:8063/search verify that you can use the sort dropdown to change the order of the items 
- Go to a channel page with search enabled (search_filter set in the database/admin), for example http://localhost:8063/c/department/biology. Verify that all dropdowns also still work

### DIFF
--- a/frontends/mit-open/src/pages/SearchPage/SearchPage.tsx
+++ b/frontends/mit-open/src/pages/SearchPage/SearchPage.tsx
@@ -11,6 +11,7 @@ import {
   Typography,
   PlainList,
   Skeleton,
+  SimpleSelect,
 } from "ol-components"
 import { MetaTags } from "ol-utilities"
 
@@ -69,6 +70,25 @@ const AGGREGATIONS: LRSearchRequest["aggregations"] = [
   "offered_by",
 ]
 
+const SORT_OPTIONS = [
+  {
+    label: "Relevance",
+    key: "",
+  },
+  {
+    label: "New",
+    key: "new",
+  },
+  {
+    label: "Popular",
+    key: "-views",
+  },
+  {
+    label: "Upcoming",
+    key: "upcoming",
+  },
+]
+
 const ColoredHeader = styled.div`
   background-color: #394357;
   height: 150px;
@@ -80,6 +100,17 @@ const SearchField = styled(SearchInput)`
   width: 100%;
 `
 
+export const StyledDropdown = styled(SimpleSelect)`
+  margin: 8px;
+  min-width: 140px;
+`
+
+const SortContainer = styled.div`
+  ${({ theme }) => theme.breakpoints.up("sm")} {
+    float: right;
+    padding-right: 12px;
+  }
+`
 const FacetStyles = styled.div`
   * {
     color: ${({ theme }) => theme.palette.secondary.main};
@@ -263,7 +294,6 @@ const SearchPage: React.FC = () => {
     },
     [setSearchParams],
   )
-
   const facetManifest = useFacetManifest()
   const onFacetsChange = useCallback(() => {
     setPage(1)
@@ -278,6 +308,7 @@ const SearchPage: React.FC = () => {
     currentText,
     setCurrentText,
     setCurrentTextAndQuery,
+    setParamValue,
   } = useResourceSearchParams({
     searchParams,
     setSearchParams,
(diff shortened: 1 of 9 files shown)


Ferdi:
What is the default sort ?

abeglova:
The default search is whatever is returned by open search. "relevance" if there is a text query but no particular order if there isn't

abeglova:
I'm not sure what sort makes sense if there is no filter

Ferdi:
I think we should rename the options to

Relevance
New
Popular
Upcoming

and make "new" the default sorting when there no search term

abeglova:
@Ferdi , sure i can make that change. "Relevance" is confusing when there is no text since it sorts by an unknown order. Does it make sense to to just add a sort by new clause to all searches where there is no text?

Ferdi:
> Does it make sense to to just add a sort by new clause to all searches where there is no text?

I think that's a good idea.

frontends/ol-components/src/components/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,82 @@
...
+const Dropdown: React.FC<DropdownProps> = ({
+  title,
+  keyName,
+  initialValue,
+  isMultiple,
+  onChange,
+  options,
+}) => {
+  return (
+    <FormControl key={keyName}>
```
ChristopherChudzicki, line 54: I don't think FormControl adds anything here since we aren't using (and don't want) MUI's `<FormControlLabel>`. I would just get rid of it. The `key` prop definitely doesn't do anything.

Also, I would suggest:
- add a className prop and pass it to `Select`. That enables styling via `styled(Dropdown)`.

frontends/ol-components/src/components/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,82 @@
...
+      <Select
+        multiple={isMultiple}
+        displayEmpty
+        value={initialValue as string}
+        renderValue={() => {
+          return title
+        }}
+        onChange={onChange}
+        sx={{ m: 1, minWidth: 140 }}
```
ChristopherChudzicki, line 63: Could you remove this and style via `styled`? (Or allow parent to pass the `sx`, but I don't think `ol-component` should have a margin.

frontends/ol-components/src/components/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,82 @@
...
+   * value for the dropdown option
+   */
+  key: string
+  /**
+   * label for the dropdown option
+   */
+  label: string
+}
+
+const Dropdown: React.FC<DropdownProps> = ({
```
ChristopherChudzicki, line 45: It seems to me that "Dropdown" is mostly a simplified interface for MUI's Select component. I think that's good and convenient. My understanding of "Dropdown" though is that it's usually a menu, not something that stores a value—that's a Select.

 What would you think of calling it `SimpleSelect` to emphasize that (a) it's a Select component, (b) it's simpler, and (c) it's sort of parallel to our existing SimpleMenu (which is a simplified MUI Menu interface).

ChristopherChudzicki (changes requested):
My requested change is: Show `"Sort by: <value>"` in the dropdown.

As discussed in slack, I think good way to do this is allow passing `renderValue` and use something like
```ts
      renderValue={(value) => {
        const opt = options.find((option) => option.key === value)
        return `${title}: ${opt?.label}`
      }}
```

frontends/ol-components/src/components/SimpleSelect/SimpleSelect.tsx
```diff
@@ -0,0 +1,74 @@
...
+      displayEmpty
+      value={initialValue}
+      onChange={onChange}
+      className={className}
+      renderValue={renderValue}
+    >
+      {options.map((option, index) => (
+        <MenuItem
+          value={option.key.toString()}
+          key={"option-".concat(index.toString())}
```
ChristopherChudzicki, line 64: Probably better to just use `option.value`. Keys should be unique and specific to the item. Index is usually a bad choice (and ESLint should warn about it, but probably won't if you prepend a string). Index is bad because the index isn't really unique to the item—item indices change if the list is reordered items are prepended.

It's probably not an issue here, since the menu items are unlikely to change, but IMO just use `option.value`. The values **need** to be unique anyway.

frontends/mit-open/src/pages/SearchPage/SearchPage.tsx
```diff
@@ -357,6 +388,20 @@ const SearchPage: React.FC = () => {
...
+                <StyledDropdown
+                  initialValue={params.sortby || ""}
+                  isMultiple={false}
+                  onChange={(e) => setParamValue("sortby", e.target.value)}
+                  options={SORT_OPTIONS}
+                  renderValue={(value) => {
+                    const opt = SORT_OPTIONS.find(
+                      (option) => option.key === value,
+                    )
+                    return `Sort: ${opt?.label}`
```
ChristopherChudzicki, line 401: "Sort by: ..." seems more natural and is closer to the tentative designs.

ChristopherChudzicki (approved):
Approved. Left one suggestion that I don't think needs a re-review.

PR #881, opened 2024-05-08
Display loading view for search page

jonkafton:

https://github.com/mitodl/hq/issues/4131.

Adds a skeleton view while search page and field search results initially load.

A small change also included to offset the padding on the text input component so the contents don't shift on hover as we increase the border width (testable on the home page hero search).

### How can this be tested?

Navigate to `/search`. The page should briefly display the loading state:

image

Navigate to a departments page, e.g. `/c/department/media-arts-and-sciences/`. The page should briefly display the loading state:

image

### Additional Context

- We may want to give similar treatment to the Facet Display as it can display the filter topics and "offered by" title and search inputs up front with skeleton aggregation results for a gentler page update on load (though this lives in course-search-utils).

- The loading view is only displayed on initial load and not while making additional searches. This may be ok as the previous content is displayed while the new content is loaded. For this we'd need to return an `isLoading` flag from `useResourceSearchParams` (also in course-search-utils).

### DIFF
--- a/frontends/mit-open/src/pages/FieldPage/FieldSearch.tsx
+++ b/frontends/mit-open/src/pages/FieldPage/FieldSearch.tsx
@@ -6,6 +6,7 @@ import {
   Card,
   CardContent,
   PlainList,
+  Skeleton,
 } from "ol-components"
 import { getReadableResourceType } from "ol-utilities"
 
@@ -134,6 +135,10 @@ export const FieldSearchControls = styled.div`
   margin-top: 30px;
 `
 
+const StyledSkeleton = styled(Skeleton)`
+  border-radius: 4px;
+`
+
 const PAGE_SIZE = 10
 
 interface FeildSearchProps {
@@ -205,7 +210,7 @@ const FieldSearch: React.FC<FeildSearchProps> = ({
 
   const page = +(searchParams.get("page") ?? "1")
 
-  const { data } = useLearningResourcesSearch(
+  const { data, isLoading } = useLearningResourcesSearch(
     {
       ...(allParams as LRSearchRequest),
       aggregations: facetNames as LRSearchRequest["aggregations"],
@@ -221,6 +226,7 @@ const FieldSearch: React.FC<FeildSearchProps> = ({
           <GridColumn variant="main-2-wide-main">
             <AvailableFacetsDropdowns
               facetMap={facetManifest}
+              isLoading={isLoading}
               activeFacets={allParams as Facets & BooleanFacets}
               onFacetChange={setParamValue}
               facetOptions={(name) => data?.metadata.aggregations?.[name] ?? []}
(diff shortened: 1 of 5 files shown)


ChristopherChudzicki (changes requested):
Two comments:

1. Apologies for not including this in the issue description—Could you include the loading state in FieldSearch, too, please? (Which ought to be renamed "ChannelSearch". "Channel" is the general concept for individual department/topic/platform pages.) Click on a department in https://mitopen-rc.odl.mit.edu/departments and you'll go to a "Department channel page". The general idea is that

2. Separately—it occurs to me that <CardTemplate /> could probably be slightly reworked so that we could just pass skeletons to the title, image, etc if we wanted a slightly fancier loader. (E.g., passing a skeleton as `title` should work now, except the TS type would have to change from `string` to `React.ReactNode`; img would need a slight change.) Main advantage might be a more standard way to control the height.

(1) is a requested change; (2) maybe not worth doing, but thought for future.

jonkafton:
> 1. Apologies for not including this in the issue description—Could you include the loading state in FieldSearch, too, please?

No problems, those are now added.

> Separately—it occurs to me that could probably be slightly reworked so that we could just pass skeletons to the title, image, etc if we wanted a slightly fancier loader. 

Are you thinking to pass into a ResourceCard, for example? It would be good to create adjacent loading versions, e.g. ResourceCardLoading, that return load views that represent their internal layout rather than just blocking out, though it could look busy - I think that's one for the design team.

ChristopherChudzicki:
> Are you thinking to pass into a ResourceCard, for example? It would be good to create adjacent loading versions, e.g. ResourceCardLoading, that return load views that represent their internal layout rather than just blocking out, though it could look busy - I think that's one for the design team.

That is what I was thinking. The card uses CardTemplate, so presumably you could do
```
const LoadingCard = () => {
 <CardTemplate
    title={<Skeleton ... />}
    image={<Skeleton ... />} // these skeletons would all be slightly different props, I asusume.
    details={<Skeleton ... />}
  />
}
```
This is roughly the approach I took in https://github.com/mitodl/mit-open/blob/f8fae44143aa1064c127c8ff29fd0ac666176a4e/frontends/ol-components/src/components/learning-resources/ExpandedLearningResourceDisplay/ExpandedLearningResourceDisplay.tsx#L83 (which displays in the drawer when you click a card title)

ChristopherChudzicki (approved):

jonkafton:
> > Are you thinking to pass into a ResourceCard, for example? It would be good to create adjacent loading versions, e.g. ResourceCardLoading, that return load views that represent their internal layout rather than just blocking out, though it could look busy - I think that's one for the design team.
> 
> That is what I was thinking. The card uses CardTemplate, so presumably you could do
> 
> ```
> const LoadingCard = () => {
>  <CardTemplate
>     title={<Skeleton ... />}
>     image={<Skeleton ... />} // these skeletons would all be slightly different props, I asusume.
>     details={<Skeleton ... />}
>   />
> }
> ```

That could work if the skeleton placeholders fill the space correctly. I suspect they'd need some styling.

PR #885, opened 2024-05-09
Apply free filter to all resources

mbertrand:
Closes https://github.com/mitodl/hq/issues/4202

Applies the `free` api/search filter to all learning resources even if they are always free, not just courses and programs.

### How can this be tested?
http://localhost:8063/api/v1/podcasts/?free=True should return the same results as http://localhost:8063/api/v1/podcasts/

http://localhost:8063/api/v1/podcasts/?free=False should return 0 results

http://localhost:8063/api/v1/learning_resources_search/?resource_type=podcast&free=True should return the same results as http://localhost:8063/api/v1/learning_resources_search/?resource_type=podcast

http://localhost:8063/api/v1/learning_resources_search/?resource_type=podcast&free=False should return 0 results

### DIFF
--- a/learning_resources/filters.py
+++ b/learning_resources/filters.py
@@ -106,9 +106,9 @@ class LearningResourceFilter(FilterSet):
 
     def filter_free(self, queryset, _, value):
         """Free cost filter for learning resources"""
-        queryset = queryset.exclude(runs__isnull=True)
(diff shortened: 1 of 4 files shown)


abeglova (approved):
LGTM

PR #887, opened 2024-05-09
Featured API endpoint

mbertrand:
Closes https://github.com/mitodl/hq/issues/4183

- Adds an API endpoint for "featured" resources that queries for all resources in the "featured_list" of each of 5 offeror channels, sorted by the position of the resource in each featured list.  The endpoint allows for the same filtering options as the learning_resources api endpoint.
- Adds a management command (for dev purposes only) that auto-generates a featured list for each of the 5 offerors and populates it with x number of published courses from the relevant offeror.

### How can this be tested?
- Run the necessary management commands if necessary to get courses from each of the 5 offerors (adding required .env values too)
  ```
  backpopulate_prolearn_data
  backpopulate_ocw_data --skip-contentfiles
  backpopulate_xpro_data
  backpopulate_mitxonline_data
  ```
- Run `backpopulate_resource_channels`
- Run `populate_featured_lists`
- Go to http://localhost:8063/api/v1/featured/, there should be a total of 50 results (10 from each of 5 offerors).  
- Check that the result ordering coincides with the position of the resource from its respective featured learning path (you can see those at http://localhost:8063/learningpaths).  So the first 5 results should be the 1st resource from each of the 5 featured lists, the next 5 results should be the 2nd resource from each of the lists, etc.
- Try out some filtering options like `?certification=true`, `professional=true`, `free=true`.  The results should be filtered appropriately.

### DIFF
--- a/learning_resources/management/commands/populate_featured_lists.py
+++ b/learning_resources/management/commands/populate_featured_lists.py
@@ -1,8 +1,8 @@
 """Management command to create dev-only featured lists for offeror channels"""
 
-import os
 import sys
 
+from django.conf import settings
 from django.contrib.auth.models import User
 from django.core.management import BaseCommand
 
@@ -32,7 +32,7 @@ class Command(BaseCommand):
 
     def handle(self, *args, **options):  # noqa: ARG002
         """Create a Favorites userlist for every active user"""
-        if os.environ.get("MITOPEN_ENVIRONMENT") == "production":
+        if settings.ENVIRONMENT == "production":
             self.stderr.write("This command is only available in dev/rc environments")
             sys.exit(1)
 
--- a/learning_resources/management/commands/populate_featured_lists.py
+++ b/learning_resources/management/commands/populate_featured_lists.py
@@ -32,8 +32,8 @@ class Command(BaseCommand):
 
     def handle(self, *args, **options):  # noqa: ARG002
         """Create a Favorites userlist for every active user"""
-        if not os.environ.get("DEV_ENV"):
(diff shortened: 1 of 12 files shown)


shanbady (changes requested):
Looks good. just wondering from the description - is the /learningpaths view supposed to be filterable via ?professional=true and free=true?

learning_resources/management/commands/populate_featured_lists.py
```diff
@@ -0,0 +1,93 @@
...
+            "resource_count",
+            nargs="?",
+            type=int,
+            default=10,
+            help="Set the number of courses per featured list (default is 10)",
+        )
+
+    def handle(self, *args, **options):  # noqa: ARG002
+        """Create a Favorites userlist for every active user"""
+        if os.environ.get("MITOPEN_ENVIRONMENT") == "production":
```
shanbady, line 35: seems like this could be read from settings.ENVIRONMENT instead of directly from the environment

mbertrand: good point, updated

mbertrand:
>  is the /learningpaths view supposed to be filterable via ?professional=true and free=true?
All learning resources including learning paths should be filterable by the options in the `LearningResourceFilter` class, even if they're all going to be the same value (for instance, all podcasts and videos are free).

shanbady (approved):
looks good. just one other thing (non-blocker) i was wondering is if it makes sense to back-populate resource channels for whoever runs this instead of throwing the error
Screenshot 2024-05-14 at 9 02 59 AM

PR #889, opened 2024-05-09
use neue-haas-grotesk font

ChristopherChudzicki:
https://github.com/mitodl/hq/issues/4205

Adds Neue Haas Grotest fonts to MIT Open.

See https://ordinary-kitty.surge.sh/?path=/docs/smoot-design-typography--docs

### How can this be tested?

**NOTES:** 

1. Look around the site. It should look similar but should be using a new font.
2. In particular, check that the search page looks good.
3.  You could check that the fonts in http://localhost:6006/?path=/docs/smoot-design-typography--docs match https://www.figma.com/file/Eux3guSenAFVvNHGi1Y9Wm/MIT-Design-System?type=design&node-id=104-822&mode=design&t=6B5Qy7dacJnxYeVE-0
4. Try adding `{ fontWeight: 500 }` (the value can be any number) to an object. You should get an ESLint warning.

### Other Notes

1. Browsers resolve fonts by matching font descriptors (e.g., `font-family`, `font-weight`, and `font-style`) to font declarations. Generally, normal and bold-weight fonts are separate fonts, not the same character set with different thicknesses. To see the actual font being rendered by Chrome, you can use the "Computed" panel in the "Elements" tab:
 
    Screenshot 2024-05-13 at 11 49 51 AM
2. **Figma:** Does not show the actual font for designs (e.g., "Neue Haas Grotesk Display Pro 75 Bold  -  56px") but instead shows
    1. The "name" if it is a named style (e.g., H1, S3, or Body2).
    2. CSS properties. _Except, these are inaccurate for font weight. See note on `.eslintrc`._

### DIFF
--- a/frontends/.eslintrc.js
+++ b/frontends/.eslintrc.js
@@ -79,6 +79,7 @@ module.exports = {
       },
     ],
     quotes: ["error", "double", { avoidEscape: true }],
+    "no-restricted-syntax": ["error", "Literal[value=fontWeight]"],
   },
   overrides: [
     {
--- a/frontends/.eslintrc.js
+++ b/frontends/.eslintrc.js
@@ -87,6 +87,12 @@ module.exports = {
         message:
           "Do not specify `fontWeight` manually. Prefer spreading `theme.typography.subtitle1` or similar. If you MUST use a fontWeight, refer to `fontWeights` theme object.",
       },
+      {
+        selector:
+          "Property[key.name=fontFamily][value.raw=/Neue Haas/], TemplateElement[value.raw=/Neue Haas/]",
+        message:
+          "Do not specify `fontFamily` manually. Prefer spreading `theme.typography.subtitle1` or similar. If using neue-haas-grotesk-text, this is ThemeProvider's default fontFamily.",
+      },
     ],
   },
   overrides: [
--- a/frontends/.eslintrc.js
+++ b/frontends/.eslintrc.js
@@ -79,7 +79,15 @@ module.exports = {
       },
     ],
     quotes: ["error", "double", { avoidEscape: true }],
-    "no-restricted-syntax": ["error", "Literal[value=fontWeight]"],
+    "no-restricted-syntax": [
+      "error",
+      {
+        selector:
+          "Property[key.name=fontWeight][value.raw=/\\d+/], TemplateElement[value.raw=/font-weight: \\d+/]",
+        message:
+          "Do not specify `fontWeight` manually. Prefer spreading `theme.typography.subtitle1` or similar. If you MUST use a fontWeight, refer to `fontWeights` theme object.",
+      },
+    ],
   },
   overrides: [
     {
--- a/frontends/.eslintrc.js
+++ b/frontends/.eslintrc.js
@@ -81,6 +81,12 @@ module.exports = {
     quotes: ["error", "double", { avoidEscape: true }],
     "no-restricted-syntax": [
       "error",
+      /**
+       * See https://eslint.org/docs/latest/rules/no-restricted-syntax
+       *
+       * The selectors use "ES Query", a css-like syntax for AST querying. A
+       * useful tool is  https://estools.github.io/esquery/
+       */
       {
         selector:
           "Property[key.name=fontWeight][value.raw=/\\d+/], TemplateElement[value.raw=/font-weight: \\d+/]",
--- a/frontends/mit-open/.storybook/preview-head.html
+++ b/frontends/mit-open/.storybook/preview-head.html
@@ -1,0 +1,5 @@
+<!--
+      Font files for Adobe Neue Haas Grotesk.
+      WARNING: This is linked to chudzick@mit.edu's Adobe account.
+  -->
+<link rel="stylesheet" href="https://use.typekit.net/lbk1xay.css" />
--- a/frontends/mit-open/public/index.html
+++ b/frontends/mit-open/public/index.html
@@ -3,16 +3,17 @@
   <head>
     <meta charset="UTF-8" />
     <meta name="viewport" content="width=device-width, initial-scale=1" />
-    <link
-      rel="stylesheet"
-      type="text/css"
-      href="https://fonts.googleapis.com/css?family=Roboto:300,400,400i,500,700"
-    />
+    <!--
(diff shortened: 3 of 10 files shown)


frontends/mit-open/.storybook/preview-head.html
```diff
@@ -0,0 +1,5 @@
+<!--
+      Font files for Adobe Neue Haas Grotesk.
+      WARNING: This is linked to chudzick@mit.edu's Adobe account.
+  -->
+<link rel="stylesheet" href="https://use.typekit.net/lbk1xay.css" />
```
pdpinch, line 5: How can you tell that it's associated with your email and not with MIT's site license?

Would you feel more comfortable using the same package as www.mit.edu, https://use.typekit.net/uby1wjv.css ?

ChristopherChudzicki: @pdpinch I don't really feel great using https://use.typekit.net/uby1wjv.css either. I think that's a link for an adobe "Web Fonts Project" (like the one I made) but who manages it? What if they change it? Or delete the account and swap to a different one? 

Regarding
> How can you tell that it's associated with your email and not with MIT's site license?
Well, I *think* it's both:
- My Adobe account is via MIT through Touchstone login. So in that sense it's the MIT site license.
- but I believe `lbk1xay` It's associated with me, because I created the project
    Screenshot 2024-05-13 at 1 03 00 PM

I'm not entirely sure it would disappear if my touchstone account were removed, but it seems reasonable to assume.

You probably found this, but I put related info on https://github.com/mitodl/hq/issues/4205#issuecomment-2103029701

ChristopherChudzicki: @pdpinch Re
> @pdpinch I don't really feel great using https://use.typekit.net/uby1wjv.css (the stylesheet for https://mit.edu) either. I think that's a link for an adobe "Web Fonts Project" (like the one I made) but who manages it? What if they change it? Or delete the account and swap to a different one? 

On the other hand, maybe that link is intended for internal MIT use outside the homepage. I haven't looked for documentation on that (within MIT) but it seems reasonable, and if we could find out more info about that, then that would be a great solution.

frontends/.eslintrc.js
```diff
@@ -79,6 +79,27 @@ module.exports = {
       },
     ],
     quotes: ["error", "double", { avoidEscape: true }],
+    "no-restricted-syntax": [
```
ChristopherChudzicki, line 82: TLDR: I added a few ESlint rules to encourage engineers to use named font styles (e.g., `theme.typography.body1`) rather than specifying weight and font-family.

The reasons are:
1. Our "design system" has a few named styles (`H1...H5, `Subtitle1...Subtitle4`, `body1...body4`, `buttonLarge, button, buttonSmall`). The designs should always be using these and we should be using them in the code, too. This helps with consistency and if we need to tweak one down the road, it will be easy.
2. But also... **Figma CSS is incorrect for font weight**.
    - For example, our subtitle font is is "Neue Haas Grotesk Display  65 Medium".
        - Adobe says this is `font-family: neue-haas-grotesk-display`, `font-weight: 600`. (See screenshot below).
        - Figma thinks this is `font-family: Neue Haas Grotesk Display Pro`, `font-weight: 500`. (See screenshot below).
    - The correspondence of css font descriptors (weight, style) to font glyphs is provided by a stylesheet. The same font glyphs can be called `font-weight: 500` by one stylesheet and `font-weight: 600` by another stylesheet. Figma does not have any way, as far as I can tell, to establish the `weight <--> font` correspondence. It seems to just guess based on heuristics like "semi-bold = 600, bold = 700, heavy = 800, black = 900`, etc.
        - this seems like a common issue with figma, e.g., https://forum.figma.com/t/how-should-we-communicate-web-font-css-specs-when-inspect-mode-is-wrong/19355 
    - Adobe provides the font stylesheet, so Adobe's values are the one we should trust. [^1]

**Adobe:**
Screenshot 2024-05-13 at 1 19 08 PM

**Figma:**
Screenshot 2024-05-13 at 1 18 47 PM Screenshot 2024-05-13 at 1 18 36 PM




[^1]: The adobe-provided stylesheet is https://use.typekit.net/lbk1xay.css. It includes a bunch of `@font-face` declarations and URLS to the fonts. I guess we could manually edit https://use.typekit.net/lbk1xay.css, but I don't think we should, since the format of the URLs in https://use.typekit.net/lbk1xay.css could change. Whereas the stylesheet https://use.typekit.net/lbk1xay.css itself is clearly public and intended for use.

frontends/mit-open/.storybook/preview-head.html
ChristopherChudzicki: I'll add—using the "Web Project" fonts stylesheet i created is intended as an interim solution.

Assuming we merge this, I'll open an issue based on https://github.com/mitodl/hq/issues/4205#issuecomment-2103029701 for finding a better alternative.

gumaerc (approved):
👍 

This looks good, but please do create the issue you mentioned regarding getting the direct Adobe link hooked up to an account that could potentially be administered by anyone. Even if you don't plan on leaving, dealing with issues relating to the font embed on Adobe's side will rely on your account. If you're out sick or on vacation, we shouldn't have to get in touch with you to log in and fix something. This should be tied to a generic "open learning" account with the credentials for it stored in Keybase so the whole team can access it. If there is a better place to let new developers looking at our Figma designs know that the font weights may not be accurate, we should do that. At a bare minimum if we are enforcing "no custom typography rules" then we should be very strict about making sure the typograhy names in Figma match what we have in the theme. Looking at Figma it seems we have adjusted the names to be more accurate, but my point is that we should never be in a situation where a developer looks at a design and sees a typography name that doesn't exist.

ChristopherChudzicki:
> 👍
> 
> This looks good, but please do create the issue you mentioned regarding getting the direct Adobe link hooked up to an account that could potentially be administered by anyone.

- https://github.com/mitodl/hq/issues/4237

frontends/mit-open/.storybook/preview-head.html
ChristopherChudzicki: The only technical guidelines I found on https://brand.mit.edu/typography about font inclusion were, more or less, "follow Adobe's procedure via your touchstone account"

PR #892, opened 2024-05-10
 Give video/podcast/learning_path resources a default learning format of ["online"]

mbertrand:
Related to https://github.com/mitodl/hq/issues/3972

Makes the default `learning_format` == `['online']` for all non-program/course resources

### How can this be tested?
The following urls should return the same number of results after the new migration is run:

http://localhost:8063/api/v1/learning_resources/?resource_type=video,video_playlist,learning_path,podcast,podcast_episode&learning_format=online

http://localhost:8063/api/v1/learning_resources/?resource_type=video,video_playlist,learning_path,podcast,podcast_episode


learning_resources/models.py
```diff
@@ -179,7 +180,7 @@ class LearningResource(TimestampedModel):
     )
     learning_format = ArrayField(
         models.CharField(max_length=24, db_index=True),
-        default=list,
+        default=[LearningResourceFormat.online.name],
```
rhysyngsun, line 183: This should be a callable instead because otherwise the value is mutable and shared everywhere.
```suggestion
        default=lambda: [LearningResourceFormat.online.name],
```

rhysyngsun (changes requested):
Code review.

rhysyngsun (approved):
LGTM

PR #896, opened 2024-05-10
Replace react-dotdotdot with CSS

ChristopherChudzicki:
Relates to
- https://github.com/mitodl/hq/issues/4205 particularly comment https://github.com/mitodl/hq/issues/4205#issuecomment-2105413877
- https://github.com/mitodl/course-search-utils/issues/75

This PR replaces `react-dotdotdot` with CSS. For the search page, the motivation is described in https://github.com/mitodl/course-search-utils/issues/75. Elsewhere... `react-dotdotdot` seems not to work with our line heights and `neue-haas-grotesk`. I do not entirely understand why. That said, this css-based solution has good browser support and other advantages. E.g., it doesn't literally truncate the text string, just the displayed value. That seems desirable for screenreaders, particularly in the case of search page facets.

[Storybook](https://shy-guide.surge.sh/?path=/docs/ol-components-truncatetext--docs)

Screenshot 2024-05-10 at 7 55 18 PM

Screenshot 2024-05-10 at 7 55 52 PM

### How can this be tested?
1. Reinstall js dependencies. If you're running the app in docker, you'll want to `docker compose build watch`.
2. View http://localhost:6006/?path=/docs/ol-components-truncatetext--docs in several browsers
3. View the search page. Long facets should be truncated to one line. You might need to narrow your browser window.


frontends/ol-components/src/components/TruncateText/TruncateText.tsx
```diff
@@ -0,0 +1,28 @@
...
+  lineClamp?: number | "none"
+}
+
+const truncateText = (lines?: number | "none") =>
+  css({
+    display: "-webkit-box",
+    WebkitBoxOrient: "vertical",
+    overflow: "hidden",
+    textOverflow: "ellipsis",
+    WebkitLineClamp: lines,
```
ChristopherChudzicki, line 17: This says webkit, but works in all slightly modern browsers beyond IE. This will not work in IE, but will fallback should be somewhat graceful—no text truncation, all text visible. This seems acceptable.

In contrast, the behavior of `react-dotdotdot` with our font + line height does not fail gracefully: when set to `clamp={x}`, it seems generally to clamp `x-1`. In particular, that means clamping to 1 line clamps to zero.

frontends/ol-components/src/components/TruncateText/TruncateText.stories.tsx
```diff
@@ -0,0 +1,42 @@
...
+  },
+}
+
+export default meta
+
+type Story = StoryObj<typeof TruncateText>
+
+export const None: Story = {
+  args: {
+    lineClamp: "none",
```
gumaerc, line 30: A bit of a nitpick, but the "none" example here in Storybook still says "This is a long text that should be truncated after a few lines" when it fact because `lineClamp` is set to "none," it should not be truncated.

gumaerc (approved):
This looks good, and the logic surrounding the change is sound. I just have one suggestion for the Storybook, and as a reminder https://github.com/mitodl/course-search-utils/pull/104 needs to be merged and released so the version reference can be updated here before merging this PR.

PR #898, opened 2024-05-13
Switch django migrations to release phase

rhysyngsun:

Closes https://github.com/mitodl/hq/issues/4227

This switches when we run the migrations from the `bin/post_compile` script (which is now removed since that's all it was doing) to the release phase script that was already present for the cache table generation

### How can this be tested?

- We'll need to test this in a heroku deployment after merging.


shaidar (approved):

gumaerc (approved):
👍

PR #899, opened 2024-05-13
Null start dates for OCW course runs

mbertrand:
Closes https://github.com/mitodl/hq/issues/4154

Makes start date null for OCW courses

### How can this be tested?
If you have existing OCW resources, they should all have a null start date after the migration is run.
If you run `./manage.py backpopulate_ocw_data --course-name <ocw_course> --skip-contentfiles`, that course should still have a null start date afterward.


shanbady (approved):
looks good

PR #906, opened 2024-05-14
Fix 500 error loading attestations without a cover image

jkachel:

Closes mitodl/hq#4249

The cover image field used a `URLField` in the serializer, but the field is optional; if the field is None, this generates an error. This update moves that back to a `SerializerMethodField` so we can return None if there's no cover image.

### How can this be tested?

Automated tests should pass. 

The testing scenario in 4249 should also work (i.e., create an attestation without a cover image and try to load it).

### DIFF
--- a/testimonials/serializers.py
+++ b/testimonials/serializers.py
@@ -11,7 +11,11 @@ class AttestationSerializer(serializers.ModelSerializer):
     avatar = serializers.URLField(source="avatar.url")
     avatar_small = serializers.URLField(source="avatar_small.url")
     avatar_medium = serializers.URLField(source="avatar_medium.url")
-    cover = serializers.URLField(source="cover.url")
+    cover = serializers.SerializerMethodField()
+
+    def get_cover(self, attestation) -> str | None:
+        """Get the cover image URL"""
+        return attestation.cover.url if attestation.cover else None
 
     class Meta:
         """Meta options for the serializer"""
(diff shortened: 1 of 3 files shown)


testimonials/serializers.py
```diff
@@ -11,7 +11,11 @@ class AttestationSerializer(serializers.ModelSerializer):
     avatar = serializers.URLField(source="avatar.url")
     avatar_small = serializers.URLField(source="avatar_small.url")
     avatar_medium = serializers.URLField(source="avatar_medium.url")
-    cover = serializers.URLField(source="cover.url")
```
shanbady, line 14: if you set it to the [following](https://www.django-rest-framework.org/api-guide/fields/#allow_null) it should have the same effect 
```
cover = serializers.URLField(source="cover.url",
        allow_null=True,
        read_only=True,
        help_text="Get the cover image URL")
```

shanbady (changes requested):

testimonials/serializers.py
jkachel: It does not; I still get a 500 error with the change made. This is trying to reference an attribute in the resulting FieldFile object so a null here means it's trying to pull from an incomplete object.

shanbady (approved):
interesting. I did manage to get it to work by switching to an imagefield which might be worth considering but i will leave that up to you since what you have here works fine too
```
cover = serializers.ImageField(
        use_url=True, 
        allow_null=True,
        read_only=True,
        help_text="Get the cover image URL")
```

PR #907, opened 2024-05-14
Update profile API with onboarding data fields

rhysyngsun:
Blocked by (and hence based on): https://github.com/mitodl/mit-open/pull/895

Part of https://github.com/mitodl/hq/issues/4184

This updates the profile API to have additional fields we need for the personalization/onboarding flow. Since we've decided to require a confirmed email this will just use standard authentication.

### How can this be tested?

You should be able to go to http://open.odl.local:8063/api/v0/profiles/mitodl/ and update the new fields on here with the exception of Interests and Goals which DRF doesn't support in its browseable API. Swagger-ui does support these though, so you can test it over there too.


mbertrand (changes requested):
LGTM but some unit tests need to be updated to include the new fields.

mbertrand:
fwiw I was able to update "Interests" from the DRF page using JSON, for example:

```json
{
    "interests": [
        "electrical-engineering",
        "information-technology"
    ]
}
```

mbertrand (approved):
👍

PR #912, opened 2024-05-15
Fix flaky python test

mbertrand:
Closes https://github.com/mitodl/hq/issues/4262

Sorts content tags by name, which should hopefully avoid the intermittent python unit test error for learning resource serializerd data.

### How can this be tested?
Unit tests should pass.


rhysyngsun (approved):
LGTM

PR #916, opened 2024-05-15
Add learner testimonials homepage UI

jkachel:

Closes mitodl/hq#4253

Adds learner testimonials to the homepage.

- Desktop screenshots

Screenshot 2024-05-31 at 2 47 34 PM

- Mobile width screenshots

Screenshot 2024-05-31 at 2 48 01 PM

### How can this be tested?

Add some testimonials to the system. The content doesn't matter so much, but be aware that there's only so much room on the cards (especially in mobile mode) so you should keep the quote to one or two short sentences. If you set a publish date for the testimonial, you may not see it on the homepage.

View the homepage. You should see the testimonials appear on the page and you should be able to page through them. This should work in desktop or mobile views. 

### Additional Context

Testimonials also appear in the channel page but there will be a separate PR for that.


frontends/mit-open/src/pages/HomePage/TestimonialsSection.tsx
```diff
@@ -0,0 +1,284 @@
...
+                pageLeftIcon={<RiArrowDropLeftLine />}
+                pageRightIcon={<RiArrowDropRightLine />}
+                buttonAlignment="center"
+                buttonVariant="inverted"
+                buttonSize="large"
+                wrapMode="wrap"
+                scrollDistance={"slide"}
+              >
+                <TestimonialFadeLeft>
+                  <div></div>
```
ChristopherChudzicki, line 247: Could the gradient but on `TestimonialFadeLeft` itself rather than on the child div? I didn't understand why it was this way.

frontends/mit-open/src/pages/HomePage/TestimonialsSection.tsx
```diff
@@ -0,0 +1,284 @@
...
+  },
+  [theme.breakpoints.down("md")]: {
+    display: "none",
+  },
+})
+
+const TestimonialsSection: React.FC = () => {
+  return (
+    <Section>
+      <Container id="hamster-noises">
```
ChristopherChudzicki, line 228: Squeak Squeak.

`<Container> ... </Container>` is https://mui.com/material-ui/react-container/ and ensures the contents have max width at various breakpoints.

However, for this carousel, we DO want fully true width. So the carousel should not be inside `Container`.

Note: Changed on my experimental `slick` branch.

ChristopherChudzicki (changes requested):
@jkachel The API changes look great and overall the carousel looked good. I did notice a few issues, some of which we discussed in slack:

1. Looping (clarified with Steve) the carousel should loop infinitely.
2. Gradients... The gradients seem to be covering the central card at some widths. They should not.
3. The carousel should be full-width, not constrained to the central page area. (I.e., move it outside of `Container`.)

Here is a screenshot illustrating (2) and (3) to some extent.

Screenshot 2024-06-04 at 3 54 31 PM


### React Slick
As discussed in slack, I think we should use `react-slick` for the testimonial carousel. I tried it out on this branch https://github.com/mitodl/mit-open/compare/jkachel/4253-interior-testimonials...jkachel/4253-homepage-testimonials-cc?expand=1 and worked pretty well:
- has a centered mode
- has infinite looping
- was easy to position the buttons.

I do think we should keep nuka around for now—it has a "fixed-width" card mode that I'm not sure react-slick has good support for. (It may, via "variable width" as you pointed out in slack.)

Here's a screenshot from my branch:

Screenshot 2024-06-04 at 3 58 36 PM

jkachel:
Some notes about the changes that are in here:
- I adjusted the card width to be variable rather than fixed - this allows Slick Carousel to determine the width of the card and ensures they're in the center.
- I did rearrange the settings for the Slick component so they're in a separate JS object. This is mostly to support using the `responsive` setting, so we can adjust the display when on smaller screens (and I wanted to keep all the config together). 
- There was an unused import in `Button` that I cleaned up.

FWIW, this is what it looks like with a long name/title in place:
Desktop
image

Mobile
image

frontends/mit-open/src/pages/HomePage/TestimonialsSection.tsx
```diff
@@ -0,0 +1,254 @@
...
+  },
+
+  "div.testimonial-quote-closer": {
+    textAlign: "right",
+    width: "100%",
+  },
+})
+
+const OverlayContainer = styled.div({
+  position: "relative",
```
ChristopherChudzicki, line 126:
```suggestion
  position: "relative",
  maxWidth: "1900px",
  margin: "0 auto",
```
Per this discussion https://mitodl.slack.com/archives/C03K5HYGPT9/p1717533212840589?thread_ts=1717523918.589909&cid=C03K5HYGPT9, let's add a max width beyond our usual 1272px.

ChristopherChudzicki (approved):
👍 Looks and works great!

One small css request I tried it out and sent a screenshot to Bilal, but could you double check it, too? IMO no need to re-review that change.

PR #920, opened 2024-05-16
Fix schema issue that was breaking redoc

rhysyngsun:

N/A

This resolves an error that was causing Redoc to fail to load. Specifically the `$ref` in the `channel_type` field wasn't correct (`ChannelType` instead of `#/components/schema/ChannelType)

### How can this be tested?

Run `docker compose up` and go to redoc: http://open.odl.local:8063/api/v0/schema/redoc/

The page should load without error.


mbertrand (approved):
👍

PR #941, opened 2024-05-21
Fixes button styles to match design

jonkafton:

Relates to [Homepage: "Browse by Topics" #4185](https://github.com/mitodl/hq/issues/4185)

"See all" button styles are awry in Release Candidate https://github.com/mitodl/mit-open/pull/940:

Current:

image

With fix:

image

### DIFF
--- a/frontends/ol-components/src/components/ThemeProvider/typography.ts
+++ b/frontends/ol-components/src/components/ThemeProvider/typography.ts
@@ -141,6 +141,7 @@ const globalSettings: ThemeOptions["typography"] = {
     fontStyle: "normal",
     fontSize: pxToRem(14),
     lineHeight: pxToRem(14),
+    textTransform: "none",
   },
   buttonSmall: {
     fontFamily: "neue-haas-grotesk-text, sans-serif",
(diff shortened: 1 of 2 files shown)


frontends/ol-components/src/components/ThemeProvider/typography.ts
```diff
@@ -141,6 +141,7 @@ const globalSettings: ThemeOptions["typography"] = {
     fontStyle: "normal",
     fontSize: pxToRem(14),
     lineHeight: pxToRem(14),
+    textTransform: "none",
```
ChristopherChudzicki, line 144: In https://github.com/mitodl/mit-open/pull/889/files#diff-2cc4a93727691551d62dc05f2ed43bc01fbe2a7a632861ab97e7b72471408037R59 I added `...typography.button` to our medium button. `typography.button` inherited an extra `text-transform: uppercase` from MUI, which also has a `typography.button`. **That was the only inherited style**. And now it's fixed.

We're using this, rather than `typography.buttonMedium` or something because we want MUI buttons to look good, too. MUI uses their own button for things like tabs and pagination controls.

ChristopherChudzicki (approved):

PR #942, opened 2024-05-21
fix install and storybook

ChristopherChudzicki:
Followup to https://github.com/mitodl/mit-open/pull/936

Fixes the storybook startup command.

### How can this be tested?
1. `rm -rf node_modules`
2. `docker compose up` with `COMPOSE_PROFILES=backend,frontend` in your env file.
3. Check http://localhost:8063/ (frontend) and http://localhost:6006 (storybook). They should be visible.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -98,7 +98,7 @@ services:
     entrypoint: ["/bin/sh", "-c"]
     command:
       - |
-        yarn install
+        yarn install --immutable
         yarn workspace mit-open storybook & yarn workspace mit-open watch:docker
     environment:
       NODE_ENV: ${NODE_ENV:-development}
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -95,8 +95,11 @@ services:
       - frontend
     working_dir: /src
     image: node:20.13
-    command: >
-      /bin/sh -c 'yarn install && yarn storybook & yarn workspace mit-open watch:docker'
+    entrypoint: ["/bin/sh", "-c"]
+    command:
+      - |
+        yarn install
+        yarn workspace mit-open storybook & yarn workspace mit-open watch:docker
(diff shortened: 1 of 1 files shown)


docker-compose.yml
```diff
@@ -95,8 +95,11 @@ services:
       - frontend
     working_dir: /src
     image: node:20.13
-    command: >
-      /bin/sh -c 'yarn install && yarn storybook & yarn workspace mit-open watch:docker'
```
ChristopherChudzicki, line 99: I do not understand why, but it seems like the `yarn storybook & yarn workspace mit-open watch:docker` were running BEFORE install finished, which was throwing an error **if `node_modules` was empty**.

I didn't catch it in https://github.com/mitodl/mit-open/pull/936 because my node_modules wasn't empty.

ChristopherChudzicki: The error was because `/bin/sh` is left-associative with `&&` and `&`.

```sh
/bin/sh -c 'sleep 5 && echo A & echo B'
```
prints B immediately, then does sleep and prints A.

docker-compose.yml
```diff
@@ -95,8 +95,11 @@ services:
...
+    entrypoint: ["/bin/sh", "-c"]
+    command:
+      - |
+        yarn install --immutable
+        yarn workspace mit-open storybook & yarn workspace mit-open watch:docker
```
rhysyngsun, line 102: I'd suggest splitting this to another line too so it's more obvious you're sending storybook to the background as opposed to `&&` - it looked like the latter at a glance having them on one line.
```suggestion
        yarn workspace mit-open storybook &
        yarn workspace mit-open watch:docker
```

rhysyngsun (approved):
LGTM but one suggestion for clarity of what's going on

PR #945, opened 2024-05-21
Homepage News and Events section

jonkafton:

[Homepage News & Events #4254](https://github.com/mitodl/hq/issues/4254)

Adds the "Browse by Topics" section to the homepage:

[Desktop designs](https://www.figma.com/design/Eux3guSenAFVvNHGi1Y9Wm/MIT-Design-System?node-id=2606-57499&m=dev)
[Mobile designs](https://www.figma.com/design/Eux3guSenAFVvNHGi1Y9Wm/MIT-Design-System?node-id=241-4258&m=dev)

Desktop:
image

Mobile:
image

### How can this be tested?

Run the front end:

```
yarn watch:rc
```

- Load the homepage at http://localhost:8062/

- Ensure the "Experience MIT" section renders according to designs for desktop and mobile.

- Ensure images, titles and dates render.

- Ensure links navigate to sources.


frontends/mit-open/src/pages/HomePage/NewsEventsSection.tsx
```diff
@@ -0,0 +1,320 @@
...
+      <StoryPublished>
+        Published: {formatDate(item.news_details?.publish_date)}
+      </StoryPublished>
+    </StoryCard>
+  )
+}
+
+const NewsEventsSection: React.FC = () => {
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  const { data: news }: any = useNewsEventsList({
```
jonkafton, line 241: Seeing an issue with these with `Type 'FeedItem' is not assignable to type 'NewsFeedItem'.`

The return type on the `useNewsEventsList()` query does not pick up on the discriminator https://github.com/mitodl/mit-open/blob/6bbdf4abaa160cd5b16dd530d4792a9aedc244c7/frontends/api/src/generated/v0/api.ts#L618-L620

_Possibly_ due to discrepancy between `feed_type` on the request params and `resource_type` on the `FeedItem`?

gumaerc (changes requested):
Everything design-wise looks correct on both mobile and desktop. I did come across some images not loading, and I'm unsure if it's a data issue or something else:

image

Looking at the source for one of them, it looks like this:

```html
<img class="css-shydlf">
```

So, looking at the JSX and seeing `<StoryImage src={item.image?.url} alt={item.image?.alt} />`, it looks like `image` is missing from the API response in this case. I'm not sure if this is potentially a local data issue with my site, but if there is something I can check let me know.

I also looked into the type issue you are having below and came up with a potential solution:

frontends/mit-open/src/pages/HomePage/NewsEventsSection.tsx
jonkafton: That worked great, thanks @gumaerc. 

The only issue now is that the change to the `PaginatedFeedItemList` interface is in the generated API code. Do you know how we can make that change in the python source?
https://github.com/mitodl/mit-open/blob/6bbdf4abaa160cd5b16dd530d4792a9aedc244c7/frontends/api/src/generated/v0/api.ts#L1225

jonkafton:
> I did come across some images not loading, and I'm unsure if it's a data issue

Yeah, some the new events images are `null` (at the moment, at least): https://mitopen-rc.odl.mit.edu/api/v0/news_events/?feed_type=news

gumaerc (approved):
👍

PR #946, opened 2024-05-21
Topics Listing Page

ChristopherChudzicki:
- Closes https://github.com/mitodl/hq/issues/4186

This PR:
- Adds the topics listing page
- Updates our `<Chip />` component to align better (both styles and terminology) with https://www.figma.com/design/Eux3guSenAFVvNHGi1Y9Wm/MIT-Design-System?node-id=104-1187&m=dev

Screenshot 2024-05-23 at 1 34 15 PM

Screenshot 2024-05-23 at 1 09 55 PM

Screenshot 2024-05-23 at 1 10 16 PM

### How can this be tested?
1. View http://localhost:8063/topics/. You should:
    - see root-level topics, along with child "subtopics", displayed as above.
    - Clicking a topic or subtopic should take you to the corresponding channel page
2. Check that you can navigate to the topics page via:
    - the "By topics" link in upper-left nav drawer (click "Explore MIT")
    - the "See All" link on homepage
3. I also made some changes to our "Chip" component to better align with the [designs](https://www.figma.com/design/Eux3guSenAFVvNHGi1Y9Wm/MIT-Design-System?node-id=104-1187&m=dev). Review these:
    - Visit the storybook pages for Chips http://localhost:6006/?path=/docs/smoot-design-chip-pill--docs ... They should look like designs.
        - If you're running in docker, storybook should start automatically. It crashes sometines, and can be restarted with `yarn storybook`. 
    - View the homepage. There are chips right below the searchbar they should look good.
    - The date chip on course cards should still look similar to before (though course cards are not finalized).
- Check the departments page still looks good. I updated it to use the same Banner component for header that topics page uses. The Figma designs for the two pages use the same component, so they should update in sync.


frontends/mit-open/src/pages/HomePage/BrowseTopicsSection.tsx
```diff
@@ -119,23 +100,20 @@ const BrowseTopicsSection: React.FC = () => {
...
+                  <RiArrowRightLine />
+                </TopicBoxContent>
+              </TopicBox>
+            )
+          })}
         </Topics>
-        <SeeAllButton edge="rounded" size="large">
-          See all (coming soon)
+        <SeeAllButton href="/topics/" edge="rounded" size="large">
+          See all
```
jonkafton, line 116: :tada:

I did notice though that the scroll position is preserved through route navigation, so the Topics Listing page loads half way down when we click this. I'd expect scroll to top to be default behavior, but having had a Google, seems we need to roll our own, e.g. https://dev.to/kunalukey/scroll-to-top-when-route-changes-reactjs-react-router-3bgn. (Also not directly related to this PR)

frontends/mit-open/src/components/RootTopicIcon/RootTopicIcon.tsx
```diff
@@ -0,0 +1,54 @@
+import {
```
jonkafton, line 1: I've added a PlatformIcons component in my local branch. I may combine these into an Icons component, if no objections.

jonkafton (approved):

frontends/mit-open/src/components/RootTopicIcon/RootTopicIcon.tsx
jonkafton: Ignore. They're logos not icons, so should be separate (though perhaps for all logos and also a component for all icons?).

PR #949, opened 2024-05-22
Ab/horizontal channel facets

abeglova:

- [ ] Desktop screenshots


ChristopherChudzicki (commented):
@abeglova I left one large suggestion. Ferdi mentioned you had a bit more work that you wanted to do on this, but that there was a bit of rush to get it out. Does my suggestion line up with what you wanted to do?

Also, if you prefer to do any improvements as a followup PR so that this is shareable via RC, that's good too. Let me know.

PR #951, opened 2024-05-22
side nav updates

gumaerc:
Closes https://github.com/mitodl/mit-open/issues/950

This PR updates the nav drawer to be in line with the current designs as of the time of writing. Spacing and hover states were updated, and icons were added for each item.

image
image

### How can this be tested?
 - Spin up `mit-open` on this branch
 - Visit the home page at http://localhost:8063/
 - Open the nav drawer by clicking the menu button in the header
 - Look at the designs linked in the attached issue and make certain that all of the design is implemented, checking hover states as well


gumaerc:
@steven-hatch @mbilalmughal In the indicated hover styles, the nav icon is shown changing from one color to another. Unfortunately, this won't be possible with SVG icons directly embedded into the project. Libraries like Material Icons and Remix Icons appear to the browser as text with a font, so you can use CSS to set the `color` property, thus changing the color. This isn't possible with an SVG embedded with an `img` tag. There are complex ways to do it with the `filter` rule (https://stackoverflow.com/questions/22252472/how-can-i-change-the-color-of-an-svg-element) but I wouldn't feel great utilizing something like this. Our theme stores colors in hex, and a `filter` rule doesn't take hex and would need to be converted. If it needs to be done, I can add code to the project to generate a filter rule based on a hex color code. It would be best though if we could simply use icons from Remix instead.

gumaerc:
@mbilalmughal I took your advice and used opacity to make the icons lighter on hover, that's a good idea that hadn't crossed my mind.

jonkafton (approved):
Looks great. My only observation is that the svg icons are dark, then lighter on hover in Figma. Can we switch to `opacity: 1` by default and `0.7` on hover. (edit: and seen thread above after commenting!)

PR #955, opened 2024-05-24
Do not ingest prolearn courses/programs from the past

mbertrand:
Closes https://github.com/mitodl/hq/issues/4362

Adjusts the prolearn pipeline to skip over any courses that don't have a start date in the future.

### How can this be tested?
- On the main branch, run `./manage.py backpopulate_prolearn_data`.  Then go to https://mitopen.odl.mit.edu/search?offered_by=see&sortby=start_date - you will see courses with start dates in the past.
- Switch to this branch, restart containers, and repeat the above.  This time the courses with past start dates should not show up.


shanbady (approved):
looks good

PR #957, opened 2024-05-24
Clean up resource descriptions

mbertrand:

Closes https://github.com/mitodl/hq/issues/4355

Strips away any HTML tags in the `description` or `full_description` fields that don't match a list of allowed tags.

### How can this be tested?
- Run `./manage.py backpopulate_prolearn_data`
- Go to http://localhost:8063/search?offered_by=mitpe&q=%22Leadership+and+Innovation%22, click on the 1 result.
- The drawer description should look like this:

Screenshot 2024-05-28 at 3 43 47 PM

and definitely not like this:

Screenshot 2024-05-28 at 3 43 00 PM

### Additional Context
I decided not to include links as allowed HTML tags @Ferdi, let me know if you feel otherwise.

`full_description` is only present (and different from `description`) for edx courses, so I got rid of the code that assigns it the same value as `description` for some of the other sources.

### DIFF
--- a/learning_resources/etl/utils.py
+++ b/learning_resources/etl/utils.py
@@ -17,7 +17,6 @@ from subprocess import check_call
 from tempfile import TemporaryDirectory
 
 import boto3
-import markdown2
 import rapidjson
 import requests
 from django.conf import settings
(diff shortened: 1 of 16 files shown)


learning_resources/etl/utils.py
```diff
@@ -690,3 +693,12 @@ def parse_certification(offeror, runs_data):
             if (availability and availability != AvailabilityType.archived.value)
         ]
     )
+
+
+def clean_data(data: str) -> str:
+    """Remove unwanted html tags from text"""
+    if data:
+        return nh3.clean(
+            data, tags=ALLOWED_HTML_TAGS, attributes=ALLOWED_HTML_ATTRIBUTES
```
jonkafton, line 702: Should we add `clean_content_tags` to remove tag contents where they don't make sense without the tag behavior? 

Thinking e.g.
```
Please see more detail <a href="http://url.com">here</a>
```
is better as `"Please see more detail"` than `"Please see more detail here"`.

Also may authors attempting to add styles?

In the tests the <style> contents are also inside html tags (actually xml CDATA), so are getting removed.

If we had plain styles, e.g.

```
<style>body: {background: blue;}</style>
```

We probably want to remove completely rather than displaying the contents.

jonkafton (approved):

PR #958, opened 2024-05-24
Learning resource drawer design updates

jonkafton:

Closes [Learning Resource Drawer #4312](https://github.com/mitodl/hq/issues/4312)

Implements the design features and resource run selection for various resource types:

**Courses:** 
https://www.figma.com/design/Eux3guSenAFVvNHGi1Y9Wm/MIT-Design-System?node-id=2636-61736&m=dev

**Programs:**
https://www.figma.com/design/Eux3guSenAFVvNHGi1Y9Wm/MIT-Design-System?node-id=2643-66866&m=dev

**Courses / Programs, multiple runs:**
https://www.figma.com/design/Eux3guSenAFVvNHGi1Y9Wm/MIT-Design-System?node-id=2643-69359&m=dev

**Videos:**
https://www.figma.com/design/Eux3guSenAFVvNHGi1Y9Wm/MIT-Design-System?node-id=2643-69514&m=dev

image

-------

image

-------

image

-------

image

### How can this be tested?

Run locally against RC with `yarn watch:rc` and load localhost:8062 or local stack with `docker compose up nginx web watch` and load localhost:8063.

Select learning resources from the "Featured Courses" and "Media" sections on the homepage to open the drawer.

Different resource types have display variations. Verify that courses, programs, podcasts and videos render according to designs.

### Additional Context

- Courses and programs share the default view, with different treatment for podcasts and videos. The remaining resource types learning paths, podcast episodes and video playlists render as default, pending designs.
- Share button is to be added in a subsequent PR.
- "Similar Learning Resources" accordion is to be added in a subsequent PR.


frontends/ol-components/src/components/LearningResourceExpanded/LearningResourceExpanded.tsx
```diff
@@ -0,0 +1,425 @@
...
+          {run.instructors?.length
+            ? run.instructors.map(({ full_name: name }) => name).join(", ")
+            : "-"}
+        </InfoValue>
+      </InfoItem>
+      <InfoItem>
+        <RiGlobalLine />
+        <InfoLabel>Languages:</InfoLabel>
+        <InfoValue>
+          {run.languages?.length ? run.languages.join(", ") : "-"}
```
jonkafton, line 305: These are returned on the API in ISO language / region codes (e.g. `en-us`). Do we have a map? We could use this https://github.com/mozilla/language-mapping-list (though they capitalize the region, `en-US`). Ideally the human readable form would come from the API.

frontends/ol-components/src/components/LearningResourceExpanded/LearningResourceExpanded.tsx
```diff
@@ -0,0 +1,425 @@
...
+  }
+  return (
+    <InfoItems>
+      <Typography variant="subtitle2" component="h3">
+        Info
+      </Typography>
+      <InfoItem>
+        <RiMoneyDollarCircleFill />
+        <InfoLabel>Cost:</InfoLabel>
+        <InfoValue>{run.prices?.[0] || "-"}</InfoValue>
```
jonkafton, line 285: Prices are returned on the API as an array, e.g. `["0.00", "300.00", "149.00", "300.00"]`. Which to display? Are they always USD? Should we display `"0.00"` as `Free`?

Also, prices are on both the learning resource and child runs. On the parent resource they are numbers (e.g. `[0, 75, 300, 49, 50, 149]`), though strings on the runs `["0.00", "300.00", "149.00", "300.00"]`.

frontends/ol-components/src/components/LearningResourceExpanded/LearningResourceExpanded.tsx
```diff
@@ -0,0 +1,425 @@
...
+        <Skeleton variant="text" width="100%" />
+        <Skeleton variant="text" width="100%" />
+        <Skeleton variant="text" width="100%" />
+        <Skeleton variant="text" width="100%" />
+        <Skeleton variant="text" width="100%" />
+        <Skeleton variant="text" width="33%" />
+      </>
+    )
+  }
+  return <Description>{resource.description}</Description>
```
jonkafton, line 238: Some descriptions contain HTML - do we need to display as HTML. Not a good idea of course due to the XSS vulnerability, so ideally sanitized on the backend. Others (most so far) are plain text. We are handling `\n` newlines with CSS, but not rendering HTML in this PR.

**Example:** 
https://mitopen-rc.odl.mit.edu/api/v1/learning_resources/14387/

Renders as:
image

frontends/ol-components/src/components/LearningResourceExpanded/LearningResourceExpanded.test.tsx
```diff
@@ -0,0 +1,143 @@
...
+    const resource = factories.learningResources.resource({
+      resource_type: resourceType,
+    })
+
+    setup(resource)
+
+    const section = screen
+      .getByRole("heading", { name: "Topics" })!
+      // eslint-disable-next-line testing-library/no-node-access
+      .closest("section")!
```
jonkafton, line 93: Is there a better alternative here for finding a heading, then targeting the parent section for tests against its contents?

frontends/ol-components/src/components/LearningResourceExpanded/LearningResourceExpanded.tsx
ChristopherChudzicki: Matt is working on cleaning these up. See 
- https://github.com/mitodl/hq/issues/4355

The plan is the render the raw HTML—`dangerouslySetInnerHtml={{ __html: description }}` should be fine since. the HTML is being sanitized on the backend.

frontends/ol-components/src/components/Logo/Logo.tsx
```diff
@@ -0,0 +1,112 @@
...
+export const PLATFORMS = [
+  {
+    code: PlatformEnum.Ocw,
+    name: "OCW",
+    image: "mit-ocw-logo-square.png",
+  },
+  {
+    code: PlatformEnum.Edx,
+    name: "edX",
+    image: "edx_logo.png",
```
jonkafton, line 13: These are using existing image in the codebase where available, though some are low quality and don't lend themselves to the design format for the learning resource drawer.

Do we have a set of these with a fixed aspect ratio, or could we put together? 2:1 to 3:1 seems about right. @steven-hatch


image

image

image

Ferdi:
-  OCW Courses should be `cost : Free`.  (do we use 'cost' or 'price' @steven-hatch ?)
-  When one of the attributes is missing value (like language), instead of displaying a missing value we should not display that attribute at all

General note, probably not for this PR --  The theory here is that all the common metadata for all learning resources are displayed in the card.  In the drawer, we display all metadata. For example,  I know sloan / MIT PE have course metadata like "duration" and "Time commitment" . We should display those (@mbertrand do we have those around )

mbertrand:
@Ferdi no we do not have extra source-specific data like duration or time commitment.  Not saved in the database, not represented in the serialized api data.

frontends/ol-components/src/components/LearningResourceExpanded/LearningResourceExpanded.tsx
ChristopherChudzicki: https://www.npmjs.com/package/iso-639-1 is already installed in the repo. I would use that.

frontends/ol-components/src/components/LearningResourceExpanded/LearningResourceExpanded.test.tsx
ChristopherChudzicki: I think this is the best way. Playwright does have nicer locators for this sort of thing.

We can globally disable ` eslint-disable-next-line testing-library/no-node-access`. I think it's been annoying.

frontends/ol-components/src/components/Logo/Logo.tsx
ChristopherChudzicki: Incidentally, I think that's the old edX logo. I believe they use 

edx logo

now

frontends/ol-components/src/components/Logo/Logo.tsx
```diff
@@ -0,0 +1,112 @@
+import React from "react"
+import { PlatformEnum } from "api"
+
+export const PLATFORMS = [
```
ChristopherChudzicki, line 4: Any reason for this to be an array? Object might be more convenient for lookups, and if we do
```ts
const PLATFORMS: Record<PlatformEnum, PlatformObject> = {

}
```
we'll get TS warnings if we miss platforms.

frontends/ol-components/src/components/Logo/Logo.tsx
steven-hatch: 2:1 to 3:1 seems right. @mbilalmughal feel free to weigh in but the idea is to get the height consistent between all logos. This height should be the same as the button found on same line, to the right. If a logo occupies the height and runs into other elements then we would need to reduce width. I dont see this happening, but OpenCourseWare would be the one to give us trouble because its so wide. 

Also, please let me know if you need clean SVGs for logos. We have access to MIT Brand files for logos.

jonkafton:
> My vote is to keep this simple and go with 'Price'.

Updates to "Price"

frontends/ol-components/src/components/LearningResourceExpanded/LearningResourceExpanded.tsx
```diff
@@ -0,0 +1,492 @@
...
+    )
+  }
+
+  const isVideo = resource && resource.resource_type === ResourceTypeEnum.Video
+
+  return (
+    <Container padTop={isVideo}>
+      <DateSection hide={isVideo} />
+      <ImageSection resource={resource} config={imgConfig} />
+      <CallToActionSection resource={resource} />
```
ChristopherChudzicki, line 480: We may need to get the URL from the runs... Both the resource itself and runs (for courses + programs) have a url. 

However, on my local data, run URLs are either null or always the same as the resource URL. Will check RC/prod after I get off this bus.

cc @mbertrand

frontends/ol-components/src/components/RoutedDrawer/RoutedDrawer.tsx
```diff
@@ -1,18 +1,25 @@
...
 const closeSx: React.CSSProperties = {
   position: "absolute",
-  top: "0px",
-  right: "0px",
+  top: "16px",
+  right: "22px",
 }
 
+const CloseIcon = styled(RiCloseLargeLine)`
+  &&& {
```
ChristopherChudzicki, line 17: Does &&& increase the precedence? Do we need it?

frontends/ol-components/src/components/LearningResourceExpanded/LearningResourceExpanded.stories.tsx
ChristopherChudzicki, line 1: Do you mind adding a story for the other resource types? At least Podcast episode, podcast

ChristopherChudzicki (changes requested):
This is looking nice. Some minor comments. Request is to not show info items when there is no value.

frontends/ol-components/src/components/RoutedDrawer/RoutedDrawer.tsx
jonkafton: It does and we do. We have the `IconContainer` emotion classname and `& .MuiSvgIcon-root` in (Button.tsx#L127](https://github.com/mitodl/mit-open/blob/612c4fad7e419102031b3ab3dbc549093f0f5a8b/frontends/ol-components/src/components/Button/Button.tsx#L127)

Produces e.g. `.css-1qy38u1 .MuiSvgIcon-root`.

`&&&` Gives is the precedence e.g. `.css-r92gtm.css-r92gtm.css-r92gtm`. We actually only need 2 here `&&` but a bit safer and unintended change easy to miss. We may want all close icons to be this size though?

frontends/ol-components/src/components/LearningResourceExpanded/LearningResourceExpanded.tsx
```diff
@@ -0,0 +1,492 @@
...
+    return null
+  }
+
+  const price = platformCode === PlatformEnum.Ocw ? "Free" : run.prices?.[0]
+  return (
+    <InfoItems>
+      <Typography variant="subtitle2" component="h3">
+        Info
+      </Typography>
+      <InfoItem>
```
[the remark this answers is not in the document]
jonkafton: I've now removed the labels where no value (and the whole section where no values).

frontends/ol-components/src/components/LearningResourceExpanded/LearningResourceExpanded.tsx
```diff
@@ -0,0 +1,524 @@
...
+
+  if (
+    [displayPrice, level, instructors, languages].every(
+      (element) => element === null,
+    )
+  ) {
+    return null
+  }
+
+  return (
```
ChristopherChudzicki, line 409: I'm noticing that Programs and Videos seem to be missing their metadata (e.g., duration). It's not handled in the InfoSection component.

Screenshot 2024-05-30 at 11 19 01 AM

Additionally, I think it would be really **good if we could add more metadata without modifying the component code** . I'm a little worried this component will become a big heap of if / else blocks.

IMO:
- what metadata gets shown should come from a config object
- as we add more metadata, we update the config object
- the component itself should be pretty simple

I'm thinking something like this:
```tsx
type InfoSectionProps = {
  resource: LearningResource
  run?: LearningResourceRun
}
const InfoSection = ({ resource, run }: InfoSectionProps) => {
  const infoItems = INFO_MAP[resource.resource_type]
    .map(({ label, value }) => ({ label, value: value(resource, run) }))
    .filter((x): x is { label: string; value: string } => !!x.value)
  
  if (infoItems.length === 0) return
  return (
    // render the items
  )
```

and a config object:
```ts

type InfoExtractor<T extends ResourceTypeEnum = ResourceTypeEnum> = (
  resource: LearningResource & { resource_type: T },
  run?: T extends
    | typeof ResourceTypeEnum.Course
    | typeof ResourceTypeEnum.Program
    ? LearningResourceRun
    : undefined,
) => string | null | undefined
type InfoMap = Record<
  ResourceTypeEnum,
  { label: string; value: InfoExtractor }[]
>

const extractPrice: InfoExtractor = (resource, run) =>
  resource.platform?.code === PlatformEnum.Ocw
    ? "Free"
    : run?.prices?.[0] && `$${run?.prices?.[0]}`
const extractLevel: InfoExtractor = (resource, run) => run?.level?.[0]?.name
const extractInstructors: InfoExtractor = (resource, run) =>
  run?.instructors
    ?.filter((instructor) => instructor.full_name)
    .map(({ full_name: name }) => name)
    .join(", ")
const extractDuration = (resource: LearningResource) => {
  if (resource.resource_type === ResourceTypeEnum.Video) {
    return resource.video.duration
  }
  if (resource.resource_type === ResourceTypeEnum.PodcastEpisode) {
    return resource.podcast_episode.duration
  }
  return null
}

const INFO_MAP: InfoMap = {
  [ResourceTypeEnum.Course]: [
    { label: "Price", value: extractPrice },
    { label: "Level", value: extractLevel },
    { label: "Instructors", value: extractInstructors },
  ],
  [ResourceTypeEnum.Program]: [
    { label: "Price", value: extractPrice },
    { label: "Level", value: extractLevel },
    { label: "Instructors", value: extractInstructors },
  ],
  [ResourceTypeEnum.Video]: [{ label: "Duration", value: extractDuration }],
  [ResourceTypeEnum.PodcastEpisode]: [
    { label: "Duration", value: extractDuration },
  ],
  [ResourceTypeEnum.Podcast]: [],
  [ResourceTypeEnum.VideoPlaylist]: [],
  [ResourceTypeEnum.LearningPath]: [],
}
```

ChristopherChudzicki (changes requested):
I think we're missing the metadata about videos. I left a suggestion for organizing that.

frontends/ol-components/src/components/LearningResourceExpanded/LearningResourceExpanded.tsx
ChristopherChudzicki: @jonkafton Following up with what I said in Slack. Per Ferdi's comment:

> General note, probably not for this PR -- The theory here is that all the common metadata for all learning resources are displayed in the card. **In the drawer, we display all metadata**.

Might be better to do the config as an array. (I do still think it should be a config object, not handled in the component) so that if a resource has a price, we always show the price, if it has a certificate, we always show that info. If it's a podcast with a certificate...that's weird. Probably a data issue that should be addressed on the backend.

jonkafton: That's pushed, following your suggestion for item config above @ChristopherChudzicki. We have an array of item config (label, icon and value selector function). All items are tried across learning resource types and will be displayed if they return a value. We can add discriminator logic for resource types (and anything else) inside the selector functions.

Adds duration, "offered by" and "number of courses". "Date Posted" is missing though - I couldn't match that on anything in the VideoResource schema.

ChristopherChudzicki (changes requested):
Ack, didn't catch this before. Topics section is showing even if no topics. (Might be inherited issue from old version of the drawer.) Sorry didn't catch that earlier.

Other than that, all looks good.
Screenshot 2024-05-30 at 7 59 34 PM

ChristopherChudzicki:
@jonkafton Thanks for changing Subjects to Topics 😁

frontends/ol-components/src/components/LearningResourceExpanded/LearningResourceExpanded.tsx
```diff
@@ -0,0 +1,396 @@
...
+        <Skeleton variant="text" width="100%" />
+        <Skeleton variant="text" width="100%" />
+        <Skeleton variant="text" width="100%" />
+        <Skeleton variant="text" width="33%" />
+      </>
+    )
+  }
+  return (
+    <Description
+      dangerouslySetInnerHTML={{ __html: resource.description || "" }}
```
ChristopherChudzicki, line 264:
```suggestion
      /**
       * Resource descriptions can contain HTML. They are santiized on the
       * backend during ETL. This is safe to render.
       */
      dangerouslySetInnerHTML={{ __html: resource.description || "" }}
```

ChristopherChudzicki (approved):
👍

PR #960, opened 2024-05-24
Upgrade NukaCarousel to v8

jkachel:

n/a

Upgrades NukaCarousel from v7 to v8. 

v8 deprecates a number of props, reworks some method signatures, and moves some configuration (including for page size) to CSS; this updates the TabbedCarousel and Carousel components to handle those changes. It should generally work the same as it did before but there may (will) be some minor styling differences.

### How can this be tested?

Automated tests should pass.

The carousels on the homepage should work as expected; clicking the left and right buttons should cycle through the pages and the buttons should automatically deactivate when at the start or end of the list.

Note that there are some mocks that were moved so they apply globally - specifically, mocks for `matchMedia` and `ResizeObserver`, since NukaCarousel uses these now and thus anything that's got a carousel in it will need them to have their tests pass. The mock for `ResizeObserver` came from the CKEditor component, and that was lifted up.


jkachel:
Python tests are failing but that seems to be an upstream thing - this code is just frontend changes.

ChristopherChudzicki (approved):

PR #967, opened 2024-05-28
Fix detect-secrets baseline file

rhysyngsun:

N/A

This updates the `detect-secrets` baseline file by running `detect-secrets scan --baseline .secrets.baseline`. There are apparently some files that are putting this checker into a weird state where it will detect these changes, fail, but not actually update the baseline file.

### How can this be tested?

- Go to the `main` branch and verify when you run `pre-commit run detect-secrets -a` that you see it fail with error output but no change to `.secrets.baseline`
- Checkout this branch and verify that  `pre-commit run detect-secrets -a` succeeds with no error output and no file changes.

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -1,5 +1,5 @@
 {
-  "version": "1.5.0",
+  "version": "1.4.0",
   "plugins_used": [
     {
       "name": "ArtifactoryDetector"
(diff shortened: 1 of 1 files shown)


shanbady (changes requested):
@rhysyngsun I'm not seeing errors when running the precommit check on main (or on your branch)

ChristopherChudzicki:
> @rhysyngsun I'm not seeing errors when running the precommit check on main (or on your branch)

That said, the files corresponding to the entries that @rhysyngsun removed from secrets baseline don't exist, so it's good to have then gone.

PR #971, opened 2024-05-29
Randomize featured api order by offeror, keep sorting by position

mbertrand:
Related to https://github.com/mitodl/hq/issues/4183

Keeps sorting featured resources by their positions in their respective offeror featured lists, but randomizes the order of offerors within each position.

### How can this be tested?
- Run the following (you can probably skip the backpopulate ones if you've run them before):
```
./manage.py update_offered_by
./manage.py backpopulate_ocw_data. --skip-contentfiles
./manage.py populate_featured_lists
```

Go to  http://localhost:8063/, note the content of the featured list carousel, and then refresh repeatedly.  The first 6 resources should always be the same, but their order should differ.  Ditto for the next set of 6, etc.

### DIFF
--- a/learning_resources/views.py
+++ b/learning_resources/views.py
@@ -2,6 +2,7 @@
 
 import logging
 from hmac import compare_digest
+from random import shuffle
 
 import rapidjson
 from django.conf import settings
@@ -810,9 +811,6 @@ class VideoPlaylistViewSet(BaseLearningResourceViewSet):
 
 
 @extend_schema_view(
-    list=extend_schema(
-        description="Get a paginated list of featured resources",
-    ),
     retrieve=extend_schema(
         description="Retrieve a single featured resource",
     ),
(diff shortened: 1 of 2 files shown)


learning_resources/views.py
```diff
@@ -845,6 +845,6 @@ def get_queryset(self) -> QuerySet:
             .filter(parents__parent_id__in=featured_list_ids)
             .filter(published=True)
             .annotate(position=F("parents__position"))
-            .order_by("position")
+            .order_by("position", "?")
```
ChristopherChudzicki, line 848: A potential issue is that using `?` with `order_by` interferes with pagination: `/api/v1/featured/?limit=10&offset=0` and `/api/v1/featured/?limit=10&offset=10` can now overlap. The frontend is currently only requesting a single page of results, so maybe this isn't a realistic issue.

Could the randomizing happen in Python after fetching from db?

ChristopherChudzicki (changes requested):
This is working for the carousels, but I do think there's an issue with pagination.

IMO, it would be nice for pagination to work. If that ends up being a pain, we could do the randomization on frontend.

learning_resources/views.py
mbertrand: Refactored it to do the randomizing in python

ChristopherChudzicki (approved):
👍

PR #980, opened 2024-05-30
search pages code sharing

abeglova:

None

This reduces repeated code between the search page and field page.

### How can this be tested?
Go to http://localhost:8063/search
Go to http://localhost:8063/c/offeror/xpro

Verify that both pages look the same


frontends/mit-open/src/pages/SearchPage/SearchPage.tsx
```diff
@@ -417,95 +164,19 @@ const SearchPage: React.FC = () => {
...
-                    page={page}
-                    onChange={(_, newPage) => setPage(newPage)}
-                  />
-                </PaginationContainer>
-              </ResourceTypeTabs.TabPanels>
-            </GridColumn>
-          </ResourceTypeTabs.Context>
-        </GridContainer>
-      </Container>
+      {SearchDisplay({
```
ChristopherChudzicki, line 167: **Requested Change:** Here and in the FieldPage, render this as JSX:

```
<SearchDisplay
        searchParams={searchParams}
/>
```

etc. 

I don't have a really deep understanding of this, but basically I think JSX gets transpiled into [createElement](createElement), whereas simply calling the function doesn't. And that can lead to some unexpected behaviors.

See for example [Don't call a React function component](https://kentcdodds.com/blog/dont-call-a-react-function-component) for some of the consequences. Near the end of the article the author says:

> Oh, and it's notable to mention that sometimes it will "work" to call function components. Like so ... o it will "work," but not the way you'd expect and it could behave in unexpected ways as you make changes. So just render it normally.

I think this is one of those cases. I **believe** a concrete consequence would be seen if you put a `console.log` int he SearchPage component so that you can see how many times it would be rendered.
- with  jsx`< SearchDisplay />` any hooks that use state (react query hooks do internally) are registered to the that component instance. When state changes, ONLY that component re-renders
- with a function call `SearchDisplay({...})` there is no component instance, so hooks are registered with the parent. When state changes, `SearchPage` would re-render.
    - This particular consequence isn't really a big deal—SearchDisplay is almost all of SearchPage, so re-rendering one is almost the same as re-rendering the other. But it's just an example of the unexpected effects.

ChristopherChudzicki (changes requested):
Left one simple requested change. Otherwise 👍

ChristopherChudzicki (approved):

PR #985, opened 2024-05-31
Fix celerybeat schedule

mbertrand:
Closes https://github.com/mitodl/hq/issues/4412

FIxes one of the new scheduled celerybeat tasks

### How can this be tested?
- Temporarily change the last scheduled task in the `CELERY_BEAT_SCHEDULE` of main/settings_celery.py to a schedule of every 2 minutes (`"schedule": 120).
- Start containers.
- Within a couple of minutes, you should see `INFO/ForkPoolWorker-8] Sending daily subscription emails for channel_subscription_type` in the logs.  The task may fail at first if opensearch isn't done starting up.

### DIFF
--- a/learning_resources_search/tasks.py
+++ b/learning_resources_search/tasks.py
@@ -199,6 +199,7 @@ def send_subscription_emails(self, subscription_type, period="daily"):
     """
     Send subscription emails by percolating matched documents
     """
+    log.info("Sending %s subscription emails for %s", period, subscription_type)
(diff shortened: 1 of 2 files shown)


shanbady (approved):
This must have been the crux of the issue I was seeing. it works now.

shanbady:
I dont think it will be an issue when deploying but one thing i noticed locally is a flood of events when i restarted docker with this fix in place (presumably a backlog of failed tasks)

PR #978, opened 2024-05-30
remove package-lock.json

ChristopherChudzicki:
### What are the relevant tickets?
This should fix failing heroku build https://dashboard.heroku.com/apps/mitopen-rc/activity/builds/c4a66b24-d88e-4aed-9a89-ff244ec36607

### Description (What does it do?)
This PR removes `package-lock.json`. That lockfile is for installing node packages via `npm`. We install via `yarn`.

It was added in https://github.com/mitodl/mit-open/commit/f421964cae1457e3f6e095a6f26385069390ccde ... That renovate commit was coauthored, so maybe we added it accidentally.

### How can this be tested?
CI should pass.

### Additional Context

PR #986, opened 2024-05-31
Error if using npm to install

ChristopherChudzicki:
Followup to #978 

This PR makes an error if you run `npm install` instead of `yarn install`.  

### How can this be tested?
1. Run `docker compose exec watch bash`.
2. Try running `docker compose exec watch npm install`. It should error with a reasonably helpful error message.
3. Try running `docker compose exec watch npm yarn install`. It should work.

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -1,0 +1,1 @@
+engine-strict = true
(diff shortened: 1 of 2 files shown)


gumaerc (approved):
👍

PR #992, opened 2024-05-31
Insert learning_path_parents/user_list_parents values into search results

mbertrand:
Closes https://github.com/mitodl/hq/issues/4439

If the user is authenticated, inserts values for `learning_path_parents` and `user_list_parents` into search results

### How can this be tested?
- As an admin user, create a learning path and a user list and add some resources to each (click on the buttons shown below on each resource) based on a particular search filter like http://localhost:8063/search/?offered_by=ocw&topic=Engineering
- Refresh the search url above and check the raw response for the search results.  Both the `user_list_parents` and `learning_path_parents` fields should have a list of serialized relationships (id,parent,child) based on your selections above.
- As a non-admin user, create a user list and add some resources to it based on the same search results.
- Refresh the search url above and check the raw response for the search results.  The `user_list_parents` field should have a list of serialized relationships (id,parent,child) based on your selections above.  The `learning_path_parents` field should be an empty list.
- Log out and refresh the same search url.   Both the `user_list_parents` and `learning_path_parents` fields should be empty lists for all results.

### DIFF
--- a/learning_resources_search/serializers.py
+++ b/learning_resources_search/serializers.py
@@ -13,19 +13,27 @@ from rest_framework.utils.urls import replace_query_param
 
 from learning_resources.constants import (
     DEPARTMENTS,
+    GROUP_STAFF_LISTS_EDITORS,
     LEARNING_RESOURCE_SORTBY_OPTIONS,
     CertificationType,
     LearningResourceFormat,
+    LearningResourceRelationTypes,
     LearningResourceType,
     LevelType,
     OfferedBy,
     PlatformType,
 )
-from learning_resources.models import LearningResource
+from learning_resources.models import (
+    LearningResource,
+    LearningResourceRelationship,
+    UserListRelationship,
+)
 from learning_resources.serializers import (
     ContentFileSerializer,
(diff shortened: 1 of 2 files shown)


learning_resources_search/serializers.py
```diff
@@ -451,8 +459,58 @@ def get_previous(self, instance) -> str | None:
...
+                or request.user.is_superuser
+                or request.user.groups.filter(name=GROUP_STAFF_LISTS_EDITORS).first()
+                is not None
+            ):
+                # Get learning path parents for all returned resources
+                learning_path_parents = LearningResourceRelationship.objects.filter(
+                    child__id__in=hit_ids,
+                    relation_type=LearningResourceRelationTypes.LEARNING_PATH_ITEMS.value,
+                ).values("id", "child_id", "parent_id")
+                for parent in learning_path_parents:
```
shanbady, line 491: this method is a bit lengthy. non blocker but i would consider breaking this up into seperate methods.

shanbady (approved):
looks good. left a minor comment to consider

PR #1001, opened 2024-06-03
Adds learner testimonials component for interior pages

jkachel:

Closes mitodl/hq#4457

This adds a component that can be included on any page to display testimonials, and then adds that to the `FieldPage` component so it should display on the field search pages (which includes offeror pages, I think).

Screenshots are of the component on the homepage since my local search is broken. 

Desktop:

image

Mobile:

image

### How can this be tested?

You will need to configure your app with testimonial data. This can be anything; you just need records in there.

You should see the testimonial card on the offeror pages - for example, http://od.odl.local:8063/c/offeror/ocw/ . You should only see testimonials that are attached to that particular offeror. You should be able to paginate through them, and the design should match the design in Figma. 

You should also see the testimonial card on the search pages as well; there, it should show all the testimonials. 

### Additional Context

- I added a bit of margin on top of the pager buttons - this isn't in the Figma but they were sort of getting stuck too close to the name block.
- You can also test this by adding the testimonial card to the homepage (which is what I did for testing, since my search is broken). Look in `mit-open/src/pages/FieldPage/FieldPage.tsx` to see how but you really just have to drop the component in anywhere in there.

### DIFF
--- a/frontends/mit-open/src/page-components/TestimonialDisplay/TestimonialDisplay.tsx
+++ b/frontends/mit-open/src/page-components/TestimonialDisplay/TestimonialDisplay.tsx
@@ -1,0 +1,197 @@
+import React from "react"
+
+import { ActionButton, styled, theme } from "ol-components"
+import { useTestimonialList } from "api/hooks/testimonials"
+import { Attestation } from "api/v0"
+import { RiArrowRightLine, RiArrowLeftLine } from "@remixicon/react"
+import Slider from "react-slick"
+
+type TestimonialDisplayProps = {
+  channels?: number[]
+  offerors?: string[]
+}
+
+type InternalTestimonialDisplayProps = {
+  attestation: Attestation
+}
+
+const QuoteContainer = styled.section(({ theme }) => ({
+  backgroundColor: theme.custom.colors.darkGray2,
+  color: theme.custom.colors.white,
+  overflow: "auto",
+  padding: "16px 0 24px 0",
+}))
+
+const QuoteBlock = styled.div(() => ({
+  justifyContent: "center",
+  alignItems: "center",
+  width: "100%",
+  maxWidth: "1328px",
+  margin: "0 auto",
+  padding: "0 22px",
+}))
+
+const QuoteLeader = styled.div(({ theme }) => ({
+  width: "100%",
+  height: "32px",
+  fontSize: "60px",
+  lineHeight: "60px",
+  color: theme.custom.colors.silverGray,
+}))
+
+const QuoteBody = styled.div(({ theme }) => ({
+  width: "100%",
+  display: "flex",
+  [theme.breakpoints.down("md")]: {
+    flexDirection: "column",
+  },
+}))
+
+const AttestationBlock = styled.div(({ theme }) => ({
+  width: "auto",
+  flexGrow: "5",
+  ...theme.typography.h5,
+  justifyContent: "top",
+}))
+
+const AttestantBlock = styled.div(({ theme }) => ({
+  width: "248px",
+  display: "flex",
+  [theme.breakpoints.down("md")]: {
+    marginTop: "24px",
+  },
+}))
+
+const AttestantAvatar = styled.div({
+  marginRight: "12px",
+  img: {
+    objectFit: "cover",
+    borderRadius: "50%",
+    width: "40px",
+    height: "40px",
+    boxShadow:
+      "0px 2px 4px 0px rgba(37, 38, 43, 0.10), 0px 2px 4px 0px rgba(37, 38, 43, 0.10)",
+  },
+})
+
+const AttestantNameBlock = styled.div(({ theme }) => ({
+  flexGrow: "1",
+  width: "auto",
+  color: theme.custom.colors.lightGray2,
+}))
+
+const AttestantName = styled.div(({ theme }) => ({
+  ...theme.typography.subtitle1,
+  whiteSpace: "nowrap",
+  color: theme.custom.colors.white,
+  lineHeight: "125%",
+}))
+
+const ButtonsContainer = styled.div(({ theme }) => ({
+  display: "flex",
+  justifyContent: "right",
+  margin: "4px auto 0",
+  gap: "16px",
+  [theme.breakpoints.down("md")]: {
+    marginTop: "16px",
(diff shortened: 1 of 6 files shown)


frontends/mit-open/src/page-components/TestimonialCard/TestimonialCard.tsx
```diff
@@ -0,0 +1,184 @@
...
+}))
+
+const InteriorTestimonialCard: React.FC<InternalTestimonialCardProps> = ({
+  resource,
+}) => {
+  return (
+    <QuoteBody>
+      <AttestationBlock>
+        {resource?.quote ??
+          "You have three powerful forces coming together again, the HBCUs, and MIT, Merlot. The need is worldwide for people of all races, underrepresented populations to be able to access information, and to tap in to rich cultural resources to help them in their academic journey."}
```
ChristopherChudzicki, line 107: `quote`, `attestant_name`, `title` are all required, right? Let's drop the fallbacks.

frontends/mit-open/src/pages/FieldPage/FieldPage.tsx
```diff
@@ -32,6 +33,9 @@ const FieldPage: React.FC = () => {
     channelType && (
       <FieldPageSkeleton name={name} channelType={channelType}>
         <p>{fieldQuery.data?.public_description}</p>
+        <TestimonialCard
+          offerors={channelType === "offeror" ? [name] : undefined}
+        />
```
ChristopherChudzicki, line 38: I believe that only offeror ("unit") pages should have Testimonial. Other channel pages, like "Departments" do not.

Figma for Department Detail pages:
Screenshot 2024-06-11 at 1 19 20 PM

But this is showing all testimonials for department pages (and topic pages).

Screenshot 2024-06-11 at 1 20 10 PM

frontends/mit-open/src/page-components/TestimonialCard/TestimonialCard.tsx
```diff
@@ -0,0 +1,184 @@
...
+    },
+  ]
+
+  if (!data) return null
+
+  return (
+    <QuoteContainer>
+      <QuoteBlock>
+        <QuoteLeader>“</QuoteLeader>
+        <Slider
```
ChristopherChudzicki, line 148: When there is only 1 testimonial, I don't think we should use a slider. We don't want buttons, and it seems to cause problems (more on this in a sec).

So I suggest: `{data.results.length === 1 ? : < InteriorTestimonialCard /> : <Slider ... />}`


### Carousel misbehaves with a single slide

When there's only one testimonial, I get:

Screenshot 2024-06-11 at 11 43 35 AM

Crazy. Works great for 2+ slides.

frontends/mit-open/src/page-components/TestimonialCard/TestimonialCard.tsx
```diff
@@ -0,0 +1,184 @@
...
+  const responsive = [
+    {
+      breakpoint: theme.breakpoints.values["md"],
+      settings: {
+        adaptiveHeight: true,
+      },
+    },
+  ]
+
+  if (!data) return null
```
ChristopherChudzicki, line 142: Let's also bail early if no testimonials.

```suggestion
  if (!data) return null
  if (data.results.length === 0) return null
```

frontends/mit-open/src/page-components/TestimonialCard/TestimonialCard.tsx
ChristopherChudzicki, line 1: Small comment... To me, the testimonial display does not seem like "Cards". Cards are usually narrow and displayed with many visible at once.

I'd call this TestimonialDisplay or something.

Or TestimonialCarousel (even though I suggest we not use a carousel if there's only 1).

ChristopherChudzicki (changes requested):
The testimonials look great. I did notice some issues with their display when there's only a single testimonial, and it seems like we are displaying it more places than we should.

jkachel:
These should all be resolved now.

frontends/mit-open/src/pages/FieldPage/FieldPage.tsx
```diff
@@ -32,6 +33,9 @@ const FieldPage: React.FC = () => {
     channelType && (
       <FieldPageSkeleton name={name} channelType={channelType}>
         <p>{fieldQuery.data?.public_description}</p>
+        {channelType === "offeror" ? (
+          <TestimonialDisplay offerors={[name]} />
+        ) : null}
```
ChristopherChudzicki, line 38: Could you rebase? (May be easier to squash and then rebase).

This should now be `channelType === "unit"`.

ChristopherChudzicki (changes requested):

frontends/mit-open/src/page-components/TestimonialDisplay/TestimonialDisplay.tsx
```diff
@@ -0,0 +1,197 @@
+import React from "react"
+
+import { ActionButton, styled, theme } from "ol-components"
+import { useTestimonialList } from "api/hooks/testimonials"
+import { Attestation } from "api/v0"
```
ChristopherChudzicki, line 5:
```suggestion
import type { Attestation } from "api/v0"
```
IMO `type` imports are nice when you only need the type. The are 100% guaranteed to have no impact at runtime (whereas `import { Attestation } from "api/v0"` can have an impact even if you're only importing a type, if `api/v0` has side effects.)

Also makes intent clear.

ChristopherChudzicki (approved):
If you have time, I think it would be great to add a few tests to this. Some things I would test:
1. `TestimonialDisplay.test.tsx`
    - displays nothing if no testimonials
    - displays data if 1
    - displays carousel if many
2. `FieldPage.test.tsx`
    - Displays a carousel if channel type is unit
    - otherwise does not. 
        - "Otherwise does not" can be tricky. Could give the testimonials a `data-testid`. Or could spy on `TestimonialDisplay` and assert not called.

jkachel:
Merging this now because there's some stuff that needs to change a bit (the attestant stuff needs to be pulled into its own component for the homepage) - will add tests while I do that. (Also, we changed "offeror" to "unit" so need to do that too.)

PR #1003, opened 2024-06-03
Update profile fields to align to LR data

rhysyngsun:

Closes https://github.com/mitodl/hq/issues/4382

- Switches `Profile.interests` (renamed to `Profile.topic_interests`) to be a relation to `LearningResourceTopic`.
- Switches `Profile.course_format` (renamed to `Profile.learning_format`) to use the constants `LearningResourceLearningFormat`.

### How can this be tested?

Go to `/api/v0/profiles/USERNAME/` and verify you can patch your profile for the updated fields.

For example, assuming `1` is an id for a top level topic:
```json
{
    "topic_interests": [1]
}
```


mbertrand (changes requested):
Works great, just noticed one small issue with the profile form in django admin.

rhysyngsun:
@mbertrand Addressed that issue

mbertrand (approved):
👍

PR #1006, opened 2024-06-04
Add nullalbe offerors and channels to the testimonials model/API

jkachel:

https://github.com/mitodl/hq/issues/4457

After some discussion with Steve and Bilal, we'll need the testimonial attestations to be attached to offerors* in the system, so we can filter them for display on the offeror pages. This adds the field and updates the filter to allow for that.

In addition, this sets the channel and offeror fields to be blankable. (Channel definitely needed to be nullable - offeror maybe not so much but we can change that if it makes sense to.)

* I am using offeror here - we're going to change this to unit at some point but it's called "offeror" at the time of writing this PR so sticking with that.

### How can this be tested?

Automated tests should pass.

You should now see an Offerors field in Django Admin, and you should be able to set it. You should also be able to leave it blank.
You should now be able to leave the Channels field blank in Django Admin.
When searching, you should be able to filter by Offeror. (This is perhaps easiest done through the DRF interface but you can do it through either Swagger or Redoc too.)
You should see the additional options in Swagger, Redoc, and OpenAPI.


shanbady (approved):
looks good

PR #1012, opened 2024-06-05
Style tab components to match figma

ChristopherChudzicki:
Adds a component for use in
- https://github.com/mitodl/hq/issues/4456
- https://github.com/mitodl/hq/issues/4360

Adds a version of TabList that uses our tertiary button.

Screenshot 2024-06-05 at 11 20 12 AM

### How can this be tested?
1. View http://localhost:6006/?path=/docs/smoot-design-tabs--docs and the associated stories. Try thing slike:
    - changing tabs
    - navigating with keyboard
    - shrinking screen.

### DIFF
--- a/frontends/ol-components/src/components/TabButtons/TabButtonList.tsx
+++ b/frontends/ol-components/src/components/TabButtons/TabButtonList.tsx
@@ -1,0 +1,68 @@
+import React from "react"
+import MuiTab from "@mui/material/Tab"
+import type { TabProps } from "@mui/material/Tab"
+import MuiTabList from "@mui/lab/TabList"
+import type { TabListProps } from "@mui/lab/TabList"
+import styled from "@emotion/styled"
+import { Button, ButtonLink } from "../Button/Button"
+import type { ButtonLinkProps, ButtonProps } from "../Button/Button"
+import { css } from "@emotion/react"
+import type { Theme } from "@mui/material/styles"
+
+const TabButtonList: React.FC<TabListProps> = styled(MuiTabList)([
+  {
+    minHeight: "unset",
+    ".MuiTabs-indicator": {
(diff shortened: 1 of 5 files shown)


frontends/ol-components/src/components/TabButtons/TabButtonList.tsx
```diff
@@ -0,0 +1,75 @@
...
+    ".MuiTabs-scroller": {
+      display: "flex",
+    },
+  },
+])
+
+const tabStyles = ({ theme }: { theme: Theme }) =>
+  css({
+    ":focus-visible": {
+      outlineOffset: "-1px",
```
ChristopherChudzicki, line 38: Scrollable tabs include a scroll container with `overflow:hidden` that, without some extra padding, can occlude the browser's focus outline. Rather than adding padding (which would affect layout), I think we should move the focus outline inwards a little bit.

frontends/ol-components/src/components/TabButtons/TabButtonList.tsx
```diff
@@ -0,0 +1,75 @@
...
+)
+
+const TabButton = (props: TabProps<"button">) => (
+  <MuiTab {...props} component={TabButtonInner} />
+)
+const TabButtonLink = (props: TabProps<"a">) => (
+  <MuiTab {...props} component={TabLinkInner} />
+)
+
+export { TabButtonList, TabButton, TabButtonLink }
```
ChristopherChudzicki, line 75: The naming here leaves something to be desired. Originally I just used TabList, but as @gumaerc's work on profile page shows, we may sometimes want MUI's TabList component for other purposes.

gumaerc (approved):
👍 Looks and works great

PR #1015, opened 2024-06-05
Learning Resource Card

jonkafton:

[Learning Resource Cards #4356](https://github.com/mitodl/hq/issues/4356)

Implements the Learning Resource Card "card" variants, sizes "4bloc" and "6bloc" in Figma.

https://www.figma.com/design/Eux3guSenAFVvNHGi1Y9Wm/MIT-Design-System?node-id=827-6168&m=dev

The "list" variants are to follow in subsequent PR.

Provides a base `Card` component and `LearningResourceCard` in `ol-components`. The base `Card` is also now in use by the cards in the Homepage "News and Events" section.

image

image

### How can this be tested?

This has been applied to the homepage carousels in the "Featured Courses" (medium/4bloc variant) and "Media" (small/6bloc variant) sections.

- Load the homepage and confirm the card match the designs in Figma.
  - The title should link to the resource URL.
  - Resources with a next start date should display the start date in the footer.
    - The best run start date is used if not on the parent resource.
    - OCW courses show "As taught in: \<semester\> \<year\>".
  - Certificated resources should display a "Certificate" badge.
  - Authenticated learning path editors should have an action button that opens the "add to learning path" dialog.
  - Authenticated users should have an action button that opens the "add to user list" dialog.
  - As resources are loading, the skeleton load state should be displayed and broadly reflect the loaded contents.

- A LearningResourceCard component has also been added to `ol-components` and should render correctly in Storybook (`yarn run storybook` and navigate to http://localhost:6006/?path=/story/ol-components-learningresourcecard--course).

- A base Card component has also been added to `ol-components` and should render correctly in Storybook (small and medium variants) - http://localhost:6006/?path=/story/ol-components-card--medium.

- The base Card component

### Additional Context

Developer notes:

- This adds a base Card component to ol-components, intended to replace the template pattern, that provides the base styles to render a card. It provides "slots" for calling components to inject content into, ie.:

```
<Card>
  <Card.Image src={url} />
  <Card.Info>
    Info here
  </Card.Info>
  <Card.Title>
    Title here
  </Card.Title>
<Card>
```

  It also provides a `Card.Content` slot that will occupy the full `Card` for the panel styles only.

- The `Card` component provides `link` prop for cards that are links in their entirety - this is in use by the Homepage "News and Events" cards.

- I added some adjustments to the carousel width calculations so the cards align with the page container width (1280px), though the cards are now fixed width (302px and 194px including borders), so do not expand to parent width and do not need widths applied by the carousel. These aren't working on mobile where they're squashed, though @ChristopherChudzicki is looking at the carousel at the moment so is in his good hands.

- The LearningResourceCard and LearningResourceCardTemplate are still in use for the list variants (by ListDetailsPage and SearchPage), so are left in the mit-open workspace. These will be refactored out as the list variant is implemented.

- This renames the MUI Card export from `ol-components` to `MuiCard`. I didn't want to call our card anything other than `Card` and am not sure we should relinquish all the good component names to MUI. We could prefix all the MUI exports and I see the benefit in having a single entrypoint for all components, though my preference would be not to re-export library components at all. I see greater benefit in it being clear in the consuming code that one of our components vs a library component is being imported and I would import directly from MUI - it's unlikely we'd add a new component library to the project unnoticed.


frontends/mit-open/package.json
```diff
@@ -10,7 +10,7 @@
   "scripts": {
     "watch": "API_BASE_URL=http://localhost:8063 NODE_ENV=development ENVIRONMENT=local webpack serve",
     "watch:docker": "API_BASE_URL=http://nginx:8063 NODE_ENV=development ENVIRONMENT=local webpack serve",
-    "watch:rc": "API_BASE_URL=https://mitopen-rc.odl.mit.edu/ NODE_ENV=development ENVIRONMENT=local webpack serve",
+    "watch:rc": "API_BASE_URL=https://api.mitopen-rc.odl.mit.edu/ NODE_ENV=development ENVIRONMENT=local webpack serve",
```
gumaerc, line 13: This change seems unrelated to this PR. Can we break it out into a separate one for tracking purposes, and in case it needs to be reverted?

gumaerc (approved):
👍 

Everything here looks good, although I did notice a few changes that are unrelated to cards, and would be better broken out into a separate PR for tracking purposes or reversion. It's not 100% necessary though.

I feel like I should also mention that we had a conversation about this PR outside of Github on the topic of deprecating the current use of the re-exported `Card` from Material UI. We decided to push a change that does not require a size to be specified on `Card`, and it now does not restrict the size by default. So, in a future PR we can refactor uses of `MuiCard` to:

```jsx
<Card>
  <Card.Content>
    ...
  </Card.Content>
</Card>
```

This is a very basic example and other existing uses of `Card` might need to do something different, but it should at least now be a drop-in replacement.

ChristopherChudzicki:
@jonkafton Could we make the cards 300 and 192 (not 302 and 194)? The page width is actually supposed to be 1272, which i'm going to fix in carousel PR.
Screenshot 2024-06-06 at 11 18 00 AM

frontends/mit-open/package.json
jonkafton: PR here: https://github.com/mitodl/mit-open/pull/1022

PR #1017, opened 2024-06-05
Profile-based search filter preferences

mbertrand:
Closes https://github.com/mitodl/hq/issues/4386

Adds a field to the profile serializer specifying which search filters to use, based on profile preferences

### How can this be tested?
In django admin, go to the `Profile` object for your user, assign values for topic interests, certification desired, and learning format fields.  Check the API response at `http://localhost:8063/api/v0/profiles/<username>` and verify that the `preference_search_filters` attribute has appropriate values.


rhysyngsun (approved):
LGTM

PR #1021, opened 2024-06-05
Purge the fastly cache on deploy

blarghmatey:
N/A

Adds a step to purge the Fastly cache on deploy to RC and Production

### DIFF
--- a/.github/workflows/release-candidate.yml
+++ b/.github/workflows/release-candidate.yml
@@ -55,6 +55,13 @@ jobs:
           heroku_app_name: "mitopen-rc"
           heroku_email: ${{ secrets.HEROKU_EMAIL }}
           branch: release-candidate
+
+      - name: Purge Fastly cache
+        uses: jcoene/fastly-purge-action@master
+        with:
+          api_key: "${{ secrets.FASTLY_API_KEY_RC }}"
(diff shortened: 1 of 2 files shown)


.github/workflows/release-candidate.yml
```diff
@@ -55,6 +55,13 @@ jobs:
           heroku_app_name: "mitopen-rc"
           heroku_email: ${{ secrets.HEROKU_EMAIL }}
           branch: release-candidate
+
+      - name: Purge Fastly cache
+        uses: jcoene/fastly-purge-action@master
+        with:
+          api_key: "${{ secrets.FASTLY_API_KEY }}"
+          service_id: "${{ secrets.FASTLY_SERVICE_ID }}"
```
Ardiea, line 63: These need to include the `env_var_suffix` as apporpirate to each environment like the Embedly and AWS secrets.

Ardiea (changes requested):

.github/workflows/release-candidate.yml
blarghmatey: Good point. I scoped the key to work for both RC and Production, so that can stay as is, but I'll update the service ID to include that suffix.

Ardiea (approved):

PR #1024, opened 2024-06-06
Fix bug with onboarding steps not saving

rhysyngsun:

N/A

This fixes a bug introduced when I fixed an infinite rendering loop by adding a check for updated value equality. As part of this a variable reference that was originally shadowing a higher scope one got renamed but the usage didn't.

### How can this be tested?

To test, verify you can step through the onboarding flow (`htttp://localhost:8063/onboarding`) and see a payload with data in the PATCH requests


gumaerc (approved):
👍 

image

PR #1029, opened 2024-06-07
Add profile edit page

rhysyngsun:

Closes https://github.com/mitodl/hq/issues/4565
Fixes https://github.com/mitodl/hq/issues/4504

This adds the profile edit page under the dashboard.

- It significantly refactors the field implementations for onboarding so that they can be reused on this page.
- It moves `ChoiceBox` into `ol-components` and makes it more flexible and configurable so that it can be used on both pages.
- I also made the onboarding steps optional since that bug had been reported and I was in that code changing things and the profile edit page needed to behave the same.

Screenshot 2024-06-14 at 16-29-42 User Home
Screenshot 2024-06-14 at 16-30-27 User Home
Screenshot 2024-06-14 at 16-30-15 User Home

### How can this be tested?

- You should be able to navigate to the profile edit page under the dashboard and edit your profile there.
- You should still be able to successfully complete the onboarding flow (`/onboarding`).


frontends/ol-components/src/index.ts
```diff
@@ -10,6 +10,8 @@
  *    functionality.
  */
 
+export { useTheme } from "@mui/material/styles"
```
shanbady, line 13: This doesnt appear to be necessary

shanbady (changes requested):
when I change values for either of these select boxes the label doesnt appear to update (hitting save still updates them to whatever they are set to)
Screenshot 2024-06-17 at 12 55 31 PM

rhysyngsun:
@shanbady addressed those issues

shanbady (approved):
looks god

PR #1032, opened 2024-06-07
Add a setting for CSRF_COOKIE_DOMAIN

rhysyngsun:

N/A

This adds a settting for `CSRF_COOKIE_DOMAIN` so that we can have django set the cookie on a shared domain for doing cross domain CSRF requests via cookie.

### How can this be tested?

It's difficult to reproduce the bug, but you can confirm the behavior works locally by setting `CSRF_COOKIE_DOMAIN=.odl.local`. You should then see this cookie get set on `.odl.local` when loading the app locally.

### DIFF
--- a/app.json
+++ b/app.json
@@ -70,6 +70,10 @@
       "description": "CSAIL courses base URL",
       "required": false
     },
+    "CSRF_COOKIE_DOMAIN": {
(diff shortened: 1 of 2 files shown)


shanbady (approved):
looks good

PR #1039, opened 2024-06-10
Fixing authentication issue, and fixing some filtering and test issues

jkachel:

n/a

- Explicitly `AllowAny` - otherwise this requires authentication, which it shouldn't
- Remove filtering for published and default to published only - this API doesn't need to/shouldn't show unpublished items
- Fix tests to remove testing for published filter
- Fix tests to include testing for offeror search
- Fix tests to use the unauthenticated client rather than `user_client`, which is authenticated

### How can this be tested?

Automated tests should pass.

You should be able to hit the API with an unauthenticated session (say, in a private browsing window). The API is at http://od.odl.local:8063/api/v0/testimonials/ for local deployment.

You should only see published attestations from the API - these are attestations with either no publish date or a publish date in the past. You should no longer be able to filter by the publish status.

This should have no effect on the existing code - the testimonials are (now) only displayed on the homepage and it does not do any filtering.

### DIFF
--- a/testimonials/views.py
+++ b/testimonials/views.py
@@ -2,11 +2,11 @@
 
 from django.db.models import Q
 from drf_spectacular.utils import extend_schema, extend_schema_view
-from rest_framework.permissions import AllowAny
 from rest_framework.viewsets import ReadOnlyModelViewSet
 
 from learning_resources.views import LargePagination
 from main.filters import MultipleOptionsFilterBackend
+from main.permissions import AnonymousAccessReadonlyPermission
 from main.utils import now_in_utc
 from testimonials.filters import AttestationFilter
 from testimonials.models import Attestation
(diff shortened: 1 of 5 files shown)


testimonials/views.py
```diff
@@ -29,7 +29,10 @@ class AttestationViewSet(ReadOnlyModelViewSet):
...
 
     serializer_class = AttestationSerializer
-    queryset = Attestation.objects.all()
+    queryset = Attestation.objects.filter(
+        Q(publish_date__isnull=True) | Q(publish_date__lte=now_in_utc())
+    ).all()
     pagination_class = LargePagination
     filter_backends = [MultipleOptionsFilterBackend]
     filterset_class = AttestationFilter
+    permission_classes = [AllowAny]
```
mbertrand, line 38: The permission class `main.permissions.AnonymousAccessReadonlyPermission` has been used for other read-only views to allow anonymous access (like `TopicViewSet`, `OfferedByViewset`), but this works too.

mbertrand (approved):
LGTM, just one comment. 👍

PR #1042, opened 2024-06-10
adding configurable csrf settings and including withXSRFToken in axio…

### LINKED ISSUE #1041: CRSF errors in live environments
### Expected Behavior
We should not encounter any crsf errors when doing things like subscribing to lists or subscribing to searches when logged in

### Current Behavior
Currently certain actions on the site encounter a 403 forbidden due to invalid csrf tokens

### Steps to Reproduce
1. login on rc
2. attempt to subscribe to an offeror page (https://mitopen-rc.odl.mit.edu/c/offeror/ocw)  and see it fail with a 403 in web inspector

### Possible Solution
some combo of csrf related settings in settings.py and some settings to axios.ts that can fix it.

shanbady:
Fixes #1041

Adds some configurable cors and csrf settings so we can attempt to fix errors on our live environments

### How can this be tested?
We will need to deploy this to a live environment by setting the newly added settings in the environment


main/settings.py
```diff
@@ -151,11 +151,22 @@
...
-CORS_ALLOW_CREDENTIALS = True
 
-CSRF_TRUSTED_ORIGINS = get_list_of_str("CSRF_TRUSTED_ORIGINS", [])
+CORS_ALLOW_CREDENTIALS = get_bool("CORS_ALLOW_CREDENTIALS", True)  # noqa: FBT003
+SECURE_CROSS_ORIGIN_OPENER_POLICY = get_string(
+    "SECURE_CROSS_ORIGIN_OPENER_POLICY",
+    "same-origin",
+)
+
+CSRF_COOKIE_SECURE = get_bool("CSRF_COOKIE_SECURE", False)  # noqa: FBT003
```
mbertrand, line 161: I think it would be safer to have the default be `True` here, just in case we accidentally forget to set it on RC and prod.  Can ensure it's True locally in docker by adding `CSRF_COOKIE_SECURE: "False"` to the `x-environment` section of `docker-compose.yml`

mbertrand (changes requested):
LGTM, just one suggested change

blarghmatey (approved):

main/settings.py
mbertrand: Can you also add this line to the `x-environment` section of docker-compose.yml?

```
  CSRF_COOKIE_SECURE: ${CSRF_COOKIE_SECURE:-False}
```

Issue #384, opened 2024-01-09
Storybook stories for learning resource cards


jonkafton:
### Description/Context

Add Storybook stories for our front end learning resource `page-components`. These are quite configurable will benefit from a render playground and usage notes.

We currently have stories for our `ol-components` publishing to GitHub Pages at https://mitodl.github.io/mit-open/storybook. The additional stories will accompany them here.

### Plan/Design

Write *.stories.tsx and *.mdx files for `LearningResourceCard` and `LearningResourceCardTemplate`. The [ButtonLink](https://github.com/mitodl/mit-open/tree/019d74fbb83f73ee555149f50c8dae788201bfc7/frontends/ol-components/src/components/ButtonLink) stories can be used as a reference.

See also the Storybook documentation:
- [What's a story](https://storybook.js.org/docs/get-started/whats-a-story)
- [MDX](https://storybook.js.org/docs/writing-docs/mdx)

PR #1054, opened 2024-06-10
Learning Resource cards, list view

jonkafton:

Closes [Learning Resource Cards #4356](https://github.com/mitodl/hq/issues/4356)
Closes [Storybook stories for learning resource cards #384](https://github.com/mitodl/mit-open/issues/384)

Implements the Learning Resource Card "list" variants for desktop and mobile.

https://www.figma.com/proto/Eux3guSenAFVvNHGi1Y9Wm/MIT-Design-System?page-id=107%3A2479&node-id=3375-49853&viewport=-15774%2C-1090%2C0.55&t=cX0nylncBbGUadPG-1&scaling=min-zoom&starting-point-node-id=2197%3A31345

Applies them to: 
- the Search page at `/search`
- the Learning Paths page at `/learningpaths`

Also updates the `Card` component in its link variant to reflect changes to the hover design and behavior, - https://www.figma.com/design/Eux3guSenAFVvNHGi1Y9Wm/MIT-Design-System?node-id=827-6168&m=dev:

- Whole card is clickable
- No shadow by default
- Shadow on hover
- No title underline on hover

Search Page:

image
image

image

------

Learning Paths page:

image

------

Card hover updates:

image
image

### How can this be tested?

Load the Search and Learning Path pages at `/search` and `/learningpaths` and ensure the list items display correctly on desktop and at mobile screen widths
- Click anywhere on card to navigate
- Default image is shown where none on learning resource
- Item count shown in learning path resources
- Certificate badge shown where resource is certificated
- Action buttons display correctly and open dialogs when clicked
   - Search page
     - Add to user list for authenticated user
     - Add to learning path for learning path editor
   - Learning Paths
      - Edit/delete menu 
- Skeleton load state displays correctly

### Additional Context

- Moves action buttons out of the parent Link container in the DOM structure and places them with css to avoid the HTML spec constraint that a button cannot be nested inside an anchor.
    - There's an issue that the action buttons can overlap the footer info. We can solve this with javascript, though it's a bit messy and could be better solved by design - min width on the container, another breakpoint between mobile and desktop, limits on the text displayed (display 3 letter month, display data only).
   
image

- The LearningResourceCard page-component (and LearningResourceCardTemplate) is still in use by the Learning Resource Details page. We'll need to refactor sortability across before we can remove it for this one.

- Adds a button variation for now that matches the design with `edge="none"`, `color="secondary"`. There are [CardActions button designs](https://www.figma.com/design/Eux3guSenAFVvNHGi1Y9Wm/MIT-Design-System?node-id=3152-18866&t=vAm3BtVv8nketftd-0) that we haven't yet modelled in code.

image


Ferdi:
@jonkafton We have some courses (most of the MITx courses) that are free but you pay for a certificate.  I think we have a specific design for that . Can you pls provide a screenshot of one of those ?

jonkafton:
> @jonkafton We have some courses (most of the MITx courses) that are free but you pay for a certificate. I think we have a specific design for that . Can you pls provide a screenshot of one of those ?

@Ferdi An example here from MITx:

image

On the data on both we have `prices: [0]`, `free: true` and `certification: true`.

Similar from Sloan:
image

On this one we have `prices: [0]`, `free: false` and `certification: true`.

There is a design where the price is shown as $$$:

image

Ferdi:
@jonkafton thanks.  how about a screenshot for any of the courses in this list https://mitopen-rc.odl.mit.edu/search/?certification_type=completion&offered_by=mitx

@steven-hatch i think `$$$` could be confusing because the number of `$` usually reflects the price (`$` is cheap and `$$$` is expensive).  One suggestion is to replace it with "Paid"

Unrelated to Jon's PR, we have some data problems here @pdpinch @mbertrand 
`prices: [0], free: true and certification: true`
 Should never be the case -- We might not have prices for programs, but that does not mean they are free.
 
 Sloan courses / programs are never free

jonkafton:
> how about a screenshot for any of the courses in this list https://mitopen-rc.odl.mit.edu/search/?certification_type=completion&offered_by=mitx

@Ferdi here's a result from that search that's labelled as free, with certificate:

image

Here we have `free: true`, though `prices: [0, 129, 29]`. At the moment when we have an array of prices we select the first and ignore the rest.

frontends/mit-open/src/pages/LearningPathListingPage/LearningPathListingPage.tsx
```diff
@@ -60,49 +68,27 @@ const EditListMenu: React.FC<EditListMenuProps> = ({ resource }) => {
...
-      resource={list}
-      imgConfig={imgConfigs["row-reverse-small"]}
-      footerActionSlot={canEdit ? <EditListMenu resource={list} /> : null}
-      onActivate={onActivate}
-    />
-  )
-}
-
 const LearningPathListingPage: React.FC = () => {
+  const isMobile = !useMuiBreakpointAtLeast("md")
```
ChristopherChudzicki, line 88: Mentioning: I have been avoiding using these. I'm not sure what the SSR situation is for `useMuiBreakpoint`. Where we can, I think CSS is a better solution, though I understand changing props via mobile vs desktop can't be done with CSS.

frontends/ol-components/src/components/Button/Button.tsx
```diff
@@ -127,6 +127,15 @@ const ButtonStyled = styled.button<ButtonStyleProps>((props) => {
       // Pill-shaped buttons... Overlapping border radius get clipped to pill.
       borderRadius: "100vh",
     },
+    edge === "none" && {
+      border: "none",
+      ":hover:not(:disabled)": {
+        "&&": {
+          backgroundColor: "inherit",
+        },
+      },
```
ChristopherChudzicki, line 136: What's this for? It seems surprising to me. I wouldn't expect `edge` to affect the background color.

- With this style, some variants hover strangely with `edge="none"` (fill, tertiary)
- Without this...the new cards seem to look OK? and buttons seem to hover OK?

frontends/ol-components/src/components/Card/ListCard.tsx
```diff
@@ -0,0 +1,200 @@
...
+  height: ${theme.typography.pxToRem(40)};
+  ${theme.breakpoints.down("md")} {
+    ${{ ...theme.typography.body3 }}
+    height: ${theme.typography.pxToRem(32)};
+  }
+
+  overflow: hidden;
+  margin: 0;
+
+  @supports (-webkit-line-clamp: 2) {
```
ChristopherChudzicki, line 91: I saw you used `TruncateText` elsewhere. Should we use that here, possibly adding the `@supports` to it?

frontends/ol-components/src/components/Card/ListCard.tsx
```diff
@@ -0,0 +1,200 @@
...
+
+  Children.forEach(children, (child) => {
+    if (!isValidElement(child)) return
+    if (child.type === Content) content = child.props.children
+    else if (child.type === Image) imageProps = child.props
+    else if (child.type === Info) info = child.props.children
+    else if (child.type === Title) title = child.props.children
+    else if (child.type === Footer) footer = child.props.children
+    else if (child.type === Actions) actions = child.props.children
+  })
```
ChristopherChudzicki, line 164: I'd not seen this pattern before...what is its purpose? Is it so that child order doesn't matter? I.e., so that
```
// 1
<ListCard>
  <ListCard.Content>...</ListCard.Content>
  <ListCard.Actions>...</ListCard.Actions>
</ListCard>
// 2
<ListCard>
  <ListCard.Actions>...</ListCard.Actions>
  <ListCard.Content>...</ListCard.Content>
</ListCard>
```
are treated the same?

frontends/ol-components/src/components/LearningResourceCard/LearningResourceListCard.tsx
```diff
@@ -0,0 +1,323 @@
...
+          <LoadingView isMobile={isMobile} />
+        </ListCard.Content>
+      </ListCard>
+    )
+  }
+  if (!resource) {
+    return null
+  }
+  return (
+    <ListCard href={href || `?resource=${resource.id}`} className={className}>
```
ChristopherChudzicki, line 270: **Requested Change:** This `href` does not behave correctly for the search page—it replaces the search query parameters rather than augmenting them.

This does work
```
const { search } = useLocation()
// then later

<ListCard
  href={href || `${search}${search ? "&" : "?"}resource=${resource.id}`}
  className={className}
>
```

**HOEVER:** It feels strange to me that `ol-components` is aware of what route the resource drawer is at. (Route-awareness belongs to page-components).

IMO, let's just specify the href via the parent. (Although, I think having a page-component with the button actions and href builtin would be reasonable. That's what we used to do, with LearningResourceCardTemplate and LearningResourceCard. Don't love the "template" language, though. )

ChristopherChudzicki (changes requested):
This is looking really nice. I left a few comments and one requested change—the current `href` breaks search page.

### Spacing issue
@steven-hatch @jonkafton I also noticed a spacing issue on mid-size screens. That said, I'm not sure if the solution here is
1. change the cards,
2. change the search page
3. make the medium breakpoint to a larger value

I do not think this spacing issue needs to be resolved in this PR.

Screenshot 2024-06-13 at 4 24 04 PM

Uploading Screen Recording 2024-06-13 at 4.45.14 PM.mov…

frontends/ol-components/src/components/Button/Button.tsx
jonkafton: The idea is that when edge is "none" we're on mobile, so no hover state. It does happen though that if you are on desktop and shrink your window, the hover area doesn't look good.

image

I'm anticipating this will be removed from here / updated in favor of the Card Action Buttons, though we don;t seem to have a variant of these for the edge-less mobile variant https://www.figma.com/design/Eux3guSenAFVvNHGi1Y9Wm/MIT-Design-System?node-id=4026-61418&m=dev

image

@steven-hatch ?

frontends/ol-components/src/components/Card/ListCard.tsx
jonkafton: Yes, the pattern allows us to
- Provide slots that can be used without respect to the sequence
- Insert slots into a nested structure (example we might want to provide more granular slots inside the footer for e.g. start date, learning format)
- Give different treatment depending on the component, e.g. here we use image props if an image is provided - the image doesn't have children.

This RFC with [open PR](https://github.com/reactjs/rfcs/pull/223) goes into the detail: https://github.com/nihgwu/rfcs/blob/neo/slots/text/0000-slots.md

frontends/ol-components/src/components/Card/ListCard.tsx
jonkafton: Yes!.. pushed

frontends/ol-components/src/components/LearningResourceCard/LearningResourceListCard.tsx
jonkafton: Yes, intended as a default, though I agree it should be controlled by the caller (and total miss on replacing the whole search!).

I've added a LearningResourceDrawer hook to provide the href.

frontends/mit-open/src/pages/LearningPathListingPage/LearningPathListingPage.tsx
jonkafton: It's not really needed here - I've replaced with css.

ChristopherChudzicki (approved):
👍 👍

PR #1056, opened 2024-06-11
search prefs learning format as list

mbertrand:
Closes https://github.com/mitodl/hq/issues/4386

Changes the `learning_format` field inside `profile.preference_search_filters` to be a list.

### How can this be tested?
In django admin, go to the `Profile` object for your user, assign values for topic interests, certification desired, and learning format fields.  Check the API response at `http://localhost:8063/api/v0/profiles/<username>` and verify that the `preference_search_filters` attribute has appropriate values, and `learning_format` is a list (if that profile field is not null).


gumaerc (approved):
👍

PR #1060, opened 2024-06-11
make slick fail more gracefully when parent width unconstrained

ChristopherChudzicki:
A followup to https://github.com/mitodl/hq/issues/4360 .

Our carousel needs a container with some width constraint. On main, if there is no width constraint, the carousel enters an infinite loop. (This behavior does not occur on the live site, but was stumbled upon as part of the dashboard page.)

This PR makes the carousel fail more gracefully: it no longer enters an infinite loop and has a fallback max width.

### How can this be tested?
If you want to see the carousel crash on main, add `<div style={{display: "flex"}}>` wrapper to any carousel on the homepage. Flex items are not allowed, by default, to shrink smaller than their content, and the carousel will resize itself larger and larger and larger. 

Try the same on this branch. It won't crash.


gumaerc (approved):
👍 This fixes the issue

PR #1064, opened 2024-06-11
Add ordering to testimonials, add featured testimonials view

jkachel:

Closes mitodl/hq#4534

This adds a few things to the testimonials code:
- You can now specify an order for testimonials. Testimonials are listed in order of specified order, then last updated date when listing them out.
- There is now a Featured Testimonials API. This lists out one applicable testimonial per offeror that has a testimonial.
   - The order the testimonials appear in the final output is shuffled. 
   - If there's more than one featured testimonial in the given offeror, the applicable testimonials for the offeror are shuffled and one is chosen.
   - If a testimonial is both in the featured spot and tagged with more than one offeror, it ensures that the testimonial is not duplicated (so if you have one in ocw and see, for instance, and it appears in ocw, it won't also appear in see).

This also switches the homepage testimonials section to use the Featured Testimonials API and adds the requisite hooks for it, since it's a new API. 

A testimonial is considered featured if its `position` attribute is set to 1. _This is also the default_, so existing testimonials will be set to 1.

### How can this be tested?

Automated tests should pass.

You will need a batch of testimonials in the system. There's a bunch now in RC if you want to test with more-or-less real data. I would recommend at least 9 - that seemed like a good amount to cover all the test cases - but you can test with more or less of them.

Test that testimonials appear in order:
- Set your testimonials so they have a known order. Some should overlap.
- Load the regular testimonials API: http://od.odl.local:8063/api/v0/testimonials/
- Items should load in order of position. The overlapping items should load in order of updated_on descending.
- If you filter by channel or offeror, the sorting should remain.

Test that featured testimonials appear correctly:
- Set all testimonials to have _one_ offeror. Do this in a batch - for my testing, I had three sets of three set to ocw, bootcamps, and see. 
- Set _at least two_ testimonials in each batch to be in position 1, and the rest to be in some other position (2 is fine).
- Load the featured testimonials list API: http://od.odl.local:8063/api/v0/featured_testimonials/
- You should only see featured testimonials in the list (i.e., ones in position 1).
- You should only see _one_ testimonial per offeror, and you should not see more testimonials than there are offerors that have testimonials. That is, if you've only assigned testimonials to 3 offerors, you should only get 3 testimonials back.
- The order of the testimonials should be randomized.
- If you've assigned more than one testimonial to the featured spot for an offeror, the selected testimonial for the offeror should also be randomized - you may need to refresh a few times to see this.
- If you've assigned a testimonial to the featured spot and also to more than one offeror, you should never see that testimonial more than once.

On the homepage:
- You should only see featured testimonials displayed in the carousel.
- The order of those testimonials should change, and (if you've assigned >1 to the featured spot) you may see different ones for each offeror spot.


jkachel:
Simplified this a whole lot and opened a new PR: https://github.com/mitodl/mit-open/pull/1067

PR #1069, opened 2024-06-12
Update education options and add to schema

rhysyngsun:

Fixes https://github.com/mitodl/hq/issues/4537 

- Updates the options to match other products
- Updates our schema generation per-value descriptions to enums.
  - This uses a well-known schema extension `x-enum-descriptions` that openapi-generator supports (you can see as much with the addition of per-value doc comments).
  - NOTE: some typescript generated description enums are empty. I'm treating this as a known issue because there's 2 complications to fixing this I don't want to address in this PR:
    - There are enums being generated inline that are nested within API response declarations in the schema and I don't want to invested in supporting those because they mostly relate to enums that we already have defined, the schema generator just doesn't have the information it needs for this.
    - The generator templates use mustache, which is logic-less, so I can't conditionally exclude enums that don't have descriptions
- Removes hardcoded `LABELS` definitions in the onboarding flow to use the new generated descriptions.
- Adds a migration to empty out the education choices because none of the old values are valid anymore.

### How can this be tested?

- You should still be able to use the onboarding flow and the dropdown for education should reflect the new options.


shanbady (changes requested):
Looks like there is a missing migration and a broken test. otherwise looks and works well

shanbady (approved):
looks good

PR #1074, opened 2024-06-12
Add pytest-xdist and use it for CI builds

rhysyngsun:

N/A

Python tests have been taking an increasing amount of time to run, definitely locally and I suspect github actions are suffering as well. This adds the plugin `pytest-xdist` which allows the tests to run in parallel. Each test worker gets its own db so they don't stomp on each other.

### How can this be tested?

You can run this a few ways (you don't need to test them all, but test at least one `-n` option):

- `pytest` - same as before, everything is run serially
- `pytest -n auto` - run 1 process per physical cpu
- `pytest -n logical` - run 1 process per logical cpu
- `pytest -n N` - run N processes

On a daily basis you probably want to use the `auto` option since it won't consume all your cpu, and for me at least I didn't see enough of a difference to justify it. In my testing I saw these run times:

`pytest` - 506.12s (0:08:26)
`pytest -n auto` - 214.72s (0:03:34)
`pytest -n logical` - **206.06s (0:03:26)**

**Testing on GitHub Actions:**

`pytest` - 475.74s (0:07:55)
`pytest -n auto` - 303.72s (0:05:03)
`pytest -n logical` - **266.84s (0:04:26)**

### DIFF
--- a/learning_resources_search/serializers_test.py
+++ b/learning_resources_search/serializers_test.py
@@ -565,7 +565,7 @@ def test_serialize_bulk_learning_resources(mocker):
 @pytest.mark.django_db()
 @pytest.mark.parametrize(
     "resource_type",
-    set(LearningResourceType.names()) - {LearningResourceType.course.name},
+    sorted(set(LearningResourceType.names()) - {LearningResourceType.course.name}),
 )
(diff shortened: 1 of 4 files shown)


mbertrand (approved):
👍

PR #1084, opened 2024-06-14
Update OCW unit name in offerors.json

pdpinch:
The unit name is "MIT OpenCourseWare"

none

"OpenCourseWare" => "MIT OpenCourseWare" 

### How can this be tested?

After this is changed, I expect the breadcrumbs on the OCW unit page will say "MIT OpenCourseWare" instead of "OCW"

### DIFF
--- a/learning_resources/data/channel_templates/offerors.json
+++ b/learning_resources/data/channel_templates/offerors.json
@@ -7,7 +7,7 @@
     "banner_background": "/static/images/unit_banners/mitx.jpg"
   },
   "ocw": {
-    "name": "OpenCourseWare",
+    "name": "MIT OpenCourseWare",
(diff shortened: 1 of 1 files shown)


shanbady (approved):
Approving this change. As a side-note - offerors.json is local. As it stands I have to ssh into each env post-deploy and manually run a django command to see this change live:
  - its might be worth getting rid of the json and instead baking these values into a migration
  - updates to values can also be handled by newer data migrations
  - for local environments re-running migrations or ```docker compose restart``` updates to the latest values

pdpinch:
Thanks for the explanation. I expect one more round of changes (likely
significiant) and then we won’t touch this again.

A migration makes sense to me. I'll leave the decision to the next person
who works in this, instead of editing json.

PR #1085, opened 2024-06-14
Sort run prices on save; make learning resource prices equal "next run" prices

mbertrand:
Closes https://github.com/mitodl/hq/issues/4585

- Modifies the calculation of `LearningResource.prices` to return the prices of the resource's next run (if any).
- Adds a line to the `etl.loaders.load_run` function to sort the list of prices before saving.

#### LearningResourceRun.prices sort
- On the main branch, if you haven't already, run `backpopulate_mit_edx_data` (you will need the `EDX_API_` settings from heroku).  Check that the following returns some unsorted run prices:
  ```python
  for run in LearningResourceRun.objects.filter(published=True, prices__isnull=False):
      if run.prices != sorted(run.prices):
          print(run.id, run.learning_resource.etl_source, run.learning_resource.resource_type, run.prices)
  ```
- Switch to this branch, run `backpopulate_mit_edx_data` again.  This time, if you run the above code, no runs should be returned

#### LearningResource.prices
Go to `http://localhost:8063/api/v1/courses` and check that the resource `prices` field is the same as for the next run of that resource.

### Additional Context
This PR won't update existing runs with unsorted price, but the ETL pipelines that will do so run every day.

### DIFF
--- a/frontends/api/src/generated/v1/api.ts
+++ b/frontends/api/src/generated/v1/api.ts
@@ -603,11 +603,11 @@ export interface CourseResource {
    */
   certification_type: CourseResourceCertificationType
   /**
-   * Returns the prices for the learning resource
-   * @type {Array<number>}
+   *
+   * @type {Array<string>}
    * @memberof CourseResource
    */
-  prices: Array<number>
+  prices: Array<string>
   /**
    *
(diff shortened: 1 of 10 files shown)


jkachel (approved):
LGTM 👍

PR #1086, opened 2024-06-14
Expose thenew user login url as an environment var

rhysyngsun:

Fixes https://github.com/mitodl/hq/issues/4586

Exposes the onboarding url setting as an environment variable so we can set it to the frontend domain.

### How can this be tested?

You can set this locally to a dummy value and verify it gets picked up.


shanbady (approved):
looks good.

PR #1094, opened 2024-06-14
Mb/mitxonline price zero

mbertrand:
https://github.com/mitodl/hq/issues/4606

Always makes sure there is a $0 in run prices from MITx Online, openedx courses.

### How can this be tested?
Run the following, with required .env settings from heroku RC:
```
./manage.py backpopulate_mit_edx_data
```

Check that the imported runs have a `prices` list containing at least one 0.


abeglova (approved):
This works! I did notice that prices are sometimes strings and sometimes numbers depending on the offeror. We should definitely be consistent but that can be a follow up pr

PR #1097, opened 2024-06-17
Fix populate_featured_lists mgmt command

mbertrand:
N/A, related to 

Fixes a bug in the `populate_featured_lists` mgmt command introduced by https://github.com/mitodl/mit-open/pull/1031

- [ ] Desktop screenshots

### How can this be tested?
Run `./manage.py backpopulate_resource_channels --all` followed by `./manage.py populate_featured_lists`, they should both complete successfully.


shanbady (changes requested):
change looks good and works but we should add a test that will catch this error/exercises that code path

shanbady (approved):

PR #1103, opened 2024-06-17
Homepage Stories & Events layout fixes

jonkafton:

Closes https://github.com/mitodl/hq/issues/4477

1. Card default and hover state design updates were done in previous PR (https://github.com/mitodl/mit-open/pull/1054).

2. Fixes the bottom horizontal alignment on the News and Events Homepage section cards.

3+4. vertically centers the title in the event cards.

image

### How can this be tested?

Load the Homepage and check that the above items are addresses and the layout matches [the designs](https://www.figma.com/design/Eux3guSenAFVvNHGi1Y9Wm/MIT-Design-System?node-id=3852-51239&m=dev).

### Additional Context

- Also addresses issue where the footer text and action buttons collide in the small variant cards. Not solved necessarily, but "Start:" label removed and month abbreviated to 3 letters (small mode only).

Previously:
image

- Adds the facility to pass styles to override the Card slots.

### DIFF
--- a/frontends/mit-open/src/pages/HomePage/NewsEventsSection.tsx
+++ b/frontends/mit-open/src/pages/HomePage/NewsEventsSection.tsx
@@ -73,7 +73,6 @@ const EventsContainer = styled.section`
   flex-direction: column;
   align-items: flex-start;
   gap: 12px;
-  align-self: stretch;
 `
 
 const StoryCard = styled(Card)<{ mobile: boolean }>`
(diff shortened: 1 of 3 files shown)


shanbady (changes requested):
style changes Look good. one discrepancy i saw was that the news events did not have any action buttons as shown in the figma. might be outside the scope of this ticket but is this expected?

jonkafton:
> one discrepancy i saw was that the news events did not have any action buttons as shown in the figma. might be outside the scope of this ticket but is this expected?

To my knowledge news/events don't have action buttons (the learning resource "add to user list" and admin "add to learning path" don't apply).

shanbady (approved):
in that case looks like there is just the test to fix before merging

PR #1104, opened 2024-06-17
show courses and programs first in default sort

abeglova:
closes https://github.com/mitodl/hq/issues/4578
makes backend changes needed for https://github.com/mitodl/hq/issues/4620

This pr adds a new "is_learning_material"  boolean to the learning resources search index and api. This is true for Podcast, Podcast Episode, Video, Video Playlist. 

It also updates the default sort when there is no text filter to  [is_learning_material, -created_on] instead of just -created_on so that courses and programs are shown before "learning materials"

### How can this be tested?

recreate your search index by running manage.py recreate_index --all ` this will take a while if you have a lot of content files

go to http://localhost:8063/search
go to http://localhost:8063/search?sortby=new in a different tab

Verify that the default sort ("best match") shows the same resources as sort by new but with everything except for courses and programs excluded.

Try searching for text term. The default sort should sort by relevance

Verify that http://localhost:8063/api/v1/learning_resources_search/?is_learning_material=true, http://localhost:8063/api/v1/learning_resources_search/?is_learning_material=false and http://localhost:8063/api/v1/learning_resources_search/?aggregations=is_learning_material gives the expected results and http://localhost:8063/api/v1/schema/redoc/#tag/learning_resources_search/operation/learning_resources_search_retrieve looks good

### DIFF
--- a/learning_resources_search/serializers.py
+++ b/learning_resources_search/serializers.py
@@ -38,6 +38,8 @@ from learning_resources.serializers import (
 from learning_resources_search.api import gen_content_file_id
 from learning_resources_search.constants import (
     CONTENT_FILE_TYPE,
+    COURSE_TYPE,
+    PROGRAM_TYPE,
 )
 from learning_resources_search.models import PercolateQuery
 from learning_resources_search.utils import remove_child_queries
@@ -79,6 +81,8 @@ def serialize_learning_resource_for_update(
     return {
         "resource_relations": {"name": "resource"},
         "created_on": learning_resource_obj.created_on,
+        "is_learning_material": learning_resource_obj.resource_type
+        not in [COURSE_TYPE, PROGRAM_TYPE],
         **serialized_data,
     }
 
@@ -175,6 +179,7 @@ LEARNING_RESOURCE_AGGREGATIONS = [
     "professional",
     "free",
     "learning_format",
+    "is_learning_material",
 ]
 
 CONTENT_FILE_AGGREGATIONS = ["topic", "content_feature_type", "platform", "offered_by"]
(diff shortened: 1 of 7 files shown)


learning_resources_search/serializers.py
```diff
@@ -79,6 +81,8 @@ def serialize_learning_resource_for_update(
     return {
         "resource_relations": {"name": "resource"},
         "created_on": learning_resource_obj.created_on,
+        "is_learning_material": learning_resource_obj.resource_type
+        not in [COURSE_TYPE, PROGRAM_TYPE],
         **serialized_data,
     }
```
ChristopherChudzicki, line 87: These three fields:
1. resource_relations
2. created_on
3. is_learning_material

are "extra". None of them show up in http://localhost:8063/api/v1/learning_resources/.

(1) and (2) do not show up in http://localhost:8063/api/v1/learning_resources_search/, but (3) does. Do you know why?

Mostly asking out of curiosity. If removing (3) is easy, that might be worthwhile since then the response formats will match exactly, which has been the goal. That said, having (3) is not a problem—it's "extra" and it's not usable on the frontend since it doesn't show up in the openapi (good).

ChristopherChudzicki (approved):
👍 Looks good to me.

I left one question.

learning_resources_search/serializers.py
abeglova: I hid the is_learning_material from the results. It's pretty redundant anyway. I guess it's a bit weird to filter by something that's not a field in the results but keeping the the output the same is good and cluttering the results of both with a field that's pretty specific to our ui seems unnessesary

PR #1111, opened 2024-06-18
two separate search inputs

abeglova:
closes https://github.com/mitodl/hq/issues/4632

Prior to this PR the search inputs on the front page and search page was the same component. https://github.com/mitodl/mit-open/pull/1104 updated the input. However we do not want to update the input on the search page so we need two separate components

frontends/mit-open/src/page-components/SearchDisplay/SearchInput.tsx is mostly the latest version of frontends/ol-components/src/components/SearchInput/SearchInput.tsx

and 

frontends/mit-open/src/pages/HomePage/SearchInput.tsx is mostly the version frontends/ol-components/src/components/SearchInput/SearchInput.tsx before my latest pr

Screenshot 2024-06-17 at 9 49 23 PM

Screenshot 2024-06-17 at 9 49 44 PM

Screenshot 2024-06-17 at 9 48 40 PM

Screenshot 2024-06-17 at 9 48 59 PM

### How can this be tested?
Go to http://localhost:8063/
Go to http://localhost:8063/search/

The search inputs should work and look like the designs

https://www.figma.com/design/Eux3guSenAFVvNHGi1Y9Wm/MIT-Design-System?node-id=3375-49855&m=dev


ChristopherChudzicki (changes requested):
@abeglova This restores the homepage nicely. I do think we change some things, though. In particular, we should use the `Input`'s `size` property and move the "primary CTA ('call to action')" button into `ol-components`.

ChristopherChudzicki:
Arg. I submitted my review without submitting the main comment... stand by.

frontends/mit-open/src/page-components/SearchDisplay/SearchInput.tsx
ChristopherChudzicki, line 1: I want to request some overall changes here. Basically, the request is
1. use existing component library styles (e.g., the "size" property)
2. Add the "primary CTA (call to action) button" (the rounded button to right of input) to the component library.

In this file, the `Input` component is manually styled pretty heavily—we set height, border, breakpoint, padding, icon size, etc. These things should be handled in the component library (and largely are already).

If you look in Figma, you'll see the input has size "XL (hero)" (or "mobile hero" on mobile) 
Screenshot 2024-06-18 at 9 50 12 AM

Compare with http://localhost:6006/?path=/docs/smoot-design-input--docs (This is storybook. It should be running. If it's not, run `yarn storybook` in the watch container.)

So to size this input, we should pass `size="hero"` to the underlying input, rather than styling the input manually in the file. That way:
- If designs for "hero" input update (and we implement the updates), then the search page would update automatically. *Desirable and expected by designers.*
- The input currently has size "medium" (the default) with manual overrides. Updating the medium size **should not** affect search page, but currently will.
- Using `size="hero"`, you should have to do much less manual styling.
- `size="hero"` automatically adjusts for mobile screens.
    - Although, that currently happens at `sm` screens not `md` screens. You could change `Input.tsx` a little to make that customizable.

Additionally, I would request that you add the "Primary CTA (call to action)" button to `ol-components` rather than implemetning it here, and add an example of its usage in storybook. (I'm happy to help with this.)

You can see that this is a component we intend to re-use here: https://www.figma.com/design/Eux3guSenAFVvNHGi1Y9Wm/MIT-Design-System?node-id=104-1762&m=dev

The primary CTA button should work well with `size="default"` or `size="mobile"`. 

The above changes will also be using when building out the search box for the channel pages.

ChristopherChudzicki (approved):
@abeglova Let's align the search input separately, as you suggested.

PR #1112, opened 2024-06-18
Update the conditional for the marketing image chooser

jkachel:

Closes mitodl/hq#4635

The `generateMarketingImageSrc` uses a do..while to re-select a random marketing image if the image that we just displayed to the user is the one that gets selected again. This also needed to drop out of the loop if there wasn't a last marketing image, so now it does.

### How can this be tested?

Automated tests should pass. No real front-end change should be seen - the homepage should load and your testimonials images should be randomized and shouldn't repeat back-to-back (they will still repeat in the list as a whole).


ChristopherChudzicki (approved):
Approving this because it fixes the infinite loop in our tests.

It does have an issue which james is going to fix in a followup PR.

https://github.com/mitodl/mit-open/assets/9010790/d2327402-afc8-46d2-be33-21ddf03c38a9

PR #1113, opened 2024-06-18
Fixes/simplifies marketing image selector

jkachel:

Closes mitodl/hq#4636

Refactors the marketing image selection code to be simpler and to avoid selecting the same image back-to-back (with assistance from @ChristopherChudzicki).

### How can this be tested?

Automated tests should pass. You should be able to load the homepage and the marketing images should not repeat back-to-back.


gumaerc (approved):
👍

PR #1120, opened 2024-06-18
switch default sort to use popular instead of created on

abeglova:
None

Ferdi asked for the default sort to sort by popularity instead of new

So the sort is now by is_learning_material (false first) then popularity

### How can this be tested?
Go to http://localhost:8063/search verify that the sort order is the same as http://localhost:8063/search?sortby=-views with everything that isn't a course or program excluded 

If you search by text, the search should continue to be by relevance


shanbady (approved):
looks good

PR #1121, opened 2024-06-18
Updates to page titles

jonkafton:

Closes [Page titles in Unified/Open #4510](https://github.com/mitodl/hq/issues/4510)

- Adds a build variable, `SITE_NAME`, defaults to "MIT Open"
- Appends "MIT Open" (currently) page title part to all page titles.
- Separates title parts with ` | `
- Updates titles according to https://github.com/mitodl/hq/issues/4510

Also updates titles:

- Edit article page from "Editing: <article_name>" to "<article_name> | Edit | MIT Open"
- User dashboard from "User Home" to "Your MIT Learning Journey | MIT Open"
- User lists page from "User Lists" to "My Lists | MIT Open"
- Any titles prefixed "MIT Open", now suffixed.

### How can this be tested?

- Navigate through site and confirm page title updated in tab and appears in history and bookmarks.

- Lighthouse report passes title checks (post RC release? not tested)

### DIFF
--- a/frontends/jest.jsdom.config.ts
+++ b/frontends/jest.jsdom.config.ts
@@ -1,6 +1,8 @@
 import { resolve } from "path"
 import type { Config } from "@jest/types"
 
+process.env.SITE_NAME = "MIT Open"
+
 /**
  * Base configuration for jest tests.
  */
--- a/frontends/api/src/clients.ts
+++ b/frontends/api/src/clients.ts
@@ -24,6 +24,11 @@ import {
 
 import axiosInstance from "./axios"
 
+// const BASE_PATH =
+//   process.env.ENVIRONMENT === "local"
+//     ? ""
+//     : process.env.MITOPEN_AXIOS_BASE_PATH?.replace(/\/+$/, "") ?? ""
+
 const BASE_PATH = process.env.MITOPEN_AXIOS_BASE_PATH?.replace(/\/+$/, "") ?? ""
 
 const learningResourcesApi = new LearningResourcesApi(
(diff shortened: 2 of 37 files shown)


frontends/jest.jsdom.config.ts
```diff
@@ -1,6 +1,8 @@
 import { resolve } from "path"
 import type { Config } from "@jest/types"
 
+process.env.SITE_NAME = "MIT Open"
```
shanbady, line 4: wondering if this can just fall back to the [default](https://github.com/mitodl/mit-open/pull/1121/files#diff-cf829b1ff5e1c8d9a75406c84d056628f3406931f3e539a198db34b85a7936e5R54) since I dont think we explicitly set environment variables even for testing purposes

shanbady (changes requested):
works as intended. left a minor nitpick

frontends/jest.jsdom.config.ts
jonkafton: The idea for the tests is that we mock the process.env value by setting it explicitly here, then expect it in the tests, both hardcoded.

We could not hardcode, instead provide on environment, and then check e.g. ``expect(document.title).toBe(`About Us | ${process.env.SITE_NAME}`)`` in the tests, though not sure it adds anything above mocking the environment for testing.

shanbady: thats true. i raised it since i didnt see us setting an env variable directly elsewhere  - dont think its a blocker. approving.

shanbady (approved):

PR #1122, opened 2024-06-18
Working search filter on topics with commas

mbertrand:
Closes https://github.com/mitodl/hq/issues/4013

- Converts `LearningResourcesSearchRequestSerializer` fields from a `StringArray` to a standard `ListField`
- Makes changes to db rest api's to handle query parameter values containing commas.
- Removes ability to have comma-delimited query parameter values (like `?topic=Biology,Chemistry`; must be `?topic=Biology&topic=Chemistry`)

#### Backend
The following should work and return results (if you have imported OCW courses):
http://localhost:8063/search/?topic=Atomic,%20Molecular,%20Optical%20Physics
http://localhost:8063/search/?topic=Atomic%2C+Molecular%2C+Optical+Physics&topic=Epidemiology
http://localhost:8063/api/v1/courses/?topic=Atomic,%20Molecular,%20Optical%20Physics
http://localhost:8063/api/v1/courses/?topic=Atomic%2C+Molecular%2C+Optical+Physics&topic=Epidemiology

#### Frontend
The following page should work and return search results (assuming you've run the mgmt command `backpopulate_resource_channels --all`:  http://localhost:8063/c/topic/atomic-molecular-optical-physics/

Update a channel with a search filter containing multiple values for certain facets:
```python
from channels.models import FieldChannel
FieldChannel.objects.filter(unit_detail__unit__code="ocw").update(
  search_filter="offered_by=ocw&topic=Biology&topic=Physics&department=7&department=8"
)
```
Go to http://localhost:8063/c/unit/ocw and make sure the search request returns results using the same topic and department parameters as above.

### DIFF
--- a/frontends/mit-open/src/pages/FieldPage/FieldPage.test.tsx
+++ b/frontends/mit-open/src/pages/FieldPage/FieldPage.test.tsx
@@ -46,7 +46,14 @@ const setupApis = (
   const urlParams = new URLSearchParams(fieldPatch?.search_filter)
   const subscribeParams: Record<string, string[] | string> = {}
   for (const [key, value] of urlParams.entries()) {
-    subscribeParams[key] = value.split(",")
+    if (
+      subscribeParams[key] !== undefined &&
+      Array.isArray(subscribeParams[key])
+    ) {
+      subscribeParams[key] = [...subscribeParams[key], value]
+    } else {
+      subscribeParams[key] = [value]
+    }
   }
   subscribeParams["source_type"] = "channel_subscription_type"
   const subscribeResponse = isSubscribed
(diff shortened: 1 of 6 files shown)


pdpinch:
Will this require any changes in the front-end?

ChristopherChudzicki:
> Will this require any changes in the front-end?

I think the answer is "No", but we should definitely be careful with this this.

- The API client is using exploded parameters (i.e., `a=1&a=2&a=3`) and almost everything is using the API client.
- The search page already uses exploded parameters, e.g.,  https://mitopen-rc.odl.mit.edu/search/?certification_type=completion&certification_type=professional&topic=Engineering&topic=Mechanical+Engineering
    - Which is to say, course-search-utils is using exploded parameters.
- There a few hard-coded links to the search page, but they don't have arrays in the query params.
    - e.g., the home page links to `/search?certification=true` 

We should also check subscription queries and channel filters, but I don't think any of them use multiple parameters right now. cc @shanbady

mbertrand:
@pdpinch I don't believe so, the front end should work the same as before.

shanbady:
@ChristopherChudzicki @mbertrand I recall having to adapt the query param parser to work for all array type params (a=test&a=3 , a[]=test, a=test,2) I would still double check that this works

abeglova (approved):
lgtm

abeglova (approved):

PR #1130, opened 2024-06-20
fixing sort method for panel detail display

shanbady:

Fixes https://github.com/mitodl/hq/issues/4665

This fixes the sorting of information displayed in the info panel on unit pages (https://mitopen-rc.odl.mit.edu/c/unit/mitx/)

In particular - Offerings should always be first and more information should be last.

### How can this be tested?
1. Checkout this branch
2. visit a unit page /c/unit/mitx/ etc.
3. see that offerings is always first and more information (if displayed) should always show up last

How this should behave is that everything should be sorted in natural order - we specify an 'order' of -1 for offerings which is followed by everything else which is 0 and then more information should always be last since it is >0. You can try playing with [the order parameter](https://github.com/mitodl/mit-open/blob/fdaa66a2bd46c0a49245ed424b3d07744864063e/frontends/mit-open/src/page-components/ChannelDetails/ChannelDetails.tsx#L106) and verify it behaves as we expect

### DIFF
--- a/frontends/mit-open/src/page-components/ChannelDetails/ChannelDetails.test.tsx
+++ b/frontends/mit-open/src/page-components/ChannelDetails/ChannelDetails.test.tsx
@@ -40,4 +40,28 @@ describe("ChannelDetails", () => {
     screen.getByText(certifications.join(" | "))
     screen.getByText(contentTypes.join(" | "))
   })
+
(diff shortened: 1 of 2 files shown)


jkachel (approved):
LGTM 👍

PR #1137, opened 2024-06-21
Search facet checkbox and label styles

jonkafton:

Closes https://github.com/mitodl/hq/issues/4654

Styles the search facet check box and label to align with designs.

image

### How can this be tested?

Navigate to the search page. Default, hover and checked styling should match https://www.figma.com/proto/Eux3guSenAFVvNHGi1Y9Wm/MIT-Design-System?page-id=107%3A2479&node-id=3375-49853&viewport=-21276%2C125%2C0.84&t=x1opLeUv6Qp1anj1-1&scaling=min-zoom&starting-point-node-id=2197%3A31345

### Additional Context

This change didn't call for a checkbox in ol-components as we can't provide a component into `@mitodl/course-search-utils` and are styling from outside it. We'll want to reuse the styles when we add one (add to list dialog is likely the next item).

### DIFF
--- a/frontends/mit-open/src/page-components/SearchDisplay/SearchDisplay.tsx
+++ b/frontends/mit-open/src/page-components/SearchDisplay/SearchDisplay.tsx
@@ -214,16 +214,14 @@ export const FacetStyles = styled.div`
       .facet-count {
         font-size: 12px;
         padding-left: 3px;
-        color: ${({ theme }) => theme.custom.colors.silverGray};
+        color: ${({ theme }) => theme.custom.colors.silverGrayDark};
         float: right;
       }
     }
 
-    .facet-visible.checked,
-    .facet-visible:hover {
-      .facet-label label {
-        color: ${({ theme }) => theme.custom.colors.darkGray2};
-      }
+    .facet-visible.checked .facet-label label,
+    .facet-visible .facet-label label:hover {
+      color: ${({ theme }) => theme.custom.colors.darkGray2};
     }
 
     .facet-more-less {
--- a/frontends/mit-open/src/page-components/SearchDisplay/SearchDisplay.tsx
+++ b/frontends/mit-open/src/page-components/SearchDisplay/SearchDisplay.tsx
@@ -155,13 +155,6 @@ export const FacetStyles = styled.div`
     }
   }
 
-  .facet-visible.checked,
-  .facet-visible:hover {
-    .facet-label label {
-      color: ${({ theme }) => theme.custom.colors.darkGray2};
-    }
-  }
-
   .facet-visible {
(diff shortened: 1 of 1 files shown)


frontends/mit-open/src/page-components/SearchDisplay/SearchDisplay.tsx
```diff
@@ -273,6 +282,7 @@ export const FacetStyles = styled.div`
     padding-bottom: 12px;
     padding-top: 10px;
 
+    /* stylelint-disable no-descending-specificity */
```
jonkafton, line 285: Lint rule errors either way round when this block is moved to above
```
[mit-open]:  211:7  ✖  Expected selector ".facets .facet-visible label" to come before selector ".multi-facet-group .facet-visible .facet-label label"                no-descending-specificity
[mit-open]:  235:7  ✖  Expected selector ".facets .facet-visible .facet-count" to come before selector ".multi-facet-group .facet-visible .facet-label .facet-count"  no-descending-specificity
```

ChristopherChudzicki:
@jonkafton This looks good. I noticed two small issues:

Screenshot 2024-06-20 at 4 47 03 PM

Figma also has the "Free" count left-aligned, but I'm 90% sure that's outdated and it should be right-aligned like the others.

ChristopherChudzicki (changes requested):
@jonkafton This looks good. I noticed two small issues:

Screenshot 2024-06-21 at 10 31 41 AM

Figma also has the "Free" count left-aligned, but I'm 90% sure that's outdated and it should be right-aligned like the others.

Pretty sure both of the above should be simple. If for any reason they aren't, we can merge as-is.

jonkafton:
> @jonkafton This looks good. I noticed two small issues:
> 
> Screenshot 2024-06-21 at 10 31 41 AM
> Figma also has the "Free" count left-aligned, but I'm 90% sure that's outdated and it should be right-aligned like the others.
> 
> Pretty sure both of the above should be simple. If for any reason they aren't, we can merge as-is.

Good spot on the "Free" color - Changed pushed for that one.

The counts don't change color on hover in the [Figma mock up](https://www.figma.com/proto/Eux3guSenAFVvNHGi1Y9Wm/MIT-Design-System?page-id=107%3A2479&node-id=3375-49853&viewport=-21276%2C125%2C0.84&t=x1opLeUv6Qp1anj1-1&scaling=min-zoom&starting-point-node-id=2197%3A31345&show-proto-sidebar=1). I also noticed that the count is not clickable, so it's not correct to give it a hover state. I pushed a change to only apply a hover state to clickable checkbox and label.

ChristopherChudzicki (approved):
LGTM 👍

PR #1144, opened 2024-06-21
Adds a separate pane for the filter CTAs, adds an apply button on mobile

jkachel:

Closes mitodl/hq#4599

In the filter drawer on mobile, there should be a CTA section that includes an Apply Filters and a Clear All button if the learner has selected any of the options in the drawer, and it should appear below the title and close box. This adds those. (Currently, there is a Clear All button but it is set in the title area, so that one just moves down a bit.) 

image

### How can this be tested?

Open the drawer while in tablet or mobile mode (width < 900px). You should see the buttons once you select something in the pane. They should be hidden if you don't have anything selected.

### DIFF
--- a/frontends/mit-open/src/page-components/SearchDisplay/SearchDisplay.tsx
+++ b/frontends/mit-open/src/page-components/SearchDisplay/SearchDisplay.tsx
@@ -393,27 +393,12 @@ const StyledDrawer = styled(Drawer)`
 `
 
 const MobileFacetSearchButtons = styled.div`
-  width: 100%;
+  display: flex;
   gap: 12px;
-`
 
-const MobileApplyFiltersButton = styled(Button)`
-  background-color: ${({ theme }) => theme.custom.colors.mitRed};
-  color: ${({ theme }) => theme.custom.colors.white};
-  padding: 12px;
-  border-radius: 4px;
-  width: 144px;
-  margin-right: 12px;
-  box-shadow:
-    0 2px 4px 0 #25262b1a,
-    0 3px 8px 0 #25262b1f;
-`
-
-const MobileClearAllButton = styled(Button)`
-  background-color: white;
-  padding: 12px;
-  border-radius: 4px;
-  width: 144px;
+  & > button {
+    flex: 1;
+  }
 `
 
 const MobileDrawerCloseButton = styled(Button)`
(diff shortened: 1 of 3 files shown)


frontends/mit-open/src/page-components/SearchDisplay/SearchDisplay.tsx
```diff
@@ -392,10 +392,28 @@ const StyledDrawer = styled(Drawer)`
...
+  width: 100%;
+  gap: 12px;
+`
+
+const MobileApplyFiltersButton = styled(Button)`
+  background-color: ${({ theme }) => theme.custom.colors.mitRed};
+  color: ${({ theme }) => theme.custom.colors.white};
+  padding: 12px;
+  border-radius: 4px;
+  width: 144px;
```
ChristopherChudzicki, line 405: In addition to my comment above... try to avoid explicit widths. Figma often specifies them because Steve and Bilal tend to design for a few specific screen sizes. Ideally,  try to match figma at full width, but keep roughly the same look and feel at other widths.

In particular, a fixed width here doesn't work great on some phones:

Screenshot 2024-06-25 at 10 28 22 AM
Screenshot 2024-06-25 at 10 28 28 AM

*Figma designs are for 375px screens, and the iphone SE is 375px, so you might be surprised the 144px didn't work well. I was surprised, too. I guess there's a slight deviation in Figma's drawer vs our implementation. Figma shows 300px content space. Our drawer has `max-width: 332px; padding-left/right: 16px, width: 85%`. The `width: 85%` is the difference, making the drawer a little smaller than 300px.*  Might be worth aligning that with Figma, but probably not here.

**Anyway:** A flex-based layout for the buttons would be a better way to match Figma at the 300px content width, but scale appropriately on slightly smaller phones.

frontends/mit-open/src/page-components/SearchDisplay/SearchDisplay.tsx
```diff
@@ -392,10 +392,28 @@ const StyledDrawer = styled(Drawer)`
...
+  margin-right: 12px;
+  box-shadow:
+    0 2px 4px 0 #25262b1a,
+    0 3px 8px 0 #25262b1f;
```
ChristopherChudzicki, line 409: The `<Button variant="primary" />` is missing its shadow. That's a bug. IMO, don't change it here. Just live with no shadow. I will create a separate issue for that.

ChristopherChudzicki (changes requested):
Left some suggestions / requests that I think will simplify this a bit.

frontends/mit-open/src/page-components/SearchDisplay/SearchDisplay.tsx
ChristopherChudzicki: https://github.com/mitodl/hq/issues/4700

frontends/ol-components/src/components/Button/Button.tsx
```diff
@@ -133,6 +139,17 @@ const ButtonStyled = styled.button<ButtonStyleProps>((props) => {
         color: colors.silverGray,
       },
     },
+    variant === "noBorder" && {
+      backgroundColor: colors.white,
+      color: colors.darkGray2,
+      border: "none",
+      ":hover:not(:disabled)": {
+        backgroundColor: tinycolor(colors.darkGray1).setAlpha(0.06).toString(),
```
ChristopherChudzicki, line 147: I didn't see a hover state specified in figma. This looks fine, but I did leave a note for steve/bilal jkachel/4599-mobile-filter-ctas

ChristopherChudzicki (approved):
👍

PR #1152, opened 2024-06-24
fix django hijack route

shanbady:

Fixes #https://github.com/mitodl/hq/issues/4383

Fixes user hijacking functionality

### How can this be tested?
1. checkout this branch
2. restart your nginx container via ```docker compose restart ngix```
3. go to the user admin page /admin/auth/user/
4. click "hijack" on some user and see that it successfully authenticates as them

### Additional Context
I'm not entirely certain if we need to also update some routes in fastly to include /hijack or not. This may also require additional tweaking since we access admin from api.* on the live servers so we will have to see how this functions once deployed

### DIFF
--- a/config/nginx.conf
+++ b/config/nginx.conf
@@ -32,7 +32,7 @@ server {
         return 204;
     }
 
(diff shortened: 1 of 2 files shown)


mbertrand (approved):
Fixes hijack locally! 👍 

Not seeing a frontend indicator that hijack is in use as in open-discussions (screenshot below), but that can be a separate issue/PR.

Screenshot 2024-06-24 at 1 45 16 PM



@blarghmatey might have insight on whether or not fastly might need route changes to support this, but approving since this is at minimum working for local docker containers.

PR #1154, opened 2024-06-24
Use course_description_html field for OCW courses

mbertrand:
Closes https://github.com/mitodl/hq/issues/4592

Alters the OCW pipeline to extract descriptions from the `course_description_html` field.

### How can this be tested?
- Set `AWS_` and `OCW_` values to the same as RC in your .env file.
- Run the following:
  > ./manage.py backpopulate_ocw_data --skip-contentfiles --course-name res-21g-3003-marguerite-de-roberval-a-web-based-approach-to-teaching-a-renaissance-heroine-fall-2023 --overwrite
- When done, check that the description for the course contains a few `<p>` tags but no links or other HTML elements.


gumaerc (approved):
👍

PR #1167, opened 2024-06-25
fix content file search

abeglova:
closes https://github.com/mitodl/hq/issues/4697 

This PR fixes the current error on api/v1/content_files_search by makes user_list_parents and learning_path_parents 
something that is added to /learning_resources_search results only and not /content_file_search

Also renames the LearningResourceSearchResponse to LearningResourcesSearchResponse since the view is called /learning_resources_search and the other serializer is called LearningResourcesSearchRequest

NA

### How can this be tested?
Go to http://localhost:8063/api/v1/content_file_search/. Verify that it does not throw an error both when you are logged in and when you are not logged in 

Go to http://localhost:8063/api/v1/learning_resources_search/.  If you don't have them already create some user lists and learning paths and add some resources into the lists. Verify that user_list_parents and learning_path_parents are set when  /learning_resources_search/ returns resources that are part of a user list. Log out.  Verify that user_list_parents and learning_path_parents are set to [] for all resources.

### DIFF
--- a/learning_resources_search/serializers.py
+++ b/learning_resources_search/serializers.py
@@ -457,6 +457,29 @@ class SearchResponseSerializer(serializers.Serializer):
     def get_count(self, instance) -> int:
         return instance.get("hits", {}).get("total", {}).get("value")
 
+    def get_metadata(self, instance) -> SearchResponseMetadata:
+        return {
+            "aggregations": _transform_aggregations(instance.get("aggregations", {})),
+            "suggest": _transform_search_results_suggest(instance),
+        }
+
+
+class PercolateQuerySerializer(serializers.ModelSerializer):
+    """
+    Serializer for PercolateQuery objects
+    """
+
+    class Meta:
+        model = PercolateQuery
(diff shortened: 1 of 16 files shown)


learning_resources_search/serializers.py
```diff
@@ -457,6 +457,25 @@ def get_previous(self, instance) -> str | None:
...
 
+    def get_metadata(self, instance) -> SearchResponseMetadata:
+        return {
+            "aggregations": _transform_aggregations(instance.get("aggregations", {})),
+            "suggest": _transform_search_results_suggest(instance),
+        }
+
+
+class PercolateQuerySerializer(serializers.ModelSerializer):
+    class Meta:
```
mbertrand, line 468: Maybe add a simple docstring like "Serializer for PercolateQuery objects"

mbertrand (approved):
Works great, 👍

learning_resources_search/serializers.py
abeglova: I didn't update this code - it's only showing up as new because stuff around it moved. But I'll add a docstring

PR #1168, opened 2024-06-25
fix flaky test

ChristopherChudzicki:
Closes https://github.com/mitodl/hq/issues/4707

Fixes https://github.com/mitodl/hq/issues/4707. One additional change, see testing notes.

- [ ] Desktop screenshots

### How can this be tested?
1. Tests should pass. *Note that commit [9a86ebf](https://github.com/mitodl/mit-open/pull/1168/commits/9a86ebfcbd6441a9309db118110d3449b76b1137) includes running the originally flaky test 1000 times; they passed.*
2. Check that the "Featured Courses" carousel on unit pages still displays.
3. Check that the "Featured courses" carousel on unit pages only makes a single API call
    - Compare with https://mitopen-rc.odl.mit.edu/c/unit/ocw ... if you view the network tab, you'll see two API calls:
        - `featured/?limit=12`, which occurs while the channel is loading
        - then `featured/?limit=12&offered_by=ocw` after the channel has loaded

### DIFF
--- a/frontends/mit-open/src/pages/FieldPage/FieldPage.test.tsx
+++ b/frontends/mit-open/src/pages/FieldPage/FieldPage.test.tsx
@@ -1,5 +1,5 @@
 import { assertInstanceOf } from "ol-utilities"
-import { urls, factories } from "api/test-utils"
+import { urls, factories, makeRequest } from "api/test-utils"
 import type { FieldChannel } from "api/v0"
 import type { LearningResourceSearchResponse } from "api"
 import {
@@ -7,7 +7,6 @@ import {
   screen,
   setMockResponse,
   within,
-  act,
   waitFor,
 } from "../../test-utils"
 import FieldSearch from "./FieldSearch"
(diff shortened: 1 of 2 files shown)


frontends/mit-open/src/pages/FieldPage/FieldPageSkeleton.tsx
```diff
@@ -311,6 +311,7 @@ const FieldSkeletonProps: React.FC<FieldSkeletonProps> = ({
             <FeaturedCoursesCarousel
               title="Featured Courses"
               config={FEATURED_RESOURCES_CAROUSEL}
+              isLoading={field.isLoading}
```
ChristopherChudzicki, line 314: The carousel normally displays a loading state while the API call specified by its config is in flight.

Adding `isLoading` delays any API call from being initiated until `true`. In this case, after the field has loaded.

frontends/mit-open/src/pages/FieldPage/FieldPage.test.tsx
```diff
@@ -35,11 +34,7 @@ const setupApis = (
     field,
   )
   setMockResponse.get(
-    urls.learningResources.featured({ limit: 12, platform: ["ocw"] }),
-    factories.learningResources.resources({ count: 0 }),
-  )
-  setMockResponse.get(
-    urls.learningResources.featured({ limit: 12 }),
+    expect.stringContaining(urls.learningResources.featured()),
```
ChristopherChudzicki, line 37: The flakiness was being caused by the fact that the API url on line 38 wasn't always correct—it depends on the `search_filter`. In some tests it was *always* wrong, but the test would pass before any api call would be attempted, meaning we never got an error. 

To fix this, I made the mocking more flexible... now it just needs to contain `urls.learningResources.featured()`. Since this is much more flexible, I added an assertion to one of the tests that the correct API call is actually made.

abeglova (approved):
👍

PR #1170, opened 2024-06-25
rename field to channel

gumaerc:
Closes https://github.com/mitodl/mit-open/issues/734

This PR renames `FieldChannel` to `Channel`, along with all related code mentioning "field" or "fields" to "channel" or "channels."

### How can this be tested?
Realistically, the entire app should be smoke tested but I'm not sure what else to recommend here as it's such a massive rename. First of all, when you spin up the app the migrations should apply successfully. Beyond that, the code should be gone through with care to ensure that there haven't been any unintended side effects.


frontends/mit-open/src/common/urls.ts
```diff
@@ -28,15 +28,17 @@ export const DEPARTMENTS = "/departments/"
 export const TOPICS = "/topics/"
 
 export const FIELD_VIEW = "/c/:channelType/:name" as const
```
mbertrand, line 30: Change this to `CHANNEL_VIEW`

mbertrand (changes requested):
Overall looks great, everything I tried still works, just a couple comments.  This could probably use another set of eyes so I'll ask on slack if someone else can take a look at it too.

channels/urls.py
```diff
@@ -5,29 +5,29 @@
 
 from channels.views import (
     ChannelByTypeNameDetailView,
-    FieldChannelViewSet,
+    ChannelViewSet,
     FieldModeratorDetailView,
```
shanbady, line 9: should this be ChannelModeratorDetailView?

channels/urls.py
```diff
@@ -5,29 +5,29 @@
...
     FieldModeratorListView,
```
shanbady, line 10: same here

channels/models.py
```diff
@@ -82,7 +82,7 @@ def __str__(self):
         return self.title
 
 
-class FieldChannel(BaseChannel, TimestampedModel):
+class Channel(BaseChannel, TimestampedModel):
```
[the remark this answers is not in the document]
gumaerc: I'm going to opt to leave this for now because I'm more afraid of the migration this will create than the others. I realize `BaseChannel` isn't actually used anywhere, so there are no instances of `BaseChannel` objects on their own to migrate, but if I took the fields from `BaseChannel` and put them on `Channel`, wouldn't that still trigger `makemigrations` to pick up that the fields have been removed and re-added? I'm afraid of data loss in that scenario, but maybe that concern is unfounded and Django is smart enough to realize we're just getting rid of the base class?

shanbady (changes requested):

channels/models.py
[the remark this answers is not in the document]
gumaerc: Actually, scratch that, I just tried it and it doesn't generate a migration, so I will go ahead and do this.

shanbady (changes requested):
Just noticed on the frontend there are several components that have a "field" prop which is a reference to a Channel type. EditChannelAppearanceForm has a reference like this ```
type FormProps = {
  field: Channel
}```

gumaerc:
@mbertrand @shanbady Thanks for looking at this. I went through all the files referencing the word "field" in any way again with a fine toothed comb and I think I found all the remaining references. If y'all can just give it one last quick onceover I think we can merge this thing.

mbertrand (commented):
Found [another](https://github.com/mitodl/mit-open/pull/1170/files#diff-2e3e472ef3e92d351230bbae48bf6066634203fc201a99d59905a2f5b371993eR25) (`fieldName`), maybe change to `channelName` or just `name`:

```typescript
const useChannelDetail = (channelType: string, fieldName: string) => {
  return useQuery({
    ...channels.detailByType(channelType, fieldName),
  })
}
```

shanbady (changes requested):
I did see that FieldAvatar.tsx and FieldPageSkeleton.tsx still refer to a "field" (channel) prop and there is a comment about field-oriented pages. it seems the generated api generated/v0/api.ts has docstrings referring to "Field Moderators". 

Also - this might be more effort than its worth but should we be renaming Field* components such as FieldPageSkeleton.tsx , FieldSearch.tsx, FieldPage.tsx, FieldAvatar.tsx, FieldMenu.tsx to Channel* ?

gumaerc:
@shanbady You might want to make sure you're looking at the right branch, as I renamed all of those pages yesterday (`FieldPage.tsx`, etc.)

shanbady (approved):
@gumaerc it seems like my ide was caching some stuff. looks good on my end.

mbertrand (approved):
👍

PR #1179, opened 2024-06-26
Removing nginx serving of the frontend for local setups

rhysyngsun:

Closes https://github.com/mitodl/hq/issues/4602
Closes https://github.com/mitodl/hq/issues/4611

- All nginx serving of frontend is removed (except for the e2e tests as a special case).
- Updated the nginx setup in docker-compose to use the same nginx.conf that production uses. This implementation was lifted from [MITx Online](https://github.com/mitodl/mitxonline/tree/main/nginx)
- Changed the configuration of `env_file` so that it's split up between frontend vs. backend so it's generally cleaner and not as noisy.
- Renamed the environment variables for the API and app urls so they have more obvious names.

### How can this be tested?

- You should be able to still use your existing `.env` file as I left support for that in place
- You can also split it up into `env/frontend.local.env` and `env/backend.local.env`. There's also `env/shared.local.env`, but that is mainly for the `MITOPEN_API_BASE_URL` and `MITOPEN_APP_BASE_URL` vars.
- You can also remove a lot of vars from your `.env` file(s) that are now defined by default in the checked in `.env` files. A lot of time we just ended up leaving what was in `.env` and this should hopefully make that file a lot less noisy.
- You should be able to do all the normal things you can do:
  - Run the app via `docker compose up` with the different combinations of `COMPOSE_PROFILES` set.
    - If you're running both profiles, you'll now access the app in your browser via port `8062` (e.g. http://open.odl.local:8062/). The API and django-admin are still on port `8063`.
  - Run one-off containers via `docker compose run --rm web bash`, etc.
  - Run the tests on both frontend and backend.


docker-compose.yml
```diff
@@ -60,11 +47,16 @@ services:
...
+    build:
+      context: ./nginx
     ports:
       - "8063:8063"
     links:
       - web
+    environment:
+      PORT: 8063
+      NGINX_WORKERS: 1
+      NGINX_UWSGI_PASS: "web:8011"
```
gumaerc, line 59: This seems incorrect, shouldn't it be `NGINX_UWSGI_PASS: "web:8061"`?

docker-compose.yml
```diff
@@ -60,11 +47,16 @@ services:
...
     volumes:
       - ./config/nginx.conf:/etc/nginx/conf.d/web.conf
```
gumaerc, line 61: With the changes here, this should be changed to:

```yaml
volumes:
  - ./config:/etc/nginx/templates
  - ./frontends/mit-open/build:/src/frontends/mit-open/build
```

Although, I think we can also remove that volume mount for the frontend build now right?

nginx/Dockerfile
```diff
@@ -0,0 +1,22 @@
...
+
+# Logs are configured to a relatic path under /etc/nginx
+# but the container expects /var/log
+RUN mkdir -p /etc/nginx/logs && ln -sf /var/log/nginx /etc/nginx/logs/
+
+# erb unfortunately needs a whole ruby install
+RUN apt-get update && apt-get install -y ruby
+
+# this gets run automatically by the nginx container's entrypoint
+COPY 20-compile-nginx-conf-erb.sh /docker-entrypoint.d
```
gumaerc, line 15: This script will not be executable inside the container once it's copied over. A line needs to be added after this to make it so:

`RUN chmod +x /docker-entrypoint.d/20-compile-nginx-conf-erb.sh`

nginx/20-compile-nginx-conf-erb.sh
```diff
@@ -0,0 +1,18 @@
...
+
+set -e
+
+ME=$(basename $0)
+
+auto_erb_build() {
+	template="/etc/nginx/templates/nginx.conf.erb"
+	output_path="/etc/nginx/nginx.conf"
+
+	echo >&3 "$ME: Running erb on $template to $output_path"
```
gumaerc, line 12: I had to remove this line or else I would get an error about a "bad file descriptor"

gumaerc (changes requested):
I was able to get this working, but I had to make some tweaks to the nginx configuration:

rhysyngsun:
@gumaerc I addressed your feedback and also updated the readme to reflect the new config file locations.

gumaerc (approved):
This looks good to me now, however I will leave you a reminder to put up a PR in `ol-infrastructure` to set `MITOPEN_APP_BASE_URL` and `MITOPEN_API_BASE_URL` in our various deployment environments before releasing this.

mbertrand (commented):
Using my old .env, I can't seem to login via django admin (tried via both ports/domains), I always get a CSRF error, unless I set both MITOPEN_API_BASE_URL and MITOPEN_APP_BASE_URL to localhost with different ports.  Even setting them both to `open.odl.local` with different ports leads to a CSRF error.

Renamed my old .env and tried to use `env/backend.local.env`, `env/shared.local/end`, `frontend.local.env` but it couldn't seem to find `shared.local.env` at least where `COMPOSE_PROFILES` is, kept getting the error "no service selected"

gumaerc:
@rhysyngsun I will test this again on Monday and take a closer look at the settings Matt mentioned above, but we should probably hold off merging until we know it picks up those settings properly.

rhysyngsun:
> Renamed my old .env and tried to use `env/backend.local.env`, `env/shared.local/end`, `frontend.local.env` but it couldn't seem to find `shared.local.env` at least where `COMPOSE_PROFILES` is, kept getting the error "no service selected"

`COMPOSE_PROFILES` still needs to be defined in a root `.env` file or as an exported environment variable in your shell. This is because the behavior of it being picked up in that file is a feature of `docker compose`. We just happened to be also using this file for our own `.env`. I'll update the readme to make that clear and remove `COMPOSE_PROFILES` from the `shared.local.example.env` file.

env/frontend.local.example.env
```diff
@@ -0,0 +1,2 @@
+# MITOPEN_AXIOS_WITH_CREDENTIALS=True
+# WEBPACK_ANALYZE=True
```
jonkafton, line 2: `EMBEDLY_KEY=` here as well.

`MITOPEN_AXIOS_WITH_CREDENTIALS` isn't needed as it's in the committed file. Are the .local.env files intended only for secrets (aside from where it makes sense to group related variables)? Thinking `WEBPACK_ANALYZE` can go in `frontend.env` as well.

gumaerc:
> > Renamed my old .env and tried to use `env/backend.local.env`, `env/shared.local/end`, `frontend.local.env` but it couldn't seem to find `shared.local.env` at least where `COMPOSE_PROFILES` is, kept getting the error "no service selected"
> 
> `COMPOSE_PROFILES` still needs to be defined in a root `.env` file or as an exported environment variable in your shell. This is because the behavior of it being picked up in that file is a feature of `docker compose`. We just happened to be also using this file for our own `.env`. I'll update the readme to make that clear and remove `COMPOSE_PROFILES` from the `shared.local.example.env` file.

I tested this again today and confirmed that `COMPOSE_PROFILES` needs to be set at the root level, but the other vars pull just fine from the more specific env files.

env/frontend.env
```diff
@@ -0,0 +1,4 @@
+NODE_ENV=development
+PORT=8062
+MITOPEN_AXIOS_WITH_CREDENTIALS=true
+WEBPACK_ANALYZE=True
```
mbertrand, line 4: This causes my watch container to throw an error and exit

```
WEBPACK_ANALYZE: Invalid bool input: "True"
```

mbertrand (commented):
No longer getting a CSRF error

env/frontend.env
gumaerc: > This causes my watch container to throw an error and exit
> 
> ```
> WEBPACK_ANALYZE: Invalid bool input: "True"
> ```

I get this same error

gumaerc: @rhysyngsun I think this needs to be lowercase `true` because the var is evaluated by Javascript

mbertrand (approved):
👍

PR #1180, opened 2024-06-26
Filled vs Unfilled Bookmarks

ChristopherChudzicki:
https://github.com/mitodl/hq/issues/4712

Makes userlist bookmark and learningpath icon filled vs unfilled based on membership in at least 1 list.

**Cards**
Screenshot 2024-06-26 at 6 32 13 PM

**List Cards**
Screenshot 2024-06-26 at 6 32 46 PM

**List Cards, Mobile**
Screenshot 2024-06-26 at 6 32 57 PM

### How can this be tested?

1. As a logged in user, click the bookmark button on a card to change its membership in "userlists".
    - try this throughout the app, particularly for horizontal "ListCard" (search page) and vertical cards
2. When the resource belongs to at least one list, the bookmark should be filled.
3. When the resource belongs to zero lists, the bookmark should be unfilled.
4. The stafflist button should behave similarly.
5. Try adding/removing featured courses (homepage "Featured Courses" carousel) to a **userlist**.
    - in contrast to RC, the carousel resources should maintain their order.

### DIFF
--- a/frontends/api/src/hooks/learningResources/index.ts
+++ b/frontends/api/src/hooks/learningResources/index.ts
@@ -16,7 +16,6 @@ import type {
   LearningPathResource,
   LearningPathRelationshipRequest,
   MicroLearningPathRelationship,
-  LearningResource,
   LearningResourcesSearchApiLearningResourcesSearchRetrieveRequest as LRSearchRequest,
   UserlistsApiUserlistsListRequest as ULListRequest,
   UserlistsApiUserlistsCreateRequest as ULCreateRequest,
@@ -28,10 +27,13 @@ import type {
   MicroUserListRelationship,
   PlatformsApiPlatformsListRequest,
   FeaturedApiFeaturedListRequest as FeaturedListParams,
+  PaginatedLearningResourceList,
 } from "../../generated/v1"
 import learningResources, {
   invalidateResourceQueries,
   invalidateUserListQueries,
+  updateListParentsOnAdd,
+  updateListParentsOnDestroy,
 } from "./keyFactory"
 import { ListType } from "../../common/constants"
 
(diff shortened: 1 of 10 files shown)


frontends/api/src/hooks/learningResources/index.ts
```diff
@@ -298,13 +294,21 @@ const useUserListRelationshipCreate = () => {
         UserListRelationshipRequest: params,
       }),
     onSuccess: (response, _vars) => {
-      queryClient.setQueryData(
-        learningResources.detail(response.data.child).queryKey,
-        response.data.resource,
+      queryClient.setQueriesData<PaginatedLearningResourceList>(
+        learningResources.featured({}).queryKey,
+        (old) => updateListParentsOnAdd(response.data, old),
```
ChristopherChudzicki, line 299: Regarding `updateListParentsOnAdd` and `updateListParentsOnDestroy `:

1. The featured carousel on the homepage is populated by an API that has an intentionally randomized order.
2. In general throughout the app, when data might change, we tell react query "invalidate the cached data" and it re-fetches.
3. Adding the featured courses to a userlist changes their data (it affects `resource.user_list_parents). **However**, if we invalidate the cache and refetch new, updated data, then the carousel order will change, which is strange.
    - you can see this behavior on RC / Prod, but the bookmarks won't fill / unfill, so it's less noticeable that position has changed.
4. So instead of invalidating the cache and refetching the data, we manually update the data on the frontend.

ChristopherChudzicki, line 299: Regarding `updateListParentsOnAdd` and `updateListParentsOnDestroy `:

1. The featured carousel on the homepage is populated by an API that has an intentionally randomized order.
2. In general throughout the app, when data might change, we tell react query "invalidate the cached data" and it re-fetches.
3. Adding the featured courses to a userlist changes their data (it affects `resource.user_list_parents). **However**, if we invalidate the cache and refetch new, updated data, then the carousel order will change, which is strange.
    - you can see this behavior on RC / Prod, but the bookmarks won't fill / unfill, so it's less noticeable that position has changed.
4. So instead of invalidating the cache and refetching the data, we manually update the data on the frontend. 


We **only** do this for userlists, not for learnignpaths, because...
- changing a resource's learningpath membershould can change whether it is a featured course or not, so we really do need to refetch the data.

abeglova (changes requested):
This works fantastic generally and looks great! Some small fixes are needed for smaller mobile screens

ChristopherChudzicki:
@abeglova I've tweaked the button positioning to account for button paddings. The cards should look good at 375px with one button, and may look good with 2 buttons depending on date length.

Screenshot 2024-06-28 at 2 37 18 PM

Screenshot 2024-06-28 at 2 35 13 PM

abeglova (approved):
lgtm

ChristopherChudzicki:
pre-commit.ci run

ChristopherChudzicki:
Since github says this is mergable, and pre-commit just says
> timeout waiting for merge ref

I'm going to merge this.

PR #1181, opened 2024-06-27
Swap search and login button positions

jonkafton:

Closes https://github.com/mitodl/hq/issues/4718

Switches search and login button position in header logged out view.

image

### How can this be tested?

Load any page while logged out.


rhysyngsun (approved):
LGTM

PR #1183, opened 2024-06-27
Square aspect ratio for media resource images

jonkafton:

Closes https://github.com/mitodl/hq/issues/4523

Displays media resource (podcast episodes, videos) images tn the homepage Media section carousel with square aspect ratios.

image

### How can this be tested?

Navigate to homepage. Check resource cards in the Media section have images with aspect ratio 1 and that their loading state reflects loaded proportions. Check that other resource cards are not affected.

### Additional Context

- Adds an `isMedia` prop to the `LearningResourceCard`. 
- Sets this on the Media section cards by carousel configuration

Opted to do this by configuration because although we can detect that a resource is media by its resource type (`podcast`, `podcast_episode`, `video`), if we do that at the card level we end up with these:

Screenshot 2024-06-27 at 17 45 41

### DIFF
--- a/frontends/mit-open/src/pages/HomePage/HomePage.tsx
+++ b/frontends/mit-open/src/pages/HomePage/HomePage.tsx
@@ -28,8 +28,10 @@ const FeaturedCoursesCarousel = styled(ResourceCarousel)(({ theme }) => ({
 }))
 const MediaCarousel = styled(ResourceCarousel)(({ theme }) => ({
   margin: "80px 0",
+  minHeight: "388px",
   [theme.breakpoints.down("md")]: {
     margin: "40px 0",
+    minHeight: "418px",
   },
 }))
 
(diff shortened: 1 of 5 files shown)


frontends/mit-open/src/pages/HomePage/HomePage.tsx
```diff
@@ -28,8 +28,10 @@ const FeaturedCoursesCarousel = styled(ResourceCarousel)(({ theme }) => ({
 }))
 const MediaCarousel = styled(ResourceCarousel)(({ theme }) => ({
   margin: "80px 0",
+  minHeight: "388px",
```
jonkafton, line 31: This preserves the height of the section as it was collapsing as I tabbed through the media types, though we may want to do this in the carousel itself.

ChristopherChudzicki:
@steven-hatch fyi ☝️ 

What was your expectation for media cards and course/program cards being mixed on the same carousel?

Jon's solution—make the square aspect ratio opt in, and only use it where we know all cards are media—seems reasonable.

abeglova (approved):
Works great!

PR #1185, opened 2024-06-27
setting 100px as default width for buttons

shanbady:

Closes https://github.com/mitodl/hq/issues/4717

Sets a min width of 100px for buttons

### How can this be tested?
Checkout this branch and ensure all buttons such as the login button have a minimum width of 100px

### DIFF
--- a/frontends/mit-open/src/page-components/Header/MenuButton.tsx
+++ b/frontends/mit-open/src/page-components/Header/MenuButton.tsx
@@ -1,4 +1,4 @@
-import { Button, styled } from "ol-components"
+import { styled } from "ol-components"
 import { RiMenuLine } from "@remixicon/react"
 import React from "react"
 
@@ -23,14 +23,25 @@ const MenuButtonInner = styled.div({
   alignItems: "flex-start",
 })
 
-const StyledMenuButton = styled(Button)({
+const StyledMenuButton = styled.button(({ theme }) => ({
   padding: "0",
   background: "transparent",
   "&:hover:not(:disabled)": {
     background: "transparent",
   },
   touchAction: "none",
-})
+  textAlign: "center",
+  display: "inline-flex",
+  justifyContent: "center",
+  alignItems: "center",
+  color: theme.palette.text.primary,
+  transition: `background ${theme.transitions.duration.short}ms`,
+  cursor: "pointer",
+  borderWidth: "1px",
+  borderColor: "currentcolor",
+  borderStyle: "none",
+  color: theme.custom.colors.white,
+}))
 
 interface MenuButtonProps {
   text?: string
--- a/frontends/mit-open/src/page-components/Header/MenuButton.tsx
+++ b/frontends/mit-open/src/page-components/Header/MenuButton.tsx
@@ -49,7 +49,7 @@ interface MenuButtonProps {
 }
 
 const MenuButton: React.FC<MenuButtonProps> = ({ text, onClick }) => (
-  <StyledMenuButton variant="text" onPointerDown={onClick}>
+  <StyledMenuButton onPointerDown={onClick}>
     <MenuButtonInner>
       <MenuIcon />
       {text ? <MenuButtonText>{text}</MenuButtonText> : ""}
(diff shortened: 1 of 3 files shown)


jonkafton:
We probably don't want the min width on action and and icon only buttons (e.g. on the LearningResourceCards), though they could probably use separate treatment from the main Button component.

The action buttons have their own widths, so `minWidth: "auto"` in the `actionStyles` around L305 should fix.

image

vs

image

ChristopherChudzicki (changes requested):
See Jon's comment.

Please also take a look at the relevant storybook pages... e.g., 
- http://localhost:6006/?path=/story/smoot-design-button--action-buttons-showcase
- http://localhost:6006/?path=/story/smoot-design-button--buttons-showcase

just to make sure things look reasonable.

shanbady:
> See Jon's comment.
> 
> Please also take a look at the relevant storybook pages... e.g.,
> 
> * http://localhost:6006/?path=/story/smoot-design-button--action-buttons-showcase
> * http://localhost:6006/?path=/story/smoot-design-button--buttons-showcase
> 
> just to make sure things look reasonable.

thanks. I keep forgetting we have something like this

shanbady:
@jonkafton 

just pushed a fix with your suggestion

frontends/ol-components/src/components/Button/Button.tsx
```diff
@@ -94,6 +94,7 @@ const ButtonStyled = styled.button<ButtonStyleProps>((props) => {
       ":disabled": {
         cursor: "default",
       },
+      minWidth: "100px",
```
ChristopherChudzicki, line 97: Our tab component is based on `Button` (it uses tertiary buttons). So this increases the width, of tabs, too.

@steven-hatch Do we want tabs to have min-width 100px? I can see pros/cons

Screenshot 2024-06-27 at 4 24 31 PM

Screenshot 2024-06-27 at 4 24 42 PM
Screenshot 2024-06-27 at 4 25 01 PM

Screenshot 2024-06-27 at 4 25 18 PM

Really the only con is width on mobile screens.

## Separate issue

In the screenshots above, you can see the mobile menu button in top-left corner now has wrong width.

@shanbady I'm sorry this came up... `Button` should not have been used for that in the first place.

But, could you fix it so we don't regress? Probably that menu button should be a "Snowflake". I would just apply the relevant styles to it. See figma: https://www.figma.com/design/Eux3guSenAFVvNHGi1Y9Wm/MIT-Design-System?node-id=3852-51162&m=dev

The code is in `frontends/mit-open/src/page-components/Header/MenuButton.tsx`

ChristopherChudzicki (changes requested):
There's a tab question for @steven-hatch.

@shanbady ... requested change: Could you fix the MenuButton to not use `Button`? Probably should just be a snowflake `styled.button(...)`.

frontends/ol-components/src/components/Button/Button.tsx
shanbady: Should be addressed now @ChristopherChudzicki

frontends/mit-open/src/page-components/Header/MenuButton.tsx
```diff
@@ -23,22 +23,32 @@ const MenuButtonInner = styled.div({
...
-})
+  textAlign: "center",
+  display: "inline-flex",
+  justifyContent: "center",
+  alignItems: "center",
+  color: theme.palette.text.primary,
+  transition: `background ${theme.transitions.duration.short}ms`,
+  cursor: "pointer",
+  borderWidth: "1px",
+  borderColor: "currentcolor",
```
ChristopherChudzicki, line 41: Could you remove these? They do nothing with `borderStyle: "none"`.

*In `<Button />` they do have an effect, but then `variant="text"` set borderStyle to none, so they aren't relevant for that variant.*

ChristopherChudzicki (approved):
👍 Looks good. I would suggest removing the borderWith and borderColor on the menu button, as noted.

PR #1186, opened 2024-06-27
adding command to remove old tables

shanbady:
Removes old tables that were left behind by legacy apps

This pr modifies a migration that is failing on deploy by prepending it with sql to remove conflicting tables

### How can this be tested?
1. checkout main
2. roll the channels app migration to 0008 via ```python manage.py migrate channels 0008```
3. get into the db shell ```python manage.py dbshell``` and artificially create a channels table with a conflicting relation via 

CREATE TABLE channels_channel (
    widget_list_id     integer,
    name    varchar(40)
);
alter table channels_channel
  add constraint fk_widgets_w
  foreign key (widget_list_id) 
  references widgets_widgetlist (id);

4. attempt to re-run migrations and see that it fails with an error we see on rc deploy.
5. checkout this branch
6. re-run migrations and it should succeed

### Additional Context
@rhysyngsun is looking for other tables we need to add to this migration

### DIFF
--- a/channels/migrations/0009_rename_field_to_channel.py
+++ b/channels/migrations/0009_rename_field_to_channel.py
@@ -17,6 +17,7 @@ class Migration(migrations.Migration):
     ]
 
     operations = [
+        migrations.RunSQL("DROP TABLE IF EXISTS channels_channel CASCADE"),
         migrations.RenameModel(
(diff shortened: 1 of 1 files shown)


rhysyngsun (approved):
LGTM

PR #1189, opened 2024-06-28
Scroll results into view when paginating

jonkafton:

Closes https://github.com/mitodl/hq/issues/4741

Scrolls the top of the search results to the center of the page on pagination.

https://github.com/mitodl/mit-open/assets/939376/816e3a70-5be4-46d4-8566-dda64becd94d

### How can this be tested?

Navigate to the search page or a unit page and paginate through search results.
- http://od.odl.local:8063/search/
- http://od.odl.local:8063/c/unit/ocw/

### DIFF
--- a/frontends/mit-open/src/page-components/SearchDisplay/SearchDisplay.tsx
+++ b/frontends/mit-open/src/page-components/SearchDisplay/SearchDisplay.tsx
@@ -503,7 +503,7 @@ const SearchDisplay: React.FC<SearchDisplayProps> = ({
   setSearchParams,
 }) => {
   const [searchParams] = useSearchParams()
-  const scrollHook = useRef<HTMLInputElement>(null)
+  const scrollHook = useRef<HTMLDivElement>(null)
   const activeTab =
     TABS.find((t) => t.name === searchParams.get("tab")) ??
     TABS.find((t) => t.defaultTab) ??
--- a/frontends/mit-open/src/page-components/SearchDisplay/SearchDisplay.tsx
+++ b/frontends/mit-open/src/page-components/SearchDisplay/SearchDisplay.tsx
@@ -1,4 +1,4 @@
-import React, { useMemo } from "react"
+import React, { useMemo, useRef } from "react"
 import {
   styled,
   Pagination,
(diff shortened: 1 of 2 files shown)


frontends/mit-open/src/page-components/SearchDisplay/SearchDisplay.tsx
```diff
@@ -503,6 +503,7 @@ const SearchDisplay: React.FC<SearchDisplayProps> = ({
   setSearchParams,
 }) => {
   const [searchParams] = useSearchParams()
+  const scrollHook = useRef<HTMLInputElement>(null)
```
rhysyngsun, line 506: I think this should just be `HTMLElement` because it's a `div`

rhysyngsun (changes requested):
Functionality looks good, just one piece of feedback.

frontends/mit-open/src/page-components/SearchDisplay/SearchDisplay.tsx
jonkafton: Good spot, thanks.

rhysyngsun (approved):
LGTM

frontends/mit-open/src/page-components/SearchDisplay/SearchDisplay.tsx
```diff
@@ -680,7 +681,15 @@ const SearchDisplay: React.FC<SearchDisplayProps> = ({
...
                   page={page}
-                  onChange={(_, newPage) => setPage(newPage)}
+                  onChange={(_, newPage) => {
+                    setPage(newPage)
+                    setTimeout(() => {
+                      scrollHook.current?.scrollIntoView({
+                        block: "center",
+                        behavior: "smooth",
+                      })
+                    }, 0)
```
ChristopherChudzicki, line 691: Out of curiosity... why wrap this in setTimeout?

jonkafton: Despite some investigating, all I can say for sure is that it doesn't work without it. Likely some nuance around the dom element not being available during the old render cycle.

PR #1190, opened 2024-06-28
Drawer CSS fixes

ChristopherChudzicki:
Closes https://github.com/mitodl/hq/issues/4698

This PR:
- makes the resource drawer a little narrow to match Figma. This also fixes an issue with scrollbars. 
- makes drawer close button match figma color

Screenshot 2024-06-28 at 11 37 55 AM
Screenshot 2024-06-28 at 11 38 12 AM

### How can this be tested?
1. You may want to first find a resource on `main` that erroneously has horizontal scrollbars. Click on any resource card to open the resource drawer. On `main,` a resource will have a horizontal scrollbar if it has a vertical scrollbar, so find one with long content or zoom in to make the text bigger.
2. Check the same resource on this branch. The drawer should be a bit narrower and there will not be a horizontal scrollbar. Also, the close icon in upper right should be darker.

### DIFF
--- a/frontends/ol-components/src/components/LearningResourceExpanded/LearningResourceExpanded.tsx
+++ b/frontends/ol-components/src/components/LearningResourceExpanded/LearningResourceExpanded.tsx
@@ -10,7 +10,6 @@ import {
   formatDate,
   resourceThumbnailSrc,
   getReadableResourceType,
-  DEFAULT_RESOURCE_IMG,
 } from "ol-utilities"
 import type { EmbedlyConfig } from "ol-utilities"
 import { theme } from "../ThemeProvider/ThemeProvider"
@@ -27,10 +26,7 @@ const Container = styled.div<{ padTop?: boolean }>`
   padding: 18px 32px 160px;
   gap: 20px;
   ${({ padTop }) => (padTop ? "padding-top: 64px;" : "")}
-  width: 600px;
-  ${({ theme }) => theme.breakpoints.down("md")} {
-    width: 550px;
-  }
+  width: 516px;
   ${({ theme }) => theme.breakpoints.down("sm")} {
     width: auto;
   }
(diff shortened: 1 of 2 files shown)


frontends/ol-components/src/components/LearningResourceExpanded/LearningResourceExpanded.tsx
```diff
@@ -27,10 +26,7 @@ const Container = styled.div<{ padTop?: boolean }>`
   padding: 18px 32px 160px;
   gap: 20px;
   ${({ padTop }) => (padTop ? "padding-top: 64px;" : "")}
-  width: 600px;
-  ${({ theme }) => theme.breakpoints.down("md")} {
-    width: 550px;
-  }
```
ChristopherChudzicki, line 33: The HTML for the learning resource drawer was roughly
```
<div class="MuiDrawer-paper">
  <div class="our content" />
</div>
```
The paper has a max-width of 600px. The content was exactly 600px wide. The content should be narrower according to Figma. (516px). 

Additionally, when the drawer content was too tall to fit, a vertical scrollbar would appear on the `MuiDrawer-paper`, making the paper width = `600px from content + 15ishpx from scrollbar`. So a vertical scrollbar also caused a horizontal scrollbar to appear.

frontends/ol-components/src/components/LearningResourceExpanded/LearningResourceExpanded.tsx
```diff
@@ -171,10 +166,6 @@ const ImageSection: React.FC<{
         alt={resource?.image.alt ?? ""}
       />
     )
-  } else if (resource) {
-    return (
-      <Image src={DEFAULT_RESOURCE_IMG} aspect={config.width / config.height} />
-    )
```
jonkafton, line 177: We need this to display the default image for resources that have a `null` image (or we get the loading skeleton), unless I'm missing something?

jonkafton (approved):

ChristopherChudzicki:
[pre-commit.ci](http://pre-commit.ci/) run

ChristopherChudzicki:
pre-commit.ci run

PR #1195, opened 2024-06-28
Remove "Top picks" carousel if no results

jonkafton:

Closes https://github.com/mitodl/hq/issues/4711

Removes the "Top picks for you" if there are no results from the search.

Update: This change now applies the logic to all carousels.

### How can this be tested?

Navigate to Dashboard from the user menu or at /dashboard. Testable only by mocking the `/api/v1/learning_resources_search` response or overwriting in the code.

### Additional Context

This will render the loading state first, then removes if no results, expecting that there will normally be results.

I attempted a test to check that the section is not in the document, though found it unreliable to use `waitForElementToBeRemoved()` as that depends on the element always being there first. Also got unexpected results checking for the second render cycle (ie. `isLoading` now `false`) by check and then calling `waitForElementToBeRemoved()` or `expect(queryByText("Top picks for you")).not.toBeInTheDocument()`. Info [here](https://testing-library.com/docs/guide-disappearance#waiting-for-disappearance), open to suggestions.

If we need this in other places we may want to build this into the `ResourceCarousel` itself, or provide a carousel+container wrapper in the `DashboardPage` so we don't need to duplicate the `useLearningResourcesSearch()` call.


frontends/mit-open/src/pages/DashboardPage/DashboardPage.tsx
```diff
@@ -329,6 +333,28 @@ const UserListDetailsTab: React.FC<UserListDetailsTabProps> = (props) => {
   )
 }
 
+const TopPicksCarousel: React.FC = () => {
+  const { isLoading: isLoadingProfile, data: profile } = useProfileMeQuery()
+  const config = TopPicksCarouselConfig(profile)
+  const { data, isLoading } = useLearningResourcesSearch(
+    config[0].data
+      .params as LearningResourcesSearchApiLearningResourcesSearchRetrieveRequest,
```
jonkafton, line 341: Do we have a shorter equivalent?

ChristopherChudzicki: No 😭

I import `LearningResourcesSearchApiLearningResourcesSearchRetrieveRequest as SearchRequest` sometimes.

ChristopherChudzicki (changes requested):
Discussed w/ Jon in Slack: Carousel tabs should be visible while carousel is loading, else it looks weird

https://github.com/mitodl/mit-open/assets/9010790/d47ed35e-b852-4d67-8f55-b96bf00c911d

jonkafton:
> Discussed w/ Jon in Slack: Carousel tabs should be visible while carousel is loading, else it looks weird

I've pushed the fix. The inverse isn't ideal either, though certainly the lesser evil if we assume there will generally be content.


https://github.com/mitodl/mit-open/assets/939376/4a47a2eb-2e4c-42c5-b926-003a7ac35e61

ChristopherChudzicki (approved):

PR #1197, opened 2024-06-28
stop publishing github pages every pr

ChristopherChudzicki:
Toward https://github.com/mitodl/hq/issues/4760

This PR removes our existing gh-pages publishing step. You can see at https://mitodl.github.io/mit-open/ that it's not working, but it runs on every PR. I'd like to stop running it while I debug this.

### How can this be tested?
Tests should pass

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -319,36 +319,3 @@ jobs:
         with:
           name: playwright-report
           path: e2e_testing/playwright-report
-
-  publish-pages:
-    runs-on: ubuntu-latest
-
-    environment:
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -319,36 +319,3 @@ jobs:
         with:
           name: playwright-report
           path: e2e_testing/playwright-report
-
-  publish-pages:
```
ChristopherChudzicki, line 323: I left in-place the step that builds storybook above.

rhysyngsun (approved):
LGTM

PR #1200, opened 2024-06-28
Fix storybook github pages publishing

ChristopherChudzicki:
- closes https://github.com/mitodl/hq/issues/4760

This PR:

1. fixes publishing storybook on github pages. It wasn't working before
    - it will only publish when manually run or when there are pushes to main.
2. Removes our `github-pages` react app.
    - At one point in the past, we had a react app + storybook deploying to github pages. The react app didn't have much besides a link to storybook, so we've just removed it.
3. Renames a few storybook stories.
    - Previously we had two "Categories": `ol-components/storyname` and `smoot-design/storyname`. I've moved most things under `ol-components/` to `smoot-design/` (this only affects the storybook, not the workspace name) and renamed `ol-components` category to "`old`"

### How can this be tested?
1. Run `yarn storybook` locally and compare with https://mitodl.github.io/mit-open/. They should be the same.


frontends/mit-open/package.json
```diff
@@ -42,6 +43,7 @@
     "html-webpack-plugin": "^5.6.0",
     "mini-css-extract-plugin": "^2.6.1",
     "ol-test-utilities": "0.0.0",
+    "react-refresh": "^0.14.2",
```
ChristopherChudzicki, line 46: This is a peer-dependency of another package we use, `@pmmmwh/react-refresh-webpack-plugin`. Previously we were not installing the peer dep explicitly (it was added by react-scripts, I believe). Now that the `github-pages` workspace is removed, we needed to explicitly install it.

.github/workflows/publish_pages.yml
```diff
@@ -0,0 +1,54 @@
+on:
+  # Runs on pushes targeting the default branch
+  push:
+    branches: [$default-branch, cc/publish-pages]
```
jonkafton, line 4: Remove for cleanup

jonkafton, line 4: Remove for cleanup

.github/workflows/publish_pages.yml
```diff
@@ -0,0 +1,54 @@
...
+
+      - name: Build Storybook
+        run: yarn workspace mit-open build-storybook
+
+      - name: Upload artifact
+        uses: actions/upload-pages-artifact@v3
+        with:
+          path: ./frontends/mit-open/storybook-static
+
+  # Deploy job
```
jonkafton, line 34: These comments aren't adding much. Perhaps not doing any harm; I'd remove them.

jonkafton:
PR looks good. Now we have static hosting for the front end set up, let's pick up the conversation of publishing Storybook alongside it for branch PRs and environments.

For the Storybook renaming, do you think it's worth making the distinction between components that form part of the design language - the Smoot Design components - and components that are just UI devices without any particular design - e.g. PlainList, Popover, the RoutedDrawer (arguably)?

.github/workflows/publish_pages.yml
jonkafton, line 1: Small nit - the other workflow filenames are hyphened.

jonkafton:
Might be worth updating the project README (last line) to say it's now publishing Storybook there.

ChristopherChudzicki:
@jonkafton 
> For the Storybook renaming, do you think it's worth making the distinction between components that form part of the design language - the Smoot Design components - and components that are just UI devices without any particular design - e.g. PlainList, Popover, the RoutedDrawer (arguably)?

I do think that this is a good idea and worthwhile, but mostly I want to get this deployed and shareable with Steve/Simone. I suspect we will do some further reorganizing / renaming after they've looked at it a bit. I'd like to hold off on more reorganizing till then.

jonkafton (approved):

PR #1208, opened 2024-07-01
Privacy policy updates

jonkafton:

Closes https://github.com/mitodl/hq/issues/4764

Updates copy in line with https://docs.google.com/document/d/1Zv1jECh9EmJGin_srNWYU9LvqRuYijIVtYnla9L54uk/edit.

Uses `SITE_NAME` from the environment for previus occurences of "MIT Open" or "MITx Online".

### How can this be tested?

Navigate to `/privacy` to confirm changes.


frontends/mit-open/src/pages/PrivacyPage/PrivacyPage.tsx
```diff
@@ -212,89 +210,36 @@ const PrivacyPage: React.FC = () => {
           </UnorderedList>
           <BodyText variant="body1">
             If you have concerns about any of these purposes, or how we
-            communicate with you, please contact us at
-            mitxonline-support@mit.edu. We will always respect a request by you
-            to stop processing your personal information (subject to our legal
-            obligations).
+            communicate with you, please contact us at mitopen-support@mit.edu.
```
pdpinch, line 213: Can we use the value of the setting `MITOPEN_SUPPORT_EMAIL`?

See https://github.com/mitodl/ol-infrastructure/blob/d8696e6e222d52240d1e3c2a5caf3b2f71625852/src/ol_infrastructure/applications/mitopen/Pulumi.applications.mitopen.Production.yaml#L44

ChristopherChudzicki (changes requested):
Looks good, other than Peter's request.

frontends/mit-open/src/pages/PrivacyPage/PrivacyPage.tsx
jonkafton: I've added it to the front end build config so that we can configure it on CI, but to use the infrastructure config we'd need the backend to return it on an API request.

Could be worth doing if it's likely to change much or if there are likely to be other cases where we need to dynamically update public site config or content.

ChristopherChudzicki (approved):
👍

PR #1213, opened 2024-07-01
make primary buttons shadowy, remove edge=none

ChristopherChudzicki:
Closes https://github.com/mitodl/hq/issues/4700

Makes primary buttons (and only primary buttons) have a shadow to match figma.

App *There aren't too many primary buttons in the app so far.*
Screenshot 2024-07-01 at 4 26 42 PM

Storybook:
Screenshot 2024-07-01 at 4 26 35 PM

### How can this be tested?
1. Run storybook (should be on 6006, if not, `yarn storybook`) and check:
    - http://localhost:6006/?path=/story/smoot-design-button--buttons-showcase

Buttons should be unchanged, except primary buttons should have a shadow AND only in their default state (no shadow on hover).

### DIFF
--- a/frontends/ol-components/src/components/Button/Button.stories.tsx
+++ b/frontends/ol-components/src/components/Button/Button.stories.tsx
@@ -42,7 +42,7 @@ const meta: Meta<typeof Button> = {
       control: { type: "select" },
     },
     edge: {
-      options: ["circular", "rounded", "none"],
+      options: ["circular", "rounded"],
       control: { type: "select" },
     },
     startIcon: {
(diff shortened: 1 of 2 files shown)


frontends/ol-components/src/components/Button/Button.stories.tsx
```diff
@@ -42,7 +42,7 @@ const meta: Meta<typeof Button> = {
       control: { type: "select" },
     },
     edge: {
-      options: ["circular", "rounded", "none"],
+      options: ["circular", "rounded"],
```
ChristopherChudzicki, line 45: edge="none" is not in figma; it was added for resource cards, but that is handled a little differently now (changed in #1180 )

frontends/ol-components/src/components/Button/Button.tsx
```diff
@@ -110,17 +110,22 @@ const ButtonStyled = styled.button<ButtonStyleProps>((props) => {
...
       },
       ":disabled": {
         backgroundColor: colors.silverGray,
+        boxShadow: "none",
       },
     },
     hasBorder && {
       backgroundColor: "transparent",
       borderColor: "currentcolor",
-      borderStyle: variant === "secondary" ? "solid" : "none",
```
ChristopherChudzicki, line 123: Previously, `const hasBorder = variant === "secondary" || variant === "text"`. But that was weird because `text` variant doesn't actually have a border. It was just set that way to pick up `backgroundColor: transparent`.

Applying `backgroundColor: transparent` to the text variant separately is simpler / more understandable, imo.

frontends/ol-components/src/components/LearningResourceCard/LearningResourceCard.stories.tsx
```diff
@@ -28,9 +28,10 @@ const meta: Meta<typeof LearningResourceCard> = {
       options: ["Loading", "Without Image", ...Object.values(ResourceTypeEnum)],
       mapping: {
         Loading: undefined,
-        "Without Image": makeResource({
+        "Without Image": {
+          ...makeResource(),
           image: null,
```
ChristopherChudzicki, line 33: This change is unrelated, but was something I did while opening 
- https://github.com/mitodl/hq/issues/4774

Decided to keep it.

abeglova (approved):
👍

PR #1217, opened 2024-07-02
[for discussion] try formik for onboarding

ChristopherChudzicki:

- [ ] Desktop screenshots

### DIFF
--- a/frontends/mit-open/src/pages/OnboardingPage/OnboardingPage.tsx
+++ b/frontends/mit-open/src/pages/OnboardingPage/OnboardingPage.tsx
@@ -1,6 +1,5 @@
-import React from "react"
+import React, { useId, useMemo } from "react"
 import { useNavigate } from "react-router-dom"
-import isEqual from "lodash/isEqual"
 import range from "lodash/range"
 import {
   styled,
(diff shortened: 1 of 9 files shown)


frontends/mit-open/src/pages/OnboardingPage/OnboardingPage.tsx
```diff
@@ -112,45 +115,186 @@ const Label = styled.div({
...
+    topics?.results?.map((topic) => ({
+      label: topic.name,
+      value: topic.id.toString(),
+    })) ?? []
+
+  if (!profile) {
     return null
   }
 
+  const pages = [
```
ChristopherChudzicki, line 191: `pages` could be a unction I guess, `getPage`.

Main changes:
- Single <form />` for onboarding.
- "Next" button is a submit button
- No more `React.useState<PatchedProfileRequest>({})`, since formik handles the updates
- Replaced all fields with plain `CheckboxChoiceBoxField`, `RadioChoiceBoxField`, `SimpleSelectField` instances
    - As long as all checkboxes share a common name, formik handles bundling the value into an array... nice.

This was a little bit quick and dirty. My thought is that onboarding and dashboard would share some of this:
- share constants like `GOALS_CHOICES` .... but not fields, since dashboard uses different fields for some of these
- maybe share a hook that handles initializing formik? Not that it is too complicated.

frontends/mit-open/src/pages/OnboardingPage/OnboardingPage.tsx
```diff
@@ -112,45 +115,186 @@ const Label = styled.div({
   margin: "0 0 40px",
 })
 
+const profileSchema = yup.object().shape({
```
ChristopherChudzicki, line 118: `profileSchema` isn't really doing much right now. I mostly added it out of habit.

I also made it a bit loose, though with something like
```
  goals: yup
    .array()
    .of(yup.string().oneOf(GOALS_CHOICES.map((choice) => choice.value))),
```
it will validate the membership and get the TS correct, though I'm not sure there's much point for checkboxes / radio boxes. / select.

ChristopherChudzicki:
This got incorporated into #1266

PR #1218, opened 2024-07-02
Data fixtures app for loading static fixtures

shanbady:

Closes https://github.com/mitodl/hq/issues/4687

This PR adds a "data fixtures" app that exists for the purpose of updating/loading static fixtures in both local and prod environments. for discussion around this reference the discussion topic https://github.com/mitodl/hq/discussions/4747

The idea is that this app will run its data migrations for local development as well as when deploying but not when running tests

In addition to adding the app, this PR also add initial fixture data for testimonials (copied from production)

### How can this be tested?
1. Checkout this branch
2. do a docker compose down/up and note that fixtures for attestation are loaded in via django admin. alernatively if you want to avoid doing a docker down/up you can run ```RUN_DATA_MIGRATIONS=true python manage.py migrate```


data_fixtures/models.py
```diff
@@ -0,0 +1 @@
+# Create your models here.
```
gumaerc, line 1: Are these files (models, tests, views) necessary for Django to recognize `data_fixtures` as an app? If not, can we remove them?

gumaerc (approved):
👍 

I simply did `docker compose stop / docker compose up` and my database was automatically populated with attestations. I just have a basic non-blocking question:

data_fixtures/models.py
shanbady: its not necessary. I went back and forth on this a bit. my main reason for leaving it was to serve as a reminder that this a django app but I think the apps.py is probably enough. removed the extra files

PR #1221, opened 2024-07-02
Webpack build config loads .env files for running outside of Docker

jonkafton:

Uses the `dotenv` package to read the .env files when running outside of Docker. Inside Docker, they are supplied on the  Docker Compose run config.

This is to accommodate the recent and excellent cleanup of env var handling: https://github.com/mitodl/mit-open/pull/1179

Checks that the `ENVIRONMENT` is `"local"` before reading in the .env files.

Also sets the Webpack dev server to use host 0.0.0.0.

### How can this be tested?

This also sets the `ENVIRONMENT` to `docker` for `yarn run watch:docker`. It was set to this at some point previously, though if we see issues, that would be the most likely cause.

### DIFF
--- a/frontends/mit-open/package.json
+++ b/frontends/mit-open/package.json
@@ -9,7 +9,7 @@
   },
   "scripts": {
     "watch": "NODE_ENV=development ENVIRONMENT=local webpack serve",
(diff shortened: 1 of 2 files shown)


ChristopherChudzicki (approved):
👍

PR #1222, opened 2024-07-03
resource type for learning materials

abeglova:
closes https://github.com/mitodl/hq/issues/4736

The "Resource Type" facet is displayed, but only if the selected tab is "Learning Material"

Screenshot 2024-07-03 at 10 50 12 AM

### How can this be tested?
Go to http://localhost:8062/search. There should not be a Resource Type facet displayed. 
Click on the "Learning Materials" tab. There should not be a Resource Type facet displayed and you should be able to select a type of learning material

Go to http://localhost:8062/c/topic/mathematics
Again there should be a Resource Type facet only if Learning Material is selected

### DIFF
--- a/frontends/mit-open/src/pages/SearchPage/SearchPage.tsx
+++ b/frontends/mit-open/src/pages/SearchPage/SearchPage.tsx
@@ -54,10 +54,13 @@ const SearchField = styled(SearchInput)`
   }
 `
 
+const LEARNING_MATERIAL = "learning_material"
+
 export const getFacetManifest = (
   offerors: Record<string, LearningResourceOfferor>,
+  resourceCategory: string | null,
 ) => {
-  return [
+  const mainfest = [
     {
       type: "group",
       name: "free",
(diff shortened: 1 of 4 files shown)


frontends/mit-open/src/pages/SearchPage/SearchPage.tsx
```diff
@@ -113,9 +128,17 @@ export const getFacetManifest = (
...
+  if (resourceCategory !== LEARNING_MATERIAL) {
+    mainfest.splice(1, 1)
+  }
+
+  return mainfest
 }
 
 const facetNames = [
+  "resource_type",
   "resource_category",
```
ChristopherChudzicki, line 142: I don't think `resource_category` should be a facetName, btw.
- if it is, then "Clear All" will reset the tab

I discussed this once with Steve and we decided "Clear all" should not clear the tab, since it's associated with the left-hand facets / filters.

If you change, resource_category would needed to be added o the aggregation separately.

ChristopherChudzicki (approved):
👍 Nice. I left one comment, though we've had the "Clear all" resets tab behavior for at least a few weeks, so it doesn't necessarily need to change here.

ChristopherChudzicki (approved):

PR #1236, opened 2024-07-09
Fix logout view

rhysyngsun:

Fixes https://github.com/mitodl/hq/issues/4831

- Updates the code to ensure the redirect url is absolute again.
- Removes the 404 error in favor of a redirect so this url behaves in an idempotent manner.

NOTE: I tried to add some tests for this, but they fail because we're using logout via a GET request, which is deprecated and removed in django 5+. I tried to fix *that* so we could just be done with logout for a good while but that spiraled into CSRF errors and it's not worth that level of effort at this time.

### DIFF
--- a/authentication/views.py
+++ b/authentication/views.py
@@ -4,7 +4,6 @@ from urllib.parse import urlencode
 
 from django.conf import settings
 from django.contrib.auth import views
-from django.http import Http404
 from django.shortcuts import redirect
 from social_django.utils import load_strategy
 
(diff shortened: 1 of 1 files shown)


authentication/views.py
```diff
@@ -37,25 +36,30 @@ def _keycloak_logout_url(self, user):
...
+            f"{settings.KEYCLOAK_REALM_NAME}/protocol/openid-connect/logout"
+            f"?{qs}"
+        )
 
     def get(
         self,
         request,
         *args,  # noqa: ARG002
         **kwargs,  # noqa: ARG002
-    ):  # pylint:disable=unused-argument
```
shanbady, line 50: I'm not fully understanding this change. if we absolutely need to do this ahead of django 5 could we instead define a separate post() method that does the same thing?

also worth noting - on the frontend we appear to be logging out via GET on /logout (admin has its own logout flow that is done via post)

authentication/views.py
```diff
@@ -37,25 +36,30 @@ def _keycloak_logout_url(self, user):
         qs = urlencode(
             {
                 "id_token_hint": id_token,
-                "post_logout_redirect_uri": settings.LOGOUT_REDIRECT_URL,
+                "post_logout_redirect_uri": self.request.build_absolute_uri(
```
shanbady, line 39: I can see this causing issues on a live server since self.request.build_absolute_uri will end up generating a different url depending on  if its hit from api.* (which it might if done via js) or the main hostname. I think we need to explicitly generate this from the APP_BASE_URL since redirecting back to api.mitopen.odl.mit.edu leaves the user in a broken state - some functionality does not work as intended from api.*

shanbady (changes requested):
There are a few issues i've noted which i think we should address before merging. I'm working with chris to get the infra side of things sorted out

authentication/views.py
rhysyngsun: That was an accidental commit of me trying to make POST work, I remove the changes.

authentication/views.py
rhysyngsun: The API should only ever be served from `api.*` domains. I updated the docstring to reflect the redirect path that the browser will take as the user is logged out.

shanbady (approved):
this should work using the /app route so not making this a blocker but i'll note my main concerns with using self.request.build_absolute_uri:
1. it wont work when directly accessing from /logout (has to be done via js) - seems like we are unnecessarily restricting ourselves and sometimes i do go directly to /logout instead of clicking the button. 
2. Its not clear to me what the /app endpoints supports and seems like we are necessarily making extra roundtrips: mitopen -> api.mitopen/logout -> 
3.  if we need to all of a sudden support redirects elsewhere or back to where the user previously this has the potential to get messy (we'd have to keep track of and pass some "next_url" through 3 request hops)

PR #1247, opened 2024-07-10
reindexing changes

abeglova:
closes https://github.com/mitodl/hq/issues/4798

If a recreate_index jobs is currently started before a different recreate_index job is finished, the [create_backing_index](https://github.com/mitodl/mit-open/blob/main/learning_resources_search/indexing_api.py#L446) function creates a new index and removes the reindexing tag from the first reindexing index that was created. The first reindexing index is then never deleted and left orphaned. 

This PR adds a check for an existing reindexing tag before starting a new reindexing job. In the default state, the job fails if there is an existing reindexing tag. In case something went wrong on a previous run and a current reindexing tag shouldn't exist, there is a --remove_existing_reindexing_tags option to force removing the existing reindexing tag

In addition, this pr deletes any orphaned indexes every time the recreate_index job runs

### How can this be tested?
From the command line run 
```
from learning_resources_search.indexing_api import create_backing_index
create_backing_index("program") 
create_backing_index("program") 
```
This creates an index with the reindexing tag and another orphaned index for programs

Run 
docker-compose run web ./manage.py recreate_index --program
You should see an error saying "Reindexing in progress. Reindexing indexes already exist"

Run 
docker-compose run web ./manage.py recreate_index --program --remove_existing_reindexing_tags

and the reindexing should finish

From the command line run 
```
from learning_resources_search.indexing_api import get_conn
conn = get_conn()
conn.indices.get_alias(index="*")
```

Check that there is only one index with "program" in the name

Additionally run docker-compose run web ./manage.py recreate_index --all and verify that everything looks normal

### DIFF
--- a/learning_resources_search/management/commands/recreate_index.py
+++ b/learning_resources_search/management/commands/recreate_index.py
@@ -3,6 +3,7 @@
 from django.core.management.base import BaseCommand, CommandError
 
 from learning_resources_search.constants import ALL_INDEX_TYPES
+from learning_resources_search.indexing_api import get_existing_reindexing_indexes
 from learning_resources_search.tasks import start_recreate_index
 from main.utils import now_in_utc
 
@@ -13,6 +14,13 @@ class Command(BaseCommand):
     help = "Recreate opensearch index"
 
     def add_arguments(self, parser):
+        parser.add_argument(
+            "--remove_existing_reindexing_tags",
+            dest="remove_existing_reindexing_tags",
+            action="store_true",
+            help="Overwrite any existing reindexing tags and remove those indexes",
+        )
+
         parser.add_argument(
             "--all", dest="all", action="store_true", help="Recreate all indexes"
         )
@@ -28,11 +36,9 @@ class Command(BaseCommand):
 
     def handle(self, *args, **options):  # noqa: ARG002
         """Index all LEARNING_RESOURCE_TYPES"""
+        remove_existing_reindexing_tags = options["remove_existing_reindexing_tags"]
         if options["all"]:
-            task = start_recreate_index.delay(list(ALL_INDEX_TYPES))
-            self.stdout.write(
-                f"Started celery task {task} to index content for all indexes"
-            )
+            indexes_to_update = list(ALL_INDEX_TYPES)
         else:
             indexes_to_update = list(
                 filter(lambda object_type: options[object_type], ALL_INDEX_TYPES)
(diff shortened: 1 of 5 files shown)


learning_resources_search/tasks.py
```diff
@@ -458,11 +458,26 @@ def wrap_retry_exception(*exception_classes):
...
 
 @app.task(bind=True)
-def start_recreate_index(self, indexes):
+def start_recreate_index(self, indexes, remove_existing_reindexing_tags):
     """
     Wipe and recreate index and mapping, and index all items.
     """
     try:
+        if not remove_existing_reindexing_tags:
+            existing_reindexing_indexes = api.get_existing_reindexing_indexes(indexes)
```
abeglova, line 467: We need to check both here and in the management command in case start_recreate_index is stuck behind other jobs and another indexing job is started while this task is in the queue. We could just have the check here but it's nicer for the user to know about the existing reindexing indexes right away in the normal case

learning_resources_search/tasks.py
```diff
@@ -768,7 +783,9 @@ def get_update_learning_resource_tasks(resource_type):
     ]
 
 
-@app.task(autoretry_for=(RetryError,), retry_backoff=True, rate_limit="600/m")
+@app.task(
+    acks_late=True, autoretry_for=(RetryError,), retry_backoff=True, rate_limit="600/m"
```
abeglova, line 787: I added acks_late=True here just in case, although I don't think this job failing is how orphaned indexes are generally created

learning_resources_search/management/commands/recreate_index.py
```diff
@@ -44,12 +50,24 @@ def handle(self, *args, **options):  # noqa: ARG002
...
+        if not remove_existing_reindexing_tags:
+            existing_reindexing_indexes = get_existing_reindexing_indexes(
+                indexes_to_update
             )
+            if existing_reindexing_indexes:
+                self.stdout.write(
+                    f"Reindexing in progress. Reindexing indexes already exist:"
+                    f" {', '.join(existing_reindexing_indexes)}"
+                )
+                return
```
mbertrand, line 62: If you leave this new conditional block out, and `remove_existing_reindexing_tags` is False, the user will still be notified of the problem, because any errors that occur when running the task are shown later below:

> mattbert@Matts-MacBook-Pro mit-open % docker compose exec web ./manage.py recreate_index --course
Started celery task 722af285-d989-450f-9eb8-4a9070f31f1a to index content for the following indexes: ['course']
Waiting on task...
CommandError: Recreate index errored: Reindexing in progress. Reindexing indexes already exist: discussions_local_course_67171d570c7e4386afd5b4a957669c2e

mbertrand (approved):
Works great, 👍 , just some minor nitpicks.

learning_resources_search/tasks.py
```diff
@@ -458,11 +458,26 @@ def wrap_retry_exception(*exception_classes):
...
+            if existing_reindexing_indexes:
+                error = (
+                    f"Reindexing in progress. Reindexing indexes already exist: "
+                    f"{', '.join(existing_reindexing_indexes)}"
+                )
+                log.exception(error)
+                return error
+
+        api.delete_orphaned_indexes(
+            indexes, delete_reindexing_tags=remove_existing_reindexing_tags
```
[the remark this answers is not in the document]
abeglova: I agree the conditional is not strictly necessary .

PR #1250, opened 2024-07-10
Addresses performance issues for the topics API

rhysyngsun:

Closes https://github.com/mitodl/hq/issues/4828

This addresses a performance issue created by N+1 query issues due to a hardcoded query in the serializer. I took the approach of leaving in the existing code as a fallback but otherwise annotating the query for better performance. I also put in place a query count assertion on testing this API

I saw performance improve at least an order of magnitude and it'll probably run better in deployed environments:

Before:
Screenshot 2024-07-10 at 16-23-21 Silky - Requests
After:
Screenshot 2024-07-11 at 13-08-36 Silky - Requests

### How can this be tested?

- Hit the `/api/v1/topics/` api on the `main` branch and observe that it's slow (seconds).
- Hit the same API on this branch and it should be much faster (milliseconds).
- You should be able to go to `/silk` and it requires you to be a logged in superuser to access.


gumaerc (approved):
👍 LGTM

I didn't get quite as good performance as you're showing here, but I did see a marked improvement. Since `django-silk` isn't installed on `main`, I did some basic profiling using the network tab in Chrome inspector. On `main` I was seeing response times up to and over 1 second, and on this branch I saw response times on the order of ~100ms for the topics endpoint.

image

PR #1251, opened 2024-07-10
Condensed list card components for user lists

jonkafton:

Closes https://github.com/mitodl/hq/issues/4790

image

----

image

----

image

----

image

### How can this be tested?

While logged in, navigate to the dashboard "My lists" screen at `/dashboard/#my-lists` and select a list. You should have a "My Favorites" list, or can create a new list if necessary.

If there are no items in the list, from the homepage or search page, use the card action buttons to add items to you list.

Check the list items render according to designs at desktop and mobile screen widths.

Check variations in Storybook at http://localhost:6006/?path=/story/smoot-design-cards-learningresourcelistcardcondensed--free-course

### Additional Context

This adds `Card/ListCardCondensed` base slots template and a `LearningResourceCard/LearningResourceListCardCondensed` to the ol-components library. These import heavily from their sibling `Card/ListCardCondensed` and `LearningResourceCard/LearningResourceListCard`, overriding styles only where they deviate. Relative to the existing resource list cards:
- Tighter margins and padding on the card and text contents.
- Smaller font for title.
- Action buttons have a larger gap.
- Mobile action buttons keep their border (they do not change at mobile breakpoint).
- Top left resource type is anchored top (flex-start).
- Height is not constrained. The card grows in height as longer titles wrap (I set a 4 line clamp which should rarely be hit).

Drag and sorting is to be done in a separate task.

### DIFF
--- a/frontends/mit-open/src/page-components/ResourceCard/ResourceCard.tsx
+++ b/frontends/mit-open/src/page-components/ResourceCard/ResourceCard.tsx
@@ -122,7 +122,7 @@ type ResourceListCardProps = Omit<
 const ResourceListCard: React.FC<ResourceListCardProps> = ({
   resource,
   condensed,
-  ...others
+  ...props
 }) => {
   const {
     getDrawerHref,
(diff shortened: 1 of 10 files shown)


gumaerc (approved):
👍 LGTM!

PR #1253, opened 2024-07-11
adjust department names

gumaerc:
Closes https://github.com/mitodl/hq/issues/4838

This PR updates 4 department names as noted in the above issue. `update_date` and `created_date` info was also stripped from the JSON fixtures, as it's irrelevant and confuses the `update_or_create` function that the info is passed to. The utility functions that load this data were previously matching their `update_or_create` calls based on the `name` property of the school or department, which is incorrect. They should instead be matching on their ID columns; `department_id` for departments and `id` for schools. This ensures that you can update the name and not get integrity errors like `django.db.utils.IntegrityError: duplicate key value violates unique constraint "learning_resources_learningresourcedepartmentnew_pkey"`. Another issue here was the `backpopulate_resource_channels` command, which runs `update_or_create` based on a slugified version of the department name for departments. Since the name is changing here, if the command was run as-is a new channel would be created instead of updating the existing one. This was changed to update channel based on `search_filter`, since it is based on `department_id` which is actually constant.

### How can this be tested?
 - Spin up `mit-open` on this branch
 - Run `docker compose exec web ./manage.py update_departments_schools`
 - Run `docker compose exec web ./manage.py backpopulate_resource_channels --department --overwrite`
 - Visit http://localhost:8062/departments
 - Ensure that the changed department names are listed and match what they should be as indicated in the above issue
 - Click on the changed departments and ensure that they lead to a channel page, and that the channel page has been updated with the new department name


channels/plugins.py
```diff
@@ -73,11 +73,12 @@ def department_upserted(self, department, overwrite):
             channel = None
         elif department.school and (overwrite or not channel):
             channel, _ = Channel.objects.update_or_create(
+                search_filter=f"department={department.department_id}",
                 name=slugify(department.name),
```
mbertrand, line 77: Remove this line or there will be channels for both the old and new name

Might be good to add a unit test for this too

mbertrand (changes requested):

mbertrand (approved):
👍

PR #1259, opened 2024-07-11
Make static/hash.txt served again

rhysyngsun:

N/A

This is a follow up to #1179, which removed the serving of the frontend via nginx - the side effect of this is `hash.txt` was no longer being served so our release-script bot can't detect when deployments happen anymore

### How do I test it?

Run `echo "abc" > static/hash.txt` and then `docker compose up`. You should then be able to see this dummy file at http://api.open.odl.local:8063/static/hash.txt

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -91,6 +91,7 @@ codekit-config.json
 
 # Django static
 staticfiles/
+static/hash.txt
 
 .venv
 release-notes-checklist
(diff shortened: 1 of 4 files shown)


shanbady (approved):
Appears to work. one thing to note is that the static directory "static" was not created when checking out this branch ( had to manually create it)

PR #1264, opened 2024-07-12
Take is_enrollable attribute into account for publish status of edx resources

mbertrand:
Closes https://github.com/mitodl/hq/issues/4873

For every edx run, sets the `published` value based on both `status` ("published" or not) and `is_enrollable` (must be true to be published).  The course publish status depends on whether any of its runs are published.

### How can this be tested?
- Set required backend .env variables for `EDX_API_*` and `OLL_API_*` based on RC values.
- If you have already populated MIT edX and mitxonline courses before, you might have 2 published courses returned by this search: http://open.odl.local:8062/search?platform=mitx&q=%22Linear+Elastic+Behavior%22 - one from MITx Online and another from edX.
- Run `./manage.py backpopulate_mit_edx_data` and `./manage.py backpopulate_oll_data`
- Once complete, The above search should return only one result, from MITx Online, if you've ingested those courses recently.  Otherwise it should return no results. 
- You should get very similar counts by running these commands in a shell:
  ```
    LearningResource.objects.filter(etl_source="mit_edx", published=True).count()
    
    Out[2]: 111
    
    LearningResource.objects.filter(etl_source="oll", published=True).count()

    Out[2]: 61
  ```

### DIFF
--- a/learning_resources/etl/openedx.py
+++ b/learning_resources/etl/openedx.py
@@ -220,8 +220,10 @@ def _transform_course_run(config, course_run, course_last_modified, marketing_ur
         "start_date": course_run.get("start") or course_run.get("enrollment_start"),
         "end_date": course_run.get("end"),
         "last_modified": last_modified,
-        "published": course_run.get("status", "") == "published"
-        and course_run.get("is_enrollable", False),
+        "published": (
+            course_run.get("status", "") == "published"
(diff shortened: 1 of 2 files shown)


ChristopherChudzicki (approved):
👍 Removed 21 un-enrollable courses for me!

PR #1266, opened 2024-07-12
refactor profile and onboarding

gumaerc:
Closes https://github.com/mitodl/hq/issues/4860
Closes https://github.com/mitodl/hq/issues/4804

This PR refactors `OnboardingPage` and `ProfileEditForm` to use the `formik` library for form management. As part of this process, a number of other changes were made:

 - Form controls swapped out for more generic components from `ol-components`
 - Various spacings, styles and copy updated to match the designs
 - Learning format updated to be multiple choice

image
image
image
image
image

image
image
image
image
image

image
image

### How can this be tested?
 - Spin up this branch of `mit-open`
 - Visit http://localhost:8063/onboarding
 - Follow the onboarding flow and ensure that the styles and copy match the designs
 - Visit http://localhost:8063/dashboard/#profile
 - Ensure that the styles and copy match the designs
 - Change some info on your profile and verify that the "Update" button becomes available
 - Save your profile, refresh the page and make sure the changes have persisted

### Additional Context
You may notice that the `Radio` styles don't exactly match the designs. Between this and `Checkbox`, some more extensive refactoring needs to be done that is outside the scope of this PR.


frontends/ol-components/src/components/RadioChoiceField/RadioChoiceField.tsx
```diff
@@ -44,8 +55,8 @@ const RadioChoiceField: React.FC<RadioChoiceFieldProps> = ({
   const labelId = useId()
   return (
     <FormControl className={className}>
-      <FormLabel id={labelId}>{label}</FormLabel>
-      <RadioGroup
+      <Label id={labelId}>{label}</Label>
+      <RadioGroupStyled
```
ChristopherChudzicki, line 59: Two notes on styling 

**First**
Screenshot 2024-07-16 at 12 06 41 PM

- The color here is coming from[ `theme.palette.primary.main`](https://github.com/mitodl/mit-open/blob/77752408e7b7d45f2712e7ca63ff771f74bcbe0a/frontends/ol-components/src/components/ThemeProvider/ThemeProvider.tsx#L47-L52)
    - We could main palette color to red. I briefly changed it to magenta (`rgba(255,0,255)`) just to make the effects more visible. I believe the only things using this color right now are radio buttons and Mui loading spinners.
    - Our color usage doesn't really align well with MUI, though. (Example: MUI uses `primary.main` for the color of buttons, radio groups, etc. But our buttons and our radio groups are different colors. *We don't use MUI's button component*.)
- Could also add  something like 
    ```
      ".MuiRadio-root.Mui-checked svg": {
        color: theme.custom.colors.red,
      },
    ```

**Second**
Screenshot 2024-07-16 at 12 07 55 PM

The color here is coming from use of mui's FormLabel inside mui's FormControl. I don't think we want this styling. Or, if we do, we should have it everywhere (out of the scope of this PR).

*Aside: I used to not really understand the point of FormControl. Now I see what it does better.*

**Suggestion:** Remove FormControl and/or FormLabel (replace with div and/or label).

frontends/mit-open/src/page-components/Profile/constants.tsx
ChristopherChudzicki, line 1: Seems a little odd to have have these constants in `/page-components` now. Maybe move to `common`? Could also be `.ts` instead of `.tsx`

frontends/ol-components/src/components/Checkbox/CheckboxChoiceField.tsx
```diff
@@ -0,0 +1,75 @@
...
+  values?: string[]
+  onChange?: CheckboxProps["onChange"]
+  row?: boolean
+  className?: string
+}
+
+const Container = styled.div({
+  display: "flex",
+  gap: "32px",
+})
```
ChristopherChudzicki, line 21: Re the `row` prop: This isn't a big deal right now by any means, but the breakpoint hooks will cause us trouble if we ever render react server-side to send fuller HTML pages to the client. For this reason, I try to do the breakpoint stuff in css.

Here, we could do
```tsx
import type { Breakpoint } from "@mui/system"

const Container = styled.div<{columnBelow?: Breakpoint}>(({columnBelow})=>[
  {
    display: "flex",
    gap: "32px",
    flexDirection: "row"
  },
  columnBelow && {
    [theme.breakpoints.down(columnBelow)] {
      flexDirection: "column"
    }
  }
])
```

ChristopherChudzicki (changes requested):
This is looking great. I noticed a few minor issues, mostly styling, that I commented about.

frontends/mit-open/src/page-components/Profile/constants.tsx
gumaerc: It needs to be `.tsx` because `EDUCATION_LEVEL_OPTIONS` has a `label` property with HTML in it.

frontends/ol-components/src/components/RadioChoiceField/RadioChoiceField.tsx
gumaerc: I was able to tweak the `Checkbox` styles to match the design pretty closely, but the `RadioChoiceField` control just uses the MUI `Radio` control, which we haven't customized as of yet. We do have a `Radio` component in `ol-components`, but I believe it (as well as `Checkbox`) are old. If you replace the MUI `Radio` with our version in `RadioChoiceField`, it doesn't look right at all. The label should definitely be staying black though...

frontends/ol-components/src/components/Checkbox/CheckboxChoiceField.tsx
gumaerc: The `columnBelow` thing here did not work, but just doing a standard `[theme.breakpoints.down("md")]` worked fine.

gumaerc:
@ChristopherChudzicki This is ready for another look.

ChristopherChudzicki (approved):

ChristopherChudzicki:
Looks good to me! Great to have the two forms here using formik now.

PR #1271, opened 2024-07-15
Add profile option for silky to settings

rhysyngsun:

N/A

This adds an environment variable for being able to profile API requests locally. I didn't add ti to `app.json` because we probably shouldn't be using it there (at least not without some testing).

### How can this be tested?

Set `SILKY_PYTHON_PROFILER=True` in your `env/backend.local.env`and verify that when you now run API requests that silky records cProfile results (http://api.open.odl.local:8063/silk locally, see screenshot):

Screenshot 2024-07-15 at 14-58-19 Silky - CProfile - _api_v0_news_events_


mbertrand (approved):
👍

PR #1273, opened 2024-07-16
Shanbady/clicking item routes away from list

shanbady:
closes https://github.com/mitodl/hq/issues/4879

The changes in this PR allow for hash params in the url to be preserved when navigating to and from learning resource panels. 

### How can this be tested?
1. checkout this branch
2. navigate to /search
3.  add some hash paramters to the url (/search#test=123)
4. click on a learning resource and then close out the panel
5. note that the hash parameter is preserved

### DIFF
--- a/frontends/ol-components/src/components/RoutedDrawer/RoutedDrawer.tsx
+++ b/frontends/ol-components/src/components/RoutedDrawer/RoutedDrawer.tsx
@@ -69,7 +69,7 @@ const RoutedDrawer = <K extends string, R extends K = K>(
       })
       return newSearchParams
     })
-    document.location.hash = hash.trimStart("#")
+    document.location.hash = hash.substring(1)
   }, [setSearchParams, params, hash])
 
   return (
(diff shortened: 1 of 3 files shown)


jonkafton (approved):
This looks good :+1:

frontends/ol-components/src/components/RoutedDrawer/RoutedDrawer.tsx
```diff
@@ -67,7 +69,8 @@ const RoutedDrawer = <K extends string, R extends K = K>(
       })
       return newSearchParams
     })
-  }, [setSearchParams, params])
+    document.location.hash = hash.substring(1)
```
ChristopherChudzicki, line 72: In general, in a React app, we shouldn't touch `documentation.location` (or `window.location`):
- Changes to those properties won't cause React to re-render, which could be lead to bugs
- Will usually cause a full page reload, which we don't want. *This appears not to be the case when only changing the hash, but the first point is still relevant.*

Instead use `useNavigate`. Probably you'll want something like
```ts
  const removeUrlParams = useCallback(() => {
    const newSearchParams = new URLSearchParams(location.search)
    params.forEach((param) => {
      newSearchParams.delete(param)
    })
    navigate({
      ...location,
      search: newSearchParams.toString(),
    })
  }, [navigate, location])
```

ChristopherChudzicki (changes requested):
I haven't tested this locally but I did notice one code issue.

PR #1274, opened 2024-07-16
Improve offeror API performance

mbertrand:
Closes https://github.com/mitodl/hq/issues/4884

Removes n+1 queries for offeror channel urls

Screenshot 2024-07-16 125043

### How can this be tested?
Hit the url http://api.open.odl.local:8063/api/v1/offerors at least a few times.  Check that:
  - channel_url is present and correct for each offeror
  - there are no n+1 query log warnings in the docker console
  - the response time for the api endpoint is good at http://api.open.odl.local:8063/silk/


shanbady (approved):
looks good. not a blocker but it seems like we did lose the api description for the field

PR #1275, opened 2024-07-16
Update topics code for PWT topic mappings

jkachel:

Closes mitodl/hq#4553

Makes some changes to the topics code:
- Adds topics and mappings to offeror data feeds according to the list provided in the issue referenced above
- Adds support for persisting the mapping data in the database
- Adds support for specifying the topic icon in the database
- Updates ETL pipelines to have a standardized topic mapping function, and to use the database topics stuff for resolving topics

### How can this be tested?

Automated tests should pass.

#### Topic data

You will need to run the migrations in the `data_fixtures` app to upsert the topic data into the system. **Be advised** that the migration will truncate the topics table _with cascade_ so this will also wipe out mappings and associations with learning resources. Topic channels will be invalid after this as well. You will want to backpopulate anything that's got topics in it (ocw, mitx, mitxonline, xpro, prolearn, etc) after running the migration. (However, for _just_ testing the topic data upsert stuff, you don't need to do it right away.) 

The topic data is sourced from these places (from the original issue):

> We have interpreted the vocabularies and cross-walks into 3 mappings (awaiting a final approval):
>
> * OCW->PMT Topic Map ("draft3" tab)
>     * note that this presumes that we add a topic "archaeology" under "humanities"
> * ProLearn->PMT Topic Map ("draft3" tab)
>     * note that this presumes that we add "manufacturing" and "technology innovation" topics
> * edX->PMT Topic Map ("draft3" tab)
> * OpenLearningLibrary (OLL) doesn't have any topic metadata at the moment

Then, check to make sure the data is in there properly and matches the data in the YAML file.

You can test topic upsert using a new topic or updating an existing topic in these ways:

- Via the `data_fixtures` app - create a migration that either pulls a file with the data in it, or embeds the update YAML into a string within the migration.
- Manually - Call `learning_resources.utils.upsert_topic_data_string` or  `_file`. 

If you're testing modifying an existing topic, you should specify the database ID for the topic. (It will try to match on name but that won't work if that's what you're updating.) 

#### ETL Pipelines

Clear the `learning_resources_learningresource_topics` table (which contains the maps between learning resources and topics) and run the `backpopulate_` commands. This should fill the table and you should be able to single out individual resources and confirm that their topics are mapped to Open ones properly. Not all topics will map successfully - only ones we have a mapping for will come over. We no longer create new topics if the topic didn't exist in the system.

In particular, when syncing OCW data, you should _not_ see a bunch of "Skipped adding topic" messages in the logs. (There may be a few but they should be few and far between.) OCW has a lot of topic mappings so this is a good test to make sure the mapping system is working.

### Additional Context

Since the topic maps will be different than they were, items may not be in the places you expect them - check against the `topics.yaml` file to see where things should be.

The `topics.yaml` has IDs in it that were created from a truncated and reset table (so they start at 1 and go up). For consistency, the migration also truncates the table. This requires a re-populate of a bunch of stuff (notably topic channels, and any data source that includes topic data) so some care will need to be taken to perform those tasks after this merges for both local deployments and Heroku ones.


learning_resources/utils.py
```diff
@@ -473,74 +475,108 @@ def offeror_delete_actions(offeror: LearningResourceOfferor):
...
     Returns:
     - None
     """
 
     for topic in topics:
-        lr_topic, _ = LearningResourceTopic.objects.filter(name=topic).update_or_create(
+        lr_topic, created = LearningResourceTopic.objects.filter(
+            Q(name=topic["name"]) | Q(id=topic["id"])
+        ).update_or_create(
             defaults={
```
mbertrand, line 501: What do you think of changing this to:
```python
        ).update_or_create(
            id=topic["id"],
            defaults={
```

Assuming the file will always have an id supplied, anyway.  This way the topics will be assigned the ids in the yaml file, and if running `upsert_topic_data_file` on a file that specifies an existing id and a changed name, the name will be updated on the existing topic instead of a new topic being created and the id value being ignored.

If `id` isn't always present, then maybe that arg could be used conditionally.

learning_resources/etl/prolearn_test.py
```diff
@@ -202,6 +208,7 @@ def test_prolearn_transform_programs(mock_csail_programs_data):
 
 def test_prolearn_transform_courses(mock_mitpe_courses_data):
     """Test that prolearn courses data is correctly transformed into our normalized structure"""
+    upsert_topic_data_file()
```
mbertrand, line 211: Could be done via a pytest fixture, but this works too.

mbertrand (changes requested):
Overall looks great!

learning_resources/data/topics.yaml
```diff
@@ -0,0 +1,1160 @@
...
+            - Biography
+            - Classics
+            - Comedy
+            - Comparative Literature
+            - Creative Writing
+            - Criticism
+            - Drama
+            - Fiction
+            - International Literature
+            - Nonfiction Prose *Periodic Literature
```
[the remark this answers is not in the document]
jkachel: Yes. (@pdpinch this is in the spreadsheet on row 196 - I don't know who to ask about this but it seems a bit out of place.)

learning_resources/models.py
```diff
@@ -140,6 +161,22 @@ def __str__(self):
...
+
+    topic = models.ForeignKey(
+        "LearningResourceTopic",
+        on_delete=models.CASCADE,
+        related_name="+",
+    )
+    offeror = models.ForeignKey(
+        "LearningResourceOfferor",
+        on_delete=models.CASCADE,
+        related_name="+",
```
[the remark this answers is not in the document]
jkachel: I wanted to avoid having these show up elsewhere - the only part of the app that should care about the mappings is the ETL pipelines so I didn't want mappings to show up in, say, APIs because someone used `fields = "__all__"`.

data_fixtures/migrations/0004_upsert_initial_topic_data.py
```diff
@@ -0,0 +1,36 @@
...
+
+    dependencies = [
+        ("data_fixtures", "0003_unit_page_copy_updates"),
+    ]
+
+    operations = [
+        migrations.RunSQL(
+            sql=[
+                (
+                    "TRUNCATE learning_resources_learningresourcetopic "
```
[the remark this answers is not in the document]
jkachel: I reworked the migration to cull topic channels and to delete topics using model objects.

learning_resources/utils.py
jkachel: The ID isn't always there: you may be adding a new one. However, this lead me down a bit of a rabbit hole wrt the IDs - this was using database (int) IDs and that ended up being a problem, so I refactored to add a UUID field. (Using database IDs ended up causing weirdness because inserting rows with IDs doesn't update the identity sequence, which generates database errors when you're trying to add new ones.)

jkachel:
Last few commits should address everything here.

data_fixtures/migrations/0004_upsert_initial_topic_data.py
```diff
@@ -6,7 +6,20 @@
...
+    After this migration runs, you will need to repopulate these things.
+    """
+
+    LearningResourceTopic = apps.get_model(
+        "learning_resources", "LearningResourceTopic"
+    )
+    ChannelTopicDetail = apps.get_model("channels", "ChannelTopicDetail")
+
+    LearningResourceTopic.objects.all().delete()
+    ChannelTopicDetail.objects.all().delete()
```
mbertrand, line 22: This still leaves a bunch of orphaned  `Channel` objects with `channel_type="topic"` and a null `ChannelTopicDetail`.  This should handle it, and also delete the related ChannelTopicDetail objects via cascade:

```python
Channel.objects.filter(channel_type="topic").delete()
```

learning_resources/management/commands/update_topics.py
```diff
@@ -1,42 +1,68 @@
...
 
 
 class Command(BaseCommand):
     """Update LearningResourceTopic data"""
 
-    help = "Update LearningResourceTopic data"
+    help = (
+        "Update LearningResourceTopic data from a yaml file. Optionally "
+        "dump the updated topics to a file."
+    )
```
mbertrand, line 27: I think this might be better left to the data_fixtures app, @shanbady can you chime in?

Also, in terms of the option for generating the yaml file, seems like there might be an issue if the current data in the db is incomplete?  Should the yaml file only be generated from a presumably 100% complete CSV file ?

data_fixtures/migrations/0004_upsert_initial_topic_data.py
mbertrand: Another consequence of deleting all existing topics is that any learning paths (like featured lists for OCW, xPro, etc) will have their manually assigned `topics` field set to an empty list.  Not sure how big a deal this is, as there aren't that many of them, but might be worth checking with @pdpinch.

mbertrand (commented):
Looking good, just another couple comments

learning_resources/management/commands/update_topics.py
shanbady: Yea thats probably the way to go. I feel like anything static fixture related should be consolidated under the data fixtures app

data_fixtures/migrations/0004_upsert_initial_topic_data.py
jkachel: We should run the pipelines for sources that include topic data once this gets merged anyway - would these lists get updated as part of that? (I know there's a `populate_featured_lists` for featured courses, for instance.)

learning_resources/management/commands/update_topics.py
jkachel: I removed it. The upsert APIs can be run from a shell if someone needs to process a file or string manually for whatever reason.

The upsert functions work fine with deltas, so the incoming data doesn't need to be complete.

mbertrand (commented):
Sorry for noticing this late, I often overlook podcasts & videos, but noticed there are lots more podcast episodes without topics using the new system.  Saw lots of these messages in the logs  when running `./manage.py backpopulate_podcast_data`:

> Skipped adding topic Physical Education and Recreation to resource LearningResource object (5257)
> Skipped adding topic Physical Education and Recreation to resource LearningResource object (5258)
> Skipped adding topic Business to resource LearningResource object (5258)

Still running the video pipeline but that looks better in terms of assigning topics.

mbertrand (approved):
Getting the same # podcast episodes without topics on the main branch, I'm guessing the discrepancy on prod must be due to those topics being added before we stopped adding new topics as they came in.

mbertrand:
Sorry, haven't had enough coffee, forgot to populate old topics after switching back to the main branch.  Once I did that, most podcast episodes were assigned topics.  Probably best to check with Peter/Ferdi about how important it is for most podcasts/episodes to have topics before merging.

jkachel:
Discussed briefly this morning and it doesn't seem like topics on podcasts were too big of a deal. 

I'm holding off on merging this until Monday - since this will require some manual guidance to deploy, it'd be best to do this in a controlled manner and also not block anyone else doing stuff in Open. (Plus, it's Friday so no releases anyway.)

PR #1276, opened 2024-07-16
adding migrations for copy update

shanbady:

Closes https://github.com/mitodl/hq/issues/4896

Adds a data_fixtures migration to make the copy updates as described in  https://github.com/mitodl/hq/issues/4896

### How can this be tested?
1. checkout this branch
2. run data_fixture migrations via ```RUN_DATA_MIGRATIONS=true python  manage.py migrate```
3. verify the copy has been updated on unit pages for ocw, mitpe and mitx


data_fixtures/migrations/0003_unit_page_copy_updates.py
```diff
@@ -0,0 +1,95 @@
...
+fixtures = [
+    {
+        "name": "mitpe",
+        "offeror_configuration": {
+            "value_prop": (
+                "MIT Professional Education is a leader in technology and "
+                "engineering education for working professionals pursuing "
+                "career advancement, and organizations seeking to meet modern-day "
+                "challenges by expanding the knowledge and skills of their employees. "
+                "Courses are delivered in a range of formats—in-person (on-campus "
```
abeglova, line 14: Should there be spaces around the dashes so it's `formats — in-person` instead of `formats—in-person`?

Alternatively so there are fewer different kinds of dashes `Courses are delivered in a range of formats to meet the needs of today's learners: in-person (on-campus and live virtual), online, and through hybrid approaches`

data_fixtures/migrations/0003_unit_page_copy_updates.py
```diff
@@ -0,0 +1,95 @@
...
+                "and live virtual), online, and through hybrid approaches—to "
+                "meet the needs of today's learners."
+            ),
+        },
+        "channel_configuration": {
+            "sub_heading": (
```
abeglova, line 20: is value_prop and sub_heading different for other offerors? Seems weird that the same copy is stored twice

abeglova (changes requested):
Works great back to you with a few comments

data_fixtures/migrations/0003_unit_page_copy_updates.py
shanbady: Its hard to discern in plain text but those are em dashes. once you load the fixture it displays as expected. 
Screenshot 2024-07-17 at 11 59 23 AM

data_fixtures/migrations/0003_unit_page_copy_updates.py
shanbady: This is odd and stemmed from the unit listing pages being worked alongside the unit page banners.  We should remove this value_prop field and instead derive all this from the channel configuration but I think thats worth its own effort.

abeglova (approved):
👍

PR #1278, opened 2024-07-17
Improve channel API performance

mbertrand:
Closes https://github.com/mitodl/hq/issues/4886

- Changes the channels API pagination to default (10/100) instead of large (1000/1000)
- Adds an index on `Channel.channel_type`

### How can this be tested?
- Set DEBUG=False in your backend .env file (reponse time can be >2x slower when True)
- In postman or equivalent, send multiple GET requests to the following, the majority should be speedy, ~250 ms or better (postman will show the response time and you can also check the performance with [silk](http://api.open.odl.local:8063/silk/requests/))
  - http://api.open.odl.local:8063/api/v0/channels/
  - http://api.open.odl.local:8063/api/v0/channels/unit/ocw/
  - http://api.open.odl.local:8063/api/v0/channels/unit/mitx/

### Additional Context
The `/api/v1/channels/type/unit/*` response times referenced in the ticket issue were acceptably fast locally before and after these changes.  But `/api/v1/channels/` was originally quite slow because it was returning all channels in a single response.


ChristopherChudzicki (approved):
👍

mbertrand:
@ChristopherChudzicki thanks for reviewing, but I found some n+1 queries when logged in as a non-staff user (due to `is_moderator` field).  Might try to fix that in this PR, but if it turns out to be complex I'll make a followup PR for it instead.

mbertrand:
@ChristopherChudzicki I made some changes so that there are no n+1 queries made when determining if the user is a moderator, plus added a test for that.

rhysyngsun (approved):

PR #1279, opened 2024-07-17
fix keyboard drag and drop

ChristopherChudzicki:
- Closes https://github.com/mitodl/hq/issues/4861

Makes sortable lists sortable via keyboard interactions

Note: The drag styling still is being updated as part of 
- https://github.com/mitodl/hq/issues/4859

https://github.com/user-attachments/assets/d9caa9ad-72ce-4323-8c24-2dd3f24c1ce4

### How can this be tested?
1. View a sortable list (example: dashboard page userlists or learning paths)
2. With keyboard:
    1. Tab to "Reorder" and press spacebar
    2. Tab to one of the list items and press spacebar
    3. use up/down keys to reorder
    4. Press spacebar to drop or escape to cancel the sorting

If you would like some... pretty verbose... instructions, enable a screen reader while you do the above.

### DIFF
--- a/frontends/ol-components/src/components/SortableList/SortableList.tsx
+++ b/frontends/ol-components/src/components/SortableList/SortableList.tsx
@@ -69,16 +69,17 @@ const SortableItem = <I extends UniqueIdentifier = UniqueIdentifier>(
   const handleProps: HandleProps = useMemo(
     () => ({
       ...listeners,
+      ...attributes,
       ref: setActivatorNodeRef,
       className: !props.disabled ? "ol-draggable" : undefined,
     }),
-    [setActivatorNodeRef, listeners, props.disabled],
+    [setActivatorNodeRef, attributes, listeners, props.disabled],
   )
 
   const { Component = "div" } = props
 
   return (
-    <Component ref={setNodeRef} style={style} {...attributes}>
+    <Component ref={setNodeRef} style={style}>
       <DragStyles>{props.children && props.children(handleProps)}</DragStyles>
     </Component>
   )


frontends/ol-components/src/components/SortableList/SortableList.tsx
```diff
@@ -69,16 +69,17 @@ const SortableItem = <I extends UniqueIdentifier = UniqueIdentifier>(
...
-    [setActivatorNodeRef, listeners, props.disabled],
+    [setActivatorNodeRef, attributes, listeners, props.disabled],
   )
 
   const { Component = "div" } = props
 
   return (
-    <Component ref={setNodeRef} style={style} {...attributes}>
+    <Component ref={setNodeRef} style={style}>
       <DragStyles>{props.children && props.children(handleProps)}</DragStyles>
```
ChristopherChudzicki, line 83: For some context: The reason there's a render function being uses here is that our sorting implementation supports   cases where the entire item is not draggable, but some "handle" is draggable. So the render function is used so that you can pass the listeners and attributes to the handle.

In practice, all the sort items we currently have are "whole item draggable". The handle behavior is somewhat legacy, but since a drag handle supports our current use case (whole item IS the handle) I don't see any reason to remove it.

**The bug:** The listeners and attributes need to be attached to the same item. Before we were doing something more akin to
```tsx
<div {...dragAttributes }>
  <div {...dragListeners />
</div>
```
The inner div had all the correct listeners (including keyboard listeners)!. But it had no tab index. So keyboard users had no way of actually interacting with the inner draggable div. The drag attributes and drag listeners need to be on the same div.

**Testing:** Unfortunately, I don't know of any reasonable way to test this on CI without a real browser :/

jonkafton (approved):

PR #1283, opened 2024-07-17
add staleness penalty slider

abeglova:
closes https://github.com/mitodl/hq/issues/4839

This pr adds a parameter to search which adds a penalty to the scores of older resources when there is a search tearm. A toggle that is only visible to admins on the search page that makes it possible to control the size of the penalty. Resources with upcoming runs receive no penalty. For videos and podcast episodes the penalty is calculated from the last modified date received from the source. For courses the penalty is calculated from the start date of the last run. 

The penalty is in % per year (linear). So if the penalty is 5% and the resource is 2 years old then the relevance score for the resource is multiplied 0.9 to get the order that the resource is displayed in.

Screenshot 2024-07-19 at 10 48 39 AM

### How can this be tested?
Run 
docker-compose run web ./manage.py recreate_index --all
docker-compose run web ./manage.py backpopulate_ocw_data --course-name 18-417-introduction-to-computational-molecular-biology-fall-2004

Verify that you search for "Molecular Biology" in search the course from 2004 is shown as one of the first results.

Verify that you see a toggle for "Resource Score Staleness Penalty" when you log in as an admin but not when you log in as a non-admin or are not logged in. 

Search again for "Molecular Biology". As you move the slider the course from 2005 moves further down the ranking

### Additional Context
If you try to set the staleness penalty before you recreate the index you will see no results. I'm not sure if people outside our team have admin access to prod yet. I will talk to Ferdi about whether it makes sense to break the ui component into a separate PR so it's not visible until the recreate_index job adding `resource_age_date` to the index runs

### DIFF
--- a/frontends/api/src/generated/v1/api.ts
+++ b/frontends/api/src/generated/v1/api.ts
@@ -3364,6 +3364,12 @@ export interface PercolateQuerySubscriptionRequestRequest {
    */
   professional?: boolean | null
   /**
+   * Relevance score penalty percent per year for for resources without upcoming runs. Only affects results if there is a search term.
+   * @type {number}
+   * @memberof PercolateQuerySubscriptionRequestRequest
+   */
+  yearly_decay_percent?: number | null
+  /**
    * True if the learning resource offers a certificate
    * @type {boolean}
    * @memberof PercolateQuerySubscriptionRequestRequest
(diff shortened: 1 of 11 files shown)


abeglova:
Talked to Ferdi and Peter - the slider not being operational until after the reindex is fine, as log as it only affects admins

shanbady (changes requested):
I couldnt get this interface to not* display as a non-admin user

shanbady (approved):
Looks good. Functions as intended

PR #1284, opened 2024-07-18
Remove courses under program in APIs

rhysyngsun:

Part of https://github.com/mitodl/hq/issues/4830

This removes `program.courses` under the learning resource / program APIs. It was causing a high number of N+1 errors and generally wasn't the most performant. This is instead being replaced by a `course_count` because that's the only piece of that data the frontend is currently using.


mbertrand (approved):
Looks good, definitely faster.  The featured api response (a subset of the LR api) is still pretty slow, I think the LearningResourceSerializer's use of the `LearningResource.prices` property is one of the main culprits for that.  When I return a static `[0.00]` for that property instead to test the difference, it's much faster still.

PR #1287, opened 2024-07-18
fix restricted redirect

gumaerc:
Closes https://github.com/mitodl/hq/issues/4888

This PR makes some changes to the login / routing process:

 - The `proxy` section of the Webpack configuration for `webpack-dev-server` has been removed. Since the stock env files for local dev set `MITOPEN_API_BASE_URL` and `MITOPEN_APP_BASE_URL` to sane defaults, most people shouldn't have trouble and requests should be sent to the proper URL, rather than relying on `webpack-dev-server` proxying the request.
 - Instead of throwing a `ForbiddenError` in `RestrictedRoute` when `Permissions.Authenticated` is required and a user isn't logged in, they are simply redirected to the login flow.
 - Calls to the login redirect now include the query string and hash (document fragment) so if the user is on a page like `/dashboard#profile` or on a search page with filters applied (or they are sent a link), they will be properly brought back to where they were intending to go.

### How can this be tested?
 - Make sure you have Keycloak configured to point at a valid running instance. This could be https://sso-qa.ol.mit.edu, and if you need help setting it up just ask.
 - Make sure you have a local domain set up for your instance of `mit-open`. There are the default values, but you can create your own `shared.local.env` file (see readme for more info) to customize them:
```
MITOPEN_API_BASE_URL=http://api.open.odl.local:8063
MITOPEN_APP_BASE_URL=http://open.odl.local:8062
```
 - Open an incognito window, put http://open.odl.local:8062/dashboard#profile into your URL bar and hit enter
 - You should be brought to Keycloak. Proceed to enter your login details and complete the login process.
 - Verify that after logging in, you are properly brought to http://open.odl.local:8062/dashboard/#profile

### DIFF
--- a/frontends/mit-open/src/common/urls.ts
+++ b/frontends/mit-open/src/common/urls.ts
@@ -66,6 +66,13 @@ export const login = ({
   return `${LOGIN}?next=${next}`
 }
 
+export const next = () => {
+  const path = window.location.pathname
+  const search = window.location.search
+  const hash = window.location.hash
+  return encodeURIComponent(`${path}${search}${hash}`)
+}
+
 export const DASHBOARD = "/dashboard/"
 
 export const SEARCH = "/search/"
--- a/frontends/mit-open/src/common/urls.ts
+++ b/frontends/mit-open/src/common/urls.ts
@@ -50,9 +50,11 @@ export const LOGOUT = `${process.env.MITOPEN_API_BASE_URL}/logout/`
 export const login = ({
   pathname = "/",
   search = "",
+  hash = "",
 }: {
   pathname?: string
   search?: string
+  hash?: string
 } = {}) => {
   /**
    * To include search parameters in the next URL, we need to encode them.
@@ -62,15 +64,8 @@ export const login = ({
    * There's no need to encode the path parameter (it might contain slashes,
    * but those are allowed in search parameters) so let's keep it readable.
    */
-  const next = `${window.location.origin}${pathname}${encodeURIComponent(search)}`
+  const next = `${window.location.origin}${pathname}${encodeURIComponent(search)}${encodeURIComponent(hash)}`
   return `${LOGIN}?next=${next}`
-}
-
-export const next = () => {
-  const path = window.location.pathname
-  const search = window.location.search
-  const hash = window.location.hash
-  return encodeURIComponent(`${path}${search}${hash}`)
 }
 
 export const DASHBOARD = "/dashboard/"
--- a/frontends/mit-open/webpack.config.js
+++ b/frontends/mit-open/webpack.config.js
@@ -30,7 +30,6 @@ const {
   ENVIRONMENT,
   PORT,
   MITOPEN_API_BASE_URL,
-  API_DEV_PROXY_BASE_URL,
   WEBPACK_ANALYZE,
   SITE_NAME,
   MITOPEN_SUPPORT_EMAIL,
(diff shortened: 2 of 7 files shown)


frontends/mit-open/src/common/urls.ts
```diff
@@ -66,6 +66,13 @@ export const login = ({
   return `${LOGIN}?next=${next}`
 }
 
+export const next = () => {
+  const path = window.location.pathname
+  const search = window.location.search
+  const hash = window.location.hash
+  return encodeURIComponent(`${path}${search}${hash}`)
+}
```
ChristopherChudzicki, line 74: In general, we should avoid reading from `window.location`. Use
```ts
# Can only be used inside a component.
const location = useLocation()
```
and read from that instead.

**Reason:** Changes to global variables (like `window.location`) won't cause react to re-render[^1]. So for example, suppose the login link (which is styled as a button) reads its `href` from `window.location`. Then when the user routes within the app, react will have an out-of-date `href` one the login button. 

[^1]: In the case of `window.location`, this depends on how it changes. If you change `window.location` by directly assigning to `window.location.href`, then the whole page reloads, so ... yeah, in that case, react would re-render. React router uses the [history API](https://developer.mozilla.org/en-US/docs/Web/API/History_API) to change `window.location` without doing a full refresh.

When you use `const location = useLocation()`, the `useLocation` hook gets its value from a [React Context](https://react.dev/learn/passing-data-deeply-with-context). Then related routing hooks (like `useNavigate` and `useSearchParams`) change the value of the context, which causes react to re-render.

frontends/mit-open/webpack.config.js
```diff
@@ -51,11 +50,6 @@ const {
     desc: "Base URL for API requests",
     devDefault: "",
   }),
-  API_DEV_PROXY_BASE_URL: str({
-    desc: "API base URL to proxy to in development mode",
-    default: "",
-    devDefault: process.env.MITOPEN_APP_BASE_URL,
-  }),
```
ChristopherChudzicki, line 58: **Suggestion:**
```ts
  API_DEV_PROXY_BASE_URL: str({
    desc: "API base URL to proxy to in development mode",
    default: "",
    devDefault: "",
  }),
```

frontends/mit-open/webpack.config.js
```diff
@@ -269,24 +263,6 @@ module.exports = (env, argv) => {
...
-            "/static/hijack",
-          ],
-          target: API_DEV_PROXY_BASE_URL || MITOPEN_API_BASE_URL,
-          changeOrigin: true,
-          secure: false,
-          headers: {
-            Origin: API_DEV_PROXY_BASE_URL || MITOPEN_API_BASE_URL,
-          },
-        },
-      ],
```
ChristopherChudzicki, line 289: **Suggestion:**
```ts
      proxy: API_DEV_PROXY_BASE_URL
        ? [
            {
              context: [
                "/api",
                "/login",
                "/logout",
                "/admin",
                "/static/admin",
                "/static/hijack",
              ],
              target: API_DEV_PROXY_BASE_URL,
              changeOrigin: true,
              secure: false,
              headers: {
                Origin: API_DEV_PROXY_BASE_URL,
              },
            },
          ]
        : [],
```

ChristopherChudzicki (approved):
👍 

I suggested restoring the `API_DEV_PROXY_BASE_URL` with more conservative settings.
- no default API_DEV_PROXY_BASE_URL
- only enable proxy if it's set

My thought is:
- running against RC can be useful
- API_DEV_PROXY_BASE_URL is pretty explicit about what it does

I think @jonkafton is using this sometimes; I also do. I recognize it did cause some confusion, so if you feel strongly about removing this, OK.

PR #1290, opened 2024-07-18
Save resource prices in a new database model and calculate during ETL/nightly task

mbertrand:
Part of https://github.com/mitodl/hq/issues/4759 and https://github.com/mitodl/hq/issues/4830 and 

- Adds a `prices` field to `LearningResource` (formerly a property)
- Calculates the value during course/programs ETL (`load_next_start_dates_and_prices`), and a nightly celery task.  New migration calculates the value for courses/programs that can be expected to have non-zero prices.

### How can this be tested?
- If you don't already have courses/programs with prices yet, **on the main branch** ingest some by running the following (after setting required .env variables `XPRO_*` and `MITX_ONLINE_*` based on RC values):
  ```
  ```
- Go to http://open.odl.local:8062/search/?certification_type=professional&certification_type=completion, you should get results with prices for certificates.  Take a screenshot.
- Switch to this branch.  Restart containers.  The migration should complete successfully.
- Run `./manage.py update_index --all`
- Refresh the same search url as above.  You should get the same prices back per course as before.  
- Run the same two management commands again.  They should complete successfully, and the prices in the search results should again be the same.
- Run `./manage.py backpopulate_podcast_data` too just to make sure the new code works successfully for resosurces without runs or prices.


learning_resources/migrations/0045_learningresource_next_start_date.py
```diff
@@ -14,7 +14,7 @@ def assign_next_start_date(apps, schema_editor):
     for resource in LearningResource.objects.filter(
         resource_type__in=["course", "program"]
     ).iterator():
-        load_next_start_date(resource)
+        load_next_start_date_and_prices(resource)
```
abeglova, line 17: This will break if this migration is run and there is data because the prices column is not added until 0057. I think it's better to just delete the `assign_next_start_date` part of the migration

abeglova (changes requested):
Works great. One requested change

learning_resources/migrations/0045_learningresource_next_start_date.py
mbertrand: Good catch, I didn't even notice that this migration had been affected.

mbertrand: I removed that function from the migration

abeglova (approved):
👍

PR #1293, opened 2024-07-19
Fix bug with background image obscuring search controls

jonkafton:

Fixes https://github.com/mitodl/hq/issues/4808

~~Gives the offending image container a z-index of -1 so it lives behind the search utils.~~

The previous PR https://github.com/mitodl/mit-open/pull/1286 was [reverted](https://github.com/mitodl/mit-open/pull/1291) as it introduced a worse bug than it fixed by making the search input inoperable.

This change now uses css to display the image as a background image and moves this and the gradient to the containing div.

### How can this be tested?

The bug was present at screen widths where the background image looks like this:

image

- Navigate to e.g. /search/?certification_type=none

- Confirm that the page renders as previously at all screen widths.

- Confirm the "Clear all" is clickable.

- Confirm the search input is focusable.


ChristopherChudzicki (approved):
👍  Clear facets works, search input works, background image seems better 👍

PR #1295, opened 2024-07-19
Add subscription confirmation dialog on unit pages

jonkafton:

Closes https://github.com/mitodl/hq/issues/4908

Adds a confirmation dialog to the "Follow" button on unit pages, according to [the designs](https://www.figma.com/design/Eux3guSenAFVvNHGi1Y9Wm/MIT-Design-System?m=dev&node-id=9320-130109).

Note that the designs show a dialog with fixed width buttons at equal widths with the container fixed to exactly contain them. I have discussed with @steven-hatch - this is not a specific design, so dialogs remain fixed (now at 600px width), using existing buttons (that shrink to contents, with min width 100px) that are positioned right.

image

### How can this be tested?

- Whilst logged in
  - Navigate to any unit page, e.g. `/c/unit/ocw/`
  - Click "Follow"
     - Dialog should display, matching designs (with caveat above)
     - Dialog is cancellable
       - By clicking the close X
       - By clicking the cancel button
       - By clicking anywhere on the background mask
     - On clicking "Got it!"
       - You are subscribed
       - Button remains "Following" on refresh 

### Additional Context

This uses our existing `ol-component` `BasicDialog`, now renamed to `Dialog`, and updates its styles to reflect the designs. All dialogs are therefore affected.

The existing dialogs should function ok (and look a little better), e.g. the Add to List dialogs (click any resource card bookmark icon while logged in):

image


ChristopherChudzicki (approved):
Looks good!

Left one request (aria label) but I don't think it needs a re-review, so approving 👍

jonkafton:
Holding off on merging - UX discussions ongoing.

jonkafton:
Updated to remove the "follow" confirmation dialog. The original ticket has been repurposed to apply the design updates to modals generally.

PR #1298, opened 2024-07-22
set search page size to 20

ChristopherChudzicki:
- Closes https://github.com/mitodl/hq/issues/4954

Sets search page size to 20 items per page

My screen isn't tall enough to show 20:

Screenshot 2024-07-22 at 11 27 49 AM

### How can this be tested?
1. Visit http://localhost:8062/search and http://localhost:8062/c/unit/ocw You should get 20 results per page.
    - you could count by eye and or with `document.querySelectorAll('.MuiTabPanel-root a')` in dev console.
    - also check the network tab API url ... it should have `?limit=20`
 2. Check pages 2, 3, 4, etc. The API calls should be as expected (e.g., `?offset=20&limit=20`, `?offset=40&limit=20`, etc.)


mbertrand (approved):
👍

PR #1299, opened 2024-07-22
Enable User List sorting for non Learning Path editors

jonkafton:

Fixes https://github.com/mitodl/hq/issues/4936

Gives the ItemsListingComponent `canEdit` and `showSort` props to move the conditions to determine whether a user can edit or reorder items listings to the calling components for variation depending on use and applies as follows:

|Page|Permission|canEdit|showSort|
|----|---|---|---|
|User List Details|Not Authenticated|N/A|N/A|
|User List Details|Authenticated|✓|✓|
|User List Details|Learning path editor|✓|✓|
|Learning Path Details|Not Authenticated|✗|✗|
|Learning Path Details|Authenticated|✗|✗|
|Learning Path Details|Learning path editor|✓|✓|

The change is that a user who is not a Learning Path editor can now reorder.

### How can this be tested?

Navigate to the following:
- a Dashboard "My Lists" User List Details page (e.g. "Favorites")
- a User List Details page at e.g. `/userlists/2`
- a Learning Path Details page at e.g. `/learningpaths/1`

Confirm the abilities to edit and reorder are applied according to the table above.

### DIFF
--- a/frontends/mit-open/src/pages/DashboardPage/UserListDetailsTab.tsx
+++ b/frontends/mit-open/src/pages/DashboardPage/UserListDetailsTab.tsx
@@ -4,6 +4,7 @@ import {
   useUserListsDetail,
 } from "api/hooks/learningResources"
 import { ListType } from "api/constants"
+import { useUserMe } from "api/hooks/user"
 import { manageListDialogs } from "@/page-components/ManageListDialogs/ManageListDialogs"
 import ItemsListingComponent from "@/page-components/ItemsListing/ItemsListingComponent"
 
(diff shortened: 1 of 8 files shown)


frontends/mit-open/src/pages/DashboardPage/UserListDetailsTab.tsx
```diff
@@ -28,6 +31,8 @@ const UserListDetailsTab: React.FC<UserListDetailsTabProps> = (props) => {
       items={items}
       isLoading={itemsQuery.isLoading}
       isFetching={itemsQuery.isFetching}
+      showSort={!!user?.is_authenticated}
```
jonkafton, line 34: Redundant as this page won't be routable for users who are not signed in, so just for good measure and to show intent.

ChristopherChudzicki: BTW: Userlists have a privacy setting `"privacy_level": "private" | "unlisted", where "unlisted" is meant to be. "shareable, but doesn't show up in our search index".

So it's possible this isn't completely redundant. Though I would expect we'd have a separate page for sharing a userlist (not the dashboard), and this feature is more theoretical at the moment than real.

ChristopherChudzicki (approved):
👍

PR #1303, opened 2024-07-23
Address performance of learning resource APIs

rhysyngsun:

Part of https://github.com/mitodl/hq/issues/4830

This PR formalizes how we prefetch learning resources by adding custom `QuerySet` implementations with a standard `for_serialization` method. Going forward this will be where all prefetches that are necessary to query for and serialize model objects efficiently. We should not be performing database queries in serializers going forward (aside from validation, `create()`, and `update()`).

A lot of code got moved out of serializers and into `@cached_property` on django models since this is far more robust and aligns to django's practices of how to write models.

An added benefit that I saw from this was by applying the `for_serialization` to the indexing operations I saw it go from 2k queries down to ~70. `ContentFile` indexing similarly reduced from 100-200 to a few dozen. It's not really possible to assert query counts on these because there's some variation in query counts I couldn't figure out and it's out of scope anyway.

### How can this be tested?

- The APIs should return the same data as on `main` but be much faster.
- The search indexing should similarly serialize the data the same but will probably run much faster as well.


learning_resources/serializers_test.py
```diff
@@ -266,20 +267,6 @@ def test_learning_resource_serializer(  # noqa: PLR0913
...
-    """Only published runs should be in the serializer"""
-    resource = LearningResourceFactory.create(is_course=True)
-    assert resource.runs.count() == 2
-    assert len(LearningResourceSerializer(resource).data["runs"]) == 2
-    unpublished_run = resource.runs.last()
-    unpublished_run.published = False
-    unpublished_run.save()
-    updated_data = LearningResourceSerializer(resource).data
-    assert len(updated_data["runs"]) == 1
-    assert updated_data["runs"][0]["id"] != unpublished_run.id
```
shanbady, line 279: Wondering why this test was removed since it seems like when I re-add this test it does return an unpublished run.

shanbady (changes requested):
Functionally things appear to work with these changes on my end. I left a question about a removed test.

learning_resources/serializers_test.py
rhysyngsun: Because I removed the code [here](https://github.com/mitodl/mit-open/pull/1303/files#diff-638dcb3aea0f7898675d7c87d8c639d4c572fa2a797f346a62c772db215119d7L525) that was doing the filtering as the filtering is now more appropriately being done in the [query prefetching](https://github.com/mitodl/mit-open/pull/1303/files#diff-3e244d0725c2dbf2dfd444546f36377feb3dcb0bfee9838f50a971237def1e72R289)

abeglova (approved):
All the apis I tested work as expected. Locally, reindexing took 593.61749 seconds vs
804.47014 seconds on main. One minor comment

shanbady (approved):
👍

channels/models.py
```diff
@@ -154,7 +150,10 @@ class ChannelDepartmentDetail(TimestampedModel):
         related_name="department_detail",
     )
     department = models.ForeignKey(
-        LearningResourceDepartment, null=True, on_delete=models.SET_NULL
+        LearningResourceDepartment,
+        null=True,
+        on_delete=models.SET_NULL,
+        related_name="department_details",
```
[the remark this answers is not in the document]
rhysyngsun: Makes sense

rhysyngsun:
@abeglova addressed your feedback

mbertrand (changes requested):
Overall looks great!  I did notice one issue though with the http://api.open.odl.local:8063/api/v1/schools/ api: all the channel urls are null.  Consequently none of the links on the [Departments page ](http://open.odl.local:8062/departments/) work

Ditto for the http://api.open.odl.local:8063/api/v1/departments/ endpoint

rhysyngsun:
> Overall looks great! I did notice one issue though with the http://api.open.odl.local:8063/api/v1/schools/ api: all the channel urls are null. Consequently none of the links on the [Departments page ](http://open.odl.local:8062/departments/) work
> 
> Ditto for the http://api.open.odl.local:8063/api/v1/departments/ endpoint

Good catch, I had a typo here: https://github.com/mitodl/mit-open/pull/1303/commits/2eadab130961403289e46bbd759e0dde62bd32e8#diff-3e244d0725c2dbf2dfd444546f36377feb3dcb0bfee9838f50a971237def1e72L228

I also added some prefetching to the departments and schools APIs while I was in there, continuing the pattern of `for_serialization()` methods there and also updating existing usages of `annotate_channel_url()` to instead use `for_serialization()` to future proof things.

mbertrand (approved):
👍

PR #1304, opened 2024-07-23
Database index on FeedEventDetail.event_datetime

mbertrand:
- https://github.com/mitodl/hq/issues/4883

- Adds a database index on `news_events.models.FeedEventDetail.event_datetime` 

### How can this be tested?
- Not strictly necessary but setting `DEBUG=False` in your .env file will likely speed up response times.
- Also set `SILKY_PYTHON_PROFILER=True` in your backend env file.
- If you don't have any news/events in the database yet, you can run `./manage.py backpopulate_news_events`.  You can also generate fake ones in a shell to get the numbers up (the command will likely import only a dozen or two):
  ```
  from news_events.factories import *
  source = FeedSourceFactory.create(is_event_type=True)
  FeedItemFactory.create_batch(100, source=source, is_event=True)
  ```
- In postman or equivalent, send multiple requests to http://api.open.odl.local:8063/api/v0/news_events/?feed_type=events
- Go to http://api.open.odl.local:8063/silk/ and check that the # of queries and response time seem acceptable.
- The home page should still show events ordered by event date (soonest first).

Screenshot 2024-07-23 at 8 52 41 AM


ChristopherChudzicki (approved):
👍 This sees like a good index to have when we have more news events.

I'm not sure this API is slow at the moment, though. I left a comment on the issue thread https://github.com/mitodl/hq/issues/4883#issuecomment-2245579114

PR #1305, opened 2024-07-23
No prices for archived runs or resources w/out certificates

mbertrand:
Closes https://github.com/mitodl/hq/issues/4931

Adjusts  the function`learning_resources.etl.load_runs` so that archived runs have no prices.

### How can this be tested?
- Set `EDX_API_` env values to the same as on RC
- Run `./manage.py backpopulate_mit_edx_data` (use the --api_datafile option if you already have a local json file of the mit_edx data).
- Check that the following query returns just an empty list:
  ```python
  LearningResourceRun.objects.filter(
    availability="Archived", 
    published=True, 
    learning_resource__etl_source="mit_edx"
  ).values_list("prices", flat=True).distinct()

  Out[3]: <TimestampedModelQuerySet [[]]>
  ```

### Additional Context
I thought about adding a migration to set prices = [] for published runs with "Archived" availability, but for now I decided it would be a bit redundant since the ETL scripts run every day.

### DIFF
--- a/learning_resources/etl/loaders.py
+++ b/learning_resources/etl/loaders.py
@@ -9,6 +9,7 @@ from django.contrib.auth import get_user_model
 from django.db import transaction
 
 from learning_resources.constants import (
+    AvailabilityType,
     LearningResourceFormat,
     LearningResourceRelationTypes,
     LearningResourceType,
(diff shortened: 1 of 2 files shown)


abeglova (approved):
👍

PR #1308, opened 2024-07-23
Extract department info for mitxonline from correct external API fields

mbertrand:
Assigns department ids in the mitxonline ETL pipeline based on the "departments" field in the MITx Online course/program API endpoints.

### How can this be tested?
- Set the `MITX_ONLINE_` env values based on RC
- Run `./manage.py backpopulate_mitxonline_data`
- Go to http://open.odl.local:8062/search/?resource_category=program&topic=Economics, you should get these 2 results:

Screenshot 2024-07-23 at 3 31 40 PM

### Additional Context
The MITx Online course and program API endpoints [return department info in different formats](https://github.com/mitodl/hq/issues/4970), so I wrote the function to parse them in a way that handles both.

### DIFF
--- a/learning_resources/etl/mitxonline.py
+++ b/learning_resources/etl/mitxonline.py
@@ -19,8 +19,8 @@ from learning_resources.constants import (
 )
 from learning_resources.etl.constants import ETLSource
 from learning_resources.etl.utils import (
-    extract_valid_department_from_id,
     generate_course_numbers_json,
+    get_department_id_by_name,
     parse_certification,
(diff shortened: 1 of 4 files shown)


shanbady (approved):
non-blocker but something i was curious about. It seems like mitxonline has an api for departments https://mitxonline.mit.edu/api/v2/departments/ - is it worth mapping from this vs the DEPARTMENTS constant? Also is it possible for (or how often will) the constant to diverge from live list of departments

PR #1313, opened 2024-07-24
display excerpt of resources

ChristopherChudzicki:
Adds a way to display an excerpt of the learning resource to learning resource card stories

Screenshot 2024-07-24 at 11 55 50 AM

### How can this be tested?
View, e.g., http://localhost:6006/?path=/story/smoot-design-cards-learningresourcecard--paid-course-without-certificate


gumaerc:
@ChristopherChudzicki This looks good and useful, but can we also add it to `LearningResourceListCard` and `LearningResourceListCardCondensed`? I added all of these scenarios to those as well and it would be good to see the data behind them there as well.

PR #1314, opened 2024-07-24
Reorder where the testimonial displays in the unit/offeror page to fix spacing and background

jkachel:

Closes mitodl/hq#4968

This PR does two things:
- Moves the `TestmonialDisplay` component to the `UnitChannelTemplate` from where it was in `ChannelPage` so that it's not constrained by the component it was residing in.
- Adds a minor tweak to the left/right padding so that the testimonial display is even with the header components.

Desktop:

image
image

(Green lines provided by xScope to show alignment.)

Mobile:

image
image

### How can this be tested?

Load a unit page. The testimonial display should look correct according to the design in FIgma.


jonkafton:
LGTM!

jonkafton (approved):

PR #1317, opened 2024-07-24
Convert durations to ISO8601 format (podcast episodes)

mbertrand:
Closes https://github.com/mitodl/hq/issues/4432

Converts podcast episode duration (if present) to ISO-8601 format.

### How can this be tested?
- Run `./manage.py backpopulate_podcast_data`
- It should complete successfully after which all published `PodcastEpisode` objects should have a duration that starts with "PT" or be null.
```python
from learning_resources.models import PodcastEpisode
all_eps = PodcastEpisode.objects.filter(learning_resource__published=True)
pt_eps = all_eps.filter(duration__startswith="PT)
null_eps = all_eps.filter(duration__isnull=True)
print(all_eps.count(), pt_eps.count(), null_eps.count())
assert all_eps.count() == pt_eps.count() + null_eps.count()

>1781 1744 37
```

### Additional Context
There are some python libraries to output durations in 8601 format ([isodate](https://pypi.org/project/isodate/), [isodate2](https://pypi.org/project/isodate2/)) but neither have been updated since late 2021 so I decided to roll my own function.  

I placed the `iso8601_duration` function in `etl.utils` rather than `etl.podcasts` just in case we might want to use it for something other than podcast episodes.  For example, maybe videos if we ever start ingesting them from somewhere other than Youtube (where it is already in ISO-8601 format).

### DIFF
--- a/learning_resources/etl/podcast.py
+++ b/learning_resources/etl/podcast.py
@@ -12,6 +12,7 @@ from requests.exceptions import HTTPError
 
 from learning_resources.constants import LearningResourceType
 from learning_resources.etl.constants import ETLSource
+from learning_resources.etl.utils import iso8601_duration
 from learning_resources.models import PodcastEpisode
 from main.utils import clean_data, frontend_absolute_url, now_in_utc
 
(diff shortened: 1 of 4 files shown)


shanbady (approved):
looks good. minor non-blocking suggestion

PR #1318, opened 2024-07-24
Fix channel views test

mbertrand:
Closes https://github.com/mitodl/hq/issues/4982

Fixes a flaky test in channels/views_test.py

### How can this be tested?
Run `pytest channels/views_test.py` multiple times, all tests should always pass.


jkachel (approved):
LGTM 👍

PR #1325, opened 2024-07-26
Updated designs for the unit page

jonkafton:

Closes: https://github.com/mitodl/hq/issues/4976

Applies [current designs](https://www.figma.com/design/Eux3guSenAFVvNHGi1Y9Wm/MIT-Design-System?node-id=3769-55562&m=dev) to the units page.

Unit cards are updated to the desktop card design and display the heading text in place of sub-heading text.

At mobile screen widths the designs are preserved and continue to display the subheading, pending any updates to the [designs for mobile](https://www.figma.com/design/Eux3guSenAFVvNHGi1Y9Wm/MIT-Design-System?node-id=4083-93794&m=dev).

image

---

image

### How can this be tested?

Navigate to `/units/` and confirm that the desktop designs and text updates are applied correctly.


data_fixtures/migrations/0002_unit_page_copy_updates.py
```diff
@@ -61,8 +61,9 @@
         },
         "channel_configuration": {
             "heading": (
-                "Join a powerful network of innovators and master "
-                "skills the global market needs in years to come."
+                "Offering lifelong learning opportunities that prepare engineering, "
```
shanbady, line 64: This change needs to be done as a completely new migration since this specific migration will already have been run on other environments. I would create a new blank migration via ```./manage.py makemigrations data_fixtures --empty``` (rename it to something logical) and then edit from there.

shanbady (changes requested):

data_fixtures/migrations/0002_unit_page_copy_updates.py
jonkafton: Thanks, makes sense. I have added a new [migration file](https://github.com/mitodl/mit-open/pull/1325/files#diff-ba9700e6052bb7089ed42a6a67e69b354f1db3997bc96416eb4c821935bdd242) based on the others. Is that a correct approach though? I notice the others have `# Generated by Django ...` comments. The migration gets applied ok in any case.

I've also added some fixes to the card loading styles as I noticed they were off in the process.

shanbady (approved):
Looks good 👍

PR #1326, opened 2024-07-26
urlencode search_filter

ChristopherChudzicki:
Fixes a bug exposed during https://github.com/mitodl/mit-open/pull/1323

Starting with https://github.com/mitodl/mit-open/pull/1275, some of our topic names have ampersands. The current channel creation code was creating search filters like
```json
{
   "search_filter": "topic=Education & Teaching"
}
```
The frontend expects `search_filter` to be parseable as a URLSearchParams, so ampersands were being interpreted as separate parameters: `topic=Education&topic=Teaching`.

This PR ensures that the search filter is urlencoded .

### How can this be tested?
1. Make sure that you've run the data migration from  https://github.com/mitodl/mit-open/pull/1323 and re-run the ETL pipelines to associate resources with topics.
2. **Suggested:** On `main`, go to `/topics` on the frontend. Click any topic that has an ampersand in its title. You will be navigated to the topic detail page.
    - expected: On `main`, the search results are empty.
    - compare to the API search results,  e.g., http://localhost:8063/api/v1/learning_resources_search/?topic=Education+%26+Teaching ... it should have results 
3. Now on this branch (restart docker, or at least celery!), run `./manage.py backpopulate_resource_channels --topic --overwrite` 
    - **Note:** this command only creates a channel if some resource uses that topic. So you need to make sure you're testing a topic that actually has resources associated with it.
4. Check the topic detail page from (2) again. It should have search results.

### DIFF
--- a/channels/plugins.py
+++ b/channels/plugins.py
@@ -1,5 +1,7 @@
 """Plugins for channels"""
 
+from urllib.parse import urlencode
+
 from django.apps import apps
 from django.utils.text import slugify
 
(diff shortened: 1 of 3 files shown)


jkachel (commented):
LGTM 👍 
Tested with "Business & Management" (as that had courses in it for me locally) - on main the filter has a & but in the branch it is properly urlencoded, and the topic channel page works.

jkachel (approved):
misclicked

PR #1330, opened 2024-07-29
add default yearly_decay_percent

abeglova:
 closes https://github.com/mitodl/hq/issues/5018

Sets the default yearly_decay_percent to 2.5. The parameter does not appear in the search unless it's set to something other than 2.5

### How can this be tested?
Run `docker-compose run web ./manage.py recreate_index --all` if you haven't recently
Log in as an admin.
Got to http://localhost:8062/search/?q=molecular+biology
Verify that by default the "Resource Score Staleness Penalty" slider is set to 2.5. Verify that you can change the value of the slider value and the larger the penalty the further down the the results older resources go
Log out
Verify that you get the same results at http://localhost:8062/search/?q=molecular+biology as you did as an admin with the default  "Resource Score Staleness Penalty" slider setting


shanbady (approved):
Changes look good and works. going to leave non-blocking comment - if we expect to change this frequently it might be worth it to move this constant into app settings

abeglova:
I don't think this will change - but if we need to change it I will add it into app settings

PR #1331, opened 2024-07-29
Subscription management page

shanbady:
Closes https://github.com/mitodl/hq/issues/4891

Adds a "Preferences" page to the dashboard that contains a list of channels a user has subscribed to (from which they can unsubscribe)

Screenshot 2024-07-29 at 1 59 09 PM
Screenshot 2024-07-29 at 1 59 43 PM

### How can this be tested?
1. Checkout this branch
2. Subscribe to a few channels of different types (unit, department, topic)
3. go to the settings page by navigating to the dashboard and clicking "settings"
Screenshot 2024-07-29 at 1 59 09 PM

4. Unsubscribe from topics and see that it works as expected.

### Additional context
Functionally the unsubscribe buttons on this preferences page are the same as if you unsubscribed directly from the relevant channel's page. We are just consolidating all the subscriptions in one location here.


learning_resources_search/models.py
```diff
@@ -30,3 +34,28 @@ def __str__(self):
...
     class Meta:
         unique_together = (("source_type", "original_query"),)
+
+    def original_url_params(self):
+        ignore_params = ["endpoint"]
+        query = self.original_query
+        defined_params = {
+            key: query[key] for key in query if query[key] and key not in ignore_params
+        }
+        return unquote_plus(urlencode(defined_params, doseq=True))
```
ChristopherChudzicki, line 44: I think this should be
```suggestion
        return urlencode(defined_params, doseq=True)
```
because `Channel.search_filter` will urlencoded (after #1326)

frontends/mit-open/src/pages/DashboardPage/SettingsPage.tsx
```diff
@@ -0,0 +1,96 @@
...
+const TitleText = styled(Typography)(({ theme }) => ({
+  marginTop: "20px",
+  marginBottom: "5px",
+
+  color: theme.custom.colors.darkGray2,
+  ...theme.typography.subtitle1,
+}))
+
+const SubTitleText = styled(Typography)(({ theme }) => ({
+  marginBottom: "10px",
```
ChristopherChudzicki, line 43: Curious: Did 10px come from Figma? Steve and Bilal usually use multiples of 8 essentially always use at least multiples of 4.

**Request**: Could you double check the margins/paddings/gap.

frontends/mit-open/src/pages/DashboardPage/SettingsPage.tsx
```diff
@@ -0,0 +1,96 @@
...
+const StyledLink = styled(Link)(({ theme }) => ({
+  color: theme.custom.colors.red,
+}))
+
+const TitleText = styled(Typography)(({ theme }) => ({
+  marginTop: "20px",
+  marginBottom: "5px",
+
+  color: theme.custom.colors.darkGray2,
+  ...theme.typography.subtitle1,
```
ChristopherChudzicki, line 39:
```suggestion
  ...theme.typography.h5,
```
Screenshot 2024-07-30 at 3 39 51 PM

ChristopherChudzicki (changes requested):
This generally works well thought I noticed a couple issues.

**Styling:** Doesn't quite match figma in terms of spacing and, in a couple places, font sizes. Apologies we don't have a more immediately useable ListItem for you.

**Functionality:** I noticed some issues around url encoding. Note that we started url encoding search filters for channels in #1326 because new topic names have ampersands. I noticed two things:

1. On the new subscripts page, channels subscriptions to topics with ampersands aren't compared correctly to the original search. I believe the _immediate_ solution is to NOT `unquote_plus` the search filter.
2. Topics with commas in their name can't be subscriped to.
    - this bug exists on `main`, but imo makes sense to fix here. I believe `getSearchParamMap` should NOT split on commas.

It seems to me that (1) and (2) above might both be easier if we stored `search_filter`—both for channels and for subscription queries—as objects rather than encoded strings, but that seems longer term.

frontends/mit-open/src/pages/DashboardPage/SettingsPage.tsx
```diff
@@ -0,0 +1,96 @@
...
+
+  return (
+    <>
+      <TitleText>Following</TitleText>
+      <SubTitleText>
+        All topics, academic departments, and MIT units you are following.
+      </SubTitleText>
+      <FollowList>
+        {subscriptionList?.data?.map((subscriptionItem) => (
+          <ListItem
```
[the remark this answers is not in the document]
shanbady: Adopted this change 👍

frontends/mit-open/src/pages/DashboardPage/SettingsPage.tsx
shanbady: checked and fixed

shanbady:
@ChristopherChudzicki this is ready to be looked at again

frontends/mit-open/src/pages/DashboardPage/SettingsPage.tsx
```diff
@@ -0,0 +1,131 @@
...
+    gap: "16px",
+    alignItems: "center",
+    border: `1px solid ${theme.custom.colors.lightGray2}`,
+    ":not(:last-child)": {
+      borderBottom: "none",
+    },
+    ":first-child, :last-child": {
+      borderRadius: "4px",
+    },
+  },
```
ChristopherChudzicki, line 53: Just use borderBottom for the "divider". Rest of border is on whole list.
```suggestion
    borderBottom: `1px solid ${theme.custom.colors.lightGray2}`,
    ":last-child": {
      borderBottom: "none",
    },
  },
```

ChristopherChudzicki (changes requested):
Works and looks great. Noticed a small issue with borders.

I also think it would be worth adding a simple frontend test, along the lines of 
```tsx
// SettingsPage.test.tsx
import React from "react"
import { SettingsPage } from "./SettingsPage"
import { renderWithProviders } from "@/test-utils"

describe("SettingsPage", () => {
  it("Renders user subscriptions in a list", () => {

  })

  test("Clicking 'Unfollow' removes the subscription", () => {

  })
})
```

frontends/mit-open/src/pages/DashboardPage/DashboardPage.tsx
```diff
@@ -480,6 +496,19 @@ const DashboardPage: React.FC = () => {
                     </div>
                   )}
                 </TabPanelStyled>
+                <TabPanelStyled
+                  key={TabValues.SETTINGS}
+                  value={TabValues.SETTINGS}
+                >
+                  <TitleText role="heading">Settings</TitleText>
+                  {isLoadingProfile || typeof profile === "undefined" ? (
```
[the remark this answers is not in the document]
shanbady: fixed

frontends/mit-open/src/pages/DashboardPage/SettingsPage.test.tsx
```diff
@@ -0,0 +1,66 @@
...
+  test("Clicking 'Unfollow' removes the subscription", async () => {
+    const { unsubscribeUrl } = setupApis({
+      isAuthenticated: true,
+      isSubscribed: true,
+      subscriptionRequest: {},
+    })
+    renderWithProviders(<SettingsPage />)
+
+    const followList = await screen.findByTestId("follow-list")
+    const unsubscribeButton = within(followList).getAllByText("Unfollow")[0]
```
ChristopherChudzicki, line 57: I try to use `getByRole*` for most things, since it effectively adds an assertion about the role in the accessibility treat. E.g., if we know it's an element with role link, it's almost certainly tab-navigable[^1]
```suggestion
    const unsubscribeButton = within(followList).getAllByRole("link", { name: "Unfollow" })[0]
```

[^1] I say "almost certainly" because it's probably an anchor tag. If someone did `<div role="link" />` (don't!), that would show up as a link in the browser's accessibility treat, but wouldn't be tab navigable.

ChristopherChudzicki (approved):
👍 LGTM

PR #1332, opened 2024-07-29
Avoid course overwrites in program ETL pipelines

mbertrand:
Closes https://github.com/mitodl/hq/issues/4960

Adds a new course config option, "fetch_only",  meant to be used in program ETL pipelines, that when True will prevent any courses included in the program from being upserted.  Only existing courses will be added to the program.

### How can this be tested?
- Set the `EDX_API_` and `MICROMASTERS_` env variables to the same values as RC
- Run `./manage.py backpopulate_mit_edx_data` (with `--api_datafile <mit_edx.json file>` if you've previously saved the output).
- Go to http://open.odl.local:8062/search?offered_by=mitx&q=%22Supply+Chain+Design%22 and click the top result, you should see a start date of "January 08, 2025" on both the card and drawer.
-  Run `./manage.py backpopulate_micromasters_data 
- Go back to the url above, the start date should still be there in both the card and drawer.
- Find the id for the program "Supply Chain Management".  Then go to http://api.open.odl.local:8063/api/v1/learning_resources/<program_id>/items/ and make sure that the course "Supply Chain Design" is included.

### DIFF
--- a/learning_resources/etl/loaders.py
+++ b/learning_resources/etl/loaders.py
@@ -339,11 +339,15 @@ def load_course(  # noqa: C901
         if config.fetch_only:
             # Do not upsert the course, it should already exist.
             # Just find it and return it.
-            return LearningResource.objects.filter(
+            resource = LearningResource.objects.filter(
                 readable_id=resource_id,
                 platform=platform,
                 resource_type=LearningResourceType.course.name,
+                published=True,
             ).first()
+            if not resource:
+                log.warning("No published resource found for %s", resource_id)
+            return resource
 
         if unique_field_name != READABLE_ID_FIELD:
             # Some dupes may result, so we need to unpublish resources
--- a/learning_resources/etl/loaders.py
+++ b/learning_resources/etl/loaders.py
@@ -335,6 +335,16 @@ def load_course(  # noqa: C901
             unique_field_value,
         )
         resource_id = deduplicated_course_id or readable_id
+
+        if config.fetch_only:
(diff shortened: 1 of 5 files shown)


learning_resources/etl/loaders_test.py
```diff
@@ -596,6 +596,30 @@ def test_load_course_dupe_urls(unique_url):
         assert course.published is (unique_url is False)
 
 
+@pytest.mark.parametrize("course_exists", [True, False])
+def test_load_course_no_fetch(mocker, course_exists):
+    """When no_fetch is True, course should just be fetched from db"""
```
abeglova, line 601: This should be be fetch_only and not no_fetch

learning_resources/etl/loaders.py
```diff
@@ -335,6 +335,16 @@ def load_course(
...
         resource_id = deduplicated_course_id or readable_id
+
+        if config.fetch_only:
+            # Do not upsert the course, it should already exist.
+            # Just find it and return it.
+            return LearningResource.objects.filter(
+                readable_id=resource_id,
+                platform=platform,
+                resource_type=LearningResourceType.course.name,
+            ).first()
```
abeglova, line 346: Should there be a warning if the course is not found?

abeglova (approved):
back to you with a few comments

learning_resources/etl/loaders.py
mbertrand: added

PR #1333, opened 2024-07-29
dev mode for search

abeglova:
closes https://github.com/mitodl/hq/issues/4984
closes https://github.com/mitodl/hq/issues/5023

This pr adds a dev_mode parameter to both the learning resource and content file searches. If dev_mode=True the api returns raw open search results with explain=True set. This will allow us to more easily troubleshoot search issues and understand the search results better.

It also fixes a bug that breaks searching content files by resource_id

### How can this be tested?
Search for http://localhost:8063/api/v1/content_file_search/?dev_mode=True&q=nutrition
Search for http://localhost:8063/api/v1/learning_resources_search/?dev_mode=True&q=nutrition

Verify that both searches work and you see the raw open search results and explanation.

Additionally try searching for
http://localhost:8063/api/v1/content_file_search/?resource_id=3669
this should not throw an error

Finally,
http://localhost:8062/search should work the same as previously

### Additional Context
One of the motivating reasons for this PR was to see why "Nuclear Energy: Science, Systems and Society" was showing up when you search for "nutrition". This turned out to be an ok result - that course had a two content files that addressed the consequences of nuclear poisoning on the gi track. The explain function does not actually give a good explanation when the score comes from content files, all you get is
```
 {
            "value": 8.823367,
             "description": "A match, join value 3669",
              "details": []
 },
```
to let you know that the score is coming from the join but
Search for http://localhost:8063/api/v1/learning_resources_search/?dev_mode=True&q=nutrition&resource_id=3669 returns more details about where the match is coming from

### DIFF
--- a/learning_resources_search/api.py
+++ b/learning_resources_search/api.py
@@ -21,11 +21,11 @@ from learning_resources_search.constants import (
     DEPARTMENT_QUERY_FIELDS,
     LEARNING_RESOURCE,
     LEARNING_RESOURCE_QUERY_FIELDS,
-    LEARNING_RESOURCE_SEARCH_FILTERS,
     LEARNING_RESOURCE_TYPES,
     RESOURCEFILE_QUERY_FIELDS,
     RUN_INSTRUCTORS_QUERY_FIELDS,
     RUNS_QUERY_FIELDS,
+    SEARCH_FILTERS,
     SOURCE_EXCLUDED_FIELDS,
     TOPICS_QUERY_FIELDS,
 )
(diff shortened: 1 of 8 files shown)


learning_resources_search/api.py
```diff
@@ -21,11 +21,11 @@
     DEPARTMENT_QUERY_FIELDS,
     LEARNING_RESOURCE,
     LEARNING_RESOURCE_QUERY_FIELDS,
-    LEARNING_RESOURCE_SEARCH_FILTERS,
     LEARNING_RESOURCE_TYPES,
     RESOURCEFILE_QUERY_FIELDS,
     RUN_INSTRUCTORS_QUERY_FIELDS,
     RUNS_QUERY_FIELDS,
+    SEARCH_FILTERS,
```
abeglova, line 28: Renamed because LEARNING_RESOURCE_SEARCH_FILTERS contains both Learning Resource and Content File search entries

gumaerc (approved):
👍 LGTM!

Just FYI, I didn't have the particular resource you were referencing in your instructions in my local DB, but I tested with some other resources and the explain function seems to be doing what it's supposed to do.

PR #1336, opened 2024-07-29
Show "Start Anytime" based on resource "availability" property

ChristopherChudzicki:
Closes https://github.com/mitodl/hq/issues/4743

This uses the new "availability" property on resources to display "Starts: Anytime" on resource cards.

**Vertical Cards, small**
Screenshot 2024-07-30 at 9 47 13 AM

**Vertical Cards, medium**
Screenshot 2024-07-30 at 10 11 50 AM

**Horizontal, small screen**
Screenshot 2024-07-30 at 11 05 13 AM

**Horizontal, wide screen**
Screenshot 2024-07-30 at 11 04 34 AM

**Drawer:**
Screenshot 2024-07-30 at 9 55 34 AM Screenshot 2024-07-30 at 9 56 04 AM

### How can this be tested?
1. If you haven't re-run backpopulates since #1301 , run them using RC creds.
    ```sh
    ./manage.py backpopulate_ocw_data --skip-contentfiles
    ./manage.py backpopulate_oll_data
    # The edX API has a rate limit; this may 429 if other people (or RC) has run it recently.
    ./manage.py extract_openedx_data
    ```
2. View course cards and resource drawer. If the API returns `available: "anytime"` for courses, you should see "Starts: Anytime". Otherwise, you'll see a date like before. Notes:
    - We previously showed "As taught in..." on cards for OCW courses. We do not anymore.
    - "Starts Anytime" is NOT limited to OCW (but is limited to courses and programs in the UI).
    - Card locations:
        - vertical cards:  homepage, unit channel pages, and dashboard for **vertically oriented cards** (Dashboard has small cards)
        - horizontal cards: channel page, search page
3. Check the resource drawer (click any card). If a card displays "Start anytime", it should display "As taught in..." in the resource drawer.
4. View new storybook stories: http://localhost:6006/?path=/story/smoot-design-cards-learningresourcecard--start-anytime (and dated nearby), and similar stories for the list card.

### DIFF
--- a/frontends/mit-open/.storybook/main.ts
+++ b/frontends/mit-open/.storybook/main.ts
@@ -1,5 +1,6 @@
 import { resolve, join, dirname } from "path"
 import * as dotenv from "dotenv"
+import * as webpack from "webpack"
 
 dotenv.config({ path: resolve(__dirname, "../../../.env") })
 
@@ -32,13 +33,17 @@ const config = {
   docs: {
     autodocs: "tag",
   },
-  env: (config: any) => ({
-    ...config,
-    APP_SETTINGS: {
-      PUBLIC_URL: process.env.PUBLIC_URL || "",
-      EMBEDLY_KEY: process.env.EMBEDLY_KEY || "",
-    },
-  }),
+  webpackFinal: async (config: any) => {
+    config.plugins.push(
+      new webpack.DefinePlugin({
+        APP_SETTINGS: {
+          EMBEDLY_KEY: JSON.stringify(process.env.EMBEDLY_KEY),
+          PUBLIC_URL: JSON.stringify(process.env.PUBLIC_URL),
+        },
+      }),
+    )
+    return config
+  },
 }
 
 export default config
--- a/frontends/mit-open/.storybook/preview-head.html
+++ b/frontends/mit-open/.storybook/preview-head.html
@@ -3,8 +3,3 @@
       WARNING: This is linked to chudzick@mit.edu's Adobe account.
   -->
 <link rel="stylesheet" href="https://use.typekit.net/lbk1xay.css" />
-<script>
-  window.APP_SETTINGS = {
-    embedlyKey: null,
-  }
-</script>
--- a/frontends/ol-components/src/components/Card/Card.test.tsx
+++ b/frontends/ol-components/src/components/Card/Card.test.tsx
@@ -1,0 +1,17 @@
+import { render } from "@testing-library/react"
+import { Card } from "./Card"
+import React from "react"
+
+describe("Card", () => {
+  test("has class MitCard-root on root element", () => {
+    const { container } = render(
+      <Card className="Foo">
+        <Card.Content>Hello world</Card.Content>
+      </Card>,
+    )
+    const card = container.firstChild
+
+    expect(card).toHaveClass("MitCard-root")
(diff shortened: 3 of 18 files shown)


jonkafton:
On https://github.com/mitodl/mit-open/pull/1336/commits/cf836ea213f692f8512dd6402723b95d96887985 for overriding the slot components. It does seem reasonable to provide that as an option. Perhaps the cleanest option - though until you know what it's doing, it does look a bit odd to see the FC passed as prop and then its component in the children. We're steering back to the original of just passing components as props (which I'm not keen on as a matter of style).

Some alternative options:

- Look at ways in Card.tsx to detect that a passed component is a styled slot so we can do styled(Card.Info) and use in its normal child slot position (as e.g. `<StyledInfo />`). There may be no trace and I'd rule out if can't be done reasonably cleanly. We don't want to be digging through prototypes or anything like that.

- Style from the container with nested target css, ideally using the component selector pattern for emotion generated classnames, ie. we wrap Card.Info with styled() so e.g. in LearningResourceCard can do: 

```
const LearningResourceCardContainer = styled.div`
  ${Card.Info} {
    height: 18px;
  }
`
```

We'd need a build plugin for this. I tried just now and get "Component selectors can only be used in conjunction with @emotion/babel-plugin, the swc Emotion plugin, or another Emotion-aware compiler transform." - though this would be useful across the project.

- The most straightforward, also styling from the implementing container is to give each slot in Card.tsx a classname so we can do:

```
const LearningResourceCardContainer = styled.div`
  .CardInfo {
    height: 18px;
  }
` 
```

ChristopherChudzicki:
@jonkafton Thanks for the thoughts on #cf836ea 

> Look at ways in Card.tsx to detect that a passed component is a styled slot so we can do styled(Card.Info) and use in its normal child slot position (as e.g. <StyledInfo />). There may be no trace and I'd rule out if can't be done reasonably cleanly. We don't want to be digging through prototypes or anything like that.

I believe the only way to detect that the child is a `styled(Card.Info)` would be via `__emotion_base` property on styled components. IMO this doesn't seem like part of their public API and is too tied to emotion. 

> Style from the container with nested target css, ideally using the component selector pattern for emotion ... We'd need a build plguin for this.

I gave the card roots a class `"MitCard-root"`, similar to how MUI does their "anatomy" classes. Doing similar for title, image, etc would certainly work.

I'm somewhat disinclined to do the JSX Component selector plugins because it's not clear to me that other CSS-in-JS libraries do/will support it, and ideally we'd switch from emotion to to something with zero runtime. (E.g., https://github.com/mui/pigment-css, which MUI is building for their v6). That would obviously be a very long term switch for us, so perhaps not worth worrying about.

My inclination is to use `.MitCard-title`, `.MitCard-image`, etc and style from the container.

jonkafton:
> I believe the only way to detect that the child is a `styled(Card.Info)` would be via `__emotion_base` property on styled components. IMO this doesn't seem like part of their public API and is too tied to emotion.

I agree.

> My inclination is to use .MitCard-title, .MitCard-image, etc and style from the container.

That sounds sensible for a component library, as is the prefix namespacing. I'd match the component name exactly if we're allowing uppercase, e.g. .MitCard-Info, or .MIT-Card-Info (given MIT is not an acronym); splitting hairs though!

abeglova (approved):
👍

PR #1342, opened 2024-07-30
Handle alternate unique id fields better in load_course

mbertrand:
Closes https://github.com/mitodl/hq/issues/4735

- Changes how resources are handled in `load_course` when the unique field is something other than `readable_id`, that should fix the bug described in the issue above.
- Applies some of the same logic to both courses and programs during load, via a separate function.

### How can this be tested?
- Set `PROLEARN_CATALOG_API_URL=https://prolearn.mit.edu/graphql` in your .env file
- Run `./manage.py backpopulate_prolearn_data`
- Go to the django admin page for one of the ingested prolearn courses, you can see the list at http://api.open.odl.local:8063/admin/learning_resources/learningresource/?etl_source=prolearn
- Change the readable_id to something else.
- Rerun `./manage.py backpopulate_prolearn_data`
- Refresh the admin page for the same course.  Its readable_id should now be reverted back to the original value.
- As a sanity check/smoke test, you could run other backpopulate commands too to make sure they still produce the same output.

### DIFF
--- a/learning_resources/etl/loaders.py
+++ b/learning_resources/etl/loaders.py
@@ -350,21 +350,24 @@ def load_course(  # noqa: C901
             return resource
 
         if unique_field_name != READABLE_ID_FIELD:
-            # Some dupes may result, so we need to unpublish resources
-            # with matching unique values and different readable_ids
-            for resource in LearningResource.objects.filter(
+            resource_data[READABLE_ID_FIELD] = resource_id
+            # Some dupes may result, so we should delete all but the
+            # most recently updated resource w/matching unique value
+            existing_courses = LearningResource.objects.filter(
                 **{unique_field_name: unique_field_value},
                 platform=platform,
                 resource_type=LearningResourceType.course.name,
-            ).exclude(readable_id=resource_id):
-                resource.published = False
-                resource.save()
-                resource_unpublished_actions(resource)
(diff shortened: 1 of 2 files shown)


shanbady (changes requested):
@mbertrand when trying this locally - It appears to create a new record and set the old one (with the changed readable Id) to unpublished. is that expected or is the record for which the id is initially changed supposed to revert?

mbertrand:
@shanbady That's the behavior I'd expect on the main branch, did you restart all containers after switching to this branch?

Currently I have 201 resources from prolearn.  I ran the following code to change all their readable_id values:

```python
from django.db.models import F, Value
from django.db.models.functions import Concat

LearningResource.objects.filter(etl_source="prolearn").update(
    readable_id=Concat(F("readable_id"), Value("_modified"))
)
```

and checked that all readable ids for prolearn were appended with `_modified` at http://api.open.odl.local:8063/admin/learning_resources/learningresource/?etl_source=prolearn

Then I ran `./manage.py backpopulate_prolearn_data` again.   All the readable_id values were reverted to their orignal values and the number of resources remained the same at 201.

shanbady (approved):
Looks like I hadnt restarted celery which resolved the issue. LGTM

PR #1347, opened 2024-07-31
Use resource_delete_actions instead of resource.delete directly

mbertrand:
Related to https://github.com/mitodl/mit-open/pull/1342

One line change to delete resource via `resource_delete_actions` function instead of calling `resource.delete()` directly

### How can this be tested?
Tests should still pass
Running `backpopulate_prolearn_data` should still complete successfully.

### DIFF
--- a/learning_resources/etl/loaders.py
+++ b/learning_resources/etl/loaders.py
@@ -46,6 +46,7 @@ from learning_resources.utils import (
     bulk_resources_unpublished_actions,
     load_course_blocklist,
     load_course_duplicates,
+    resource_delete_actions,
     resource_run_unpublished_actions,
     resource_run_upserted_actions,
     resource_unpublished_actions,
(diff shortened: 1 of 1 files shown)


learning_resources/etl/loaders.py
```diff
@@ -347,7 +348,7 @@ def upsert_course_or_program(
         ).order_by("-updated_on")
         if existing_courses.count() > 1:
             for course in existing_courses[1:]:
-                course.delete()
+                resource_delete_actions(course)
```
ChristopherChudzicki, line 351: `resource_delete_actions` doesn't actually delete the resource, right—it just calls the associated plugins? Shouldn't we do both?

mbertrand: it does do both, via the associated plugin function: https://github.com/mitodl/mit-open/blob/main/learning_resources_search/plugins.py#L143-L148

```python
    @hookimpl
    def resource_delete(self, resource):
        """
        Remove a resource from the search index and then delete the object
        """
        self.resource_unpublished(resource)
        resource.delete()
```

shanbady (approved):
👍

PR #1353, opened 2024-08-01
Renaming my topic update migration from 0006 to 0007

jkachel:

mitodl/hq#5080

There ended up being two 0006 migrations in data_fixtures; this renames mine to 0007 so the graph is right.

### How can this be tested?

Roll back your data_fixtures to migration 0005 before checking out this branch. Then, check it out and migrate. Everything should be as expected. (You can re-populate courses if you want and they should work OK, per https://github.com/mitodl/mit-open/pull/1344 .)


gumaerc (approved):
👍

PR #1361, opened 2024-08-02
Fix issue with User List cards not updating on edit

jonkafton:

Fixes https://github.com/mitodl/hq/issues/5100

Fixes the issue that User List cards do not update when edited.
- Invalidates the list query cache after the user list update mutation so that the useUserListList() hook triggers render of the User List page.

### How can this be tested?

While authenticated, navigate to /dashboard/#my-lists

- Edit a list item
   - Confirm that the card updates without navigation or refresh.
- Delete a list item
   - Confirm that the card is removed navigation or refresh.

Regression check:
- Click a list item to open its detail page
- Edit the item
  - Confirm that the page updates with changes

### DIFF
--- a/frontends/api/src/hooks/learningResources/index.ts
+++ b/frontends/api/src/hooks/learningResources/index.ts
@@ -32,6 +32,7 @@ import type {
 import learningResources, {
   invalidateResourceQueries,
   invalidateUserListQueries,
+  invalidateResourceWithUserListQueries,
   updateListParentsOnAdd,
   updateListParentsOnDestroy,
 } from "./keyFactory"
@@ -251,8 +252,10 @@ const useUserListUpdate = () => {
         PatchedUserListRequest: params,
       }),
     onSettled: (_data, _err, vars) => {
-      invalidateResourceQueries(queryClient, vars.id)
       queryClient.invalidateQueries(learningResources.userlists._ctx.list._def)
+      queryClient.invalidateQueries(
+        learningResources.userlists._ctx.detail(vars.id).queryKey,
+      )
     },
   })
 }
@@ -262,8 +265,9 @@ const useUserListDestroy = () => {
   return useMutation({
     mutationFn: (params: ULDestroyRequest) =>
       userListsApi.userlistsDestroy(params),
-    onSettled: (_data, _err) => {
-      queryClient.invalidateQueries(learningResources.userlists._ctx.list._def)
+    onSettled: (_data, _err, vars) => {
+      invalidateUserListQueries(queryClient, vars.id)
+      invalidateResourceWithUserListQueries(queryClient, vars.id)
     },
   })
 }
--- a/frontends/api/src/hooks/learningResources/index.ts
+++ b/frontends/api/src/hooks/learningResources/index.ts
@@ -252,6 +252,7 @@ const useUserListUpdate = () => {
       }),
     onSettled: (_data, _err, vars) => {
       invalidateResourceQueries(queryClient, vars.id)
+      queryClient.invalidateQueries(learningResources.userlists._ctx.list._def)
     },
   })
 }
@@ -261,8 +262,8 @@ const useUserListDestroy = () => {
   return useMutation({
     mutationFn: (params: ULDestroyRequest) =>
       userListsApi.userlistsDestroy(params),
-    onSettled: (_data, _err, vars) => {
-      invalidateResourceQueries(queryClient, vars.id)
+    onSettled: (_data, _err) => {
+      queryClient.invalidateQueries(learningResources.userlists._ctx.list._def)
     },
   })
 }
--- a/frontends/api/src/hooks/learningResources/keyFactory.ts
+++ b/frontends/api/src/hooks/learningResources/keyFactory.ts
@@ -155,7 +155,24 @@ const listHasResource =
           (page: PaginatedLearningResourceList) => page.results,
         )
       : data.results
+
     return resources.some((res) => res.id === resourceId)
+  }
+
+const resourceHasUserList =
+  (userListId: number) =>
+  (query: Query): boolean => {
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    const data = query.state.data as any
+    const resources: LearningResource[] = data.pages
+      ? data.pages.flatMap(
+          (page: PaginatedLearningResourceList) => page.results,
+        )
+      : data.results
+
+    return resources?.some((res) =>
+      res.user_list_parents?.some((userList) => userList.parent === userListId),
+    )
   }
 
 /**
@@ -202,6 +219,31 @@ const invalidateResourceQueries = (
   })
 }
 
+/**
+ * Invalidate Resource queries that a resource that belongs to user list appears in.
+ */
+const invalidateResourceWithUserListQueries = (
+  queryClient: QueryClient,
+  userListId: LearningResource["id"],
+  { skipFeatured = false } = {},
+) => {
+  /**
+   * Invalidate lists with a resource that is in the user list.
+   */
+  const lists = [
+    learningResources.list._def,
+    learningResources.learningpaths._ctx.list._def,
+    learningResources.search._def,
+    ...(skipFeatured ? [] : [learningResources.featured._def]),
+  ]
+  lists.forEach((queryKey) => {
+    queryClient.invalidateQueries({
+      queryKey,
+      predicate: resourceHasUserList(userListId),
+    })
+  })
+}
+
 const invalidateUserListQueries = (
   queryClient: QueryClient,
   userListId: UserList["id"],
@@ -210,6 +252,7 @@ const invalidateUserListQueries = (
     learningResources.userlists._ctx.detail(userListId).queryKey,
   )
   const lists = [learningResources.userlists._ctx.list._def]
+
   lists.forEach((queryKey) => {
     queryClient.invalidateQueries({
       queryKey,
@@ -275,6 +318,7 @@ export default learningResources
 export {
   invalidateResourceQueries,
   invalidateUserListQueries,
+  invalidateResourceWithUserListQueries,
   updateListParentsOnAdd,
   updateListParentsOnDestroy,
 }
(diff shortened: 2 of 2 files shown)


frontends/api/src/hooks/learningResources/index.ts
```diff
@@ -261,8 +262,8 @@ const useUserListDestroy = () => {
   return useMutation({
     mutationFn: (params: ULDestroyRequest) =>
       userListsApi.userlistsDestroy(params),
-    onSettled: (_data, _err, vars) => {
-      invalidateResourceQueries(queryClient, vars.id)
+    onSettled: (_data, _err) => {
+      queryClient.invalidateQueries(learningResources.userlists._ctx.list._def)
```
ChristopherChudzicki, line 266: There's a small bug here that you can experience in UI as follows:
1. Navigate to dashboard userlists
2. Create a userlist
3. navigate to dashboard "home"
4. bookmark a currently unbookmarked resource into your new list
5. navigate back to userlists
6. delete the userlist
7. navigate back to "home" ... you'll see the resource is still bookmarked.

Instead:
```suggestion
    onSettled: (_data, _err, vars) => {
        useUserListUpdate(vars.id)
```
This will invalidate the userlist and any resources that are using that userlist (otherwise their bookmarks will still show filled).

I will say...
- all this invalidation stuff is complicated. Part of me just wants to blow away the whole cache, but that doesn't seem great for some things, like sorting resources.
- we don't have a great way to test this invalidation stuff right now. The tests in `index.test.ts` are....not ideal. (neither is the filename)

frontends/api/src/hooks/learningResources/index.ts
```diff
@@ -252,6 +252,7 @@ const useUserListUpdate = () => {
       }),
     onSettled: (_data, _err, vars) => {
       invalidateResourceQueries(queryClient, vars.id)
+      queryClient.invalidateQueries(learningResources.userlists._ctx.list._def)
```
ChristopherChudzicki, line 255: You could remove the call to `invalidateResourceQueries`. It shouldn't be there. `invalidateResourceQueries` very clearly expects a `resourceId`, and userlists are not learning resources.

ChristopherChudzicki (changes requested):
Noticed one small issue.

frontends/api/src/hooks/learningResources/index.ts
ChristopherChudzicki: Realizing as we discuss in slack this invalidation isn't quite right: It invalidates the listing, but not the details. So if you make the edit on the details page, it will still be outdated.

Some options:
1. Invalidate `learningResources.userlists._def` ... This will invalidate ALL userlist queris: listing, details for list 1, details for list 2, details for list 3, etc, items for list 1, items for list 2, items for list 3, et 
2. Invalidate
    - `learningResources.userlists._ctx.list._def` the listing
        - minor comment: `_def` is sort of the parent key... `['learningResources', 'userlists', 'user_lists', 'list']`, whereas `learningResources.userlists._ctx.list(...).queryKey` will have a fifth entry that corresponds to the parameters on the listing api (e.g., pagesize, offset). 
    - `learningResources.userlists._ctx.details(vars.id).queryKey` just the details for this list, not details for other lists 
        - this will also invalidate the 

(1) is more aggressive—will cause some extra refteches, but it's also a bit simpler

jonkafton: All looking clearer now that it's not Friday night.

For the **update**, I'm clearing (option 2):
- `learningResources.userlists._ctx.list._def`, ie. `['learningResources', 'userlists', 'user_lists', 'list']`
   - covers the list query `useUserListList()` `['learningResources', 'userlists', 'user_lists', 'list', {…}]`
- `learningResources.userlists._ctx.detail(vars.id).queryKey`, e.g. `['learningResources', 'userlists', 'user_lists', 'detail', 22]`
   - matches the detail query `useUserListsDetail()` 

Updates correctly in the user list list view, detail view and "Add to list" dialogs.

For the **delete** I'm using our `invalidateUserListQueries(queryClient, vars.id)` utility, which clears:
  - `['learningResources', 'userlists', 'user_lists', 'detail', 22]`
  - `['learningResources', 'userlists', 'user_lists', 'list']`, with predicate.
 
This does not clear the association from learning resources that have been added to a list. If I'm understanding correctly, to do so we'd need to clear any query that returns that resource (`user_list_parents` is a property of the resource). On the dashboard home, for example:
 - Top picks for you - `["learningResources","search",{"resource_type":["course"],"limit":12,"learning_format":[],"sortby":"-views"}]`
 - Free courses - `["learningResources","search",{"resource_type":["course"],"limit":12,"free":true,"sortby":"-views"}]`
 - Popular - `["learningResources","search",{"limit":12,"sortby":"-views"}]`

On the home page (although I'm not sure you can currently get back to the homepage without refresh):
 - Featured Courses - e.g. `["learningResources","featured",{"limit":12}]`
 - Media e.g. `["learningResources","list",{"resource_type":["video","podcast_episode"],"limit":12,"sortby":"new"}]`

I've added an `invalidateResourceWithUserListQueries(userListId)` function that clears the following, predicated on the resource's `user_list_parents` containing the user list ID:
- `["learningResources", "detail"]`
- `['learningResources', 'list']`
- `['learningResources', 'learningpaths', 'learning_paths', 'list']`
- `['learningResources', 'search']`
- `['learningResources', 'featured']`

A thought - I'm not sure query caching is the best solution for our needs, given there's no cache normalization / concept of a keyed item that belongs to a collection that can be returned from various queries; though it does seem reasonable enough to query the cache state for our predicates. Not sure if this is a pattern or a workaround. I'm clearing the learningResource detail above, even though we don't display the bookmark in the expanded detail view - it seems sensible to clear anything that's stale in any cache regardless of what is rendered. Not an easy bug to spot though as we add more queries.

jonkafton:
All looking clearer now that it's not Friday night.

For the **update**, I'm clearing (option 2):
- `learningResources.userlists._ctx.list._def`, ie. `['learningResources', 'userlists', 'user_lists', 'list']`
   - covers the list query `useUserListList()` `['learningResources', 'userlists', 'user_lists', 'list', {…}]`
- `learningResources.userlists._ctx.detail(vars.id).queryKey`, e.g. `['learningResources', 'userlists', 'user_lists', 'detail', 22]`
   - matches the detail query `useUserListsDetail()` 

Updates correctly in the user list list view, detail view and "Add to list" dialogs.

For the **delete** I'm using our `invalidateUserListQueries(queryClient, vars.id)` utility, which clears:
  - `['learningResources', 'userlists', 'user_lists', 'detail', 22]`
  - `['learningResources', 'userlists', 'user_lists', 'list']`, with predicate.
 
This does not clear the association from learning resources that have been added to a list. If I'm understanding correctly, to do so we'd need to clear any query that returns that resource (`user_list_parents` is a property of the resource). On the dashboard home, for example:
 - Top picks for you - `["learningResources","search",{"resource_type":["course"],"limit":12,"learning_format":[],"sortby":"-views"}]`
 - Free courses - `["learningResources","search",{"resource_type":["course"],"limit":12,"free":true,"sortby":"-views"}]`
 - Popular - `["learningResources","search",{"limit":12,"sortby":"-views"}]`

On the home page (although I'm not sure you can currently get back to the homepage without refresh):
 - Featured Courses - e.g. `["learningResources","featured",{"limit":12}]`
 - Media e.g. `["learningResources","list",{"resource_type":["video","podcast_episode"],"limit":12,"sortby":"new"}]`

I've added an `invalidateResourceWithUserListQueries(userListId)` function that clears the following, predicated on the resource's `user_list_parents` containing the user list ID:
- `["learningResources", "detail"]`
- `['learningResources', 'list']`
- `['learningResources', 'learningpaths', 'learning_paths', 'list']`
- `['learningResources', 'search']`
- `['learningResources', 'featured']`

A thought - I'm not sure query caching is the best solution for our needs, given there's no cache normalization / concept of a keyed item that belongs to a collection that can be returned from various queries; though it does seem reasonable enough to query the cache state for our predicates. Not sure if this is a pattern or a workaround. I'm clearing the learningResource detail above, even though we don't display the bookmark in the expanded detail view - it seems sensible to clear anything that's stale in any cache regardless of what is rendered. Not an easy bug to spot though as we add more queries.

ChristopherChudzicki:
> I've added an invalidateResourceWithUserListQueries(userListId) function that clears the following, predicated on the resource's user_list_parents containing the user list ID:

Honestly, my inclination would be to invalidate `learningResources._def` when deleting a list. That blows away everything, but presumably deletion isn't happening very often.

> A thought - I'm not sure query caching is the best solution for our needs, given ... 

IMO the real problem changing a userlist or learningpath affects so many of our APIs. If we remove all userlist info from `/learning_resources/`, `/featured/`, `/learning_resources_search/`, etc, APIs and use a dedicated API for checking userlist membership, this all gets way simpler and more predictable.  (Something like `/userlists/membership/?ids=resource1,resource2,resource3,....`, similar to https://developer.amazon.com/docs/music/API_web_user.html#get-current-users-followed-playlists). Then:
- if you create/delete a userlist, you only need to invalidate (1)  userlist lists, (2) userlist membership, and (3) *for deletion only*, the details.
- since learning resource API responses do not vary by user, they could be cached.

ChristopherChudzicki (approved):

ChristopherChudzicki:
👍

jonkafton:
> > A thought - I'm not sure query caching is the best solution for our needs, given ...
> 
> Not caching has a variety of problems, too, like re-renders triggering API calls, and multiple components making same duplicate API calls.

Sorry, not clear - the alternative not being to not cache, but using another mechanism for preserving client state.

jonkafton:
> If we remove user_list_parents from /learning_resources/, /featured/, /learning_resources_search/, etc, APIs and use a dedicated API for checking userlist membership, this all gets way simpler and more predictable.

Yes, exactly that. The is an issue here with the API design (for this purpose at least, though may be the best design for the system overall). If the resource data didn't change depending on who was looking at it and these relations were looked up separately, the issue goes away.

PR #1364, opened 2024-08-05
Add Manufacturing topic; update upserter to make adding child topics easier

jkachel:

Closes mitodl/hq#5076

This adds the Manufacturing topic under Engineering. There are no mappings for this - in the original issue this is to cover a MicroMasters program, and it's got a topic called "Manufacturing" so would be covered by the default by-name mapping for that. 

This also adds a `parent` option to the topics yaml spec so when we're adding/changing sub-topics we can do that without including the entire tree. For example, in https://github.com/mitodl/mit-open/pull/1344 the update adds a topic under Computer Science, and has to include the Computer Science topic to do so. This could potentially be a vector for introducing errors into the topic list - if you grab topic data that's out of date then you'll overwrite changes that have been made with your new stuff. With this code, you can specify a parent UUID so you can just tell it what its parent should be.

### How can this be tested?

Topic: run the migrations for `data_fixtures` and you should see the new Manufacturing topic. It should have Engineering as its parent. It won't have any mappings or an icon.

Upserter updates: Automated tests should pass (there's one specific to this). Additionally, running the migration is a test since it uses the new code. There are changes in the `README-topics.md` too; they should make sense.

### DIFF
--- a/data_fixtures/migrations/0008_topics_update_engineering_add_manufacturing.py
+++ b/data_fixtures/migrations/0008_topics_update_engineering_add_manufacturing.py
@@ -37,7 +37,7 @@ def rollback_new_mapping(apps, schema_editor):
     Channel = apps.get_model("channels", "Channel")
 
     topic = LearningResourceTopic.objects.filter(
-        parent__topic_uuid="952604ab-ae23-45b3-a040-0e5f26fe42df", name="Manufacturing"
+        topic_uuid="4176e385-92c5-4e02-b3cc-4f34d9a4bf40",
     ).get()
 
     Channel.objects.filter(
--- a/data_fixtures/migrations/0008_topics_update_engineering_add_manufacturing.py
+++ b/data_fixtures/migrations/0008_topics_update_engineering_add_manufacturing.py
@@ -1,0 +1,59 @@
+"""
+Update the topics; add Manufacturing under Engineering
+This is intended to fix https://github.com/mitodl/hq/issues/5076 - no additional
+mappings since the topics for Principals of Manufacturing include "Manufacturing"
+so it should match on name directly.
+"""
+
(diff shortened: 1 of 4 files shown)


data_fixtures/migrations/0008_topics_update_engineering_add_manufacturing.py
```diff
@@ -0,0 +1,59 @@
...
+def rollback_new_mapping(apps, schema_editor):
+    """Remove the Manufacturing topic."""
+
+    LearningResourceTopic = apps.get_model(
+        "learning_resources", "LearningResourceTopic"
+    )
+    Channel = apps.get_model("channels", "Channel")
+
+    topic = LearningResourceTopic.objects.filter(
+        parent__topic_uuid="952604ab-ae23-45b3-a040-0e5f26fe42df", name="Manufacturing"
```
mbertrand, line 40: Not a big deal, but could this match on the topic uuid instead?

```python
topic = LearningResourceTopic.objects.filter(topic_uuid="4176e385-92c5-4e02-b3cc-4f34d9a4bf40").get()
```

mbertrand (approved):
Works as expected, just one minor comment. 👍

data_fixtures/migrations/0008_topics_update_engineering_add_manufacturing.py
jkachel: Fixed this in https://github.com/mitodl/mit-open/pull/1364/commits/17b20709226c96be0cb8da4038467d3aca5ca2b5

jkachel:
@mbertrand just a quick re-review for the rollback change - this shouldn't affect anything but just being cautious.

mbertrand (approved):
👍

PR #1368, opened 2024-08-06
Update CI to check data migrations for conflicts

jkachel:

Closes #1532

Adds `RUN_DATA_MIGRATIONS=True` in front of the poetry command to make migrations, which we use to check if there's missing migrations. This makes it include the `data_fixtures` app and thus will check for missing migrations there too (unlikely) or for conflicting migrations in that app (much more likely).

### How can this be tested?

Add a couple migrations with the same serial number to the `data_fixtures` app. (I just added a single empty one, then cut and pasted it.) 

Run `./scripts/test/detect_missing_migrations.sh` from the `web` container. It should fail.

If you commit these migrations to the branch (or to a sub-branch), it should also fail when the GitHub action runs. (Don't forget to revert the commit if you did it to this branch!)


ChristopherChudzicki (approved):
👍

PR #1374, opened 2024-08-06
Rename my 0008 to 0009 to prevent conflict

jkachel:

#1364 

There was an `0008` migration merged in so now there's a conflict - renaming mine to `0009`.

### How can this be tested?

Migrations should run successfully.


mbertrand (approved):
👍 assuming tests pass

PR #1376, opened 2024-08-06
Empty user list items view

jonkafton:

Closes https://github.com/mitodl/hq/issues/4786

Displays a view for no user list items.

image

### How can this be tested?

Log in an navigate to /dashboard/my-lists/.
- Delete all of your lists
- Validate that the empty view matches [designs](https://www.figma.com/design/Eux3guSenAFVvNHGi1Y9Wm/MIT-Design-System?node-id=5060-120326&m=dev) and that the create button functions.

### Additional Context

We have a manage.py job, `backpopulate_favorites_lists`, which gives each user am empty user list called "Favorites".

As we now have a view for no items, is this the default and do we want to remove this job?

We can run this with `--delete` to remove for all users (deletes lists with title "Favorites").


ChristopherChudzicki (changes requested):
Looking good. I did notice two things; "Issue 1" seems more important to me.

### Issue 1:
`"Create New List to start!"` seems like an awkward phrase to me, though I realize this matches Figma.

@mbilalmughal or @steven-hatch I recognize that "Create New List" is bold in the designs (below) and matches the button text, which might be desirable, but could I suggest  
- Option 1: Delete "Create New List to start!" ... The button below is enough
- Option 2: "Create a new list to start!" (maybe without bold, since it doesn't match button text anymore)

### Issue 2
I don't think this should block us, especially if you're working on something else important, but in Figma there's a `16px` gap that is `32px` in this implementation
Screenshot 2024-08-08 at 8 18 56 PM

mbilalmughal:
@ChristopherChudzicki I'm not sure about the text as Steve suggested it, but your suggestions also make sense., and I think the spacing needs to be corrected since the bottom space should be slightly larger than the top. 

The only small suggestion I have is to make the text a bit bigger, 14 change to 16px (see attached image)
cc: @steven-hatch 

image

steven-hatch:
Issue 1 
Thanks Chris. I like reiterating user action text with text in buttons. This was intentional but maybe not necessary. Lets go with "Option 1: Delete "Create New List to start!" ... The button below is enough"

Issue 2
I would like to get the 16px in there. This may appear minor, but gives balance to the empty state and groups it with instructions on top, and action on bottom. If it stays as is, everything is uniform, and appears thrown together. Changing the spacing will give it that extra design polish UI is looking for.

Note: I have intentions to reuse this empty state. Next week I will be pushing for better empty states for search experience and I will be using this as a starting point. Not related to this ticket, but just wanted to be transparent that this may be a component.

steven-hatch:
@mbilalmughal I dont want to change the design. @ChristopherChudzicki please proceed without any font size changes. We can discuss this as a team if further changes are needed. Thanks!

ChristopherChudzicki (approved):
👍

PR #1377, opened 2024-08-06
show featured resources first

abeglova:
closes https://github.com/mitodl/hq/issues/4986

This updates the default sort to show featured resources first. The resources are shown in order of their position in channel featured lists, with the order of resources with the same position randomized.  The order (including the random part) is hard coded in the index and updated each night

For non-featured resources, courses/programs should continue to be shown before learning resources.

The default sort only affects the order if there is no search term

### How can this be tested?
Check out the branch in https://github.com/mitodl/mit-open/pull/1381 which includes everything except the search query change so that the front end does not break during reindexing

Go to http://open.odl.local:8062/search/ and ensure that search works as expected 

Go to http://open.odl.local:8062/search/ and ensure that search works as expected 

Check out this branch 

Go to http://open.odl.local:8062/search/ and check that you see resources from the front end featured carousel first. The order will be slightly different from the carousel order since the order is randomized for resources that have the same position with in the featured list for different units

Run
docker-compose run web ./manage.py update_featured_rank
docker-compose run web ./manage.py update_index --courses
docker-compose run web ./manage.py backpopulate_mitxonline_data

After each command go to http://open.odl.local:8062/search/  and ensure that the featured resources remain the first results

**This pr should not be released until https://github.com/mitodl/mit-open/pull/1381 is released and the index is updated**

### DIFF
--- a/learning_resources_search/api.py
+++ b/learning_resources_search/api.py
@@ -9,7 +9,6 @@ from opensearch_dsl import Search
 from opensearch_dsl.query import MoreLikeThis, Percolate
 from opensearchpy.exceptions import NotFoundError
 
-from learning_resources.constants import LEARNING_RESOURCE_SORTBY_OPTIONS
 from learning_resources.models import LearningResource
 from learning_resources_search.connection import (
     get_default_alias_name,
@@ -21,6 +20,7 @@ from learning_resources_search.constants import (
     DEPARTMENT_QUERY_FIELDS,
     LEARNING_RESOURCE,
     LEARNING_RESOURCE_QUERY_FIELDS,
+    LEARNING_RESOURCE_SEARCH_SORTBY_OPTIONS,
     LEARNING_RESOURCE_TYPES,
     RESOURCEFILE_QUERY_FIELDS,
     RUN_INSTRUCTORS_QUERY_FIELDS,
@@ -39,7 +39,7 @@ log = logging.getLogger(__name__)
 
 LEARN_SUGGEST_FIELDS = ["title.trigram", "description.trigram"]
 COURSENUM_SORT_FIELD = "course.course_numbers.sort_coursenum"
-DEFAULT_SORT = ["is_learning_material", "-views"]
+DEFAULT_SORT = ["featured_rank", "is_learning_material"]
 
 
 def gen_content_file_id(content_file_id):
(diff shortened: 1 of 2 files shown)


learning_resources_search/indexing_api.py
```diff
@@ -43,17 +43,54 @@
 User = get_user_model()
 
 
+def clear_featured_rank(rank, clear_all_greater_then):
```
mbertrand, line 46: Maybe a typo, `clear_all_greater_than`  (than instead of then)?

learning_resources_search/indexing_api.py
```diff
@@ -43,17 +43,54 @@
...
+    """
+    Make a request to ES to set featured_rank to null for documents with
+    rank equal to rank
+
+    Args:
+        rank (int): the rank
```
mbertrand, line 52: add `clear_all_greater_than` in docstring

learning_resources_search/indexing_api.py
```diff
@@ -43,17 +43,54 @@
 User = get_user_model()
 
 
+def clear_featured_rank(rank, clear_all_greater_then):
+    """
+    Make a request to ES to set featured_rank to null for documents with
+    rank equal to rank
```
mbertrand, line 49: `equal to or greater than` ?

learning_resources_search/serializers.py
```diff
@@ -130,6 +132,20 @@ def serialize_learning_resource_for_update(
...
         ]
+
+    if learning_resource_obj.in_featured_lists > 0:
+        featured_rank = (
+            LearningResourceRelationship.objects.filter(
+                child_id=learning_resource_obj.id, parent__channel__isnull=False
+            )
+            .order_by("position")
+            .first()
+            .position
```
mbertrand, line 143: Can `featured_rank` also be calculated via an `annotate` call in the `ViewSet` query `for_search_serialization` function instead?  re https://github.com/mitodl/mit-open/pull/1303:

> We should not be performing database queries in serializers going forward (aside from validation, create(), and update()).

mbertrand (changes requested):
Works as expected, did notice one thing, maybe not a big deal: before I ran ` ./manage.py update_featured_rank`, only the four position # 1 featured resources were showing up first in the search results.  After running the command, all the featured resources showed up first in search results.

learning_resources_search/serializers.py
abeglova: I can play around with `Case` and `Subquery` to try to make it work but I don't think it would just make the code more complicated without any benefits to performance. Even though the function is called `serialize_learning_resource_for_update` , it's not really a serializer in the sense that it returns data to a user that is waiting for it, it's just used to convert a learning resources into the format expected by open search.  featured_rank is only calculated for the double digit number of resources that are actually in a featured list so this does also really slow down the recreate index job

mbertrand (approved):
👍

PR #1380, opened 2024-08-07
Update topic boxes to support multiple lines

jkachel:

Closes mitodl/hq#5016

This does a handful of things:
- Adjusts the styling of the topic boxes on the homepage to better match the Figma design
- Adjusts the icons for "Data Science, Analytics & Computer Technology" (to RiLineChartLine) and "Innovation & Entrepreneurship" (to RiSpaceShipLine)
- Updates the BrowseTopicsSection and TopicsListingPage to include icon data so they display correctly
- Updates the Django Admin for topics to show the UUID, mark the fields as read-only

Tablet view (desktop is very similar but this shows the wrapping more effectively):
Screenshot 2024-08-07 at 11 52 03 AM
onhover
image

### How can this be tested?

For best results, put your browser in responsive mode and grab a tablet-sized preview. (Screenshots are at 994x800 resolution. In Firefox, grabbing "iPad" also disables hover because it enables touch emulation.) The design of the boxes in the Browse by Topic section should match the design (and also the notes in the issue). Additionally, the icons for Data Science and Innovation should be correct.

Also:
- The topics on the listing page should have icons and they should be the correct ones (i.e. they should match the homepage).
- Go into Django Admin and load a Learning Resource Topic. The fields should be read-only and you should now see the UUID for the topic.

### DIFF
--- a/data_fixtures/migrations/0010_topics_update_icons_for_innovation_data_science.py
+++ b/data_fixtures/migrations/0010_topics_update_icons_for_innovation_data_science.py
@@ -1,0 +1,60 @@
+"""
+Update the topics; update the icons for Data Science, Analytics & Computer
+Technology and the Innovation & Entrepreneurship topics.
+
+This is just an icon revision so doing it directly.
+"""
+
+from django.db import migrations
+
+
+def _update_icons(model, topic, icon):
+    """Perform the update."""
+
+    topic = model.objects.get(topic_uuid=topic)
+    topic.icon = icon
+    topic.save()
+
+
+def update_topic_icons(apps, schema_editor):
+    """Update topic icons."""
+
+    LearningResourceTopic = apps.get_model(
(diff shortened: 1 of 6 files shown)


jonkafton (approved):
Looks good. 

I thought about dynamically importing the icons, as we basically have a hard coded whitelist in the front end that we probably don't need. @remixicon/react doesn't tree-shake, but also does not have a default export; not worth the fuss if these won't change much (dynamically import unknown named import?).

jkachel:
With regard to the whitelist: the icons _can_ change, but really only via a migration in `data_fixtures`, and if we're going that route then it's reasonable to expect the engineer making the change to update the icon whitelist too. (Part of this PR is marking all the topic fields read-only in the Django Admin to prevent this sort of change being done by a non-engineer.)

PR #1381, opened 2024-08-07
show featured resources first minus breaking changes

abeglova:
closes https://github.com/mitodl/hq/issues/4986

This updates the default sort to show featured resources first. The resources are shown in order of their position in channel featured lists, with the order of resources with the same position randomized.  After the featured parameter, the order continues to be courses/programs before learning resources.

This only affects the order if there isn't a search term

**This pr is part of https://github.com/mitodl/mit-open/pull/1377**

**This pr should be merged and released before This pr is part of https://github.com/mitodl/mit-open/pull/1377 to avoid breaking search during reindexing.** 

### How can this be tested?
Follow the instructions in https://github.com/mitodl/mit-open/pull/1377


mbertrand (approved):
👍 based on review of #1377

PR #1383, opened 2024-08-07
Only process best/next run contentfiles, and make sure other run contentfiles are deindexed

mbertrand:
- Closes https://github.com/mitodl/hq/issues/5160

- Only ingests/indexes contentfiles for one run per course (the next or most recent run)
- Makes sure that contentfiles for other runs in the course are deleted and deindexed.
- Adds a new dev-only boolean setting `SKIP_TIKA` to make local ingestion of contentfiles quicker.

### How can this be tested?
- Set `SKIP_TIKA=True` to your backend .env file and also add the following with values equal to RC:
  ```
  XPRO_CATALOG_API_URL
  XPRO_COURSES_API_URL
  XPRO_LEARNING_COURSE_BUCKET_NAME
  ```
- Run `./manage.py backpopulate_xpro_data` followed by `./manage.py backpopulate_xpro_files`. Both should complete successfully.
- Run the following code, you should get thee same results (unless a new run is added to the course between now and when you review it):
  ```python
   course = LearningResource.objects.get(readable_id="course-v1:xPRO+MLx1")
   for run in course.runs.all():
       print(run.run_id, run.published, run.content_files.count())
  > course-v1:xPRO+MLx1+R10 True 139
  ```
- Change the `run_id` of that run to something else and the date to the far past, and create a fresh run with the old run_id and a more recent date:
  ```python
   LearningResourceRun.objects.filter(
       run_id="course-v1:xPRO+MLx1+R10"
  ).update(run_id="course-v1:xPRO+MLx1+R00010", start_date="2010-01-01")
  LearningResourceRun.objects.create(learning_resource=course, run_id="course-v1:xPRO+MLx1+R10", published=True, start_date="2023-10-01")
  ```
-Run only `./manage.py backpopulate_xpro_files` again, followed by the first code block above.  This time you should get the folliowing result, with the original run now having 0 contentfiles and the new one having 139:
  ```
  course-v1:xPRO+MLx1+R00010 True 2010-01-01 00:00:00+00:00 0
  course-v1:xPRO+MLx1+R10 True 2023-01-01 00:00:00+00:00 139
  ```


abeglova (approved):
works great

PR #1386, opened 2024-08-07
Fix frontend sentry_env and release version

ChristopherChudzicki:
Followup for https://github.com/mitodl/hq/issues/5109

This fixes two things:
- The bash script for getting verison was using `\d`, which caused errors on github actions and docker. Now uses `[0-9]`
- Fixed typo, `MITOPEN_ENV_RC -> MITOPEN_ENVIRONMENT_RC` (and _PROD)

### How can this be tested?
To some extent, need to release this. Here's what I did:

1. Run `./scripts/get_version/sh` in docker (it used to fail in docker)
2. View this github action... https://github.com/mitodl/mit-open/actions/runs/10291094957/job/28482598435 I made it to test the changes here. You can see that version is now picked up correctly.


frontends/mit-open/webpack.config.js
```diff
@@ -39,6 +39,7 @@ const {
   EMBEDLY_KEY,
   CKEDITOR_UPLOAD_URL,
   SENTRY_DSN,
+  SENTRY_ENV,
```
ChristopherChudzicki, line 42: In addition to having the wrong var name in the github action, I neglected to actually pass it into the app.

scripts/get_version.sh
```diff
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 
 # Get VERSION from settings.py
-grep -Eo -m 1 'VERSION\s+= .*(\d+)\.\d+\.\d+.*' main/settings.py |
```
ChristopherChudzicki, line 4: `\d` seems not to work on docker / github actions

.github/workflows/release-candidate.yml
```diff
@@ -44,7 +44,7 @@ jobs:
           POSTHOG_PROJECT_ID: ${{ secrets.POSTHOG_PROJECT_ID_RC }}
           POSTHOG_PROJECT_API_KEY: ${{ secrets.POSTHOG_PROJECT_API_KEY_RC }}
           SENTRY_DSN: ${{ secrets.SENTRY_DSN_RC }}
-          SENTRY_ENV: ${{ secrets.MITOPEN_ENV_RC }}
+          SENTRY_ENV: ${{ secrets.MITOPEN_ENVIRONMENT_RC }}
```
blarghmatey, line 47: Honestly, for these values since we already have the workflows split by environment, we should probably just hard-code it in the YAML.

ChristopherChudzicki: That was my initial thought, except Django also uses these values. Using an env var avoids "Are django and frontend both on same page about prod vs production, rc vs qa vs release-candidate"

scripts/get_version.sh
feoh: \d isn't supported in bash. You can use [:digit:] if you want, but what you have is just fine :)

[Cite](https://stackoverflow.com/questions/17968020/how-to-match-digits-in-regex-in-bash-script)

feoh (approved):

scripts/get_version.sh
ChristopherChudzicki: Thanks! Good to know.

PR #1388, opened 2024-08-08
Rename env var prefix MITOPEN_ to MITOL_

rhysyngsun:

Part of https://github.com/mitodl/hq/issues/5154

Renames the environment variables to the more non product specific `MITOL_`.

Temporary code added to `manage.py` to use the legacy prefix if variables are present.

### How can this be tested?

- Checkout the first commit and verify you can `docker compose up` without errors 
- Checkout the second commit and verify the same


.github/workflows/ci.yml
```diff
@@ -123,7 +123,7 @@ jobs:
       - name: Webpack build
         run: yarn run build
         env:
-          MITOPEN_API_BASE_URL: https://api.mitopen-test.odl.mit.edu
+          MITOL_API_BASE_URL: https://api.mitopen-test.odl.mit.edu
```
Ardiea, line 126: Do we want to update these URLs while we're here or is that too many moving parts at the moment?

rhysyngsun: It's a dummy value but I can update it.

Ardiea: Well this one, yes. But in the production and rc files they are real. Thinking about it maybe they should come from ol-inf with much of the other config.

rhysyngsun: That's going to depend on the timing, if I update those, the next build that goes out the frontend will try to connect to those immediately and things will break hard if they're not completely in place. I think maybe the answer is that they should come from a secret, not from a static value.

rhysyngsun: Actually I'm inclined to say that changing the url values is out of scope for this issue. I'll leave the one I already touched but the others should get dealt with separately because otherwise it's comingling two different kinds of changes (variable name and value changes).

rhysyngsun:
@Ardiea I addressed your feedback in that one line you commented, but I think anything more is out of scope as noted [here](https://github.com/mitodl/mit-open/pull/1388#discussion_r1710229408).

feoh (approved):

blarghmatey (approved):
This all looks good from a semantic perspective but I'll defer to someone else for some functional testing.

.github/workflows/ci.yml
Ardiea: Sounds good!

Ardiea (approved):

.github/workflows/release-candidate.yml
```diff
@@ -44,10 +44,10 @@ jobs:
           POSTHOG_PROJECT_ID: ${{ secrets.POSTHOG_PROJECT_ID_RC }}
           POSTHOG_PROJECT_API_KEY: ${{ secrets.POSTHOG_PROJECT_API_KEY_RC }}
           SENTRY_DSN: ${{ secrets.SENTRY_DSN_RC }}
-          SENTRY_ENV: ${{ secrets.MITOPEN_ENV_RC }}
-          MITOPEN_AXIOS_WITH_CREDENTIALS: true
-          MITOPEN_API_BASE_URL: https://api.mitopen-rc.odl.mit.edu
-          MITOPEN_SUPPORT_EMAIL: mitlearn-support@mit.edu
+          SENTRY_ENV: ${{ secrets.MITOL_ENV_RC }}
+          MITOL_AXIOS_WITH_CREDENTIALS: true
+          MITOL_API_BASE_URL: https://api.mitopen-rc.odl.mit.edu
```
Ardiea, line 49: Okay, with the infrastructure where it is now, this vault SHOULD be changed at this point.

PR #1391, opened 2024-08-09
fix prettier and eslint in pre-commit

gumaerc:
Closes https://github.com/mitodl/hq/issues/5191

This PR fixes issues with `prettier` and `eslint`, more specifically discrepancies between the way they are run in pre-commit locally vs in Github Actions. The [deprecated prettier pre-commit hook](https://github.com/pre-commit/mirrors-prettier) is causing issues with inconsistent output vs what you get when running it locally, so it has been replaced with a `local` directive that locally runs the installed version specified in our `package.json`. The hook is now skipped in pre-commit CI and instead `fmt-check` is being run in Github Actions directly during the Javascript tests. The eslint configuration was modified, because the newest versions seem to throw an error when you run `pre-commit run --all` that it can't find the eslint configuration. What I did was point to the config file in pre-commit, and then fix a resulting issue with path resolution excluding `ol-components` from the rule restricting `@mui` imports.

### How can this be tested?
 - Clone this branch locally and make sure the app has been run at least once so your `watch` container has run `yarn install`, or if you are running the frontend outside docker have run `yarn install` at least once
 - Run `pre-commit run --all` and ensure all checks pass


gumaerc:
pre-commit.ci run

.pre-commit-config.yaml
```diff
@@ -20,16 +21,17 @@ repos:
...
+        entry: npx prettier --write
+        language: node
         types_or:
           [javascript, jsx, ts, tsx, json, scss, sass, css, yaml, markdown]
-        args:
-          - --no-config
-          - --no-semi
       - id: prettier
+        entry: npx prettier --write --plugin=prettier-plugin-django-alpine "**/*.html"
+        language: node
```
blarghmatey, line 34: With moving it to the local hook you don't have to have this as a separate step anymore, you can just rely on the prettier config for loading the plugin and addressing all of the file types.

gumaerc: You're correct about this, although I did have to add `html` to the `types_or` argument for it to actually check django templates

PR #1392, opened 2024-08-09
cache learning resources search api view

shanbady:

Closes #https://github.com/mitodl/hq/issues/5162

This PR uses the django cache to cache the learning resources search api endpoint in addition to clearing out the caches at the end of each backpopulate task.

### How can this be tested?
1. checkout this branch
2. before doing anything, open up the django shell and view the current cache keys (should be empty)
```
cache = caches['redis']
print(cache.keys('views.decorators.cache.cache_page.search.*'))
```
3. in learning_resources_search/views.py add some visible logging such as ```print("TESTING 123", "\n" * 20)``` to the get() method in LearningResourcesSearchView - this is so that we can visibly observe if the page is being served from the cache or not.
4. go to a search page and start browsing around - observe that on the first load of any page + url params the logging is visible but on subsequent loads it is served from the cache and does not output any print statements.
5. go back to the django shell and get the cache keys and observe there are entries for the search pages:
```
cache = caches['redis']
print(cache.keys('views.decorators.cache.cache_page.search.*'))

```

### Additional Context
This is the first step into caching all anonymous views. Once we deploy this and validate, we can apply this approach to the remaining endpoints


learning_resources_search/views.py
```diff
@@ -63,6 +64,7 @@ class LearningResourcesSearchView(ESView):
 
     permission_classes = ()
 
+    @method_decorator(cache_page(60 * 60 * 24, cache="redis", key_prefix="search"))
```
abeglova, line 67: The length of time that the cache is kept alive worries me 24hrs is a long time- although if we clear the cache after every task that updates the index, the actual age of the caches should be less than a couple hours, probably? Also are we worried about memory usage for storing the caches?

abeglova (changes requested):
The search serializer is not anonymous
https://github.com/mitodl/mit-open/blob/main/learning_resources_search/serializers.py#L637

This breaks favorite/user list functionality which is user specific. If you log in and go to http://open.odl.local:8062/search in this pr and add a resource to a user list or learning path, the  icon does not turn red to show that the resource is part of a list. And if the view is cached with a user with the resource in a list and that user logs out, the resource remains shown as part of the list after the user logs out. The cache  should only be used for anonymous users



Additionally there are a bunch of other tasks that change the index that this pr doesn't clear the cache after. For example the search recreate_index and update_index jobs as well as import_all_mit_edx_files and import_all_oll_files. There are others as well

abeglova:
You'll need to rebase and clear the cache after tasks.learning_resources_search.update_featured_rank as well

shanbady:
@abeglova this is ready for another look. I made the cache serve up only for anon users and added cache clearing to re-index, update index and update_featured_rank. The content file tasks all call get_content_tasks -> get_content_files which does end up clearing the cache.

regarding the 24 hours - this was what we decided on during discussion. I think the reasoning was that nothing changes until the pipelines run. I moved it to a constant in settings so we can easily tweak it afterwards. In regards to the memory usage - that is something to be seen. I dont have any numbers to go off of but if it does pose an issue we could switch backends or increase the instance.

main/settings.py
```diff
@@ -620,7 +620,7 @@ def get_all_config_keys():
 MIDDLEWARE_FEATURE_FLAG_COOKIE_MAX_AGE_SECONDS = get_int(
     "MIDDLEWARE_FEATURE_FLAG_COOKIE_MAX_AGE_SECONDS", 60 * 60
 )
-
+SEARCH_PAGE_CACHE_DURATION = 60 * 60 * 24
```
abeglova, line 623: Can you change this to `get_int("SEARCH_PAGE_CACHE_DURATION", 60 * 60 * 24)` so we have the option of overriding with a variable in the settings - could be useful to be able to reduce the duration quickly if we experience memory issues with the caches

abeglova (approved):
One small suggestions - otherwise looks good

PR #1394, opened 2024-08-09
Preliminary ETL pipeline for Sloan

mbertrand:
Closes https://github.com/mitodl/hq/issues/4732

Related PR: https://github.com/mitodl/ol-infrastructure/pull/2625

- Adds a new ETL pipeline to ingest courses and programs directly from Sloan's own API feed instead of Prolearn's.
- Excludes Sloan courses from being ingested by the Prolearn ETL pipeline.

### How can this be tested?
- Set the following values in your .env file:
  ```
   PROLEARN_CATALOG_API_URL=https://prolearn.mit.edu/graphql
   SEE_API_ENABLED=True
   SEE_API_URL
   SEE_API_ACCESS_TOKEN_URL
   SEE_API_CLIENT_ID
   SEE_API_CLIENT_SECRET
  ```
-  On the **main** branch, run `./manage.py backpopulate_prolearn_data`
-  When it is done, there should be learning resources with `etl_source=prolearn` and `offered_by=see`.  They will all be courses and not programs.  http://api.open.odl.local:8063/admin/learning_resources/learningresource/?o=2&offered_by__code__exact=see
- If you don't already have a populated featured learning path for Sloan, run `./manage.py populate_featured_lists` to create one

- Switch to this branch.  Run `./manage.py backpopulate_sloan_data` followed by `./manage.py backpopulate_prolearn_data`)  
- If you go to `http://open.odl.local:8062/learningpaths` and open the Professional Education featured learning path, it will be empty.  Run the following to reassign the old featured courses from prolearn to. their new equivalents
```
./manage.py transfer_list_resources course title prolearn see
```
- Once complete, you should have courses in the Sloan featured learning path again.  These commands will need to be run on RC/production as well to retain the current featured lists.

To test the `SEE_API_ENABLED` flag:
- Remove it from your backend env file or set it to `False`
- Run `./manage.py backpopulate_sloan_data`, it should not ingest any courses
- Run `./manage.py backpopulate_prolearn_data`, it should ingest (and set to published) Sloan courses


learning_resources/etl/sloan.py
```diff
@@ -0,0 +1,215 @@
...
+    """
+    return {
+        "run_id": run_data["CO_Title"],
+        "start_date": parse_datetime(run_data["Start_Date"]),
+        "end_date": parse_datetime(run_data["End_Date"]),
+        "title": course_data["Title"],
+        "url": course_data["URL"],
+        "availability": RunAvailability.current.value,
+        "published": True,
+        "prices": [run_data["Price"]],
```
mbertrand, line 144: I forgot to add instructors, something like this should work:
```python
"instructors": [{"full_name": name.strip()} for name in run_data["Faculty_Name"].split(",")]
```

learning_resources/etl/sloan_test.py
```diff
@@ -84,3 +94,60 @@ def test_sloan_extract_programs(
...
+            "certification",
+            "certification_type",
+            "professional",
+            "published",
+            "learning_format",
+            "topics",
+            "course",
+            "runs",
+        ]
+    )
```
mbertrand, line 153: Would be good to test for specific values for some of these, like:
```python
assert transformed["image"] == {
    "url": <expected url>,
    "alt": <expected alt string>,
    "description": <expected description>"
 }
 ```
 
 Could do so in separate tests instead like `test_parse_image`, `test_parse_topics` etc if that is easier.

Comment (shanbady) [reply learning_resources/etl/sloan_test.py:161]: added tests for both.

File: learning_resources/etl/sloan.py (hunk 1 above)
Comment (shanbady) [reply learning_resources/etl/sloan.py:147]: Added instructors to the struct

File: learning_resources/etl/sloan.py [hunk 3]
```diff
@@ -0,0 +1,219 @@
...
+    """
+    Parses a datetime string
+
+    Args:
+        value(str): the datetime in string format
+
+    Returns:
+        datetime: the parsed datetime
+    """  # noqa: D401
+    return parse(value).replace(tzinfo=UTC) if value else None
```
mbertrand, line 100: This might need to be adjusted a bit, it is showing dates a day earlier than expected (example, a course starting on 10/29/2024 is shown with a start date of 10/28/2024).  Maybe try something like this:

```python
from datetime import UTC
from zoneinfo import ZoneInfo

return parse(value).replace(tzinfo=ZoneInfo("US/Eastern")).astimezone(UTC)
```

ChristopherChudzicki (commented):
👍 This looks good. I did notice that `availability` does not get populated on courses. At this point—I know this is "preliminary"—are we expecting it to be?

Currently all `etl_source="see"` have `availability=null`. As far as the frontend is concerned (at the moment, anyway), that's effectively the same as `"dated"`.

mbertrand:
@ChristopherChudzicki good catch, I added "availability" with a default value of "date".  Also one more unit test and a check for missing settings.

ChristopherChudzicki (approved):
👍 The availability stuff looks good. I got 4 anytime courses, all of them show up as "Self paced" on the sloan site. The others I checked all showed up as date-based (except one, below).

  I did notice that every course has a url with https://executive.mit.edu/, except one. The course "Making AI Work: Machine Intelligence for Business and Society" at https://www.getsmarter.com/products/mit-sloan-making-ai-work-online-course?focused=true. Regarding this course... Not sure if there's anything weird about this course otherwise. (Does it have a different platform?)

learning_resources/etl/prolearn.py
```diff
@@ -29,6 +29,10 @@
 # List of query fields for prolearn, deduced from its website api calls
 PROLEARN_QUERY_FIELDS = "title\nnid\nurl\ncertificate_name\ncourse_application_url\ncourse_link\nfield_course_or_program\nstart_value\nend_value\ndepartment\ndepartment_url\nbody\nbody_override\nfield_time_commitment\nfield_duration\nfeatured_image_url\nfield_featured_video\nfield_non_degree_credits\nfield_price\nfield_related_courses_programs\nrelated_courses_programs_title\nfield_time_commitment\nucc_hot_topic\nucc_name\nucc_tid\napplication_process\napplication_process_override\nformat_name\nimage_override_url\nvideo_override_url\nfield_new_course_program\nfield_tooltip"  # noqa: E501
 
+SEE_EXCLUSION = (
+    '{operator: "<>", name: "department", value: "MIT Sloan Executive Education"}'
+)
+
 # Performs the query made on https://prolearn.mit.edu/graphql, with a filter for program or course  # noqa: E501
 PROLEARN_QUERY = """
```
ChristopherChudzicki, line 37: Minor suggestion: If we add another `%s`, depending on placement, breaks current usage. Maybe could do

```python
def prolearn_query(course_or_program: str, sloan_filter: bool, fields: str) -> str:
    return f"""
<query as f-string>
"""
```

ChristopherChudzicki (approved):
👍

PR #1395, opened 2024-08-10
Add accessibility linting

ChristopherChudzicki:
Closes https://github.com/mitodl/hq/issues/5155 (for now; more could probably be done).

This PR adds [eslint-plugin-styled-components-a11y](https://www.npmjs.com/package/eslint-plugin-styled-components-a11y?activeTab=readme) to lint accessibility issues. This package is a wrapper around [eslint-plugin-jsx-a11y](https://www.npmjs.com/package/eslint-plugin-jsx-a11y).

### How can this be tested?
- Linting should pass
- Try removing `aria-label` from an ActionButton; it should give you a useful error message
- Try removing an alt tag from an `img`, or any `styled.img`. You should get a useful error message.

### Additional Context
One of the main user-facing changes here is that I set many alt tags to `""` (e.g., on testimonials). This is better than no alt tag, and IMO is appropriate. These images are generally decorative, not, e.g., diagrams or a literal, physical product. And the testimonial APIs don't provide alt data, so we don't really have another choice.


frontends/ol-components/src/components/ShareTooltip/ShareTooltip.tsx
ChristopherChudzicki, line 1: There was a linting error in ShareTooltip.

We don't use this component. We have a "Share" button in our drawer designs (not implemented). I strongly suspect we would implement it a bit differently than it was here anyway. E.g., we would use MUI popover not `rc-tooltip`.

jonkafton:
> One of the main user-facing changes here is that I set many alt tags to `""` (e.g., on testimonials). This is better than no alt tag, and IMO is appropriate. These images are generally decorative

Agree and not sure if it would be feasible, or even useful, to describe image content, e.g. "a woman on a laptop concentrating"

jonkafton:
Looks good, works well. Interestingly, it didn't pick on removing the `aria-label` on `src/page-components/SearchDisplay/SearchDisplay.tsx:667` - a styled ol-component `Button`. We pass props all the way through,  but maybe a restriction on two levels of styled() wrapping?

ChristopherChudzicki:
>  Interestingly, it didn't pick on removing the aria-label on src/page-components/SearchDisplay/SearchDisplay.tsx:667 - a styled ol-component Button. We pass props all the way through, but maybe a restriction on two levels of styled() wrapping?

I think more likely a bug with [eslint-plugin-styled-components-a11y](https://github.com/brendanmorrell/eslint-plugin-styled-components-a11y) wrapper for [eslint-plugin-jsx-a11y](https://github.com/jsx-eslint/eslint-plugin-jsx-a11y)

jonkafton (approved):

PR #1404, opened 2024-08-12
adding fix for logo in email

shanbady:
This PR fixes an issue with the logo not displaying for subscription emails. This was due to the static path having an extra slash when combining it with the app url

fixes the display of the main logo in subscription emails for live environments

### How can this be tested?
testing instructions are the same as [described here](https://github.com/mitodl/mit-open/pull/1390#issue-2456564754) - validate that the mit learn logo appears. 

### Additional Context
We wont fully be able to validate until this gets deployed and we see emails from a live environment.

### DIFF
--- a/main/templates/email/email_base.html
+++ b/main/templates/email/email_base.html
@@ -265,7 +265,7 @@
               {% block logo %}
               <h1 style="text-align: center">
                 <img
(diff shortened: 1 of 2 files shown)


rhysyngsun (approved):
LGTM

PR #1408, opened 2024-08-13
Add migration to adjust mappings, make some minor changes to the dump to yaml util function

jkachel:

Closes mitodl/hq#5202

Updates the mappings as described in mitodl/hq#5202 .

Also, discovered and fixed a couple bugs in the `dump_topics_to_yaml`:
- Running with a single topic fails, because the fetch returns a single model object rather than a list of them.
- Topic data did not include the "parent" element. (This is more of an issue for a single topic - topics are usually defined by their hierarchy in the yaml file, but if you're dumping a single topic for, say, a migration like this, you _do_ need the parent so it will be in the right place afterwards.) 

### How can this be tested?

Test running `dump_topics_to_yaml` on a single topic - it should be correct, and the `parent` element should be there and set properly.

The migration should run with `RUN_DATA_MIGRATIONS=true` and the changes should take effect as expected. Rolling back the migration should also work. 

### Additional Context

As per usual, this will not update existing learning resources. To see resources mapped using these new mappings, you will need to run the appropriate backpopulate commands.

### DIFF
--- a/data_fixtures/migrations/0012_topic_mappings_adjust_sustainable_business_finance_accounting.py
+++ b/data_fixtures/migrations/0012_topic_mappings_adjust_sustainable_business_finance_accounting.py
@@ -1,0 +1,90 @@
+"""
+Update the topics:
+- Add mapping "Sustainability" to "Sustainable Business" for mitxpro, mitpe, see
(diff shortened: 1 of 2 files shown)


mbertrand (approved):
👍

PR #1409, opened 2024-08-13
Rename the variables on release workflows

rhysyngsun:

N/A

Renames a few variables that got missed in my last PR.


shanbady (changes requested):
Although i dont think this impacts the release - we probably also want to replace the old references in ci.yml while we are in here https://github.com/shanbady/mit-open/blob/main/.github/workflows/ci.yml#L82

rhysyngsun:
> Although i dont think this impacts the release - we probably also want to replace the old references in ci.yml while we are in here https://github.com/shanbady/mit-open/blob/main/.github/workflows/ci.yml#L82

I'd prefer to keep this PR to the minimum needed to get the deployments functional.

shanbady (approved):
approving to get the release out but we should follow up with removing the lingering references

PR #1411, opened 2024-08-14
Per Page metadata tags 

ChristopherChudzicki:
https://github.com/mitodl/hq/issues/5192

Adds per-page metadata tags, particularly for social media sharing.

### How can this be tested?
The intent is to support social sharing. However, we can't really test that until we get to RC. For now, just look at the metadata.

1. View the document `<head>...</head>` on various pages. You could inspect the HTML panel in browser, or run `document.head.querySelectorAll('meta[property], meta[name], link[rel="canonical"], title').forEach(e => console.log(e.outerHTML))`. You should check `<title>`, and `og:*`, `twitter:*` metadata tags.
    - **Note:** The easiest place to see `<title>` is in the browser tab title
2. Check pages:
    - these routes:
        - the departments listing, http://localhost:8062/departments
        - the topics listing, http://localhost:8062/topics
        - the units listing, http://localhost:8062/units
        - about, privacy, terms, etc
        - unit channel page, e.g., http://localhost:8062/c/unit/ocw
        - department channel page http://localhost:8062/c/department/physics
    - All these pages should have reasonable titles of form `Page Title | MIT Learn`
    - All these pages should have `og:*` and `twitter:*` tags (see https://github.com/mitodl/hq/issues/5192). **These pages all use a default `og:image` and `twitter:image:src`**. The default image is a placeholder for now, just the Learn logo.
    - Double check that `og:image` and `og:twitter:src` are full URLs (origin + path, not just path)
4. Also check the resource drawer...
    - On **any page** (with resources) open the resource drawer
    - The title should override whatever the current title is
    - `og:image` and `og:description` should be taken from the resource.


gumaerc (approved):
👍

PR #1412, opened 2024-08-14
Make MITOL_ settings optional in app.json

rhysyngsun:

N/A

Fixes Heroku failing the build because settings in `app.json` are required but they could exist under either the `MITOL_` or `MITOPEN_` prefixes.

### How can this be tested?

Can only test when we release to RC/prod.


jkachel (approved):
LGTM 👍

PR #1418, opened 2024-08-15
Expose the SESSION_COOKIE_NAME setting

rhysyngsun:

N/A

If we have this app hosted at a subdomain under another instance of the app, the high level domain session cookies will occlude the one from the subdomain, breaking sessions. The browser only sends 1 cookie per name and at least in practice sends the cookie from the highest level domain. So to address this we need to be able to change the cookie name for sessions.

### How can this be tested?

Set `SESSION_COOKIE_NAME` in your `.env` and verify that when you login your session cooke gets set with that name


ChristopherChudzicki (approved):
👍 Worked as expected when set `SESSION_COOKIE_NAME` to something new and deleted old cookie

PR #1420, opened 2024-08-15
set csrf cookie name from env var

ChristopherChudzicki:
Fixes an issue with https://github.com/mitodl/mit-open/pull/1419

This PR allows us to set the CSRF cookie name based on an environment variable. A followup to #1418 that did the same for sessionid. 

### How can this be tested?
In `shared.local.env`, set `CSRF_COOKIE_NAME=some-value` and restart your containers. 

Make an authenticated POST/PATCH request from frontend (creating a userlist, editing your profile). Check that the CSRF cookie sent in the request has expected name.

### Additional Context
- [ ] Need new env var for heroku
- [ ] and for github action


frontends/mit-learn/src/services/axios.ts
ChristopherChudzicki, line 1: This entire file should be removed, but I didn't want to do that here.
- https://github.com/mitodl/hq/issues/5224

env/shared.env
```diff
@@ -1,5 +1,6 @@
 MITOL_API_BASE_URL=http://api.open.odl.local:8063
 MITOL_APP_BASE_URL=http://open.odl.local:8062
 MITOL_SUPPORT_EMAIL=support@localhost
+CSRF_COOKIE_NAME=csrfoken-local
```
rhysyngsun, line 4: Nit:
```suggestion
CSRF_COOKIE_NAME=csrftoken-local
```

rhysyngsun (changes requested):
Small thing, will test now

rhysyngsun (approved):
LGTM assuming tests pass

PR #1422, opened 2024-08-15
Add load testing suite

rhysyngsun:

Related to https://github.com/mitodl/hq/issues/5101

This adds Locust as a load testing tool. It also removes some configuration around uwsgi processes and threads because this doesn't seem to be getting picked up properly.

### How can this be tested?
- Add `load-testing` to your `COMPOSE_PROFILES` setting and `docker compose up`
- Navigate to http://localhost:8089/
- Enter some load test settings (e.g. 100 users), verify it's running against the nginx hostname, and run it. You should see Locust running load tests.


gumaerc (approved):
This looks good and you could merge it as-is, but could we add some documentation surrounding how to set it up into the readme? In particular, the `COMPOSE_PROFILES` setting should be noted about how to turn it on as well as mentioning that it's meant to test the API, so you need to enter your API URL and not the frontend URL.

PR #1414, opened 2024-08-14
Ab/revert featured sort

abeglova:
### What are the relevant tickets?
This undoes the change to the default sort from https://github.com/mitodl/mit-open/pull/1377/files

It keeps the change that changed LEARNING_RESOURCE_SORTBY_OPTIONS to LEARNING_RESOURCE_SEARCH_SORTBY_OPTIONS. That should have been part of https://github.com/mitodl/mit-open/pull/1381. It will make it easier to check if the recreate_index job successfully updated all the data

### How to test
Checkout the last commit before featured_ranks were merged into main
 git checkout cc86e41d53052e630f2cdc44d29c9f6124b8d847

Run 
./manage.py recreate_index --podcasts so you have an index without the featured_rank column

Checkout this branch
Go to http://open.odl.local:8062/search
Verify select the "learning materials" tab
Verify that you see "podcast" under resource types and that search generally works fine

PR #1423, opened 2024-08-16
Set default sort to featured (#1414)

abeglova:
This reverts commit 2c94718182cd0b458ae5b111d458f67188116f92.

https://github.com/mitodl/hq/issues/4986

Sort by featured 

### How can this be tested?
Run ./manage.py recreate_index --all
Go to the search page. Verify that the first items you see are the same as the resources from the featured carousel on the front page
To be fair to all units the featured ranks resources with the same  order in a featured list are randomized, so the order of the featured resources will differ slightly from the featured carusel

### DIFF
--- a/learning_resources_search/api.py
+++ b/learning_resources_search/api.py
@@ -39,7 +39,7 @@ log = logging.getLogger(__name__)
 
 LEARN_SUGGEST_FIELDS = ["title.trigram", "description.trigram"]
 COURSENUM_SORT_FIELD = "course.course_numbers.sort_coursenum"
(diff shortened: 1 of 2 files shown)


ChristopherChudzicki (approved):
👍 Looks good to me. Reindex went through and I see same resources.

_Talked to Shankar... I took over the review b/c i had already started the reindexing. Neglected to assign myself as I was eating lunch at the time._

sovsey:
@abeglova Is there anything pending for this prior to merging? Thanks!

PR #1424, opened 2024-08-16
Remove remaining frontend routes

jkachel:

Closes mitodl/hq#4611

This PR:
- Removes the remaining routes for frontend URLs
- Adds a route for `/` that should bounce you back into the React app. (This is the same as `/app`.)
- Updates the error page handlers to be a single one that uses `api_view` so DRF will throw back a RESTful response

### How can this be tested?

Automated tests should pass. (There's a couple of new ones that will test the error handling too.)

Set `DEBUG=False` in your `.env` (otherwise Django's error handler will take over), then try to navigate to some bad URLs in the Django app. One of two things should happen:
- You'll get a generic 404 response when the route doesn't match anything at all.
- You'll get the normal 404 message for routes that are technically correct but return no data (i.e., pulling up a topic with a bad ID).

For example, navigating to http://api.learn.odl.local:8063/7716c6c8-1873-4857-b4eb-b4f8a1189896 should get you a generic 404 error, but navigating to http://api.learn.odl.local:8063/api/v1/topics/-9/ should give you a 404 that says that it can't find a LearningResourceTopic.

### DIFF
--- a/main/urls.py
+++ b/main/urls.py
@@ -21,7 +21,7 @@ from django.urls import include, path, re_path
 from django.views.generic.base import RedirectView
 from rest_framework.routers import DefaultRouter
 
-from main.views import FeaturesViewSet, index
+from main.views import FeaturesViewSet
 
 # Post slugs can contain unicode characters, so a letter-matching pattern like [A-Za-z] doesn't work.  # noqa: E501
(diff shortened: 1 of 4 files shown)


shanbady (changes requested):
Looks good. Only comment I had was that as far as I can tell - it seems like we can get away with removing the main-index url altogether unless we need it for some other reason

shanbady (approved):

PR #1429, opened 2024-08-19
Handle topics with colons from xpro in ETL pipeline

mbertrand:
Closes https://github.com/mitodl/hq/issues/5222

Handles topics from xpro that may  contain colons.

- [ ] Desktop screenshots

### How can this be tested?
- Run `./manage.py backpopulate_xpro_data`
- After it's done, only 4 xpro courses should not have any topics:
  ```python
  LearningResource.objects.filter(etl_source="xpro", published=True, topics__isnull=True).count()
  ````

### DIFF
--- a/learning_resources/etl/xpro.py
+++ b/learning_resources/etl/xpro.py
@@ -51,6 +51,27 @@ def _parse_datetime(value):
     return parse(value).replace(tzinfo=UTC) if value else None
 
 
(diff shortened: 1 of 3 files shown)


jkachel (approved):
LGTM 👍 
Tested with noted test and also checked in the DB; the problematic program has topics I would expect for it now.

PR #1432, opened 2024-08-19
Get availability value from xpro api

mbertrand:
Closes https://github.com/mitodl/hq/issues/5232

Assigns the value of course/program availability based on the xpro api field of the same name

### How can this be tested?
- Set `XPRO_*` values to the same as RC 
- Run `./manage.py backpopulate_xpro_data`
- Check that course/program availability field for xpro courses and programs have the same values as what's returned in the xpro API responses.


shanbady (approved):
👍

PR #1437, opened 2024-08-20
department is optional

abeglova:
None

Department is now optional in ocw and RES has been removed as a department. However the loader code currently assumes that there are departments and skips department update if there is no department in the data. This prevented the department (RES) from being removed from the courses that no longer have a department

### How can this be tested?
Run 
docker-compose run web ./manage.py backpopulate_ocw_data --course-name res-tll-004-stem-concept-videos-fall-2013  --overwrite

(or --course-name any ocw course in https://learn.mit.edu/search/?department=RES, ideally one that already exists in your local database with RES set as the db)

Go to 
http://api.open.odl.local:8063/api/v1/learning_resources_search/?q=%22stem%20concept%20videos%22 (or search for the course you used)
Verify that departments are empty for the course

### DIFF
--- a/learning_resources/etl/loaders.py
+++ b/learning_resources/etl/loaders.py
@@ -102,15 +102,17 @@ def load_departments(
     resource: LearningResource, department_data: list[str]
 ) -> list[LearningResourceDepartment]:
     """Load the departments for a resource into the database"""
-    if department_data:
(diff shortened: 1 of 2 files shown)


mbertrand (approved):
👍

PR #1439, opened 2024-08-20
Remove RES from departments

abeglova:
None

We removed RES ("Supplemental Resources") and added SP ("Special Programs") as departments. This updates the department list in  learning_resources/constants.py and fixtures/department.json

### How can this be tested?
The person testing this should have some RES courses in their local database
1) Go to http://open.odl.local:8062/search/?department=RES and remember/write down some of the course names
2) Run `docker-compose run web ./manage.py update_departments_schools`
3) Run  docker-compose run web ./manage.py update_index --courses --course_etl_source ocw
4) Go to  http://open.odl.local:8062/search/?department=RES you should not see any courses. Undo the department filter and search for some of the courses you saw in step one by name. They should be in the index.

### Additional Context
This pr adds SP into the constant in learning_resources/constants.py but does not add it to the database department fixture yet. The reason for this is that we need to update course-search-utils with the auto generated data from this pr before adding SP courses to the index or the department is not displayed correctly in the facet. Until SP is added to the database, the import for courses with department=SP fails


learning_resources/fixtures/departments.json
```diff
@@ -271,14 +271,6 @@
...
     }
   },
-  {
-    "model": "learning_resources.LearningResourceDepartment",
-    "fields": {
-      "department_id": "RES",
-      "name": "Supplemental Resources",
-      "school_id": null
-    }
-  },
```
mbertrand, line 281: I think it would be better now to add a new migration in `data_fixtures/migrations` to remove the RES department (and, eventually, to add the SP department) instead of running the `update_schools_departments` command

mbertrand (changes requested):

learning_resources/fixtures/departments.json
abeglova: Done! It worries me that we need to update this in three different places though.

mbertrand (approved):
👍

PR #1448, opened 2024-08-21
adding test for redirect view

shanbady:
Adds a very simple test so that the catch all redirect view is not accidentally removed

### How can this be tested?
1. tests should pass
2. locally if you remove /app from [main/urls.py](https://github.com/mitodl/mit-open/blob/main/main/urls.py#L58) and then run the tests (via ```pytest main/views_test.py -s -vvv```) it should fail.


ChristopherChudzicki (approved):
👍

PR #1449, opened 2024-08-21
change default og:image

ChristopherChudzicki:
For https://github.com/mitodl/hq/issues/5248

This PR changes the default social media preview image to

mit-and-logo

This image remains somewhat of a placeholder until we get final logo approval, but is better than the current default image (which is too small and has a bad aspect ratio for social previews).

### How can this be tested?
1. Load the homepage. 
2. Inspect the metatags ... Use the elements panel, or run
    ```js
    document.head.querySelectorAll('meta[property], meta[name]').forEach(e => console.log(e.outerHTML))
    ```
    in the JS console.
3. Inspect the `og:image` and `twitter:image:src` tags. They should contain
    - a full URL (protocal + origin, not just path)
    - should link to the image above


gumaerc (approved):
👍

PR #1453, opened 2024-08-21
Fix departments and schools fixtures

ChristopherChudzicki:
None

This PR fixes some fixtures. Currently, the `schools.json` and `departments.json` fixtures are not usable—the `created_on` and `updated_on` fields are missing. (They were removed in #1253.)

### How can this be tested?
1. On `main`, run `./manage.py loaddata schools`. It will fail
2. On this branch, run `./manage.py loaddata schools`. It will succeed.
3. Repeat for departments

### DIFF
--- a/scripts/run-django-dev.sh
+++ b/scripts/run-django-dev.sh
@@ -6,9 +6,7 @@ python3 manage.py collectstatic --noinput --clear
 python3 manage.py migrate --noinput
 RUN_DATA_MIGRATIONS=true python3 manage.py migrate --noinput
 
-if [[ $NODE_ENV == "development" ]]; then
-	# load required fixtures on development by default
-	python3 manage.py loaddata platforms schools departments offered_by
-fi
+# load required fixtures on development by default
+python3 manage.py loaddata platforms schools departments offered_by
 
 uwsgi uwsgi.ini --honour-stdin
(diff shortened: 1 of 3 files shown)


scripts/run-django-dev.sh
```diff
@@ -6,9 +6,7 @@ python3 manage.py collectstatic --noinput --clear
 python3 manage.py migrate --noinput
 RUN_DATA_MIGRATIONS=true python3 manage.py migrate --noinput
 
-if [[ $NODE_ENV == "development" ]]; then
-	# load required fixtures on development by default
-	python3 manage.py loaddata platforms schools departments offered_by
-fi
+# load required fixtures on development by default
+python3 manage.py loaddata platforms schools departments offered_by
```
ChristopherChudzicki, line 10: The block above was never running. We don't expose `NODE_ENV` to the web container, nor have we for a while now.

Since `run-django-dev.sh` is ...for dev ... it seems OK to just always run this line.

That said, also happy to remove this (and restore it to the README).

gumaerc: This is fine, I think. I can't imagine a scenario where you would need to preserve manual edits to any of this data in development, but it might be best to add a quick note to the "Loading fixture files" section of the readme. There's a note about how it's automatically done for you in development, so maybe just elaborate upon that and say that in development data is loaded from fixtures *every time*, which is what this would do as far as I understand.

gumaerc (approved):
Looks like this was my fault! It looks like I accidentally removed these when running `dumpdata`. I updated the department names in that PR in Django admin, then used `dumpdata` to regenerate the JSON but apparently missed that it removed these dates. Thanks for fixing that. Anyway, I just have one small suggestion:

PR #1456, opened 2024-08-22
ability to set search mode

abeglova:
closes https://github.com/mitodl/hq/issues/5181

This pr adds settings to the admin panel to allow users to select the search mode for queries. The user can select the search type from https://opensearch.org/docs/latest/query-dsl/full-text/multi-match/. Additionally if the user selects a "phrase" as the query type, the user can set the "slop" which is the number of words that are allowed between the terms in a multi word query. 

Currently this setting only affects text queries that are not enclosed in quotes.

Screenshot 2024-08-23 at 4 02 38 PM
Screenshot 2024-08-23 at 4 02 53 PM
Screenshot 2024-08-23 at 4 03 36 PM

### How can this be tested?
Go to open.odl.local:8062/search/ as an admin user
Verify that you can select search_mode from the dropdown
Verify that if you select search_mode=phrase you can set slop with the slider
Verify that the slop selector is not shown and slop is removed from the url if you select a different search mode
Verify that if there is a text query, your selections affect the results you see.

### Additional Context
Most likely we will want to change the default sort to search_mode=phrase and slop something like 5-10. This pr will allow us to play with the results on prod before making that change


learning_resources_search/api.py
```diff
@@ -207,15 +207,33 @@ def generate_learning_resources_text_clause(text):
         "query_string" if text.startswith('"') and text.endswith('"') else "multi_match"
     )
 
+    extra_params = {}
+
+    if query_type == "multi_match" and search_mode:
+        extra_params["type"] = search_mode
+
```
shanbady, line 214: can consolidate this conditional to 
```
 if query_type == "multi_match" and search_mode:
    extra_params["type"] = search_mode
    if search_mode == "phrase" and slop:
        extra_params["slop"] = slop
```

shanbady (approved):
noted a minor nit otherwise works well 👍

PR #1459, opened 2024-08-22
randomize the hero image

gumaerc:
Closes https://github.com/mitodl/hq/issues/5267

This PR adds some new hero images and sets them up to be randomized each time you load the home page.

image

### How can this be tested?
 - Load the home page
 - Refresh several times and verify that the images are randomized


shanbady (changes requested):
functional changes look good. minor note - it might be good to add a test or update an existing one that does a simple check that this element changes every refresh

shanbady:
> @shanbady I don't actually agree that we should add a test for this. I'm using lodash's `_.shuffle` for this, which is the same method we use for randomizing the testimonials. The shuffle function in lodash is already well tested itself, so I don't think it makes sense for us to test it. It also doesn't guarantee that you won't get repeats, so trying to test it opens the potential for a flaky test as it is technically possible to refresh 5, 10, 20, etc. times and get the same image every time. @ChristopherChudzicki what do you think?

thats a good point. i can move to approve.

shanbady (approved):

frontends/mit-learn/src/pages/HomePage/HeroSearch.tsx
```diff
@@ -21,6 +22,10 @@ type SearchChip = {
   icon?: React.ReactElement
 }
 
+const HERO_IMAGE_IDX = _.shuffle([1, 2, 3, 4, 5])
+
+const HERO_IMAGE_URL = `/static/images/hero/hero-${HERO_IMAGE_IDX[0]}.png`
```
ChristopherChudzicki, line 27: I'm ok skipping a test for the randomization, but some comments about testing this.

1. A global declaration + randomization may have some downsides:
    - we can't easily test this in Jest—even if we thought it was worth it—because you'd have to re-import the module to re-randomize, and module imports are generally cached. 
    - ...which leads me to: We've been discussing NextJS and server-side rendering lately. I'm not 100% sure, but I expect that this also would not randomized if served by NextJS, because the randomization is done globally on import. So you'd probably have to restart the server to randomize.
    - Both of the above could be solved by doing the randomization like `const getRandomHeroImage = () => { ... }` and throwing it in a `useEffect(getRandomHeroImage, [])`—that would run once when the component is mounted.
        - but even if you did that, maybe we would cache server-generated HTML, so....
2. Even if you made the change above, any test in Jest would feel pretty artificial to me. I'd much rather have a browser-based test. (E.g., hit "refresh" with playwright 5 times and check that the images aren't all the same. It'll fail some fraction—1/625th— of the time like Carey mentions because of repeats, but whatever.)

Which is all a longwinded way of saying... I don't think it's worth adding a test right now.

gumaerc:
For posterity, I ended up implementing @ChristopherChudzicki 's suggestion above moving the `_.shuffle` into a function instead of assigning the image URL globally. We'll still skip the test for now because of the reasons discussed above.

PR #1462, opened 2024-08-23
Carousel accessibility improvements

ChristopherChudzicki:
For https://github.com/mitodl/hq/issues/5144

This PR aims to improve screenreader experience for our carousels.

*There should be no visual changes to the site.*

### How can this be tested?
Ideally, you should test this in two ways:
- via tab navigation
- via screenreader navigation
    - With a screenreader, you _can_ tab between interactive elements. But screenreaders also support navigation via arrow keys to a wider variety of content. Additionally, if an HTML element has special behavior associated with arrow keys (e.g., via keydown listeners) that generally won't work on a screenreader.
    - If you use Mac, you could test this with Voiceover. 

#### Expected Carousel Behavior:
- Carousel slides themselves are not focusable, but might be in the browser tab order if they contain focusable content.
- Any slides that are offscreen or partially offscreen have `aria-hidden="true"`
    - Additionally, these slides should not be tab accessible. If they contain interactive elements, those elements should have `tabindex=-1`. 
- Carousel slides should have `role="group"` and and `aria-label` indicating how many slides there are.
- Carousel buttons should have reasonable `aria-label`s.

#### Test the resource carousels
These are on homepage and dashboard.
1. Without a screenreader, keyboard only. *Reminder: shift+tab = focus previous item*
    - Try tabbing through the carousel. You should be able to tab to all fully visible slides.
    - You can change pages via the carousel buttons.
        - This is currently kind of annoying to do because you need to tab backwards to before the carousel.
    - You can change pages via left/right arrow keys
        - NB: This will not work on a screenreader.
2. With a screenreader, keyboard-only:
    - Try navigating through the carousel. Listen for the things noted above (role "group", slide labels)
    - Try turning carousel pages via carousel buttons

#### Testimonial carousels
This is on the homepage and on channel pages.
1. Without a screenreader, keyboard-only:
    - You visually see the testimonial content
    - You can change pages via carousel buttons
    - NB: You **cannot** use the left/right arrow key shortcuts in this case, because the carousel has no interactive content.
2. With a screenreader, keyboard-only:
    - Screenreader should read the carousel slides... Note role "group" and aria-label indicating slide number and how many slides there are.
        - Depending on the mode your screenreader is in, this may require you to navigate through each slide. 
    - You should be able to navigate the screenreader slides with the arrow buttons, which should have reasonable labels.

### Additional Context
- This is a very good article on carousel accessibility https://dev.to/jasonwebb/how-to-build-a-more-accessible-carousel-or-slider-35lp, with an associated github repo https://github.com/Accessible360/accessible-carousel-boilerplates. Some of the bheavior / implementation here was based off of that.

### DIFF
--- a/frontends/mit-learn/src/page-components/TestimonialDisplay/TestimonialDisplay.tsx
+++ b/frontends/mit-learn/src/page-components/TestimonialDisplay/TestimonialDisplay.tsx
@@ -2,7 +2,13 @@ import React from "react"
 
 import { RiArrowRightLine, RiArrowLeftLine } from "@remixicon/react"
 import Slider from "react-slick"
-import { ActionButton, TruncateText, styled, theme } from "ol-components"
+import {
+  ActionButton,
+  TruncateText,
+  onReInitSlickA11y,
+  styled,
+  theme,
+} from "ol-components"
 import AttestantBlock from "./AttestantBlock"
 import { useTestimonialList } from "api/hooks/testimonials"
 import type { Attestation } from "api/v0"
@@ -147,6 +153,7 @@ const TestimonialDisplay: React.FC<TestimonialDisplayProps> = ({
           <>
             <Slider
               ref={setSlick}
+              onReInit={() => onReInitSlickA11y(slick)}
               infinite={true}
               slidesToShow={1}
               arrows={false}
--- a/frontends/mit-learn/src/page-components/TestimonialDisplay/TestimonialDisplay.tsx
+++ b/frontends/mit-learn/src/page-components/TestimonialDisplay/TestimonialDisplay.tsx
@@ -81,7 +81,8 @@ const QuoteBody = styled.div(({ theme }) => ({
   },
 }))
 
-const AttestationBlock = styled.div(({ theme }) => ({
+const AttestationBlock = styled.blockquote(({ theme }) => ({
+  margin: "0px",
   width: "auto",
   flexGrow: "5",
   ...theme.typography.h5,
@@ -148,7 +149,7 @@ const TestimonialDisplay: React.FC<TestimonialDisplayProps> = ({
   return (
     <QuoteContainer>
       <QuoteBlock>
-        <QuoteLeader>“</QuoteLeader>
+        <QuoteLeader aria-hidden>“</QuoteLeader>
         {data.count > 1 ? (
           <>
             <Slider
--- a/frontends/mit-learn/src/pages/HomePage/TestimonialsSection.tsx
+++ b/frontends/mit-learn/src/pages/HomePage/TestimonialsSection.tsx
@@ -185,6 +185,7 @@ const ButtonsContainer = styled.div({
 })
 
 const TestimonialTruncateText = styled(TruncateText)({
+  margin: "0px",
   textOverflow: "none",
   ...theme.typography.h4,
   fontSize: pxToRem(20), // This is a unicorn font size per the Figma design - it's not used anywhere else.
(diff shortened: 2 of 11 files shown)


frontends/mit-learn/src/pages/HomePage/TestimonialsSection.tsx
```diff
@@ -274,14 +279,14 @@ const SlickCarousel = () => {
           variant="inverted"
           onClick={slick?.slickPrev}
         >
-          <RiArrowLeftLineStyled />
+          <RiArrowLeftLineStyled aria-hidden />
```
ChristopherChudzicki, line 282: Voiceover, at least, (by default) reads `<button aria-label="Some Label"> <svg ></svg> </button>` as "Button group, Some Label". In order to be read as a button, not "button group" the svg needs aria-hidden.

*This makes me understand better way MUI's icons all come with aria-hidden by default.*

frontends/mit-learn/src/page-components/TestimonialDisplay/TestimonialDisplay.tsx
```diff
@@ -75,7 +81,8 @@ const QuoteBody = styled.div(({ theme }) => ({
   },
 }))
 
-const AttestationBlock = styled.div(({ theme }) => ({
+const AttestationBlock = styled.blockquote(({ theme }) => ({
```
ChristopherChudzicki, line 84: The blockquote / cite tags are more semantic and unrelated to the rest of the changes in this PR.

I had been hoping that my screenreader would announce the quote (and citation) with these tags. Voiceover does not. So sad.

But JAWS and NVDA (which account for something like 85% of the screenreader usage—in US? worldwide? I'm not sure—do announce semantic quote blocks. So decided to keep it.

See https://adrianroselli.com/2023/07/blockquotes-in-screen-readers.html

abeglova (approved):
Works great generally. One thing I did notice is that the arrows scrolling left and right  in the carusel were accessed separately via tab but the "All" "Free" "With Certificate" and "All" "Videos" "Podcast" button groups were a single tab item and then you navigate between them with arrows. Probably it should be consistent but not a big deal if that would be hard

ChristopherChudzicki:
@abeglova Regarding

> One thing I did notice is that the arrows scrolling left and right in the carusel were accessed separately via tab but the "All" "Free" "With Certificate" and "All" "Videos" "Podcast" button groups were a single tab item and then you navigate between them with arrows.

(tldr: This is expected and good behavior.)

In the PR description when I said
>  But screenreaders also support navigation via arrow keys to a wider variety of content. Additionally, if an HTML element has special behavior associated with arrow keys (e.g., via keydown listeners) that generally won't work on a screenreader.

That was a bit of an oversimplification. I wish I had a better reference for this, but according to this [SO post](https://stackoverflow.com/questions/70304270/while-using-screen-reader-such-as-nvda-how-to-press-left-or-right-key)
> A screen reader has two modes: browse mode (where keyboard events are sent to the screen reader) and forms mode (where keyboard events are sent to the html element).
> 
> The default setting for most screen readers is to have automatic switching between the two modes for certain types of html elements, although you can manually switch between the two modes as well.

For example, inside a radiogroup the screenreader's arrow keys behave the same way they do for non-screenreader users.

Now, regarding the tabs...

The tab elements have `role="tab"` (and are inside an element with `role="tablist"`). The behavior of elements with roles is generally specified by the WAI ARIA standard. For tabs 
> When focus is on a tab element in a horizontal tab list:
> - Left Arrow: moves focus to the previous tab. If focus is on the first tab, moves focus to the last tab. Optionally, activates the newly focused tab (See note below).
> - Right Arrow: Moves focus to the next tab. If focus is on the last tab element, moves focus to the first tab. Optionally, activates the newly focused tab (See note below).
> –from https://www.w3.org/WAI/ARIA/apg/patterns/tabs/#keyboardinteraction

Our tabs have that arrow key behavior (provided by MUI), **and** screenreaders treat tabs like a form widget, so the arrow key navigation works correctly within the tablist for both screenreader users and nonscreenreader users.

My understanding is that screenreader users tend to know when a role supports arrow key navigation and when it does not.

You might say... "OK, can we find a role that *does* support arrow key nav and for screenreaders **and** is appropriate for a carousel". Maybe. (I think some carousels use [listbox](https://www.w3.org/WAI/ARIA/apg/patterns/listbox/), which is similar to a  `<select>` element. But I believe that's intended for more for forms. More commonly—e.g., on https://www.hulu.com/—I see carousels with  use `role="group"` for the individual slides, as we have done here.)

PR #1463, opened 2024-08-23
add to list dialog updates

gumaerc:
Closes https://github.com/mitodl/hq/issues/5275

This PR alters the Add to List dialog used to add Learning Resources to both Learning Paths and User Lists. Instead of adding the resource to lists immediately upon checking the box, the user's changes are saved upon clicking the new "Save" button. New API endpoints were added to accomplish this, sending all the lists that the resource should be a part of as one request. The styling of the modal was also updated to reflect the current designs in Figma.

image
image

### How can this be tested?
 - Make sure you are logged in as a user with `is_staff` or `is_superuser` set to true
 - Make sure you have backpopulated at least some data into your local database and generated a search index that will give you some search results
 - Go to the search page at http://localhost:8063/search
 - On any search result, click the bookmark button to display the Add to User List modal
 - Use the Create New List button to create a few lists
 - Check some boxes and click the x in the upper right of the modal
 - Verify that your changes were not saved
 - Open the modal again and this time select a few lists and hit the save button
 - Verify that the bookmark icon on the resource you selected is now highlighted
 - When you click the bookmark icon again, your selections should come back up
 - Repeat this process with the list button next to the bookmark button, which should open the Add to Learning Path dialog


ChristopherChudzicki (changes requested):
The new dialog is UI is looking good. I did notice a few functionality issues.
- An issue with Featured courses ont he homepage
- A permissioning issue... Currently, if one user updates the resource membership, that seems to be affecting all other users.

A suggestion for the first issue is in the review. Overall suggestion for the API below.

This PR adds a variety of new APIs:
```
GET, POST /api/v1/learning_resources/{learning_resource_id}/relationships/
GET, POST, PATCH, DELETE /api/v1/learning_resources/{learning_resource_id}/relationships/{id}
PATCH /api/v1/learning_resources/{learning_resource_id}/relationships/set_learning_path_relationships/
PATCH /api/v1/learning_resources/{learning_resource_id}/relationships/set_user_list_relationships/
```
Did you intend to add these all?

**Request/Suggestion:** IMO, we should only add two new API endpoints:
```
PUT /api/v1/userlists/{userlist_id}/items/
PUT /api/v1/learningpaths/{learningpath_id}/items/
```
The reason I suggest these two endpoints is:
- Currently, our only data-mutating endpoints are under `/api/v1/userlists/` and `/api/v1/learning_paths/`. This structure maintains that pattern
- By adding them in  `UserListItemViewSet` / `LearningPathItemViewSet` I believe permissioning issues will resolve automatically.
- In both cases, a `GET` api already exist that retrieves the items for the userlist/learningpath.
- *minor*: PUT seems more approach that PATCH (this overwrites the existing list entirely)

Unfortunately... I don't know how to add methods to the "listing" urls like `/api/v1/userlists/{userlist_id}/items/`. But maybe we could just add 
```python 
class UserListItemViewSet(NestedViewSetMixin, viewsets.ModelViewSet):

    @action(detail=False, methods=["POST"], url_path="set")
    def set_list(self, request, *args, **kwargs):
        pass 
```
Then the endpoints would be `POST /api/v1/userlists/{userlist_id}/items/set/` etc.

learning_resources/views.py
```diff
@@ -376,6 +378,125 @@ class ResourceListItemsViewSet(NestedViewSetMixin, viewsets.ReadOnlyModelViewSet
     ordering = ["position", "-child__last_modified"]
 
 
+@extend_schema(
+    parameters=[
```
ChristopherChudzicki, line 382: There are 3 parameters here... 2 are relevant for one endpoint, 2 are relevant for the other.

I would try moving the annotation to the methods themselves.

I also *think* you can skip the annotation for the path parameter. I think DRF-spectacular will pick that up automatically.

learning_resources/views.py
```diff
@@ -376,6 +378,125 @@ class ResourceListItemsViewSet(NestedViewSetMixin, viewsets.ReadOnlyModelViewSet
...
+class LearningResourceListRelationshipViewSet(viewsets.GenericViewSet):
+    """
+    Viewset for managing relationships between Learning Resources
+    and User Lists / Learning Paths
+    """
+
+    permission_classes = (AnonymousAccessReadonlyPermission,)
+    filter_backends = [MultipleOptionsFilterBackend]
+    queryset = LearningResourceRelationship.objects.select_related("parent", "child")
+    http_method_names = ["patch"]
```
ChristopherChudzicki, line 414: *Minor comment:* PUT still seems like a mildly better fit to me since this replaces the existing relationshisps (for that resource). When I briefly tried, I had an issue making PUT work with `detail=false`, but wince you're using `detail=true`, I think it would work.

learning_resources/views.py
```diff
@@ -376,6 +378,125 @@ class ResourceListItemsViewSet(NestedViewSetMixin, viewsets.ReadOnlyModelViewSet
...
+
+    def get_serializer_class(self):
```
ChristopherChudzicki, line 416: Do you know if `get_serializer_class` matters here? I would expect not since both of your methods call a specific serializer + have DRF-spectacular annotations. (More of a django question for my own education.)

learning_resources/views.py
```diff
@@ -376,6 +378,125 @@ class ResourceListItemsViewSet(NestedViewSetMixin, viewsets.ReadOnlyModelViewSet
...
+    @action(detail=True, methods=["patch"], name="Set Learning Path Relationships")
+    def learning_paths(self, request, *args, **kwargs):  # noqa: ARG002
+        """
+        Set Learning Path relationships for a given Learning Resource
+        """
+        if not is_learning_path_editor(request):
+            return Response(
+                {"error": "You do not have permission to edit learning paths"},
+                status=403,
+            )
```
ChristopherChudzicki, line 471: I'm fairly certain you can pass permission_classes to an action decorator, so you could pass `permission_classes=[HasLearningPathPermissions]` to the decorator, but this is good too.

learning_resources/views.py
gumaerc: You're right in that you can usually omit the `id` path param, but in this case `drf-spectacular` mislabels it so I'm just specifying it to fix that:

"A unique integer value identifying this learning resource relationship." vs "id of the learning resource"

learning_resources/views.py
gumaerc: Without it you get this error during the OpenAPI spec check:

```
/home/runner/work/mit-open/mit-open/learning_resources/views.py:381: Error [LearningResourceListRelationshipViewSet]: exception raised while getting serializer. Hint: Is get_serializer_class() returning None or is get_queryset() not working without a request? Ignoring the view for now. (Exception: 'LearningResourceListRelationshipViewSet' should either include a `serializer_class` attribute, or override the `get_serializer_class()` method.)
```

learning_resources/views.py
```diff
@@ -376,6 +379,129 @@ class ResourceListItemsViewSet(NestedViewSetMixin, viewsets.ReadOnlyModelViewSet
...
+        ],
+        responses={200: UserListRelationshipSerializer(many=True)},
+    )
+    @action(detail=True, methods=["patch"], name="Set User List Relationships")
+    def userlists(self, request, *args, **kwargs):  # noqa: ARG002
+        """
+        Set User List relationships for a given Learning Resource
+        """
+        learning_resource_id = kwargs.get("pk")
+        user_list_ids = request.query_params.getlist("userlist_id")
```
ChristopherChudzicki, line 430: **Suggestion:**  Throw a 403 here if one of the userlists does not belong to user
Something like
```python
if (UserList.objects.filter(pk__in=user_list_ids).exclude(author=request.user).exists()):
    msg = f"You do not have permission to edit all of {user_list_ids}"
    raise PermissionDenied(msg)
```
and maybe add a test for the 403 error.

I think this is worthwhile and doesn't seem too complicated

learning_resources/views_userlist_test.py
```diff
@@ -283,3 +283,25 @@ def test_get_resource_user_lists(client, user, is_author, is_unlisted):
             assert item.get("child") == course.learning_resource.id
     else:
         assert items_json == []
+
+
+def test_set_userlist_relationships(client, user):
+    """Test the userlists endpoint for setting multiple userlist relationships"""
```
ChristopherChudzicki, line 289: **Request:** I think it's definitely worth testing "User X changes relationship does not affect other users' relationships" (earlier behavior in this PR.)

**Suggestion:** To this test, add something like
```python 
userlist_other_author = factories.UserListFactory.create()
assert userlist_other_author.author != userlists[0].author # I like this for sanity 🤷
factories.UserListRelationshipFactory.create(
    parent=userlist_other_author, child=course.learning_resource
)
# Then at the end, assert child still belongs to userlist_other_author
```

ChristopherChudzicki (changes requested):
The behavior and implementation all look good 👍 👍 I made one request regarding a test that I think should be added.

gumaerc:
@ChristopherChudzicki all set

learning_resources/views_userlist_test.py
```diff
@@ -283,3 +283,70 @@ def test_get_resource_user_lists(client, user, is_author, is_unlisted):
...
+    )
+    url = reverse(
+        "lr:v1:learning_resource_relationships_api-userlists",
+        args=[course.learning_resource.id],
+    )
+    client.force_login(user)
+    client.patch(url)
+    assert (
+        UserListRelationship.objects.filter(child=course.learning_resource).count() == 3
+    )
```
ChristopherChudzicki, line 352: 👍 👍 This fails if the `parent__author=request.user` filter is missing. That's what I was hoping for. Ty!

ChristopherChudzicki (approved):
👍 

I did just notice that all of our dialogs use 28px on the bottom, but Dialog.tsx is using 40px on the bottom.

We could add to this PR. Otherwise I'll make a followup immediately after you merge this.

Screenshot 2024-08-29 at 10 44 35 AM

PR #1475, opened 2024-08-28
Switched imagekit caching from in-memory to redis

rhysyngsun:

N/A, follow up to https://github.com/mitodl/mit-open/pull/1472

This switches the in-memory cache for imagekit over to a redis cache since the in-memory cache seems to negatively impact performance when things expire / when the server first starts.

It also adds support for the `TIMEOUT` setting to `FallbackCache` since I needed that to have these values have a finite TTL in redis. This required a bit of a rework of how `get()` works - it calls `set()` on the caches that were missed. This seemed counterintuitive because typically in a cache you'd call `get()` and it'd have no side effects, however, we need to populate those caches when we have partial misses. If all caches miss then most likely the code calling `get()` will compute the value and then call `set()`, but if `get()` returns a value, `set()` is typically never called because we just got the value from that cache.

### How can this be tested?

- Go to http://api.open.odl.local:8063/api/v0/testimonials/?format=json, verify you get a response with image urls.
- To verify the data has been cached in redis, run: 
```bash
docker compose exec redis redis-cli -n 4 KEYS ":1:imagekit:*"
```

### DIFF
--- a/main/cache/backends.py
+++ b/main/cache/backends.py
@@ -1,5 +1,5 @@
 from django.core.cache import caches
-from django.core.cache.backends.base import BaseCache
+from django.core.cache.backends.base import DEFAULT_TIMEOUT, BaseCache
 
 
 class FallbackCache(BaseCache):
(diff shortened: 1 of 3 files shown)


mbertrand (approved):
LGTM 👍

PR #1477, opened 2024-08-28
Remove the in-memory cache for imagekit

rhysyngsun:

N/A

This removes the in-memory cache for imagekit, which is causing some performance issues. I will be following this up with another PR to add in redis caching, but this is currently holding up releases so this PR simply removes the troublesome caching.

### How can this be tested?

You should be able to hit the testimonials API without issue: api.open.odl.local:8063/api/v0/testimonials/?format=json


mbertrand (approved):
👍

PR #1492, opened 2024-08-30
Setup sentry tracing and profiling

rhysyngsun:

To-do from https://github.com/mitodl/hq/discussions/5347

This adds integration with Sentry's profiling and tracing features. This adds distributed tracing to the application so we can trace and profile requests from the frontend all the way through the backend.

I also removed newrelic and silk as Sentry's functionality replaces both of those and does it better.

### How can this be tested?

- In `env/shared.local.env` add (sentry DSN you can get from the RC settings):
```
SENTRY_DSN=
SENTRY_TRACES_SAMPLE_RATE=1.0
SENTRY_PROFILES_SAMPLE_RATE=1.0
```
- `docker compose up` and navigate around the site, you should see this data showing up in the "Profiles" / "Traces" / "Performance" pages in Sentry.


shanbady (approved):
one thing I was curious about is what the sample rate should be in production environments but otherwise looks and works well.

PR #1507, opened 2024-09-03
Cache unit page counts

shanbady:

Closes https://github.com/mitodl/hq/issues/5246

Currently the "programs" and "courses counts displayed on the unit page(/units/) comes from opensearch aggregations which has been determined to be expensive for just getting counts of things that change infrequently. This PR moves the counts off to their own endpoint which is then cached and invalidated along with learning resources (which get invalidated when new resources are fetched).

### How can this be tested?
1. checkout this branch and make sure you have learning resources and units loaded
2. visit /units/ and see that counts show up.
3. go into django admin and delete some resources associated with a unit
4. go back to /units/ and note that the counts do not changed (they are fetched from cache)
5. try this as a logged in and logged out user (cache behavior should be the same)

### DIFF
--- a/channels/serializers.py
+++ b/channels/serializers.py
@@ -160,6 +160,47 @@ class ChannelBaseSerializer(ChannelAppearanceMixin, serializers.ModelSerializer)
         exclude = ["published"]
 
 
+class ChannelCountsSerializer(serializers.ModelSerializer):
+    lists = serializers.SerializerMethodField()
+    counts = serializers.SerializerMethodField(read_only=True)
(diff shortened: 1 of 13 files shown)


mbertrand (commented):
Looks great, only question is regarding this:

> which get invalidated when new resources are fetched

I'm not seeing that, if I add new OCW courses, the OCW count remains the same.  Should the cache get invalidated when courses are added?

mbertrand (approved):
Cache is invalidated now after adding more courses, 👍

PR #1508, opened 2024-09-03
Ingest OLL courses from a spreadsheet/csv

mbertrand:
Closes https://github.com/mitodl/hq/issues/5350

Ingests OLL courses from a spreadsheet - either the one added in this PR or directly from google sheets, depending on kwargs.

### How can this be tested?
Run `./manage.py backpopulate_oll_data` followed by `backpopulate_oll_files`, they both should complete without errors.


main/settings_course_etl.py
```diff
@@ -57,13 +57,8 @@
 MITX_ONLINE_PROGRAMS_API_URL = get_string("MITX_ONLINE_PROGRAMS_API_URL", None)
 MITX_ONLINE_COURSES_API_URL = get_string("MITX_ONLINE_COURSES_API_URL", None)
 
+
 # Open Learning Library settings
-OLL_API_URL = get_string("OLL_API_URL", None)
-OLL_API_ACCESS_TOKEN_URL = get_string("OLL_API_ACCESS_TOKEN_URL", None)
```
shanbady, line 62: we should remove other references to these settings. There are some in env/codespaces.env and one in learning_resources/etl/conftest.py

learning_resources/etl/oll.py
```diff
@@ -1,46 +1,186 @@
...
+    Returns:
+        dict: list of normalized course topics data
     """
-    Return the OpenEdxConfiguration for edX.
+    return [
+        # One topic name from GSheets is slightly incorrect
+        {"name": topic.replace("Educational Policy", "Education Policy")}
+        for topic in [
+            course_data["MITxO Primary Child"],
+            course_data["MITxO Secondary Child "],
```
shanbady, line 85: Is the trailing space on this field intentional? I also didnt see this column in the sheet - is this supposed to be "MITxO Primary Parent"?

shanbady (changes requested):
commands worked for me. left some minor comments

learning_resources/etl/oll.py
mbertrand: "MITxO Secondary Child " is in the online sheet (but hidden) with a trailing space in the column title, so I left it that way too in the included csv file to keep them compatible.  If you download the sheet as a csv manually ("File"->"Download"->"csv"), it will be there.

mbertrand: The parent topic for each will be assigned automatically by MIT Learn during ingestion, so only the child is needed.

For example, [according to this spreadsheet](https://docs.google.com/spreadsheets/d/1Euu4ezPKeYkiTThvmxB5vFONFNG4fOULNW42F6qLuFI/edit?pli=1&gid=431773359#gid=431773359), "Machine Learning" is a child of "Data Science, Analytics & Computer Technology" while "Physics" is a child of "Math and Science", and all 4 show up for the OLL course "Mechanics: Kinematics and Dynamics".

Screenshot 2024-09-04 at 10 39 45 AM

main/settings_course_etl.py
mbertrand: Good catch, removed

pdpinch:
This might be too late, but I just got Joe Martis to add a column of secondary topics to https://docs.google.com/spreadsheets/d/1bWS-mqmuVraseMfY2YRu6gRnhXnjAS2N8CLiw4Tl8PE/edit?gid=0#gid=0

mbertrand:
@pdpinch There already were secondary parent/child topic columns in the spreadsheet, they were just hidden away when viewed online but were present in the downloaded csv.  Does this new one take precedence?  The values definitely aren't the same as the hidden secondary topic column.

pdpinch:
I never figured out where those hidden ones came from. The new ones were deliberate.

mbertrand:
I'll make a follow-up PR to use the new column and ignore the old ones (if they are still there).

PR #1510, opened 2024-09-03
Add LearningResource.delivery field (eventually to replace learning_format)

mbertrand:
Partially addresses https://github.com/mitodl/hq/issues/5126

- Adds a `delivery` field to `LearningResource` with a default value the same as for `learning_format`
- Modifies the OCW and Sloan ETL pipelines to populate the values correctly.  The default value is fine for all other sources.
- Updates the search index to include the `delivery` field.

### How can this be tested?
- Ingest these OCW courses if you haven't done so already - you'll need to assign `OCW_` env settings to the same as RC.
```
./manage.py backpopulate_ocw_data --course-name 16-810-engineering-design-and-rapid-prototyping-january-iap-2007 --skip-contentfiles
./manage.py backpopulate_ocw_data --course-name res-7-006-7-03x-genetics --skip-contentfiles
```
) 
- Run `docker compose up`,  the new migration should complete.
- Check that the 16.810 course has a `delivery` value of `["online", "offline"]`, while the 7.006 course has a value of just `["online"]`
- You can run the mgmt commands again, though this will temporarily make 7.006 include "offline" too until https://github.com/mitodl/ocw-hugo-themes/pull/1445 is merged and a mass publish is run.

### Additional Context
The `learning_format` field is not removed in this PR because doing so would cause complications with several related repos (course-search-utils, open-api-clients).  After both of these repos are updated to include `delivery`, then `learning_format` can be removed, and the front end can be updated to use `delivery` instead.


learning_resources_search/serializers.py
```diff
@@ -239,6 +239,7 @@ def to_representation(self, obj):
     "professional",
     "free",
     "learning_format",
+    "delivery",
```
abeglova, line 242: Search aggregated by delivery does not work RN you need to add `"delivery": FilterConfig("delivery.code"),` to SEARCH_FILTERS in learning_resource_search/constants.py for delivery to be a valid aggregation option.

Additionally you should add delivery to LearningResourcesSearchRequestSerializer

abeglova (changes requested):
Looks like you started to add delivery to the search api but did not finish. Other than that, works great

abeglova (approved):

PR #1516, opened 2024-09-04
Fixes for successful build

jonkafton:

Relates to: https://github.com/mitodl/hq/issues/5244

- Removes unnecessary Client Component wrappers and "use client" directives.
- Set `<Suspense />` boundaries to client render components that `useSearchParams()` around:
   - all of the Search page.
   - the Learning Resource drawer.
   - the dashboard carousels.

### How can this be tested?

Can now run a successful build with `yarn build`.


frontends/main/src/app/search/page.tsx
```diff
@@ -8,7 +8,11 @@ export const metadata: Metadata = getMetadata({
 })
 
 const Page: React.FC = () => {
-  return <SearchPage />
+  return (
+    <Suspense>
```
ChristopherChudzicki, line 12: For our future selves...

NextJS has two types of rendering (that can be mixed):
- static ... renders pages at build time
- dynamic ... renders pages at request time

At the moment, most of our pages are static because we don't do much at request time.[^1] Within a static page, `useSeachParams` must be wrapped in a Suspense boundary. Content inside the suspense boundary won't be rendered until the client gets it. Necessary because the statically-rendered pages don't know what to render, since they don't have the search params.

[^1]: Two places we are currently doing something on the server is homepage carousel data prefetching and metadata generation.

ChristopherChudzicki (approved):
👍 One request...

Could you remove `frontends/main/src/libraries`?

PR #1526, opened 2024-09-06
change search mode default

abeglova:
https://github.com/mitodl/mit-open/pull/1456

This pr changes the dafault search mode to "phrase search" with slop=6. This gives fewer and better results than our current search

### How can this be tested?
Go to the search. Verify that everything still works as expected.


mbertrand (commented):
Search still works, but I didn't notice any difference in results when searching `?q=Biology` and switching between "best_fields" or "phrase" or adjusting slop values.  Is there a particular query that would reveal differences?

PR #1540, opened 2024-09-09
Update docker-compose config to support OS cluster

rhysyngsun:

Part of https://github.com/mitodl/hq/issues/5179

This updates the docker compose setup to support running OpenSearch in a 3-node cluster locally.

### How can this be tested?

- You should be able to run `docker compose up` as usual without any additional configuration.
- Test the 3-node cluster:
  - `docker compose stop` if it's running
  - Set `export OPENSEARCH_CLUSTER_TYPE=cluster` in your shell
  - `docker compose up` - you should see 3 opensearch nodes instead of 1
  - Run `manage.py recreate_index --all` as the new cluster uses separate volumes to avoid breaking things if you switch back and forth.
  - You should be able to search as with a single node cluster.


shanbady (changes requested):
codespaces appears to have issues. I think you may need to adjust the service name in the [devcontainer.json](https://github.com/mitodl/mit-open/blob/shanbady/cache-remaining-views/.devcontainer/devcontainer.json#L7)

shanbady (changes requested):
@rhysyngsun still appears to be some issue - 
This is a working codespace note the ports visible and that the uwsgi process is bound to port 8061, the api server is exposed at 8063 and frontend at 8062.

Screenshot 2024-09-12 at 10 47 05 AM

On this branch there are only 2 ports that show up and uwsgi process is bound to 8063 instead of 8061

Screenshot 2024-09-12 at 10 36 09 AM

If you can get the port binding to look the same as in the first screenshot I think we will have a working setup. I recall not having to explicitly exposing or binding to 8061 (that somehow seemed to automatically happen) but the 8063 I had to bind via the PORT env var.

shanbady (approved):
👍

PR #1555, opened 2024-09-12
add caching to remaining views

shanbady:

Closes https://github.com/mitodl/hq/issues/5450

This PR adds caching to the remaining server side views. We cache for all users on views that don't have any user specific info using "cache_page_for_all_users" decorator otherwise we use "cache_page_for_anonymous_users" to only cache for anonymous users

### How can this be tested?
1. checkout this branch. have data loaded
2. as you browse around the first page view will cache api calls (most of them for anonymous users only but some for all users)
3. make note of performance differences between main and the cached version
4. to clear all the caches we can manually run 
```
from main.utils import clear_search_cache
clear_search_cache()
```

there are some further instructions (specific to search cache but can now be applied to remaining api views) [here](https://github.com/mitodl/mit-open/pull/1392)

### Additional Context
Some key things in this PR for reviewer to double check:
1. The cache mechanism used for each view is appropriate (to cache for all users vs just anonymous)
2. Are the resources in the cached view refreshed appropriately (every 24 hours on loading new resources). Are there any views we should remove this caching from or tweak when it is cleared.

The cache is by default now set to the "DummyCache" - we can explicitly enable and disable on a per test basis if the point of the test is to test something with caching enabled as done [here](https://github.com/mitodl/mit-open/pull/1555/files#diff-53d113070f251b6fa10ae4a451b0c5ef650923e54ebe3c5fd4157961c1b2f2cfR488)

### DIFF
--- a/articles/views.py
+++ b/articles/views.py
@@ -8,7 +8,7 @@ from rest_framework.permissions import IsAdminUser
 from articles.models import Article
 from articles.serializers import ArticleSerializer
 from main.constants import VALID_HTTP_METHODS
-from main.utils import cache_page_for_all_users
+from main.utils import cache_page_for_all_users, clear_search_cache
 
 # Create your views here.
 
@@ -48,3 +48,11 @@ class ArticleViewSet(viewsets.ModelViewSet):
     )
     def list(self, request, *args, **kwargs):
         return super().list(request, *args, **kwargs)
+
+    def create(self, request, *args, **kwargs):
+        clear_search_cache()
+        return super().create(request, *args, **kwargs)
+
+    def destroy(self, request, *args, **kwargs):
+        clear_search_cache()
+        return super().destroy(request, *args, **kwargs)
--- a/articles/views.py
+++ b/articles/views.py
@@ -1,3 +1,5 @@
+from django.conf import settings
+from django.utils.decorators import method_decorator
 from drf_spectacular.utils import extend_schema, extend_schema_view
 from rest_framework import viewsets
 from rest_framework.pagination import LimitOffsetPagination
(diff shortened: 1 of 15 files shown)


learning_resources/views.py
```diff
@@ -994,6 +1066,11 @@ def get_queryset(self) -> QuerySet:
             .distinct()
         )
 
+    @method_decorator(
```
mbertrand, line 1069: There was some deliberate results randomization added to this featured list view so as not to give any apparent preference to any particular unit.  Might be good to double-check with Ferdi to see if it's okay if that randomization is cached so all anonymous users see the same result for 24 hours before it's randomized again.

mbertrand (commented):
Looks great, just one comment about caching of the featured resources view (check with Ferdi about that one).

Also, the news_events/views/FeedItemViewSet and FeedSourceViewSet could use caching.

learning_resources/views.py
mbertrand: PS and clear the cache after the news_events tasks or mgmt command are run.

mbertrand:
PS and clear the cache after news_events tasks or mgmt command are run

articles/views.py
```diff
@@ -37,3 +40,11 @@ class ArticleViewSet(viewsets.ModelViewSet):
...
     permission_classes = [IsAdminUser]
     http_method_names = VALID_HTTP_METHODS
+
+    @method_decorator(
+        cache_page_for_all_users(
+            settings.SEARCH_PAGE_CACHE_DURATION, cache="redis", key_prefix="search"
+        )
+    )
+    def list(self, request, *args, **kwargs):
+        return super().list(request, *args, **kwargs)
```
mbertrand, line 50: Probably want to clear this cache when a new article is  created, edited, or deleted

Create (no UI yet, just via api): http://api.open.odl.local:8063/api/v1/articles/

Editing UI (after created above): http://open.odl.local:8062/articles/<aricled_id>/edit

learning_resources/views.py
shanbady: double checked with ferdi. He said we can leave this as-is and have a separate ticket to handle the randomization on the frontend.

PR #1558, opened 2024-09-13
always use dfs_query_then_fetch

abeglova:
https://github.com/mitodl/hq/issues/5351

Programs are consistently not shown in search results with a text search because scores are calculated separately on each index and not consistent  between indexes. Setting search_type="dfs_query_then_fetch' fixes the issue. In https://github.com/mitodl/mit-open/pull/1518 i added a search parameter to use that search type because I was worried that we would get worse search performance. I tested search on rc and search_type="dfs_query_then_fetch' does not seem to affect performance, so we can just use it for all searches.

### How can this be tested?
verify that search works as expected 
verify that http://open.odl.local:8062/search/?q=Machine+Learning returns a program on the first or second page

### DIFF
--- a/frontends/api/src/generated/v1/api.ts
+++ b/frontends/api/src/generated/v1/api.ts
@@ -3623,12 +3623,6 @@ export interface PercolateQuerySubscriptionRequestRequest {
    */
   dev_mode?: boolean | null
   /**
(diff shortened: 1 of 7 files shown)


shanbady (approved):
A program showed up on the 4th page for me but I assume thats due to the difference in the data loaded on my local.

PR #1511, opened 2024-09-03
display sub topics on topic channels

gumaerc:
### What are the relevant tickets?
Closes https://github.com/mitodl/hq/issues/5295

### Description (What does it do?)
This PR adds link chips for sub topics to the top of topic channel pages. The topics API endpoint was modified to accept a new `parent_topic_id` parameter. This allows you to pass the ID of a topic and get the direct child topics in return.

### Screenshots (if appropriate):
image
image

### How can this be tested?
 - Spin up this branch of `mit-learn`
 - Visit the topics listing page at http://localhost:8062/topics
 - Click on the header for any of the topics
 - Ensure that the subtopic chips are properly displayed beneath the banner
 - Click the subtopic chips and ensure you are brought to the proper topic page

### Additional Context
The spacing below the topics and above the search input may look incorrect, but it will be adjusted as part of some coming changes to the channel page search box.

PR #1560, opened 2024-09-13
[NextJS] Fix tests, re-enable on CI

ChristopherChudzicki:
Closes https://github.com/mitodl/hq/issues/5414

This fixes our Jest tests for NextJS branch and re-enables them on CI

### How can this be tested?
1. Run the app locally with `docker compose up` or `yarn watch`. It should look good.
2. Tests should pass.


frontends/main/src/app-pages/ChannelPage/ChannelPage.tsx
```diff
@@ -11,21 +12,74 @@ import type {
...
+  const topicsQuery = useLearningResourceTopics({
+    parent_topic_id: [parentTopicId],
+  })
+  const totalSubtopics = topicsQuery.data?.results.length ?? 0
+  const subTopics = topicsQuery.data?.results.filter(
+    propsNotNil(["channel_url"]),
+  )
+  return (
+    totalSubtopics > 0 && (
+      <SubTopicsContainer>
```
ChristopherChudzicki, line 57: The subtopic work Carey added in #1511 seems to have been accidentally removed in https://github.com/mitodl/mit-open/pull/1525. (But not the test, thankfully!). This adds it back.

frontends/main/src/app-pages/AboutPage/AboutPage.test.tsx
```diff
@@ -1,20 +1,10 @@
...
-    setMockResponse.get(urls.userMe.get(), {
-      [Permissions.Authenticated]: true,
-    })
-
-    renderTestApp({
-      url: commonUrls.ABOUT,
-    })
-    await waitFor(() => {
-      expect(document.title).toBe("About Us | MIT Learn")
-    })
```
ChristopherChudzicki, line 17: I removed this and some other metadata tests since the metadata is now handled by NextJS.

It probably is worth having metadata tests, but they should be added in https://github.com/mitodl/hq/issues/5410

frontends/main/src/app-pages/ErrorPage/NotFoundPage.test.tsx
```diff
@@ -4,7 +4,7 @@ import { renderWithProviders, screen } from "@/test-utils"
 import { HOME } from "@/common/urls"
 import NotFoundPage from "./NotFoundPage"
 
-test("The NotFoundPage loads with meta", async () => {
+test.skip("The NotFoundPage loads with meta", async () => {
```
ChristopherChudzicki, line 7: This is the only test to which I added a `skip`, since it will be changed anyway in [ISSUE TO BE CREATED]

frontends/main/src/common/urls.ts
```diff
@@ -1,8 +1,10 @@
+import invariant from "tiny-invariant"
+
 const generatePath = (
   template: string,
   params: Record<string, string | number>,
 ): string => {
-  return template.replace(/:(\w+)/g, (_, key) => {
+  return template.replace(/\[(\w+)\]/g, (_, key) => {
```
ChristopherChudzicki, line 7: Marking params as with `[...]` is much more convenient for use with `next-router-mock`... see https://github.com/mitodl/mit-open/pull/1560/files#diff-22d69d0f979fef7f8d9d0b0cd55dd7a00c6c3f4302ac7d542f79220acee4f905R18

frontends/main/src/common/urls.ts
```diff
@@ -14,16 +16,16 @@ export const HOME = "/"
 
 export const ONBOARDING = "/onboarding"
 
-export const LEARNINGPATH_LISTING = "/learningpaths/"
-export const LEARNINGPATH_VIEW = "/learningpaths/:id"
+export const LEARNINGPATH_LISTING = "/learningpaths"
```
ChristopherChudzicki, line 19: NextJS by default canonicalizes removes trailing `/` from paths.

If we prefer to keep them, we can configure that with `next.config.js` (which affects the behavior of NextJS server and of the `Link` component). But it's slightly annoying since the jest tests aren't aware of next.config.js.

frontends/ol-test-utilities/src/mocks/nextNavigation.test.ts
```diff
@@ -0,0 +1,48 @@
+import { renderHook } from "@testing-library/react"
+import { nextNavigationMocks } from "./nextNavigation"
+import mockRouter from "next-router-mock"
+import { createDynamicRouteParser } from "next-router-mock/dynamic-routes"
+
+const { usePathname, useSearchParams } = nextNavigationMocks
+
+describe("Mock Navigation", () => {
```
ChristopherChudzicki, line 8: Added for sanity as I debugged an issue with the mock navigation stuff.

gumaerc (approved):
👍 

Looks good, just one small nitpick:

PR #1564, opened 2024-09-16
Prevent unnecessary refreshes and add sort order

abeglova:
closes https://github.com/mitodl/hq/issues/5454

This pr fixes two related bugs:

1) The search page refreshes if you go to a different tab or just leave the page open for too long
2) If there is no search term, the search results don't have a determined order after the featured resources on the first few pages so that if you refresh https://rc.learn.mit.edu/search/?page=10 a few times, for example the results change. The results are the same from each node but not the same between nodes so if the query is picked up by a different node, the results change. This means that if a user is paging through resources there will be some duplicates and some resources that aren't shown.

For a user this combination of bugs is annoying because if the user clicks through to a learning resource then comes back, they lose their search results

### How can this be tested?
- Verify that both the home page and the search page still load everything

- To test 1) Go to http://open.odl.local:8062/search/?page=10. Go to a different tab and stay there for a minute. Come back to your original tab. The page should not refresh
- Unfortunately 2) can't be tested locally, since locally there is only one opensearch node locally. It can be tested on rc

### DIFF
--- a/frontends/mit-learn/src/services/react-query/react-query.ts
+++ b/frontends/mit-learn/src/services/react-query/react-query.ts
@@ -16,7 +16,8 @@ const createQueryClient = (): QueryClient => {
   return new QueryClient({
     defaultOptions: {
       queries: {
-        staleTime: 30 * 1000,
+        refetchOnWindowFocus: false,
+        staleTime: Infinity,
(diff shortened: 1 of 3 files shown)


ChristopherChudzicki (approved):
👍 Looks good to me.

PR #1565, opened 2024-09-16
add SignupPopover to learning resource drawer bookmark button

gumaerc:
Closes https://github.com/mitodl/hq/issues/5505

This PR adds some functionality that was missed in https://github.com/mitodl/mit-open/pull/1537. If you are logged out, the bookmark button should still show in the learning resource drawer. When clicking this button, the same "signup popover" that shows when you click the bookmark button on the card show show.

image
image

### How can this be tested?
 - Make sure you have at least a few learning resources in your database
 - Open an incognito window or log out
 - Visit the search page
 - Click on any result to open the learning resource drawer
 - Within the drawer, click the bookmark button
 - You should see a popover as seen in the screenshots above
 - Click "Sign Up"
 - Log in or create a new account
 - Verify that you are returned to the page with the learning resource drawer open, and you can now click the bookmark button and add the resource to a list


frontends/mit-learn/src/page-components/ResourceCard/ResourceCard.tsx
```diff
@@ -124,6 +134,7 @@ const ResourceListCard: React.FC<ResourceListCardProps> = ({
   condensed,
   ...props
 }) => {
+  const loc = useLocation()
```
shanbady, line 137: It seems like there is a lot of redundancy between ResourceCard and ResourceListCard that could potentially be abstracted/consolidated.

frontends/ol-components/src/components/LearningResourceExpanded/InfoSection.tsx
```diff
@@ -86,6 +87,10 @@ const Certificate = styled.div`
     height: 16px;
   }
 `
+// The SignupPopover needs to display over the top of the resource drawer
+const StyledSignupPopover = styled(SignupPopover)({
+  zIndex: 9999,
```
shanbady, line 92: it would be good to avoid the hardcoded zIndex if at all possible (realizing this might be the only workaround for now)

shanbady (approved):
works well. left some minor nitpicks. it would be good to resolve the redundant code issue if possible but not a blocker

gumaerc:
@shanbady Thanks for the review. I've addressed your two concerns above.

PR #1567, opened 2024-09-16
Remove learning_format, use delivery for search filter/facet

mbertrand:
Part 2 of https://github.com/mitodl/hq/issues/5126, depends on https://github.com/mitodl/course-search-utils/pull/123

- Uses an updated version of course-search-utils (https://github.com/mitodl/course-search-utils/pull/123), ~~which needs to be merged before this one.  Once it is merged and released, then this PR will be updated to use the new release version.  Currently it is pinned to the course-search-utils PR commit~~.
- Removes the `learning_format` field from the `LearningResource`  model
- Renames `learning_format` to `delivery` in the `Profile` model
- Removes "learning_format" from the ETL pipelines 
- Updates the frontend to use "delivery" instead of "learning_format" for filtering and facets

### How can this be tested?
- Run some searches, with and without using the "Format" facet, everything should work as expected.
- Update your preferred format for your user profile, everything should work as expected.


shanbady (changes requested):
Seems like [this factory attribute](https://github.com/mitodl/mit-open/blob/ad1ef032902b2f3b76b132afb6c9d979cfa48d4a/frontends/api/src/test-utils/factories/learningResources.ts#L156) still remains but otherwise looks good

mbertrand:
@shanbady that "format" factory attribute is for `LearningResourceOfferor.format` which is not quite the same thing as `LearningResource.learning_format`.  I made a bunch of changes based on feedback from Simone, because some more planning has to go into how the switchover to "delivery" will be presented on the front end:

- Removed all "offline" values from courses via a migration and updated the OCW ETL pipeline to not add it unless `OCW_OFFLINE_DELIVERY=True` (default=`False`)
- Reverted "Format"->"Delivery" frontend text changes
- Updated `package.json` and `yarn.lock` to use the official new version of `course-search-utils`

shanbady (approved):

abeglova (approved):
lgtm

PR #1576, opened 2024-09-18
Signup popover owns url

ChristopherChudzicki:
Followup to https://github.com/mitodl/mit-open/pull/1565

This PR has no functionality change, but makes some rebasing / conflicts easier on the NextJS branch. This PR tweaks two changes in introduced in https://github.com/mitodl/mit-open/pull/1565. Two changes introduced there were:
1. moving SignupPopover to `ol-components`
    - this had the undesirable consequence that SignupPopover no longer owns the signup url, which seems appropriate. Additionally, many more components are accessing `useLocation`, which caused some conflicts on the `nextjs` branch. 
2. introducing a `MUI_DRAWER_Z_INDEX` constant.
    - this is more simply accessed via `theme.zIndex.drawer`

Change (1) was made so that `LearningResourceExpanded` could use the signup popover `. This PR moves that logic to the `LearningResourceDrawer`, which is more similar to how we handle the signup popover for learning resource cards and is also how we handle the "add to list dialogs".

### How can this be tested?
1. As an unauthenticated user, check that the signup popover opens when...
    - user clicks bookmark button on cards OR in resource drawer
    - user clicks subscription button on channel pages
2. Check that the add to list dialogs still open appropriately.


abeglova (approved):
👍

PR #1582, opened 2024-09-18
Codespace opensearch service fix

shanbady:
Fixes an issue where the opensearch service is not started when creating a codespace due to a descrepency with copose versions and how we now inherit compose configurations

### How can this be tested?
1. No need to check out any code! create a codespace directly off of this PR - You can click this in the bottom right to follow the build progress:
Screenshot 2024-09-18 at 2 59 34 PM

2. Let it finish initializing (A few minutes after you see 3 ports show up in the ports tab - you will want the initialization script to complete as well - a blue "Finished configuring codespace." should show up at the bottom of the terminal tab)
3. In the ports tab - set port 8063 to "public"
4. Visit the running frontend at the 8062 address - You should be able to go to the search page and see some results display

### DIFF
--- a/docker-compose.codespaces.yml
+++ b/docker-compose.codespaces.yml
@@ -19,3 +19,20 @@ services:
       file: docker-compose.apps.yml
       service: celery
     env_file: env/codespaces.env
+
+  opensearch-node-mitopen-1:
+    extends:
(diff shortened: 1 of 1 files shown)


rhysyngsun (approved):
LGTM

PR #1583, opened 2024-09-18
[NextJS] Load correct font and a few other global things

ChristopherChudzicki:
Closes https://github.com/mitodl/hq/issues/5473

This PR:
- loads the correct font for MIT Learn
- Adds a noindex tag, based on MITOL_NOINDEX
- Loads APPZI based on `NEXT_PUBLIC_APPZI_URL`.

### How can this be tested?
1. View any page in the app and inspect text. Check that the Neue Haas Grotesk font is being used. 
    - Checking what font is actually used is a little tricky. Here's how you can do it in Chrome dev tools.
    Screenshot 2024-09-18 at 5 12 15 PM

2. In your env file, set `NEXT_PUBLIC_APPZI_URL="https://w.appzi.io/w.js?token=Q2pSI"` *(this is not a secret env var, hence NEXT_PUBLIC and client side availability)*. Restart NextJS / Docker. The Appzi feedback form should appear on enter-right, just like prod.
3. Set `MITOL_NOINDEX=true` in your env file. and restart NextJS / Docker. The site should now include a noindex tag in `<head>`

<!--- Uncomment and add steps to be completed before merging this PR if necessary
### Checklist:
- [ ] Check with Jon about `NEXT_PUBLIC_*` values for docker container.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -128,6 +128,12 @@ jobs:
 
       - name: Build Next.js frontend
         run: yarn workspace main build
+        env:
+          NEXT_PUBLIC_ORIGIN: https://ci.learn.mit.edu
+          NEXT_PUBLIC_MITOL_API_BASE_URL: https://api.ci.learn.mit.edu
+          NEXT_PUBLIC_CSRF_COOKIE_NAME: cookie-monster
+          NEXT_PUBLIC_SITE_NAME: MIT Learn
+          NEXT_PUBLIC_MITOL_SUPPORT_EMAIL: help@ci.learn.mit.edu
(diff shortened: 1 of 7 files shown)


shanbady (approved):
approving - pending conflict resolution with nextjs

shanbady (changes requested):
It seems the frontend doesnt come up in a codespace. its possible that some of the env variables need to get added to the env/codespaces.env

ChristopherChudzicki:
> It seems the frontend doesnt come up in a codespace. its possible that some of the env variables need to get added to the env/codespaces.env

Probably is env var issue. I think this is worth fixing, but we should handle it in a separate PR.

shanbady (approved):
I'll create a separate issue for getting the codespaces to work. I'm guessing some of it may have to do with the hostname settings etc

PR #1591, opened 2024-09-20
[NextJS] configure query client like on `main`

ChristopherChudzicki:
Closes https://github.com/mitodl/hq/issues/5543

This PR adds most of the query config from `main` branch to the NextJS app.

### How can this be tested?
1. Load the homepage with network tab open. API requests should be made once, not over and over.
2. When authenticated, featured courses on homepage should have correct filled-in / unfilled bookmark icons.


frontends/main/src/app/page.tsx
```diff
@@ -21,26 +16,8 @@ export async function generateMetadata({
   })
 }
 
-const Page: React.FC = async () => {
-  const queryClient = getQueryClient()
-
-  /*
-   * Prefetch the first tab (All) of the carousel
-   */
-  await queryClient.prefetchQuery(
```
ChristopherChudzicki, line 30: Getting rid of this because with a nonzero staletime, the featured course bookmarks are problematic: the resources **are** user-dependent, so they can't really be fetched on the server, unless they are immediately refetched, which we don't want.

We can re-enable this once https://github.com/mitodl/hq/issues/5159 is done

jonkafton (commented):
:white_check_mark:

PR #1592, opened 2024-09-20
[NextJS] Make noindex tag opt-out (like main) rather than opt-in

ChristopherChudzicki:
Closes  https://github.com/mitodl/hq/issues/5565

This PR makes the `<meta name="robots" content="noindex">` the default. I.e., indexing is opt-in instead of opt-out. This is the same as what we currently do on `main`. Note that currently all releases of Learn should have noindex, per our "soft launch" status.

### How can this be tested?
1. With MITOL_NOINDEX undefined, check that you **do** get a noindex tag.
2. With `MITOL_NOINDEX=false`, check that you **do not** get a noindex tag.

### DIFF
--- a/frontends/main/src/app/page.tsx
+++ b/frontends/main/src/app/page.tsx
@@ -13,12 +13,9 @@ export async function generateMetadata({
 }: {
   searchParams: { [key: string]: string | string[] | undefined }
 }): Promise<Metadata> {
-  const robots =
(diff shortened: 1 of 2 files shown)


jonkafton (commented):
:white_check_mark:

jonkafton (commented):
:white_check_mark:

PR #1596, opened 2024-09-23
Add additional event capturing for some interactions

jkachel:

Closes https://github.com/mitodl/hq/issues/5520
Closes https://github.com/mitodl/hq/issues/5283
https://github.com/mitodl/hq/issues/5254

We need some events to fire when certain events happen in the frontend. These include:
- When a learner adds a course to a list (bookmarks it)
- When a learner adjusts their search parameters, including term

These aren't necessarily captured in an intuitive way automatically by PostHog, so this adds explicit events to capture these interactions. 

- For list adds, this will capture a `lr_add_to_list` event. This contains the same data as `lrd_view` (so, basic resource info) plus the type of list the resource was added to (either a user list or a learning path).
- For search query changes, this captures a `search_updated` event. This captures:
   - No extra data for facet changes.
   - An `isEnter` flag if the learner has adjusted their search term - this is True if they have submitted via Enter key, or False if they clicked the submit button/icon.

### How can this be tested?

If you don't have your own PostHog account set up, you can either set one up for this or use the MIT Learn Test project in the OL PostHog account.

> [!NOTE]
> If you're rolling your own, it is imperative that you also set up [this transformation](https://github.com/PostHog/posthog-app-url-parameters-to-event-properties) in Data Pipelines - it should exist in the list if you hit Add Transformation in PostHog. 
>
> The URL parameters to convert should be set to `q,resource_type,certification_type,delivery,department,topic,offered_by,free,professional` and the Prefix should be set to `search_param_`. I also turn on Ignore case and Add to user properties but these aren't strictly required. 

Test `lr_add_to_list` by adding a course to a list. You should see the event pop up in the Activity view (note that it may take a bit of time for it to show up). 

Test `search_update` by making changes to your search - select some facets, update your term. This should result in some events being captured in PostHog.

### Additional Context

`search_update` will fire once for _each_ facet that is chosen. So, if you select multiple items in a row, you will end up with events for all of those individual selections. This is _by design_ at this point - we will revisit later if we need to but we're explicitly not debouncing it now.

In testing, I noticed that there's a bug in https://github.com/PostHog/posthog-app-url-parameters-to-event-properties with regard to query parameter arrays: it does not handle them properly, so you only see whatever the first one is from the param. The facets code appends to the list rather than prepends so this means if you choose more than one of a given type of facet you'll only ever see the first one you chose. (This is especially noticeable for Topics.) This should really be fixed upstream so will work that out separately.

### DIFF
--- a/frontends/mit-learn/src/page-components/Dialogs/AddToListDialog.tsx
+++ b/frontends/mit-learn/src/page-components/Dialogs/AddToListDialog.tsx
@@ -10,6 +10,7 @@ import {
 } from "ol-components"
 
 import { RiAddLine } from "@remixicon/react"
+import { usePostHog } from "posthog-js/react"
 
 import * as NiceModal from "@ebay/nice-modal-react"
 
@@ -67,6 +68,7 @@ const AddToListDialogInner: React.FC<AddToListDialogInnerProps> = ({
     isLoading: isSavingLearningPathRelationships,
     mutateAsync: setLearningPathRelationships,
   } = useLearningResourceSetLearningPathRelationships()
+  const posthog = usePostHog()
   const isSaving =
     isSavingLearningPathRelationships || isSavingUserListRelationships
   let dialogTitle = "Add to list"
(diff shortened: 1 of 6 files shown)


shanbady (approved):
👍

PR #1602, opened 2024-09-24
Merge latest from main

jonkafton:

Regular sync to keep the `nextjs` branch up to date with main.

- Merges the latest changes from main
- Migrates https://github.com/mitodl/hq/issues/5527 to Next.js.


abeglova (approved):
This works as long as the DEFAULT_SEARCH_ variables are not set. Setting them from the .env file locally works on the main branch but not the nextjs branch

Since that is also broken on rc/prod, that's ok. I'll make a followup pr to fix it

PR #1608, opened 2024-09-25
Add task_reject_on_worker_lost=True to finish_recreate_index, use return instead of raise for replaced tasks

mbertrand:
Closes https://github.com/mitodl/hq/issues/5615

- Adds `reject_on_worker_lost=True` to finish_recreate_index task
- Uses`return` instead of `raise` on replaced tasks

### How can this be tested?
Run `./manage.py recreate_index --all` locally, it should still work.
Otherwise, probably best tested on RC by starting a reindex there and then restarting the worker dynos.


abeglova (approved):
LGTM

PR #1609, opened 2024-09-25
nextjs codespace fix

shanbady:

Closes https://github.com/mitodl/hq/issues/5604

This PR gets the nextjs branch to work in codespaces.

### How can this be tested?
1. create a codespace on this branch  - you will need to launch this in a 4 core container since it seems this setup requires more resources for it to work:
Screenshot 2024-09-25 at 3 28 17 PM
Screenshot 2024-09-25 at 3 29 27 PM

2. Once you see 3 ports show up in the ports tab at the bottom, right click port 8063 and set it to public
3. visit port 8062 in the browser (it may not immediately load which means its still building but it eventually will show up
4. poke around the site and see that its all there. the search page should populate with resources

### Additional Context
There were a few configuration related to nextjs that had to go in. opensearch also was not initializing presumably due to low resources when using a 2 core codespace


jonkafton (approved):
This is awesome!

PR #1610, opened 2024-09-25
Add custom label for percolate query subscriptions

shanbady:
Closes https://github.com/mitodl/hq/issues/5598

This PR adds a "display label" field to percolate queries which overrides any other label for display on the frontend

Screenshot 2024-09-25 at 3 36 49 PM

### How can this be tested?
1. Checkout this branch and run migrations.
2. Subscribe/follow to a few things
3. Go to [django admin](http://open.odl.local:8063/admin/learning_resources_search/percolatequery/) and edit a percolate query you are subscribed to. Put something custom in the "Display label" field
4. go to your settings page and view your subscriptions list. note that the subscription displays your custom label


learning_resources_search/models_test.py
```diff
@@ -152,3 +152,39 @@ def test_percolate_query_search_labels(mocker, mocked_es):
...
+    original_query = {
+        "q": "testing search filter",
+        "free": None,
+        "department": ["physics"],
+        "topic": ["math"],
+        "professional": None,
+        "certification": None,
+        "yearly_decay_percent": None,
+    }
+    test_label = "new courses about cats"
```
mbertrand, line 180: `source_description` doesn't seem to be getting tested here, suggest doing something like this:

```python
@pytest.mark.django_db
@pytest.mark.parametrize("test_label", ["new courses about cats", None])
@pytest.mark.parametrize("is_channel_query", [True, False])
def test_percolate_query_display_labels(
    mocker, mocked_es, test_label, is_channel_query
):
    """
    Test that makes sure we display the display label for a percolate query if it is defined
    """
    mocker.patch(
        "learning_resources_search.indexing_api.index_percolators", autospec=True
    )
    mocker.patch(
        "learning_resources_search.indexing_api._update_document_by_id", autospec=True
    )
    channel = ChannelFactory.create(
        search_filter="department=physics", channel_type="department"
    )
    original_query = {
        "q": "testing search filter",
        "free": None,
        "department": ["physics"] if is_channel_query else None,
        "topic": None,
        "professional": None,
        "certification": None,
        "yearly_decay_percent": None,
    }
    test_label = "new courses about cats"
    query = PercolateQueryFactory.create(
        original_query=original_query,
        query=original_query,
        display_label=test_label,
    )
    assert query.source_description() == (
        test_label
        if test_label
        else channel.title
        if is_channel_query
        else original_query
    )
```

learning_resources_search/models.py
```diff
@@ -37,6 +43,8 @@ def source_label(self):
 
     def source_description(self):
         channel = self.source_channel()
+        if self.display_label:
+            return self.display_label
         if channel:
             return channel.title
         return self.original_url_params()
```
mbertrand, line 50: Q: Should something else be displayed that's more user friendly than this when display_label and channel are both None?

Screenshot 2024-09-26 at 11 07 19 AM

mbertrand: One thing I found a bit counterintuitive but is outside the scope of this PR: the `source_description` is used as the title in this list while `source_label` is shown underneath that.  I would expect the opposite.

learning_resources_search/models.py
```diff
@@ -37,6 +43,8 @@ def source_label(self):
 
     def source_description(self):
         channel = self.source_channel()
```
mbertrand, line 45: Outside of this scope's PR changes, but should this function also take the `source_type` into consideration?

```python
    def source_channel(self):
        original_query_params = self.original_url_params()
        channels_filtered = Channel.objects.filter(search_filter=original_query_params)
        if channels_filtered.exists() and self.source_type ==  CHANNEL_SUBSCRIPTION_TYPE:
            return channels_filtered.first()
        return None
```

mbertrand (changes requested):
Works great!  Made a suggestion for improving the new test.   I thought of a few minor nitpicks that might be nice-to-haves but aren't critical and aren't mentioned in the issue:

- Prepopulate `display_label` for existing channel-based percolate queries (to `channel.title`)?
- When new percolate queries for channels are created, maybe assign the channel title as the initial display_label value
- Might be nice to add `display_label` to `PercolateQueryAdmin` so they can be seen/searched in django admin:

 
Screenshot 2024-09-26 at 10 38 40 AM


Also left some other inline code comments, but those are probably out of the scope of this PR.  Just the test change should be enough for approval.

learning_resources_search/models.py
shanbady: potentially. - im currently working on another PR that involves sending actual emails for percolated searches so this might change there

learning_resources_search/models_test.py
shanbady: fixed test

learning_resources_search/models.py
shanbady: working on another pr for sending the emails I can see about changing this there.

learning_resources_search/models_test.py
mbertrand: Delete this line, it is overriding the parameterized value:
```python
test_label = "new courses about cats"
```

Also:
```python
    original_query = {"department": ["physics"]} if is_channel_query else {
        "q": "testing search filter",
        "certification": None,
        "yearly_decay_percent": None,
    }
```

mbertrand (approved):
👍

PR #1613, opened 2024-09-26
[Nextjs] nextjs errors

abeglova:
closes https://github.com/mitodl/hq/issues/5479

This pr implements error catching and permission checking on the next js app

### How can this be tested?
Go to http://open.odl.local:8062/c/unit/not-a-unit. You should see a 404 Not Found Error 

As a not logged in user:
http://open.odl.local:8062/dashboard
http://open.odl.local:8062/dashboard/my-lists
http://open.odl.local:8062/dashboard/my-lists/id
http://open.odl.local:8062/dashboard/profile
http://open.odl.local:8062/dashboard/settings

Should redirect you to the ssl page

As a logged in admin
http://open.odl.local:8062/dashboard/my-lists/80000000 should show a 404 Not Found Error 

As a logged in non-admin
http://open.odl.local:8062/dashboard/my-lists/
http://open.odl.local:8062/dashboard/my-lists/1
Should display "Not Allowed"


frontends/main/src/components/ErrorBoundary/ErrorBoundary.tsx
```diff
@@ -0,0 +1,81 @@
...
+}
+
+interface ErrorBoundaryHandlerState {
+  hasError: boolean
+  error: unknown
+  previousPathname: string
+}
+const isForbiddenError = (error: unknown) => error instanceof ForbiddenError
+
+class ErrorBoundaryHandler extends Component<
```
jonkafton, line 24: I was interested to see a class component here and realize it's so we can use the error boundary lifecycle methods. It says here there's no equivalent for function components, https://legacy.reactjs.org/docs/hooks-faq.html#how-do-lifecycle-methods-correspond-to-hooks, though that doc is legacy... and these do appear in the up to date docs: https://react.dev/reference/react/Component#catching-rendering-errors-with-an-error-boundary (on the same page as a bunch of warnings "We recommend defining components as functions instead of classes")!

Looks like we can assume this is the recommended approach until they maybe make a hook available in future versions. The latest v19-beta does still recommend a single class component error boundary "if you’d like to avoid creating class components" - https://19.react.dev/reference/react/Component#componentdidcatch.

jonkafton (approved):
Looks and works great. Small request - could you remove the RestrictedRoute and ErrorPage components from the frontends/mit-learn workspace as we're using anything left in there to indicate that it still needs to be migrated.

PR #1620, opened 2024-09-27
Remove health checks against opensearch

rhysyngsun:

N/A

This remove the healthchecks because I've been hitting an exception locally that causes OpenSearch to be considered unhealthy which then causes docker compose to shutdown the app. There's an upstream bug in OpenSearch that's been open for 3 week but no known workaround.

### How can this be tested?

You should still be able to run `docker compose up` and nothing changes


mbertrand (approved):
👍 , though when I initially started it up I got lots of connection refused messages because all the daily ETL tasks started running and opensearch wasn't ready yet for the indexing requests.

PR #1621, opened 2024-09-27
Switch UI to use profile.name instead of first/last names

rhysyngsun:

Closes https://github.com/mitodl/hq/issues/5639

This switches the frontend over to use `profile.name`. To keep this backwards compatible before the field is available and populated in Keycloak I adjusted the API to return last+first name on the profile. This is temporary until we get users filling it out and to avoid a bunch of complicated logic in the frontend.

### How can this be tested?

Test that the application displays your full name under the following conditions:
- `user.profile.name` is set, but `user.first_name` and `user.last_name` is not
- `user.profile.name` is **not** set, but `user.first_name` and `user.last_name` is
- `user.profile.name`, `user.first_name`, and `user.last_name` are set


mbertrand (approved):
👍

PR #1631, opened 2024-10-01
fix Safari MIT logo bug

gumaerc:
Follow-up to https://github.com/mitodl/mit-open/pull/1622

This PR fixes a bug with Safari where the MIT logo in the header was not being scaled properly due to the width / height only being applied to the parent and not directly to the `img` element

image

### How can this be tested?
 - Visit the home page at http://localhost:8062 in Safari
 - Ensure that the header MIT logo looks correct

### DIFF
--- a/frontends/mit-learn/src/GlobalStyles.tsx
+++ b/frontends/mit-learn/src/GlobalStyles.tsx
@@ -20,8 +20,12 @@ const pageCss = css`
   }
 
   #app-container {
-    height: calc(100vh - 60px);
-    margin-top: 60px;
+    height: calc(100vh - 72px);
+    margin-top: 72px;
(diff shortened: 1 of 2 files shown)


shanbady (changes requested):
There are a few issues I am seeing. One that only happens in safari if you visit the units page:
Screenshot 2024-10-01 at 1 16 47 PM
The other issue (less severe) is that the text in the header scrunches at a certain size
Screenshot 2024-10-01 at 1 15 25 PM

The other thing I just noticed is that it appears the header doesn't reflect logged in state (seeing this in safari only). I logged in via django admin and go back to the home page and I am not shown as logged in

shanbady (approved):
Approving to get header changes out. will create tickets for other items noticed.

PR #1634, opened 2024-10-01
Adds base infra for the Unified Ecommerce frontend

jkachel:

Closes https://github.com/mitodl/hq/issues/5611

We're going to develop the frontend for Unified Ecommerce in Learn, and this adds some necessary support stuff to that:
- Adds an `EcommercePages` folder so we can store the page components for the UE frontend
- Adds an `EcommerceFeature` component so we can easily feature flag out the UE stuff (especially now, since we'll be doing things to begin with that don't have design yet)
- Adds a `CartPage` placeholder page, just so there's _something_ to go to
- Adds routing for `/cart/` to go to the UE code - this requires a logged-in session and an enabled feature flag to access

### How can this be tested?

Automated tests should pass.

To manually test, you'll need to make sure your instance is hitting PostHog - so it needs a valid set of keys. You can either hit the "MIT Learn Test" project that's in the OL PostHog account, or you can spin up your own PostHog account, or you can spin up your own instance of PostHog locally; any of these will work. Then, you'll need to go into Feature Flags and create a flag that just releases a boolean value called `enable-ecommerce`. 

Test by loading `<site root>/cart/`:

|    | Flag On | Flag Off |
|---|---|---|
| Authenticated | Placeholder page shown | ForbiddenError |
| Not Authenticated | Redirect to login | Redirect to login |

### Additional Context

This doesn't add the API client for UE - we'll do that in a separate PR because it'll probably require some tweaking in the UE code. So, we're just focusing on getting UE some space set up to move into.


frontends/mit-learn/src/pages/EcommercePages/CartPage.test.tsx
```diff
@@ -0,0 +1,65 @@
+import { renderTestApp, waitFor, setMockResponse } from "../../test-utils"
+import { urls } from "api/test-utils"
+import * as commonUrls from "@/common/urls"
+import { Permissions } from "@/common/permissions"
+import { login } from "@/common/urls"
+import { useFeatureFlagEnabled } from "posthog-js/react"
+
+jest.mock("posthog-js/react")
+const mockedUseFatureFlagEnabled = jest.mocked(useFeatureFlagEnabled)
```
ChristopherChudzicki, line 9:
```suggestion
const mockedUseFeatureFlagEnabled = jest.mocked(useFeatureFlagEnabled)
```

frontends/mit-learn/src/pages/EcommercePages/CartPage.tsx
ChristopherChudzicki, line 1: **Minor Suggestion:** I wouldn't bother with most of the styling here:
- A lot of it isn't doing anything, really. (e.g., the `align-self: stetch` and various flexboxes yield the same layout as default `display: block` here.
- Some of this, like making the paragraph have `component="h2"` is undesirable (not a header) and might be accidentally left in later.

Clearly this is all placeholder, so it doesn't really matter, but
```tsx
    <EcommerceFeature>
      <Container>
        <MetaTags title="Shopping Cart" />
        <Breadcrumbs
          variant="light"
          ancestors={[{ href: urls.HOME, label: "Home" }]}
          current="Shopping Cart"
        />
        <Typography component="h1" variant="h3">
          Shopping Cart
        </Typography>
        <Typography>
          The shopping cart layout should go here, if you're allowed to see
          this.
        </Typography>
      </Container>
    </EcommerceFeature>
```

looks fine, IMO

frontends/mit-learn/src/page-components/EcommerceFeature/EcommerceFeature.tsx
```diff
@@ -0,0 +1,31 @@
...
+ */
+
+const EcommerceFeature: React.FC<EcommerceFeatureProps> = ({ children }) => {
+  const ecommFlag = useFeatureFlagEnabled("enable-ecommerce")
+
+  if (ecommFlag === false) {
+    throw new ForbiddenError("Not enabled.")
+  }
+
+  return <>{ecommFlag ? children : null}</>
```
ChristopherChudzicki, line 28: No need for a fragment here.
```suggestion
  return ecommFlag ? children : null
```

frontends/mit-learn/src/page-components/EcommerceFeature/EcommerceFeature.tsx
```diff
@@ -0,0 +1,31 @@
...
+ *
+ * There's a PostHogFeature component that is provided but went this route
+ * because it seemed to be inconsistent - sometimes having the flag enabled
+ * resulted in it tossing to the error page.
+ *
+ * If the feature flag for this changes, this is where this needs to be set.
+ */
+
+const EcommerceFeature: React.FC<EcommerceFeatureProps> = ({ children }) => {
+  const ecommFlag = useFeatureFlagEnabled("enable-ecommerce")
```
ChristopherChudzicki, line 22: Might be worth putting these as constants in `@/common/feature_flags` or something.

(Or a TS enum, though they are occasionally annoying.)

ChristopherChudzicki (changes requested):

ChristopherChudzicki:
This looks good. Left some very minor requests. Mostly the typo + unnecessary fragment.

jkachel:
https://github.com/mitodl/mit-open/pull/1634/commits/3a02e22a0d59508ec04db775f8e87b1edb0b9157 should address the remaining stuff. Also - it's probably not a bad idea to make the `EcommerceFeature` component generic for any feature flag, so we can gate anything, but that's probably out of scope for this PR.

ChristopherChudzicki (approved):
👍 Also out of scope for this PR, but might want to make the posthog events an enum / constants, too.

frontends/mit-learn/src/page-components/EcommerceFeature/EcommerceFeature.tsx
```diff
@@ -0,0 +1,33 @@
...
+ * Set the feature flag here using the enum, and then make sure it's also
+ * defined in commmon/feature_flags too.
+ */
+
+const EcommerceFeature: React.FC<EcommerceFeatureProps> = ({ children }) => {
+  const ecommFlag = useFeatureFlagEnabled(FeatureFlags.EnableEcommerce)
+
+  if (ecommFlag === false) {
+    throw new ForbiddenError("Not enabled.")
+  }
```
ChristopherChudzicki, line 28: _Post-merge comment_

I noticed some odd behavior with PostHog and feature flags that I thought was worth mentioning. Essentially, I think there's no reliable way—with `useFeatureFlagEnabled`—to determine whether a flag is "off" or "not yet loaded". (_Note: the `useFeatureFlagEnabled` is really simple. Here's its [source](https://github.com/PostHog/posthog-js/blob/main/react/src/hooks/useFeatureFlagEnabled.ts)._)

`useFeatureFlagEnabled` returns `boolean | undefined`. This seems like the "normal" behavior:
1. `useFeatureFlagEnabled("flag-A")` returns undefined before any flags have loaded
2. When any flags have loaded, `useFeatureFlagEnabled("flag-A")` returns `true` if the flag is on, and `false` if the flag is off or has not loaded yet.
    - "When any flags have loaded" includes:
        - bootstrapped flags
        - flags determined via API call to posthog site

So the following is possible:

Screenshot 2024-10-04 at 12 30 41 PM

In this scenario, we were attempting to:
1. If flags not loaded yet, return null
2. If flags loaded, either
    - redirect
    - or render the cart page 

But this doesn't seem to work quite right if any flags are bootstrapped, as shown in the sreenshot.

PR #1635, opened 2024-10-01
raise SystemExit as a RetryError

mbertrand:
Related to https://github.com/mitodl/hq/issues/5615

Tries to avoid the following error which occurred on RC when restarting the celery dyno:

> 2024-10-01T18:11:19.451534+00:00 app[extra_worker_performance.1]: [2024-10-01 18:11:19,446: ERROR/ForkPoolWorker-144] Task learning_resources_search.tasks.index_learning_resources[f5205655-dcae-4cff-9317-51621c9ee3df] raised unexpected: EncodeError(TypeError('Object of type SystemExit is not JSON serializable'))

### How can this be tested?
Follow the same instructions as the previous PR: https://github.com/mitodl/mit-open/pull/1615

### DIFF
--- a/learning_resources_search/tasks.py
+++ b/learning_resources_search/tasks.py
@@ -283,7 +283,7 @@ def send_subscription_emails(self, subscription_type, period="daily"):
 @app.task(
     acks_late=True,
     reject_on_worker_lost=True,
-    autoretry_for=(RetryError, SystemExit),
+    autoretry_for=(RetryError,),
     retry_backoff=True,
     rate_limit="600/m",
 )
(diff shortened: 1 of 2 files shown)


shanbady (changes requested):
looks could. seems to work from my testing but I did notice that we dont have deindex_contentfiles handling acks_late, reject_on_worker_lost or catching the SystemExit. do we need to handle this same way?

mbertrand:
> looks could. seems to work from my testing but I did notice that we dont have deindex_contentfiles handling acks_late, reject_on_worker_lost or catching the SystemExit. do we need to handle this same way?

Maybe - `deindex_contentfiles` isn't run during a recreate_index task.  But if this finally works, then it might be good to handle more tasks the same way.

shanbady (approved):

PR #1644, opened 2024-10-02
Adds PostHog back into the NextJS build

jkachel:

Closes https://github.com/mitodl/hq/issues/5411

The PostHog integration was not implemented in the NextJS build of Learn. This adds it back:

- Specifying the PostHog project API key and root URL (this is actually technically new)
- The root-level PostHogProvider, initialized only if there's an API key set
- Bootstrapping the provider with feature flag settings via `.env` if we want to do that, and customizing the prefix for those flags
- Capturing both automatic PostHog events and `lrd_view` events

Any future work that gets pulled in and uses the PostHog hook or components should now "just work" too. (Unless it's like the `lrd_view` code, which checks to see if PostHog is there or not.)

Components when not enabled:
ph-nextjs-disabled

Components and settings when enabled (this is with a test flag added too):
ph-nextjs-enabled

### How can this be tested?

Automated tests should pass. This includes a couple for LearningResourceDrawer that were marked as `skip` because there wasn't PostHog support yet.

Set the required environment variables:
- `NEXT_PUBLIC_POSTHOG_PROJECT_API_KEY` - the PostHog project API key

Set the optional environment variables: 
- `NEXT_PUBLIC_POSTHOG_API_HOST` - the PostHog API host; defaults to `https://us.i.posthog.com`. Don't change this unless you [have your own PostHog instance running](https://posthog.com/docs/self-host).
- `NEXT_PUBLIC_POSTHOG_FEATURE_PREFIX` - the prefix to use for feature flags, defaults to `FEATURE_`
- `NEXT_PUBLIC_FEATURE_some_flag_name` - customize the lower case bit, and change `FEATURE_` if you set the prefix above; set this to something you'll recognize

With _at least_ the required variable set, you should see the PostHogProvider show up in the React components list as in the screenshots. Additionally, you should start to see events flowing into PostHog as you start doing things, including `lrd_view` events (which fire when you open a learning resource drawer). 

If you've bootstrapped some feature flags, you should see those get populated in the PostHogProvider options. We don't have any configured flags right now - you can manually test this by adding something that's flagged out to an existing page, but it's probably sufficient to see the options passed into the provider correctly. 

If you want to test the API host setting, you can maybe point it to something like [catch-webhooks](https://github.com/rangle/catch-webhooks) to make sure it's trying to hit the right place.

### DIFF
--- a/frontends/main/next.config.js
+++ b/frontends/main/next.config.js
@@ -3,6 +3,22 @@ const { validateEnv } = require("./validateEnv")
 
 validateEnv()
 
+const processFeatureFlags = () => {
+  const featureFlagPrefix =
+    process.env.NEXT_PUBLIC_POSTHOG_FEATURE_PREFIX || "FEATURE_"
+  const bootstrapFeatureFlags = {}
+
+  for (const [key, value] of Object.entries(process.env)) {
+    if (key.startsWith(`NEXT_PUBLIC_${featureFlagPrefix}`)) {
+      bootstrapFeatureFlags[
+        key.replace(`NEXT_PUBLIC_${featureFlagPrefix}`, "")
+      ] = value === "True" ? true : JSON.stringify(value)
+    }
+  }
+
+  return { FEATURE_FLAGS: JSON.stringify(bootstrapFeatureFlags) }
+}
+
 /** @type {import('next').NextConfig} */
 const nextConfig = {
   webpack: (config, { webpack }) => {
@@ -13,6 +29,7 @@ const nextConfig = {
       new webpack.IgnorePlugin({
         resourceRegExp: /mockAxios\.ts/,
       }),
+      new webpack.EnvironmentPlugin(processFeatureFlags()),
     )
 
     // Do not do this. Added to fix "import type", but causes a strage issue where
--- a/frontends/main/next.config.js
+++ b/frontends/main/next.config.js
@@ -16,7 +16,7 @@ const processFeatureFlags = () => {
     }
   }
 
-  return { FEATURE_FLAGS: JSON.stringify(bootstrapFeatureFlags) }
+  return bootstrapFeatureFlags
 }
 
 /** @type {import('next').NextConfig} */
@@ -29,7 +29,6 @@ const nextConfig = {
       new webpack.IgnorePlugin({
         resourceRegExp: /mockAxios\.ts/,
       }),
-      new webpack.EnvironmentPlugin(processFeatureFlags()),
     )
 
     // Do not do this. Added to fix "import type", but causes a strage issue where
@@ -73,6 +72,10 @@ const nextConfig = {
         pathname: "**",
       },
     ],
+  },
+
+  env: {
+    FEATURE_FLAGS: JSON.stringify(processFeatureFlags()),
   },
 }
 
--- a/frontends/main/src/app/providers.tsx
+++ b/frontends/main/src/app/providers.tsx
@@ -5,38 +5,20 @@ import { getQueryClient } from "./getQueryClient"
 import { QueryClientProvider } from "@tanstack/react-query"
 import { ThemeProvider, NextJsAppRouterCacheProvider } from "ol-components"
 import { Provider as NiceModalProvider } from "@ebay/nice-modal-react"
-import { PostHogProvider } from "posthog-js/react"
+import ConfiguredPostHogProvider from "@/components/ConfiguredPostHogProvider/ConfiguredPostHogProvider"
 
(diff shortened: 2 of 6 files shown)


jkachel:
pre-commit.ci run

frontends/main/next.config.js
```diff
@@ -13,6 +29,7 @@ const nextConfig = {
       new webpack.IgnorePlugin({
         resourceRegExp: /mockAxios\.ts/,
       }),
+      new webpack.EnvironmentPlugin(processFeatureFlags()),
```
ChristopherChudzicki, line 32: I think we should avoid adding to the webpack config unless we really need to, and NextJS has an `env` key for this... Could  `processFeatureFlags ` return the bootstrapFeatureFlags, and then do:
```js
const nextConfig = {
  webpack: { ... } // leave as is... I'm pretty sure we can get rid of it entirely, but no need to futz with it here.
  env: {
    FEATURE_FLAGS: JSON.stringify(bootstrapFeatureFlags)
  }
  ...
}
```

frontends/main/src/app/providers.tsx
```diff
@@ -18,4 +31,12 @@ export default function Providers({ children }: { children: React.ReactNode }) {
...
     </QueryClientProvider>
   )
+
+  return apiKey.length > 0 ? (
+    <PostHogProvider apiKey={apiKey} options={postHogOptions}>
+      {interiorElements}
+    </PostHogProvider>
+  ) : (
+    interiorElements
+  )
```
ChristopherChudzicki, line 41: What would you think of doing something like
```tsx
const ConfiguredPHProvider: React.FC<{ children: React.ReactNode }> = ({
  children,
}) => {
  const apiKey = process.env.NEXT_PUBLIC_POSTHOG_PROJECT_API_KEY || ""

  const apiHost =
    process.env.NEXT_PUBLIC_POSTHOG_API_HOST || "https://us.i.posthog.com"
  const featureFlags = JSON.parse(process.env.FEATURE_FLAGS || "")

  const postHogOptions = {
    api_host: apiHost,
    bootstrap: {
      featureFlags: featureFlags,
    },
  }
  if (!apiKey) return children
  return (
    <PostHogProvider apiKey={apiKey} options={postHogOptions}>
      {children}
    </PostHogProvider>
  )
}
```
?

Then `ConfiguredPHProvider` can handle the conditional itself + configuration.

Partly suggesting this to encapsulate it, but partly because we could run into the same situation in future with other providers, and as-structured I think it would get awkward with more than 1.

frontends/main/src/page-components/LearningResourceDrawer/LearningResourceDrawer.tsx
```diff
@@ -23,38 +23,35 @@ import {
...
-  //   data?.id,
-  //   data?.readable_id,
-  //   data?.platform?.code,
-  //   data?.resource_type,
-  //   POSTHOG?.api_key,
-  // ])
+  const apiKey = process.env.NEXT_PUBLIC_POSTHOG_PROJECT_API_KEY
+
+  useEffect(() => {
+    if (!apiKey || apiKey.length < 1) return
```
ChristopherChudzicki, line 36: `""` is falsy, so you can just check `!apiKey`

frontends/main/validateEnv.js
```diff
@@ -21,6 +21,11 @@ const schema = yup.object().shape({
     .string()
     .oneOf(["true", "false"]),
   NEXT_PUBLIC_CSRF_COOKIE_NAME: yup.string().required(),
+  NEXT_PUBLIC_POSTHOG_PROJECT_ID: yup.string(),
+  NEXT_PUBLIC_POSTHOG_PERSONAL_API_KEY: yup.string(),
```
ChristopherChudzicki, line 25: Is `NEXT_PUBLIC_POSTHOG_PERSONAL_API_KEY` used for anything?

ChristopherChudzicki (changes requested):
Works great. Two small requests.

frontends/main/validateEnv.js
jkachel: No, and actually that shouldn't be in the build since it's the actually secret one. I pulled it (and the project ID; we don't use that either, PostHog figures it out just from the project key).

jkachel:
https://github.com/mitodl/mit-open/pull/1644/commits/90ad4861e294b221cc3fdf990f03e99ef4be773d addresses the comments. This does pull the PostHog provider stuff into its own component, which makes the `providers.tsx` a good bit  cleaner.

ChristopherChudzicki (approved):
👍

PR #1646, opened 2024-10-02
topic detail banner / subtopic logic revisions

gumaerc:
Closes https://github.com/mitodl/hq/issues/5361

This PR moves some things around relating to topic channel pages:
 - Subtopics moved from channel page body to the banner
 - New `ChannelPageTemplate` component created to give topic channels their own template
 - Clicking on Subtopics to load their channel pages now shows "Related Topics" aka their siblings as Subtopics of the parent topic
 - Action buttons moved below the channel title
 - Various style changes

image
image

### How can this be tested?
 - Ensure that you have enough data backpopulated to have some topics and courses in those topics
 - Visit http://localhost:8062/topics
 - Click one of the top level topics that contain subtopics to visit its channel page
 - Verify that the subtopics display in the manner described above and on the issue at desktop and mobile resolutions
 - Click one of the subtopic chips
 - Ensure that the "Subtopics" title above the chips now says "Related Topics"
 - Verify that you can click the 3rd tier breadcrumb and go back to the parent topic


frontends/mit-learn/src/pages/ChannelPage/TopicChannelTemplate.tsx
```diff
@@ -0,0 +1,285 @@
...
+  },
+}))
+
+const ChannelControlsContainer = styled.div(({ theme }) => ({
+  display: "flex",
+  flexDirection: "row",
+  alignItems: "end",
+  flexGrow: 0,
+  flexShrink: 0,
+  order: 2,
```
ChristopherChudzicki, line 42: `order` is usually something we should avoid, since it's visual only and doesn't affect keyboard nav. And here I don't think it does anything:

Screenshot 2024-10-08 at 10 19 17 AM

gumaerc: This got copied over from `DefaultChannelTemplate`: https://github.com/mitodl/mit-open/blob/d9ed99aa0f15f528f2ddbf9a0f963cdccfafd9a6/frontends/mit-learn/src/pages/ChannelPage/DefaultChannelTemplate.tsx#L28

My best guess is that this was done to try and control the position of the channel controls container. I agree that it doesn't do anything and have removed it from both places.

frontends/ol-components/src/components/Banner/Banner.tsx
```diff
@@ -30,11 +25,13 @@ const BannerBackground = styled.div<BannerBackgroundProps>(
     backgroundImage: backgroundDim
       ? `linear-gradient(rgba(0 0 0 / ${backgroundDim}%), rgba(0 0 0 / ${backgroundDim}%)), url('${backgroundUrl}')`
       : `url(${backgroundUrl})`,
-    backgroundSize: "cover",
```
shanbady, line 33: I think the removal of the background size cover is causing the breakage at the particular width that chris had pointed out.

shanbady (dismissed):
One thing I noticed is that the currently active breadcrumb doesnt appear to dim on this branch (only seems to be happening with topics afaik).

production:
Screenshot 2024-10-08 at 10 36 03 AM
branch:
Screenshot 2024-10-08 at 10 36 23 AM

gumaerc:
> One thing I noticed is that the currently active breadcrumb doesnt appear to dim on this branch (only seems to be happening with topics afaik).
> 
> production: Screenshot 2024-10-08 at 10 36 03 AM branch: Screenshot 2024-10-08 at 10 36 23 AM

@shanbady This is because you are viewing a subtopic, and as part of this PR the subtopic shows the parent topic breadcrumbs. The last one is clickable to return to the parent topic.

frontends/ol-components/src/components/Banner/Banner.tsx
gumaerc: Yea, @mbilalmughal and I had a back and forth about the background. The 140% size rule was for ultrawide screens. I moved the rule to only apply to `lg` and above, and for everything else reset it to default with the position centered and top aligned.

gumaerc:
@ChristopherChudzicki @shanbady This is ready for another look whenever y'all have a moment

frontends/mit-learn/src/pages/ChannelPage/ChannelPage.test.tsx
```diff
@@ -262,27 +312,59 @@ describe.each(NON_UNIT_CHANNEL_TYPES)(
...
+    const { channel } = setupApis({
+      search_filter: "topic=Physics",
+      channel_type: ChannelTypeEnum.Topic,
+    })
+    const { subTopics, subTopicChannels } = setupTopicApis(channel)
+    invariant(subTopicChannels)
+    const subTopicChannel = subTopicChannels[0]
+    const filteredSubTopics = subTopics?.results.filter(
+      (t) =>
+        t.name.replace(/\s/g, "-") !== subTopicChannel.name.replace(/\s/g, "-"),
```
ChristopherChudzicki, line 352: Could be `t.name !== subtopicChannel.title` ... topic name and channel title are equivalent. (Topic model should use title, I guess. Oh well).

gumaerc:
> It seems to me that whether the appropriate value is 140%, 120%, 180%, etc, is going to vary image by image. cover seems like a reasonable universal default, but a percentage doesn't to me.
> 
> I'd suggest if we really want this to be 140%, it should be set only on the topic page, and not site-wide.
> 
> Banner has a backgroundSize prop, though it seems currently unused. Maybe that would be a good place for this?
> 

@mbilalmughal Wanted the 140% size rule not just for topics but for any banner that falls back to using the default "steps" image. From what I understand, we changed the image itself to display better on very large screens. The `cover` rule stretches the image horizontally too much and you lose all the detail in the steps.

gumaerc:
@ChristopherChudzicki This is ready for another look, but read the comment above about the background image size rules.

ChristopherChudzicki (approved):
👍

PR #1652, opened 2024-10-04
instructor search mode

abeglova:
closes https://github.com/mitodl/hq/issues/5696

This pr updates the search so that instructor names use "best_fields" instead of "phrase" just for instructor fields. It also removes instructor.first_name as a field because it is very likely to result in incorrect matches since some first names are very common

### How can this be tested?
Run 
manage.py backpopulate_oll_data
manage.py backpopulate_mitxonline_data

Search for "Dr. Jennifer French Kamrin"

You should see the OLL version of "Introduction to Probability and Statistics" where the instructor is listed as "Jennifer French Kamrin"

You should see mitxonline "Introduction to Differential Equations" where the instructor is listed as "Jennifer French"

There are a few extra results with other Jennifers or Drs as instructors, but they are below the relevant results.

### DIFF
--- a/learning_resources_search/api.py
+++ b/learning_resources_search/api.py
@@ -289,7 +289,7 @@ def generate_learning_resources_text_clause(text, search_mode, slop):
                                     query_type: {
                                         "query": text,
                                         "fields": RUN_INSTRUCTORS_QUERY_FIELDS,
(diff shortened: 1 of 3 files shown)


shanbady (approved):
Looks good 👍

PR #1653, opened 2024-10-04
Shanbady/randomize featured resources

shanbady:

Closes https://github.com/mitodl/hq/issues/5502

This PR moves the featured resources randomization for the featured resources carousel to the frontend (since the backend results are now cached).

### How can this be tested?
1. Checkout this branch
2. make sure you have resources loaded up. Also run `python manage.py populate_featured_lists` to have featured resources
3. visit the homepage and unit pages and note that the featured resources change on each load

### DIFF
--- a/frontends/api/src/hooks/learningResources/keyFactory.ts
+++ b/frontends/api/src/hooks/learningResources/keyFactory.ts
@@ -30,7 +30,17 @@ import type {
   UserListRelationship,
   MicroUserListRelationship,
 } from "../../generated/v1"
+
 import { createQueryKeys } from "@lukemorales/query-key-factory"
+
+const shuffle = ([...arr]) => {
+  let m = arr.length
+  while (m) {
+    const i = Math.floor(Math.random() * m--)
+    ;[arr[m], arr[i]] = [arr[i], arr[m]]
+  }
+  return arr
+}
 
 const learningResources = createQueryKeys("learningResources", {
   detail: (id: number) => ({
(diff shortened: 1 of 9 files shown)


frontends/api/src/hooks/learningResources/keyFactory.ts
```diff
@@ -30,8 +30,18 @@ import type {
...
 
+const shuffle = ([...arr]) => {
+  let m = arr.length
+  while (m) {
+    const i = Math.floor(Math.random() * m--)
+    ;[arr[m], arr[i]] = [arr[i], arr[m]]
+  }
+  return arr
+}
+
```
mbertrand, line 44: The backend randomizer used to randomize results within each learning path position index, so for example, assuming 5 "featured resource" learning paths:

Carousel positions 1-5 would be the 1st course from each of the 5 paths, in random order
Carousel positions 6-10 would be the 2nd course from each of the 5 paths, in random order
etc.

This ensured that all units would be equally represented, with the top courses appearing first.  

This PR seems to make the order completely random, for instance the first time I tried it I got:
1. Position #3 course from MITx
2. Position #1 course from Professional Education
3. Position #1 course from OCW
4. Position #2 course from Sloan
5. Position #3 course from Sloan
etc

Have the specs changed so that the order can be completely random now, or should the old pseudo-random sorting still be in place?

shanbady: I just asked Ferdi - he said try to preserve the old randomization if its easy enough so I'm going to refactor this to match what it was previously there.

mbertrand (approved):
Works great! 👍 

PS just a reminder, I think Chris mentioned this morning that any frontend changes to main should also be applied to the nextjs branch.

PR #1654, opened 2024-10-04
Ignore NotFoundErrors in switch_indices function

mbertrand:
Closes https://github.com/mitodl/hq/issues/5710

Ignores NotFoundErrors when switching indices in the finish_recreate_index task

### How can this be tested?
- Tests should pass.


abeglova (approved):
lgtm

PR #1657, opened 2024-10-07
Merge `main` into nextjs branch

ChristopherChudzicki:
For https://github.com/mitodl/hq/issues/5395

Merges `main` into `nextjs` and fixes a few issues that arose.

### Reviewing:
**Suggestion:** Most of the commits on this branch are already on `main`. The merge conflicts were resolved in https://github.com/mitodl/mit-open/pull/1655/commits/c8696de47e51e99dc7f458fbcd0ee4740db3698b.

Some things were broken after the conflict resolutions, so I fixed them in the followup commit.

Most of the changes I made in the followup commit are related to making sure the header and footer look **identical** on  https://learn.mit.edu/ vs the nextjs branch. See video below, which switches between two browser tabs.

https://github.com/user-attachments/assets/55c09ced-6d44-4802-900c-1a522ff3d04c

### How can this be tested?
Run locally. You should get recent changes from `main`. Notably:

- header should look like learn.mit.edu
- Ecommerce page at `/cart`.
    - This is likely to be removed from MIT Learn and developed in a separate project
    - Accessing `/cart` requires setting a posthog flag. See https://github.com/mitodl/mit-open/pull/1634


frontends/main/src/components/MITLogoLink/MITLogoLink.tsx
```diff
@@ -1,34 +1,71 @@
...
-  </Link>
-)
+const StyledImage = styled(Image)({
+  /**
+   * Needs display: block because otherwise the image is inline, which complicates
+   * parent container height calculations.
+   *
+   * See https://stackoverflow.com/a/11126701/2747370
+   */
+  display: "block",
```
ChristopherChudzicki, line 21: On `main`, the parent link is `display: flex` which causes its children to be block. *This was the only link styling on `main`*.

On `nextjs`, that styling was removed. But images are inline elements by default, which messes up their positioning a little. So this makes them block.

frontends/ol-components/src/components/Button/Button.tsx
```diff
@@ -357,13 +362,15 @@ const actionStyles = (size: ButtonSize) => {
...
 const ActionButton = styled(
   React.forwardRef<HTMLButtonElement, ActionButtonProps>((props, ref) => (
-    <ButtonStyled ref={ref} type="button" {...props} />
+    <button ref={ref} type="button" {...props} />
   )),
-)(({ theme, size = defaultProps.size, responsive }) => {
+)((props) => {
+  const { size, responsive } = { ...props, ...defaultProps }
   return [
+    buttonStyles(props),
```
ChristopherChudzicki, line 370: Previous, `ActionButton` was:
```tsx
const ActionButton = styled(
  React.forwardRef<HTMLButtonElement, ActionButtonProps>((props, ref) => (
    <ButtonStyled ref={ref} type="button" {...props} />
  ))(...)

// and, essentially
const ActionButtonLink = ActionButton.withComponent(Link)
```

This caused ActionButtonLink to be missing the styles from ButtonStyled.

Moving the button styles to its own helper fixed that. Story added.

jonkafton (approved):
Looks good!

PR #1658, opened 2024-10-07
update og-image with new logo

ChristopherChudzicki:
Closes https://github.com/mitodl/hq/issues/5727

Updates our opengraph social preview image to use new logo.

Image should be 

image

### How can this be tested?
1. Spin up this branch. Once the page has rendered, run `$$('meta').forEach(e => console.log(e.outerHTML))` in the JS dev console and check that both `<meta property="og:image" ... />` and `<meta name="twitter:image:src ... />` have an absolute URL pointing to the image above.

You can't really check the social preview locally, unfortunately, since it relies on prerender.io to actually wiork. (For now.)

### Checklist:
- [ ] We'll need to clear the prerender.io cache after deploy


frontends/mit-learn/src/page-components/MetaTags/MetaTags.tsx
```diff
@@ -71,7 +71,7 @@ const canonicalPathname = (pathname: string) => {
 }
 
 const SITE_NAME = APP_SETTINGS.SITE_NAME
-const DEFAULT_OG_IMAGE = `${window.origin}/static/images/mit-and-logo.jpg`
```
ChristopherChudzicki, line 74: I've left the old image in place for now so social previews still work between release / prerender cache clear.. We should delete it in a followup PR. (If not sooner, it will be deleted when we switch to nextjs branch.)

shanbady (approved):
👍

PR #1660, opened 2024-10-07
Always delete past events during ETL, and filter them out from api results too just in case

mbertrand:
Closes https://github.com/mitodl/hq/issues/5730

- Always deletes an event source's expired events during ETL pipeline runs
- Filters out expired events from the news_events API feed (just in case the ETL pipeline doesn't run for awhile for whatever reason).

### How can this be tested?
- Run `./manage.py backpopulate news_events --pipelines mitpe_events_etl`
- It should run, and create a feed source for "MIT Professional Education Events" if you don't already have one, but (as of today), there should be no feed items associated with that source (all in the past):
  - http://api.open.odl.local:8063/admin/news_events/feedsource/
  - `http://api.open.odl.local:8063/admin/news_events/feeditem/?source__id__exact=<id of feed source>`
- In a django shell, create a new event feeditem with a past event date:
  ```python
  from news_events.models import *
  feed_item = FeedItem.objects.create(guid="fake_guid", source_id=<source id>, title="Expired!")
  event_detail = FeedEventDetail.objects.create(
      feed_item=feed_item, 
      event_type=["Webinar"], 
      audience=["everyone"], 
      location=["everywhere"], 
      event_datetime="2020-01-01T00:00:00Z"
  )
  ```
- Go to http://api.open.odl.local:8063/api/v0/news_events/?feed_type=events, you may see other events if you previously ingested them, but the one you created above should not be returned (it will be returned if you switch back to the main branch)
- Run `./manage.py backpopulate news_events --pipelines mitpe_events_etl` again.  The feeditem you created in the previous step should now be deleted.
  ```python
  FeedItem.objects.filter(guid="fake_guid").exists()
  > False
  ```

### DIFF
--- a/news_events/etl/conftest.py
+++ b/news_events/etl/conftest.py
@@ -65,7 +65,7 @@ def sources_data() -> SimpleNamespace:
                         }
                         if feed_type == FeedType.news.name
                         else {
-                            "event_datetime": "2024-03-15T13:42:36Z",
+                            "event_datetime": "2124-03-15T13:42:36Z",
                             **event_details,
                         },
                     },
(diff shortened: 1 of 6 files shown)


abeglova (approved):
lgtm

PR #1667, opened 2024-10-08
content file weight

abeglova:
closes https://github.com/mitodl/hq/issues/5721

This adds a slider to the the admin search section to control whether to include content file data in the search results for searches with a text query

### How can this be tested?
run ./manage.py backpopulate_ocw_data --course-name 6-006-introduction-to-algorithms-spring-2020

http://open.odl.local:8062/search/?q=dirt should return that course 

http://open.odl.local:8062/search/?q=dirt&content_file_score_weight=0  should not return it

Verify that as an admin you see "Content File Score Weight Adjustment" in the admin facet sidebar and can set it

Verify that search works normally otherwise

### DIFF
--- a/frontends/api/src/generated/v1/api.ts
+++ b/frontends/api/src/generated/v1/api.ts
@@ -3863,6 +3863,12 @@ export interface PercolateQuerySubscriptionRequestRequest {
    */
   max_incompleteness_penalty?: number | null
   /**
+   * Score weight for content file data.  1 is the default. 0 means content files are ignored
+   * @type {number}
+   * @memberof PercolateQuerySubscriptionRequestRequest
+   */
+  content_file_score_weight?: number | null
+  /**
    *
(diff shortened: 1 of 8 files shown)


mbertrand (approved):
LGTM.  I had to lower "Minimum Score Cutoff" to 4 before that course would show up at all, and then lowering "Content File Score Weight Adjustment" to 0.9 or less removed it.
If I set "Minimum Score Cutoff" to 0, then the course was returned even if "Content File Score Weight Adjustment" was also set to 0 - I think that's okay though?

abeglova:
Yeah you are right - the query as it is written now does not filter out matches from content files completely when content-file-weight is zero, just gives them low scores so without a score minimum they are still shown. Having to lower "Minimum Score Cutoff", makes sense. It must have been just over the threshold in my local index and exact scores vary depending on what other resources you have locally.

PR #1668, opened 2024-10-09
Clear Cache on Deploy

shanbady:

Closes https://github.com/mitodl/hq/issues/5743

This pr adds a django command that is run on each deploy to clear the cache.

The site-wide page cache gets invalidated when etl pipelines run and when search indexing operations complete however - we recently ran into an issue where an api endpoint added a new attribute and due to caching, the api responded with a cached response and a frontend feature that relied on this broke. 

### How can this be tested?
To test locally:
1. checkout this branch
2. browse around some pages in order to have some items in the cache
3. run the cache clear management command `python manage.py clear_cache` - you should see some output indicating the number of items cleared.

### DIFF
--- a/main/management/commands/clear_cache.py
+++ b/main/management/commands/clear_cache.py
@@ -1,0 +1,15 @@
+"""Command to generate the channel avatar SVGs"""
+
+from django.core.management.base import BaseCommand
+
+from main.utils import clear_search_cache
+
+
+class Command(BaseCommand):
+    """Command to generate the channel avatar SVGs"""
+
+    help = "Command to clear the cache"
+
+    def handle(self, *args, **options):  # noqa: ARG002
+        cache_items = clear_search_cache()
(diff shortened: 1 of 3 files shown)


gumaerc (approved):
👍 LGTM

The only thing I might suggest here; if you expect this cache clearing function to take a long time in the event that there are a lot of objects to clear, it might be worth putting this into a celery task. Otherwise, it can stay as-is. We'll have to see how this works on RC to test the Heroku component.

shanbady:
> 👍 LGTM
> 
> The only thing I might suggest here; if you expect this cache clearing function to take a long time in the event that there are a lot of objects to clear, it might be worth putting this into a celery task. Otherwise, it can stay as-is. We'll have to see how this works on RC to test the Heroku component.

I was thinking something like that as well if it takes long. Either celery task (easier) or if we are only going to be dealing with redis manually calling unlink instead of delete

PR #1673, opened 2024-10-09
prevent featured course carousel from re-randomizing

gumaerc:
Closes https://github.com/mitodl/hq/issues/5434

This PR fixes an issue where the featured course carousel on the home page would re-render (and re-randomize) if you save one of the courses to a learning path / user list.

### How can this be tested?
 - Ensure that you have sufficient data backpopulated to have courses show up in the featured course carousel on the homepage, and that you are logged in as a staff / superuser
 - Visit the home page at http://localhost:8062/
 - Add one of the courses to a user list by clicking the bookmark icon, selecting a list and then clicking Save
 - Verify that the bookmark icon is now highlighted red and the carousel does not refresh / re-randomize
 - Repeat the above, but use the list icon to add / remove a course from a learning path

### DIFF
--- a/frontends/api/src/hooks/learningResources/index.test.ts
+++ b/frontends/api/src/hooks/learningResources/index.test.ts
@@ -211,7 +211,7 @@ describe("LearningPath CRUD", () => {
   }
 
   test("useLearningpathCreate calls correct API", async () => {
-    const { path, pathUrls, keys } = makeData()
+    const { path, pathUrls } = makeData()
     const url = pathUrls.list
 
     const requestData = { title: path.title }
@@ -226,9 +226,12 @@ describe("LearningPath CRUD", () => {
 
     await waitFor(() => expect(result.current.isSuccess).toBe(true))
     expect(makeRequest).toHaveBeenCalledWith("post", url, requestData)
-    expect(queryClient.invalidateQueries).toHaveBeenCalledWith(
-      keys.learningResources,
-    )
+    expect(queryClient.invalidateQueries).toHaveBeenCalledWith([
+      "learningResources",
+      "learningpaths",
+      "learning_paths",
+      "list",
+    ])
   })
 
   test("useLearningpathDestroy calls correct API", async () => {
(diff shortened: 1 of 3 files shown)


shanbady (changes requested):
re-randomizing on saving an item appears to be resolved with this but im seeing another issue - the carousel appears to still be randomizing when creating a new list instead of selecting an existing one. the first time it randomizes and if you try to create another new list a second time, an error is thrown. 
Screenshot 2024-10-09 at 3 46 14 PM
I'm seeing the error issue on main so if its not simple it may be worth a separate ticket

gumaerc:
@shanbady Thanks for your review. I wasn't able to reproduce the error upon creating multiple lists, but I was able to reproduce randomization after creating a new list.

gumaerc:
https://github.com/user-attachments/assets/23e1815b-b12c-4c86-9e0b-7b026d40e6eb

@shanbady I was unable to reproduce this again today. I've attached a screen recording above, but I tried what you described many times and was not able to reproduce it

shanbady (approved):
re-randomization issue looks to be resolved. we are handling the learning paths exception separately.

gumaerc:
@shanbady Something else @ChristopherChudzicki pointed out is that the re-randomization of the featured list happens when you create a learning path from the dialog because of this code:

```
type LearningPathCreateRequest = Omit<
  LPCreateRequest["LearningPathResourceRequest"],
  "readable_id" | "resource_type"
>
const useLearningpathCreate = () => {
  const queryClient = useQueryClient()
  return useMutation({
    mutationFn: (params: LearningPathCreateRequest) =>
      learningpathsApi.learningpathsCreate({
        LearningPathResourceRequest: params,
      }),
    onSettled: () => {
      // Invalidate everything: this is over-aggressive, but the new resource
      // could appear in most lists
      queryClient.invalidateQueries(learningResources._def)
    },
  })
}
```

In the `onSettled` callback here, we are invalidating all learning resource queries for no reason, thus causing the carousel to re-render. As part of this PR, I will fix this because it is related.

The error message you pointed out is separate from both the original issue and the one I pointed out above, and I'm creating a separate issue for that. It's caused by selecting topics that don't have a `channel_url` while creating the learning path.

PR #1671, opened 2024-10-09
Set metadataBase url

ChristopherChudzicki:
### What are the relevant tickets?
Closes https://github.com/mitodl/hq/issues/5754

### Description (What does it do?)
Sets [metadataBase](https://nextjs.org/docs/app/api-reference/functions/generate-metadata#metadatabase)

### How can this be tested?
1. On this branch, run `$$('meta').forEach(e => console.log(e.outerHTML))` in the dev console and check that `og:image` is an absolute URL
    - check og:image on homepage
    - on channel pages, where it is determined via an api call

PR #1677, opened 2024-10-10
Make metadataBase apply to all pages, not just homepage

ChristopherChudzicki:
Followup to #1671 

This PR moves `metadadataBase` from root page to root layout.

Child pages load all their parent layouts (but not parent pages!) so the earlier positioning of metadataBase was only affecting the home page.

### How can this be tested?
1. Despite what the docs say [here](https://nextjs.org/docs/app/api-reference/functions/generate-metadata#default-value), it seems like metadataBase does not work in local dev mode. So set `NODE_ENV=production` and `NEXT_PUBLIC_ORIGIN=https://learn.mit.edu`, then run `yarn workspace main build` `yarn workspace main start`
    - you should NOT see any warnings like 
     
    > ⚠ metadataBase property in metadata export is not set for resolving social open graph or twitter images, using "http://localhost:8062". See https://nextjs.org/docs/app/api-reference/functions/generate-metadata#metadatabase 
2. Visit your local production server and run `$$('meta').forEach(e => console.log(e.outerHTML))` in the dev console
    - all pages, including homepage, should have `og:image` urls beginning with learn.mit.edu

### DIFF
--- a/frontends/main/src/app/c/[channelType]/[name]/page.tsx
+++ b/frontends/main/src/app/c/[channelType]/[name]/page.tsx
@@ -25,7 +25,6 @@ export async function generateMetadata({
     searchParams,
     title: `${channelDetails.title}`,
     description: channelDetails.public_description,
-    image: channelDetails.configuration.logo,
   })
 }
 
--- a/frontends/main/src/app/layout.tsx
+++ b/frontends/main/src/app/layout.tsx
@@ -1,5 +1,3 @@
-"use client"
-
 import React from "react"
 import Header from "@/page-components/Header/Header"
 import Footer from "@/page-components/Footer/Footer"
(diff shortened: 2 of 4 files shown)


frontends/ol-components/src/index.ts
```diff
@@ -1,3 +1,4 @@
+"use client"
```
ChristopherChudzicki, line 1: These are all client components for now because they use runtime CSS-in-JS (Emotion) which depends on react useContext.

Marking them as use-client allows importing them into server components.

frontends/main/src/app/layout.tsx
```diff
@@ -1,5 +1,3 @@
-"use client"
```
ChristopherChudzicki, line 1: `layout.tsx` needs to be a server component in order to define metadata

frontends/main/src/app/c/[channelType]/[name]/page.tsx
```diff
@@ -25,7 +25,6 @@ export async function generateMetadata({
     searchParams,
     title: `${channelDetails.title}`,
     description: channelDetails.public_description,
-    image: channelDetails.configuration.logo,
```
Comment (ChristopherChudzicki) [inline frontends/main/src/app/c/[channelType]/[name]/page.tsx:28]: This is unrelated to the other changes, but...
- we noticed the metadataBase issue because channel og:image were not working
- `og:image` cannot be an SVG, so without this change, they still won't work.

jonkafton (approved):
Works!.. in prod mode `yarn build; yarn start;`

Interesting and odd that it does not work in dev mode. You mentioned that's apparently the case on Slack, but didn't see that info while scanning the [docs](https://nextjs.org/docs/app/api-reference/functions/generate-metadata#metadatabase).

PR #1679, opened 2024-10-10
Updates when search events are triggered

jkachel:

Closes https://github.com/mitodl/hq/issues/5763

We added some triggering for search events to the app in https://github.com/mitodl/mit-open/pull/1596 . This allowed us to capture search term changes and facet selections on the search results page. However, this wasn't a complete picture of search events:
- Searches started on the homepage didn't trigger an event
- Facet changes on interior pages _besides_ the search results page (such as on a unit page) didn't trigger an event.
- Clearing selections did not trigger an event.

These changes fix that.

- Updates the homepage HeroSearch component to use the SearchField component, rather than SearchInput, so it will trigger the search_update event
- Updates the SearchInput component to not require `setPage`, so it can be used in HeroSearch.
- Adds event triggering to the relevant events in SearchDisplay (by wrapping the event handlers that are passed in for that)
- Pulls related triggering from 

### How can this be tested?

You will need a PostHog account to test manually. For best results, go to the Activity page in the sidebar and then add an event filter for `search_update`, since that's the only thing that gets triggered by this code. (Also, it helps to be patient - it can take a minute or two for PostHog to process the events and display them. But not too patient. More than a couple of minutes and it _probably_ didn't work.) 

For homepage and search results:
1. Perform a search on the homepage. You should see a `search_update`, and it should capture if the event was triggered by hitting Enter or clicking on the submit icon. 
2. On the search results page, adjust things - update the facets, change the category, update sorting, change the search term. All of these should trigger `search_update` events. 
3. On the search results page, use the Clear All button to clear facet selections. This too should trigger a `search_update` event. 

For interior pages:
1. Navigate to a unit page. This should list the resources available in the unit.
2. Change things about the results - add a term, select filters, change category, etc. You should get `search_update` events for these. 
3. Similarly, using Clear All should also trigger an event.
4. Repeat this for other pages that use the ChannelPage base - department, topics, etc. - they should all work in the same manner.

### Additional Context

The issue being tracked in https://github.com/mitodl/hq/issues/5767 is in progress still (as of writing this), so if you select >1 facet type (e.g. topic, department, etc.) you'll only ever get the first one you select. (But this should make it slightly easier to refactor our code to manually capture the data if we have to.)

### DIFF
--- a/frontends/mit-learn/src/page-components/SearchDisplay/SearchDisplay.tsx
+++ b/frontends/mit-learn/src/page-components/SearchDisplay/SearchDisplay.tsx
@@ -51,6 +51,7 @@ import type { TabConfig } from "./ResourceCategoryTabs"
 import { ResourceCard } from "../ResourceCard/ResourceCard"
 import { useSearchParams } from "@mitodl/course-search-utils/react-router"
 import { useUserMe } from "api/hooks/user"
+import { usePostHog } from "posthog-js/react"
 
 const StyledResourceTabs = styled(ResourceCategoryTabs.TabList)`
   margin-top: 0 px;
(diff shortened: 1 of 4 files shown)


shanbady (changes requested):
Seems to work well!

the one thing I noticed is that didnt work as described was getting a "search update" event from the homepage. I did however get the pageview once i hit enter.

shanbady (approved):

PR #1681, opened 2024-10-10
search defaults fix

abeglova:
closes https://github.com/mitodl/hq/issues/5706

https://github.com/mitodl/mit-open/pull/1590 was supposed to make the search setting settable from the environment. This worked for the backend but not the frontend. This creates a small api page that shares the defaults with the frontend

### How can this be tested?
Go to http://open.odl.local:8062/search/ . Everything should work as expected as both an admin user and non-logged in user.

Log in as an admin. You should be able to manipulate the admin search settings. 

Set some of the setting from your .env variable. For example
DEFAULT_SEARCH_SLOP=2
DEFAULT_SEARCH_MODE=best_fields

Restart docker. The defaults in the admin facet should be set to the settings from your .env file


shanbady (approved):
Looks good. works as intended

PR #1684, opened 2024-10-11
Shanbady/consolidate static fixtures

shanbady:
Closes https://github.com/mitodl/hq/issues/4905
Closes https://github.com/mitodl/hq/issues/5145

This PR solves the above 2 tickets related to required static data not being loaded in on fresh environments.
It does the following:
1. restores the running of data migrations in local environments (which i think was accidentally removed at some point)
2. moves the static fixture files to the data_fixtures app
3. moves the utility methods for loading static fixtures to the data_fixtures app
5. adds a method in the very initial data_fixtures data migration that manually loads static fixtures in before anything else
6. fixed a few other places where migrations could fail when using get_or_create/update without specifying defaults
7. updated the codespace's devcontainer.json to backpopulate resource channels after loading everything up

### How can this be tested?
There are two ways to try this - 
- #### in codespaces
    - spin up a codespace on this PR - make sure you spin it up via "new with options" and specify "4 core" before launching 
    Screenshot 2024-10-11 at 10 26 07 AM
    Screenshot 2024-10-11 at 10 26 17 AM

    - once the codespace is up (make sure you make port 8603 public) you should see topics show up on the channel page and there should be search data
- #### locally
    - do a `docker compose down` and `docker compose up`
    - all migrations should go through without issues
    - you should also be able to run the presviously failing commands referenced [here](https://github.com/mitodl/hq/issues/5145) with no issue
    - upon running an etl load, the topics channel page should be automatically populated with topics (since we did the static load via migration)

### Additional Context
On existing live environments like production and rc, the addition of the static fixture load will be ignored (since those migrations have already run) - on new environments this conveniently preloads the required data for us.


.devcontainer/devcontainer.json
```diff
@@ -22,6 +22,6 @@
...
   },
   "features": {},
-  "postStartCommand": "while [ \"$(python manage.py showmigrations | grep \"\\[ \\]\" | wc -l)\" -ne \"0\" ]; do echo \"waiting for migrations\"; sleep 2; done && python manage.py update_offered_by && python manage.py update_platforms && python manage.py update_departments_schools && python manage.py backpopulate_resource_channels --overwrite --all && python manage.py update_course_number_departments && python manage.py backpopulate_mitxonline_data && python manage.py backpopulate_micromasters_data && python manage.py recreate_index --all",
+  "postStartCommand": "while [ \"$(python manage.py showmigrations | grep \"\\[ \\]\" | wc -l)\" -ne \"0\" ]; do echo \"waiting for migrations\"; sleep 2; done && python manage.py update_offered_by && python manage.py update_platforms && python manage.py update_departments_schools && python manage.py backpopulate_resource_channels --overwrite --all && python manage.py update_course_number_departments && python manage.py backpopulate_mitxonline_data && python manage.py backpopulate_micromasters_data && python manage.py backpopulate_resource_channels --overwrite --all && python manage.py recreate_index --all",
```
ChristopherChudzicki, line 25: Why do we do `python manage.py backpopulate_resource_channels --overwrite --all` twice?

Currently in my dev container, I am seeing many topic channels without resources. They should be unpublished.

IMO the best solution for this is probably to channel the model's default value of `published` from True to False. But I suspect it might also be fixed by removing the second instance `python manage.py backpopulate_resource_channels --overwrite --all`.

I think:
1. First, `python manage.py backpopulate_resource_channels --overwrite --all` runs and all the channels get created
2. The resource backpopulation runs, which triggers unpublishing of many channels (since most don't have resources associated, they should not be published).
3. `python manage.py backpopulate_resource_channels --overwrite --all` runs again, and the channels all get re-created from scratch, with published=True.
    - Again, I think the default should be `published=False`

ChristopherChudzicki (approved):
👍 

This worked very well locally. I noticed one small issue on the dev container (too many channels are published).

Approving, but I do think it's worth re-considering whether the `python manage.py backpopulate_resource_channels --overwrite --all` needs to run twice.

If it does, the best solution to the issue I noticed is probably to make `published=False` the default, which IMO we should handle in a separate PR.

scripts/run-django-dev.sh
```diff
@@ -3,9 +3,15 @@
...
+
+# run initial django migrations
 python3 manage.py migrate --noinput
+
+# populate cache table
 python3 manage.py createcachetable
-python3 manage.py migrate --noinput
+
+# run ONLY data migrations
+RUN_DATA_MIGRATIONS=true python3 manage.py migrate --noinput
```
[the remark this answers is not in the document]
shanbady: Its run this way since the data migrations are intended to run once all the actual schema migrations have run. There is no elegant way to add "latest schema migration x" as a dependency of the first data migration so instead we make sure regular schema migrations are run and then the data migrations go in afterwards.

.devcontainer/devcontainer.json
shanbady: removed the duplicate call to backpopulate.  This only affects the codespace environment at the moment. If we need to address the published=False as a default I'm thinking it might be better to do that in a separate ticket

PR #1689, opened 2024-10-15
Make add to list dialog scrollable

ChristopherChudzicki:
Closes https://github.com/mitodl/hq/issues/5774

**Change:** The AddToListDialog shows more lists and is scrollable (if necessary):

Screenshot 2024-10-15 at 11 06 53 AM

**Unchanged:** Most other dialogs

Screenshot 2024-10-15 at 11 07 03 AM
Screenshot 2024-10-15 at 11 07 15 AM
Screenshot 2024-10-15 at 11 07 26 AM
Screenshot 2024-10-15 at 11 07 40 AM

### How can this be tested?
1. Create a bunch of learning paths (or userlists) and check that the AddToList dialog content scrolls. (You can open the AddToListDialog via list/bookmark button on learning resources).
2. Check that other dialogs, like the list creation (in dashboard -> my lists) is unchanged

### DIFF
--- a/frontends/mit-learn/src/page-components/Dialogs/AddToListDialog.tsx
+++ b/frontends/mit-learn/src/page-components/Dialogs/AddToListDialog.tsx
@@ -14,11 +14,7 @@ import { usePostHog } from "posthog-js/react"
 
 import * as NiceModal from "@ebay/nice-modal-react"
 
-import {
-  type LearningPathResource,
-  type LearningResource,
-  type UserList,
-} from "api"
+import type { LearningPathResource, LearningResource, UserList } from "api"
 
 import {
   useLearningResourceSetUserListRelationships,
@@ -30,6 +26,8 @@ import {
 import { manageListDialogs } from "@/page-components/ManageListDialogs/ManageListDialogs"
 import { ListType } from "api/constants"
 import { useFormik } from "formik"
+
+const LIST_LIMIT = 100
 
 const ResourceTitle = styled.span({
   fontStyle: "italic",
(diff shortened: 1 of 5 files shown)


frontends/mit-learn/src/page-components/Dialogs/AddToListDialog.tsx
```diff
@@ -202,7 +200,7 @@ const AddToLearningPathDialogInner: React.FC<AddToListDialogProps> = ({
 }) => {
   const resourceQuery = useLearningResourcesDetail(resourceId)
   const resource = resourceQuery.data
-  const listsQuery = useLearningPathsList()
+  const listsQuery = useLearningPathsList({ limit: LIST_LIMIT })
```
ChristopherChudzicki, line 203: Showing up to 100 and scrolling isn't a great solution. We should have infinite scroll or something.

This is a remediation to unblock the DLLs.

frontends/ol-components/src/components/Dialog/Dialog.tsx
```diff
@@ -23,7 +23,9 @@ const Header = styled.div`
 `
 
 const Content = styled.div`
-  margin: 28px 28px 40px;
+  margin: 28px;
```
ChristopherChudzicki, line 26: Previously, in Figma, a couple dialogs had 40px between the content and the dialog footer but most had 28px.

In THE CODE it was determined via `Dialog` vs `FormDialog`. Both used 40px margin on the content, but FormDialog also had `-12px` on the footer for an effective `28px`. This is (a) weird, and (b) **the negative margin approach did not work well with scrolling overflow.** (Which is why it's showing up in this PR).


Confirmed with @steven-hatch we want `28px` uniformly across the dialogs.

Here's an example of a noticeable change:

Screenshot 2024-10-15 at 11 17 04 AM

shanbady (approved):
👍

PR #1692, opened 2024-10-15
increase learning path listing limit to 100

ChristopherChudzicki:
Followup to https://github.com/mitodl/hq/issues/5774 / https://github.com/mitodl/mit-open/pull/1689

In https://github.com/mitodl/mit-open/pull/1689, we made the AddToListDialog scrollable to unblock DLLs editing learningpaths on production—it now shows more than 10 learning paths (max 100).

This PR does the same remediation for the listing page, `/learningpaths`—increase limit to 100.

The corresponding user-facing public UI issue is tracked via https://github.com/mitodl/hq/issues/5776

### How can this be tested?
With 10+ learning paths (but less than 100...) check that they are all visible at `/learningpaths`


shanbady (approved):
👍

PR #1697, opened 2024-10-16
Ab/search synonyms

abeglova:
closes https://github.com/mitodl/hq/issues/5781

This adds a list of synonyms to the search analyzer.

### How can this be tested?
run ./manage.py recreate_index --all

Verify that 
http://open.odl.local:8062/search?q=ai and http://open.odl.local:8062/search?q=artificial+intelligence have the same results

### DIFF
--- a/learning_resources_search/constants.py
+++ b/learning_resources_search/constants.py
@@ -90,13 +90,13 @@ SEARCH_NESTED_FILTERS = {
 
 ENGLISH_TEXT_FIELD = {
     "type": "text",
-    "fields": {"english": {"type": "text", "analyzer": "english"}},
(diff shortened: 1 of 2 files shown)


shanbady (approved):
Functionality appears to work (tested in codespace with ai example). One non-blocking suggestion I have which might make this list easier to update (and potentially expand via some automated process in the future) is to move the synonyms to a text file and reference it via "synonyms_path".

abeglova:
I think while we have so few synonyms just keeping them in the code is easier. If we decide to expand the synonyms list substantially, we should do what you suggests and keep them in a text file.

PR #1699, opened 2024-10-16
Shanbady/improve bookmark button label

shanbady:

Closes https://github.com/mitodl/hq/issues/5545

This PR changes the aria-label for the add to user list button to "Bookmark (resource type)"

### How can this be tested?
1. checkout this branch. 
2. validate that the aria label for the add to user list (little bookmark icon) on resource cards says "Bookmark (resource type)"
  -  Screenshot 2024-10-16 at 1 35 48 PM
4. optionally use a screen reader and verify it captures the aria-label

### DIFF
--- a/frontends/mit-learn/src/page-components/ResourceCard/ResourceCard.test.tsx
+++ b/frontends/mit-learn/src/page-components/ResourceCard/ResourceCard.test.tsx
@@ -7,6 +7,7 @@ import {
   expectProps,
 } from "../../test-utils"
 import type { User } from "../../test-utils"
+import { getReadableResourceType } from "ol-utilities"
 import { ResourceCard } from "./ResourceCard"
 import {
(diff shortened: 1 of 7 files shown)


gumaerc (approved):
👍 LGTM

The only thing I might suggest here, even though it's not part of the issue, is to do the same thing with the Add to Learning Path buttons. The language for the User List button is "Bookmark <resource_type>" so maybe for learning paths we could do "Add <resource_type> to Learning Path?"

shanbady:
> 👍 LGTM
> 
> The only thing I might suggest here, even though it's not part of the issue, is to do the same thing with the Add to Learning Path buttons. The language for the User List button is "Bookmark <resource_type>" so maybe for learning paths we could do "Add <resource_type> to Learning Path?"

tagging @sovsey for thoughts on this

PR #1701, opened 2024-10-16
NextJS Sentry Integration

ChristopherChudzicki:
Closes https://github.com/mitodl/hq/issues/5771

### How can this be tested?

Trigger some errors in the frontend and check that (a) appropriate UI is shown, (b) errors show up in Sentry

1. Now, set env vars... You something specific to you for the sentry environment name
    ```sh
    # in env/frontend.local.env
    NEXT_PUBLIC_SENTRY_DSN="https://8131564a9b9e31ac3cbcf7952f0fea80@o48788.ingest.us.sentry.io/4506260054540288"
    # suggest using your initials
    NEXT_PUBLIC_SENTRY_ENV=abc-test
    ```
2. Visit https://mit-office-of-digital-learning.sentry.io/issues/?end=2024-10-16T16%3A09%3A48&environment=cc-test&project=4506260054540288&start=2024-10-16T16%3A08%3A47 **and change the env filter to your environment**
3. Restart the nextjs app (e.g., restart docker) and trigger some errors.
4. **Async Errors:** Visit /sentry-example-page; click `Throw async error!` 
     - The errors should show up in your sentry environment
5. **Sync Errors:** Visit /sentry-example-page; click `Render error!`
    - synchronous errors during rendering crash the app unless they are caught by an error boundary. We catch the error with `error.ts`. You should see a generic error page. 
     - The error should show up in your sentry environment
6. **Forbidden Errors:** Visit /sentry-example-page; click `Forbidden error!`
    - You should get the Not Allowed page
     - The error should show up in your sentry environment
7.  **Root Layout Error:** 
    - Production only: This uses `global-error.ts`, which only functions when running in production mode. So to test this, you need to run NextJS locally in production mode. This should be easier than it is, but for now:
    ```
    # If not using docker, skip to `NODE_ENV...` commands below.
    docker compose stop watch
    docker compose run --rm --service-ports watch bash

    # now within the watch container:
    NODE_ENV=production yarn workspace main build
    NODE_ENV=production yarn workspace main next start
    ```
    - (if you're not using docker, run the last two `NODE_ENV...` commands in your terminal)
    - Now that you're running in production mode, click "Root Layout Error (production only)
            </button>" in the header. You should see an error page, but no header/footer. (The error occurred in the header. We can't show it.)

### Before merging
- [ ] Remove the testing page and header button
- [ ] merge `ol-infra` PR


frontends/main/next.config.js
```diff
@@ -44,6 +44,8 @@ const nextConfig = {
     return config
   },
 
+  productionBrowserSourceMaps: true,
```
ChristopherChudzicki, line 47: > During production builds, they are disabled to prevent you leaking your source on the client, unless you specifically opt-in with the configuration flag.

From [docs](https://nextjs.org/docs/app/api-reference/next-config-js/productionBrowserSourceMaps)

Seems like silly obfuscation to me.

Anyway, our source is public. I *think* keeping the source maps public will mean we don't need to upload them to sentry, but we'll only know for sure after we deploy it.

frontends/main/next.config.js
```diff
@@ -79,4 +81,41 @@ const nextConfig = {
   },
 }
 
-module.exports = nextConfig
+// Injected content via Sentry wizard below
+
+const { withSentryConfig } = require("@sentry/nextjs")
+
+module.exports = withSentryConfig(nextConfig, {
```
ChristopherChudzicki, line 88: This, and a lot of stuff below, was injected automatically by `@sentry/wizard` CLI

frontends/main/sentry.client.config.ts
```diff
@@ -0,0 +1,30 @@
...
+  ),
+  tracesSampleRate: Number(process.env.NEXT_PUBLIC_SENTRY_TRACES_SAMPLE_RATE),
+  tracePropagationTargets: process.env.NEXT_PUBLIC_MITOL_API_BASE_URL
+    ? [process.env.NEXT_PUBLIC_MITOL_API_BASE_URL]
+    : [],
+  // Add optional integrations for additional features
+  integrations: [
+    Sentry.replayIntegration(),
+    Sentry.browserTracingIntegration(),
+    Sentry.browserProfilingIntegration(),
```
ChristopherChudzicki, line 23: We were using these two already in the `mit-learn/` app. The `replayIntegration` was added by `@sentry/wizard`, so I kept it :shrug:

frontends/main/src/app/sentry-example-page/page.tsx
ChristopherChudzicki, line 1: To be removed prior to merging. (Added by `@sentry/wizard` and modified a bit, but useful for testing this PR.)

frontends/main/src/components/ErrorBoundary/ErrorBoundary.tsx
ChristopherChudzicki, line 1: This file was unnecessary... NextJS has the error boundaries builtin, we just provide relevant UI at `error.ts` pages.

I removed it because it was also a little buggy... It had this render method:
```tsx
  render() {
    if (this.state.hasError && isForbiddenError(this.state.error)) {
      return <ForbiddenPage />
    }
    return this.props.children
  }
```
which meant that the error was silently suppressed if it was not a `ForbiddenError`. (The error "should" have been re-thrown, but really we should have just used `error.tsx`.)

frontends/main/sentry.edge.config.ts
ChristopherChudzicki, line 1: I extremely doubt we will ever use this file, but `@sentry/wizard` added it, and there's no real harm in having it.

gumaerc (approved):
👍 LGTM

Just remember to remove that example page and header error button

PR #1707, opened 2024-10-17
bundle analyzer, fix search page + channel page JS sizes

ChristopherChudzicki:
 - Resolves https://github.com/mitodl/hq/issues/5818

This PR updates course-search-utils to provide an ESM and CJS output. 

Our build will use the ESM output, which allows us to tree-shake the libraries used within `course-search-utils`. (Primarily, remixicon).

**Important:** Review with https://github.com/mitodl/course-search-utils/pull/148

### How can this be tested?
1. Check that the search page / channel pages still look ~good~ great.
2. Run `NODE_ENV=production yarn workspace main build`
    - In the output, you should see that the JS load for `/search` and ` /c/[channelType]/[name]` is a bit bigger than the other routes, but not double, as it is on `nextjs` branch
    - If you want, add `ANALYZE=true` to the command above to get the webpack analysis

### DIFF
--- a/frontends/main/next.config.js
+++ b/frontends/main/next.config.js
@@ -103,29 +103,38 @@ const nextConfig = {
 // Injected content via Sentry wizard below
 
 const { withSentryConfig } = require("@sentry/nextjs")
+const withSentry = (config) =>
+  withSentryConfig(config, {
+    // For all available options, see:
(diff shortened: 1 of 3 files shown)


abeglova (approved):
lgtm

PR #1713, opened 2024-10-18
Announce search count for screen readers

ChristopherChudzicki:
Closes https://github.com/mitodl/hq/issues/5420

Adds an aria-live region that announces search result count.

### How can this be tested?
1. Visit the search page or any channel page
2. Make searches while a screenreader is active. On a Mac, you could use VoiceOver.
3. Each search should announce "<count> results".
4. Make two searches with the same count. This is easiest with 0 results. E.g., search "fooooz" then "foooozz". You should hear the count announced both times, even though the count did not change.


abeglova (approved):
👍

PR #1725, opened 2024-10-22
Split frontend/backend releases

rhysyngsun:

N/A

This splits the frontend and backend deployments because we want to be able to disable but still retain the backend ones.

### How can this be tested?

We'll test during a release.

### DIFF
--- a/.github/workflows/deploy-backend-production.yml
+++ b/.github/workflows/deploy-backend-production.yml
@@ -1,0 +1,48 @@
+name: Production Deploy (Backend)
+
+on:
+  workflow_run:
+    workflows: [CI]
+    types: [completed]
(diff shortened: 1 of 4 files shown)


ChristopherChudzicki (commented):
we had been doing the frontend release after the backend release.

They're parallel now, right?

that's OK...we just need to be more careful about releasing backend changes first, yes?

rhysyngsun:
> we had been doing the frontend release after the backend release.
> 
> They're parallel now, right?
> 
> that's OK...we just need to be more careful about releasing backend changes first, yes?
We should be being careful and separating the backend and frontend changes anyway, but if we need that guarantee then we shouldn't merge this.

PR #1726, opened 2024-10-22
Navigation Accessibility

ChristopherChudzicki:
Closes https://github.com/mitodl/hq/issues/5828

Makes navigation menu keyboard openable / closeable

The close button moved:
Screenshot 2024-10-22 at 12 32 26 PM

### How can this be tested?
1. Open and close the navigation menu with your mouse, with your keyboard.
2. Check that "Escape" closes the nav menu
3. Check that the nav menu traps focus (cycling through all the links+button should get you back to the beginning)
4. Check that clicking outside the nav menu closes it
    - even if you click on the initial trigger (the hamburger / explore MIT button)

### DIFF
--- a/frontends/main/src/page-components/Header/Header.tsx
+++ b/frontends/main/src/page-components/Header/Header.tsx
@@ -7,7 +7,6 @@ import {
   AppBar,
   NavDrawer,
   Toolbar,
-  ClickAwayListener,
   ActionButtonLink,
 } from "ol-components"
 import {
@@ -270,10 +269,8 @@ const Header: FunctionComponent = () => {
     event.nativeEvent.stopImmediatePropagation() // Prevent clicking on "Explore MIT" button from triggering the ClickAwayHandler
     toggleDrawer(!drawerOpen)
   }
-  const closeDrawer = (event: MouseEvent | TouchEvent) => {
-    if (drawerOpen && event.type !== "touchstart") {
-      toggleDrawer(false)
-    }
+  const closeDrawer = () => {
+    toggleDrawer(false)
   }
 
   return (
@@ -283,14 +280,10 @@ const Header: FunctionComponent = () => {
           <DesktopOnly>
             <StyledMITLogoLink logo="learn" />
             <LeftSpacer />
-            <MenuButton
-              text="Explore MIT"
-              onClick={toggler}
-              drawerOpen={drawerOpen}
-            />
+            <MenuButton text="Explore MIT" onClick={toggler} />
           </DesktopOnly>
           <MobileOnly>
-            <MenuButton onClick={toggler} drawerOpen={drawerOpen} />
+            <MenuButton onClick={toggler} />
             <LeftSpacer />
             <StyledMITLogoLink logo="learn" />
           </MobileOnly>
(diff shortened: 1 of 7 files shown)


frontends/ol-components/src/components/NavDrawer/NavDrawer.tsx
```diff
@@ -1,15 +1,19 @@
...
 const DrawerContent = styled.div(({ theme }) => ({
-  paddingTop: "72px",
+  paddingTop: theme.custom.dimensions.headerHeight,
   width: "366px",
   height: "100%",
   background: theme.custom.colors.white,
   borderRight: `1px solid ${theme.custom.colors.lightGray2}`,
   [theme.breakpoints.down("sm")]: {
-    height: "60px",
+    paddingTop: theme.custom.dimensions.headerHeightSm,
```
ChristopherChudzicki, line 16: The nav menu had extra top padding at mobile widths... I think the `height: 60px ` was supposed to be `paddingTop: 60px` all along. Changed to theme constants.

frontends/main/src/page-components/Header/Header.tsx
```diff
@@ -282,30 +280,19 @@ const Header: FunctionComponent = () => {
...
-            <MenuButton onClick={toggler} drawerOpen={drawerOpen} />
+            <MenuButton onClick={toggler} />
             <LeftSpacer />
             <StyledMITLogoLink logo="learn" />
           </MobileOnly>
           <Spacer />
           <UserView />
         </StyledToolbar>
       </Bar>
-      <ClickAwayListener
```
gumaerc, line 300: Removing the `ClickAwayListener` here broke the drawer on mobile:


https://github.com/user-attachments/assets/d045155f-8d9c-4246-b0a1-c10ab65ebbbf

I realize we created a close button in the drawer, but IMO clicking the menu button again should still close the drawer.

gumaerc (changes requested):
There's an issue with the drawer on mobile:

frontends/main/src/page-components/Header/Header.tsx
ChristopherChudzicki: Good catch. Should be resolved now.

Apparently ClickAwayListener attaches listeners to detect both click (mouse or touch) away and touchend (touch only).

Disabling the touch-only events fixes the issue, sort of. Click events also fire on touchscreens, but apparently 
- mousedown -> move lots -> mouse up = click event
- touchstart -> move lots -> touchend NOT trigger a click. (Movement must be fairly small.)

So maybe disabling the touch-only trigger would feel weird, so went with a different solution.

gumaerc (approved):
The latest changes work well on my Android phone (Samsung Galaxy S22) and on iOS, but still exhibit the "open close open" or "close open close" behavior in the Firefox mobile simulator. I'm okay with approving it because it's just the mobile simulator in FF that doesn't work, and it's possible that it's just a bug with the FF simulator itself and the way it simulates touch events. Otherwise this all works fine.

ChristopherChudzicki:
@gumaerc Tested this a lot more on my iphone this morning and various emulators on my mac. It seems to be working well. Very occasionally my iphone iOS safari seems to drop a click event when spamming taps very fast. This has been (a) rare, and (b) only under very spammy conditions, so I'm going to accept it for now.

PR #1728, opened 2024-10-22
New LearningResourcePrice model with frontend changes to use it

mbertrand:
Closes https://github.com/mitodl/hq/issues/5134 and https://github.com/mitodl/hq/issues/4685

- Adds a `LearningResourcePrice` model with two fields (`amount`, `currency`)
- Adds a many-to-many relation between LearningResource and LearningResourcePrice, also between LearningResourceRun and LearningResourcePrice (named `resource_prices`)
- Removes the `prices` field from LearningResource and LearningResourceRun.
- Updates QuerySet classes to avoid n+1 queries
- Removes the old `prices` field on LearningResource and LearningResourceRun
- Updates the front end to use the new field and display the correct currency symbol.
- Updates the search index schema with addition of the new `resource_prices` field and removal of the `prices` field.
- Adds a new python package [pycountry](https://github.com/pycountry/pycountry) to validate currency codes during ETL pipeline runs, and a JS package [currency-symbol-map](https://www.npmjs.com/package/currency-symbol-map) to display the correct symbol for the currency.

If price is in Euros (EUR):
Screenshot 2024-10-23 at 1 05 50 PM

If price is in Swiss francs (CHF, no symbol):

Screenshot 2024-10-23 at 1 03 54 PM

In dollars (USD):
Screenshot 2024-10-23 at 1 07 46 PM

### How can this be tested?
- If you don't already have some courses with prices, run these commands on the **main** branch:
  ```
  ```
- Log in as an admin user to avoid caching and go to `http://open.odl.local:8062/search?platform=mitxonline&platform=xpro`, you will see a mix of courses that should all either be paid with certificate (xpro) or free with a paid certificate option (mitxonline).  If you click on a course card you will see the same price in the drawer.
- Switch to this branch, run `docker compose build web celery` (new python package) and restart containers.  Make sure the new migrations complete successfully.
- Go to `http://open.odl.local:8062/search?platform=mitxonline&platform=xpro` again.  This time, you will see "Paid" instead of an actual price on the course cards because the index has not been updated.  If you click on one you should see the actual price in the drawer, because that is coming straight from the database.
- Run `./manage.py update_index --courses --programs`, then wait a minute or 2.
- Go to `http://open.odl.local:8062/search?platform=mitxonline&platform=xpro` again.  Now you should see the actual prices on the course cards.
- Run `./manage.py recreate_index --all`, it should complete without errors.
- You can test the currency by running this code:
   ```python
   from learning_resources.utils import resource_upserted_actions
   resource = LearningResource.objects.get(title="Applying Machine Learning to Engineering and Science")
   resource.resource_prices.all().update(currency="EUR")
   resource_upserted_actions(resource, False)
   ```
- Now wait a minute then go to http://open.odl.local:8062/search?q=%22Applying%20Machine%20Learning%20to%20Engineering%20and%20Science%22.   The price should show the euro symbol €1559 instead of $1559 on the course card and resource drawer.
- Check and uncheck the "Free" checkbox on the search page.  The results should change as expected.
- Running the following pipelines should still complete successfully (you'll need the appropriate env values from RC for `SEE_*` and `EDX_API_*`):
  ```
  ./manage.py backpopulate_mit_edx_data [--api_course_file=edx.json --api_program_datafile=edx_programs.json]
  ```
- Use the django debug toolbar to measure the # and speed of SQL queries for the URL http://api.open.odl.local:8063/api/v1/learning_resources/?platform=mitxonline.  It should not be significantly slower than the same url on the main branch, but there will be 2 additional queries.

Screenshot 2024-10-23 at 1 59 26 PM

vs

Screenshot 2024-10-23 at 2 01 03 PM

- Finally if you want to revert your price data before you go back to the main branch, run this and the `prices` field should be restored with values:
  ```
  ./manage.py migrate learning_resources 0070_learningresource_location
  ```

### Additional Context
- Potential TODO in a future PR: rename `resource_prices` to `prices`.  This was not done here because it would cause indexing errors until the index is fully recreated (incompatible schema for that field).
- Most data sources always have prices in USD.  Exceptions are Sloan and MIT edX, so their ETL pipelines have been adjusted to extract the currency from their API's.


frontends/ol-utilities/package.json
```diff
@@ -16,6 +16,7 @@
     "@faker-js/faker": "^9.0.0",
     "api": "workspace:*",
     "classnames": "^2.3.2",
+    "currency-symbol-map": "^5.1.0",
```
mbertrand, line 19: I added this as a convenience for getting the correct symbol for a currency code, but if preferable I could just copy the mapping of codes to symbols to our own code base instead.

pyproject.toml
```diff
@@ -80,6 +80,7 @@ ruff = "0.6.4"
 dateparser = "^1.2.0"
 uwsgitop = "^0.12"
 pytest-lazy-fixtures = "^1.1.1"
+pycountry = "^24.6.1"
```
mbertrand, line 83: For validating currency codes coming in from external API's

frontends/ol-components/src/components/LearningResourceExpanded/LearningResourceExpanded.test.tsx
```diff
@@ -169,10 +169,14 @@ describe("Learning Resource Expanded", () => {
         .getByRole("heading", { name: "Info" })!
         .closest("section")!
 
-      const price = run.prices?.[0]
+      const price = run.resource_prices?.[0]
 
       const displayPrice =
-        parseFloat(price!) === 0 ? "Free" : price ? `$${price}` : null
+        parseFloat(price.amount!) === 0
```
ChristopherChudzicki, line 175: Mentioning: We shouldn't have been using `!` here in the first place, and now it has no effect.

`!` is Typescript's [non-null assertion](https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#non-null-assertion-operator-postfix-), but `price.amount` has type `string` (already non-null) so it doesn't do anything.

We shouldn't really use it because it can lead to errors (it's silencing a type error that could be real). We have ended up using it in a few places because DRF Spectacular's types are sometimes nullable where they shouldn't be. (At least, I'm pretty sure DRF spec makes them nullable in some places it shouldn't be. Perhaps there's some reason I don't understand that it is technically correct.)

mbertrand:
Closing, replaced by https://github.com/mitodl/mit-learn/pull/1736 and https://github.com/mitodl/mit-learn/pull/1737

PR #1729, opened 2024-10-22
Middleware for SSL redirect

jonkafton:

Closes https://github.com/mitodl/hq/issues/5836

Adds a middleware function to redirect any traffic on HTTP to HTTPS.

Redirects for requests to http://mit.learn.edu are handled by the CDN, though this change blocks content from being served over HTTP if the assigned Next.js server domains are addressed directly. Very edge case as we have a CORS list, though this was highlighted while we had a whitelisted domain before the CDN was set up on RC.

### How can this be tested?

Build the app for production and test that it redirects to HTTPS
- Run `yarn build`
- Run `PORT=8062 yarn start`
- Add an `/etc/hosts` file entry to 127.0.01 that does not include the string "local", e.g. `127.0.01     testing`.
- Navigate to http://testing:8062/
- You should be redirected to https://testing:8062/ (where locally you'll see an SSL error).
- Paths and search params should be preserved.

### DIFF
--- a/frontends/main/src/middleware.ts
+++ b/frontends/main/src/middleware.ts
@@ -1,0 +1,25 @@
+import { NextResponse } from "next/server"
+import type { NextRequest } from "next/server"
+
+/* The CDNs handle SSL redirects, but this is intended to block content being served
+ * over HTTP if directly addressed at their origin server domains.
+ */
+export function middleware(request: NextRequest) {
+  const protocol = request.headers.get("x-forwarded-proto")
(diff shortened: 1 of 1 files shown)


shanbady (approved):
changes here appear to work. discovered a separate issue with local node setup which jon has cut separate issues for

jonkafton:
For attention on RC when we release: There is a possibility that this will cause a redirect loop if the `x-forwarded-proto` is not exactly `"https"`.

PR #1730, opened 2024-10-22
Search count inside label

ChristopherChudzicki:

For https://github.com/mitodl/hq/issues/5708

Puts the search count inside the `<label>` for screenreaders 

No visual change

Screenshot 2024-10-22 at 4 57 11 PM
Screenshot 2024-10-22 at 4 59 14 PM

### How can this be tested?
1. Check that the search page has no visual change
    - OK I guess there is one... the mouse cursor is `pointer` on the count now, too.
    - you might want to check that the checkboxes change color the same as prod on hover, etc
2. Check that the checkbox has count as part of its label in the accessibility tree. You could also enable a screenreader like voiceover, if you want. (See screenshot above)

### Checklist:
- [ ] Review with https://github.com/mitodl/course-search-utils/pull/151

### DIFF
--- a/frontends/main/src/page-components/SearchDisplay/SearchDisplay.tsx
+++ b/frontends/main/src/page-components/SearchDisplay/SearchDisplay.tsx
@@ -206,7 +206,7 @@ const FacetStyles = styled.div`
         float: right;
       }
 
-      &.chcked,
+      &.checked,
       &:hover {
         .facet-label .facet-text {
           color: ${({ theme }) => theme.custom.colors.darkGray2};
--- a/frontends/main/src/page-components/SearchDisplay/SearchDisplay.tsx
+++ b/frontends/main/src/page-components/SearchDisplay/SearchDisplay.tsx
@@ -140,7 +140,7 @@ const FacetStyles = styled.div`
     width: 100%;
     align-items: baseline;
 
-    label {
+    .facet-text {
       ${truncateText(1)};
       color: ${({ theme }) => theme.custom.colors.silverGrayDark};
     }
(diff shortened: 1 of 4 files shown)


frontends/main/src/page-components/SearchDisplay/SearchDisplay.tsx
```diff
@@ -205,12 +205,13 @@ const FacetStyles = styled.div`
         color: ${({ theme }) => theme.custom.colors.silverGrayDark};
         float: right;
       }
-    }
 
-    .facet-visible.checked .facet-label label,
-    .facet-visible .facet-label label:hover,
-    .facet-visible input:hover + .facet-label label {
-      color: ${({ theme }) => theme.custom.colors.darkGray2};
+      &.chcked,
```
jonkafton, line 209: `&.checked` here (impact is that we don't get the darker text when checked).

jonkafton (changes requested):
Just the one mistype on the selector.

For my curiosity, what is the intent of this change? I would have thought the count to be semi-auxiliary information that clutters the audio. I guess all information should be available to screen readers, though doesn't labels such as "With Certificate 20" break the semantics? Would `role="presentation"` help, or does that default the point?

ChristopherChudzicki:
> I would have thought the count to be semi-auxiliary information that clutters the audio. I guess all information should be available to screen readers, though doesn't labels such as "With Certificate 20" break the semantics?

I can see both viewpoints (clutter audio / useful info) but am deferring to the screenreader review on this one. This particular piece of screenreader feedback—and many others we're working on lately—come from a full-time screenreader user + accessibility expert who works for MIT.

PR #1731, opened 2024-10-23
Clean out the static frontend

jonkafton:

https://github.com/mitodl/hq/issues/5846

Removes the ./frontends/mit-learn workspace and any references. This was the entrypoint for the previous static frontend, now replaced by the Next.js entrypoint at ./frontends/main.

Also removes the E2E/browser tests as these have been inactive and have not been migrated to work with Next.js.

### How can this be tested?

Generally, all tests should pass on frontend and backend and services should build and run with Docker without error.


main/settings.py
```diff
@@ -217,7 +217,6 @@
         "BACKEND": "django.template.backends.django.DjangoTemplates",
         "DIRS": [
             BASE_DIR + "/templates/",
-            BASE_DIR + "/frontends/mit-learn/build",
```
jonkafton, line 220: Is this needed for anything?

scripts/test/python_tests.sh
```diff
@@ -12,10 +12,6 @@ function run_test {
 	poetry run "$@"
 }
 
-# Placeholder build for use in django tests.
-mkdir ./frontends/mit-learn/build
-touch ./frontends/mit-learn/build/index.html
```
jonkafton, line 17: Are these needed for anything?

channels/views_test.py
```diff
@@ -187,7 +187,7 @@ def test_patch_channel_image(client, channel, attribute):
         os.path.dirname(__file__),  # noqa: PTH120
         "..",
         "frontends",
-        "mit-learn",
+        "main",
         "public",
         "images",
         "blank.png",
```
jonkafton, line 193: I've added a blank.png at this path, but do we actually need it for this test? It's not used in the frontend.

main/settings.py
ChristopherChudzicki: It was necessary when django served the JS

scripts/test/python_tests.sh
ChristopherChudzicki: Only back when django served the JS

channels/views_test.py
ChristopherChudzicki: It's "supposed" to be the fallback for embedly. See frontends/ol-utilities/src/learning-resources/learning-resources.ts

That said:
- we don't need the /static version
- The version we do have set isn't working right now (https://rc.learn.mit.edu/static/images/blank.png) ... unsure why

ChristopherChudzicki (approved):
Approved, but could you change the url in
```
const BLANK_IMAGE = "https://rc.learn.mit.edu/static/images/blank.png"
```

to use NEXT_PUBLIC_ORIGIN

PR #1736, opened 2024-10-24
New LearningResourcePrice model

mbertrand:
PR 1 of 5 to address https://github.com/mitodl/hq/issues/5134 and https://github.com/mitodl/hq/issues/4685

- Adds a `LearningResourcePrice` model with two fields (`amount`, `currency`)
- Adds a many-to-many relation between LearningResource and LearningResourcePrice, also between LearningResourceRun and LearningResourcePrice (named `resource_prices`)
- Updates QuerySet classes to avoid n+1 queries
- Updates the search index schema with addition of the new `resource_prices` field 
- Adds a new python package [pycountry](https://github.com/pycountry/pycountry) to validate currency codes during ETL pipeline runs

### How can this be tested?
- If you don't already have some courses with prices, run these commands on the **main** branch:
  ```
  ```
- Log in as an admin user to avoid caching and go to `http://open.odl.local:8062/search?platform=mitxonline&platform=xpro`, you will see a mix of courses that should all either be paid with certificate (xpro) or free with a paid certificate option (mitxonline).  If you click on a course card you will see the same price in the drawer.
- Switch to this branch, run `docker compose build web celery` (new python package) and restart containers.  Make sure the new migrations complete successfully.
- Go to `http://open.odl.local:8062/search?platform=mitxonline&platform=xpro` again.  Everything should still look the same.  
- Click on a result and check the API response for that resource.  There should be a `resource_prices` attribute for the resource and each of it's runs, with `amount` and `currency` attributes.
- Run `./manage.py update_index --courses --programs`, then wait a minute or 2.
- Go to `http://open.odl.local:8062/search?platform=mitxonline&platform=xpro` again.  It should still look the same, and if you inspect search API results, the `resource_prices` field should be there too.
- Run `./manage.py recreate_index --all`, it should complete without errors.
- Check and uncheck the "Free" checkbox on the search page.  The results should change as expected.
- Running the following pipelines should still complete successfully (you'll need the appropriate env values from RC for `SEE_*` and `EDX_API_*`):
  ```
  ./manage.py backpopulate_mit_edx_data [--api_course_file=edx.json --api_program_datafile=edx_programs.json]
  ```
- Use the django debug toolbar to measure the # and speed of SQL queries for the URL http://api.open.odl.local:8063/api/v1/learning_resources/?platform=mitxonline.  It should not be significantly slower than the same url on the main branch, but there will be 2 additional queries.

Screenshot 2024-10-23 at 1 59 26 PM

vs

Screenshot 2024-10-23 at 2 01 03 PM

- Finally if you want to revert your price data before you go back to the main branch, run this and the `prices` field should be restored with values:
  ```
  ./manage.py migrate learning_resources 0070_learningresource_location
  ```

### Additional Context
- Most data sources always have prices in USD.  Exceptions are Sloan and MIT edX, so their ETL pipelines have been adjusted to extract the currency from their API's.
- There will be 4 subsequent PR's before https://github.com/mitodl/hq/issues/5134 is closed:
  - 2: Change frontend to use the `resource_prices` field instead of `prices`
  - 3: Modify the `LearningResource.prices` and `LearningResourceRun.prices` fields to be the same as `resource_prices`, and update schema. (Note: this will cause index updates to fail until a reindex completes)
  - 4: Change frontend to use `prices` field again
  - 5: Remove `resource_prices` from models and schemas.

### DIFF
--- a/learning_resources/migrations/0071_learningresourceprice.py
+++ b/learning_resources/migrations/0071_learningresourceprice.py
@@ -62,7 +62,6 @@ class Migration(migrations.Migration):
             name="resource_prices",
             field=models.ManyToManyField(
                 blank=True,
-                related_name="runs",
                 to="learning_resources.learningresourceprice",
             ),
         ),
@@ -71,7 +70,6 @@ class Migration(migrations.Migration):
             name="resource_prices",
             field=models.ManyToManyField(
                 blank=True,
-                related_name="resources",
                 to="learning_resources.learningresourceprice",
             ),
         ),
--- a/learning_resources/models.py
+++ b/learning_resources/models.py
@@ -439,9 +439,7 @@ class LearningResource(TimestampedModel):
     prices = ArrayField(
         models.DecimalField(decimal_places=2, max_digits=12), default=list
     )
-    resource_prices = models.ManyToManyField(
-        LearningResourcePrice, blank=True, related_name="resources"
-    )
+    resource_prices = models.ManyToManyField(LearningResourcePrice, blank=True)
     availability = models.CharField(  # noqa: DJ001
         max_length=24,
         null=True,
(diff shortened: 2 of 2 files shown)


ChristopherChudzicki (approved):
This looks good 👍 

On thing I noticed is that when I ran the indexing commands, Django told me:
```
WARNINGS:
learning_resources.LearningResource.resource_prices: (fields.W340) null has no effect on ManyToManyField.
learning_resources.LearningResourceRun.resource_prices: (fields.W340) null has no effect on ManyToManyField.
Operations to perform
```

mbertrand:
@ChristopherChudzicki Good catch, I just updated the model definition/migration for those fields

PR #1739, opened 2024-10-24
Replace plain <a> tags with Next.js <Link> components

jonkafton:

Closes https://github.com/mitodl/hq/issues/5853.

Replaces any remaining plain links to internal routes with Next.js `<Link>` components.

### How can this be tested?

- Check the nav drawer links. These should navigate without reloading the page.
  - Search links (navigating between items in “Learn” and “Discover Learning Resources”) should update the search filters without reload.
  - The nav drawer should close on selection.
 
- The header search icon should navigate without reload.

- Footer links should navigate without reload ("Accessibility" and "Contact Us" are external, but still navigate ok).

### Additional Context

The Channel detail links to external sites and login links in the header, user menu and Personalize section are the only remaining plain `<a>` tags. The login links use `<ButtonLink>` with `rawAnchor={true}`, which we perhaps don't need as Next.js `<Link>` works fine for external URLs and `target="_blank"` (e.g. the MIT logo link in the header).

### DIFF
--- a/frontends/ol-components/src/components/NavDrawer/NavDrawer.tsx
+++ b/frontends/ol-components/src/components/NavDrawer/NavDrawer.tsx
@@ -2,6 +2,7 @@ import Drawer, { DrawerProps } from "@mui/material/Drawer"
 import ClickAwayListener from "@mui/material/ClickAwayListener"
 import { FocusTrap } from "@mui/base/FocusTrap"
 import styled from "@emotion/styled"
+import Link from "next/link"
 import React, { ReactElement } from "react"
 import { RiCloseLargeLine } from "@remixicon/react"
 import { ActionButton } from "../Button/Button"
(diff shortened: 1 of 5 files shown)


frontends/ol-components/src/components/NavDrawer/NavDrawer.tsx
```diff
@@ -127,10 +128,11 @@ export interface NavItem {
   icon?: string | ReactElement
   description?: string
   href: string
+  onClick?: () => void
```
ChristopherChudzicki, line 131: This doesn't seem to be used? Let's remove it to avoid expanding the interface.

ChristopherChudzicki (approved):
👍 with the one comment

frontends/ol-components/src/components/NavDrawer/NavDrawer.tsx
jonkafton: It's newly used ([L191](https://github.com/mitodl/mit-learn/pull/1739/files#diff-2959b2acb45249092d040d1b51c53e9b4d47ff54e29b6f588972d8d6b5593c06R191)) to pass the drawer close when the link is clicked.

PR #1743, opened 2024-10-24
Shanbady/similar resources endpoint

shanbady:

Closes https://github.com/mitodl/hq/issues/4310

This PR introduces a "similar resources" endpoint which is expanded onto the learning resource detail endpoint. To get similar resources for a learning resource, add '/similar' to the learning resources detail endpoint. example <pre>http://od.odl.local:8063/api/v1/learning_resources/<b>20/similar/</b></pre>

### How can this be tested?
1. Checkout this branch. have resources loaded up. 
2. pick some random resource id ( note the details of the resource like title description etc to compare to the results)
3. visit the similar resources endpoint with this id <pre>http://od.odl.local:8063/api/v1/learning_resources/<b>{resource id}</b>/similar/</pre>
4. By default it returns 10 resources. this can be overriden by passing in the limit as a query parameter.

### DIFF
--- a/learning_resources_search/api.py
+++ b/learning_resources_search/api.py
@@ -895,7 +895,8 @@ def get_similar_resources(
     indexes = relevant_indexes([COURSE_TYPE], [], endpoint=LEARNING_RESOURCE)
     search = Search(index=",".join(indexes))
     if num_resources:
-        search = search.extra(size=num_resources)
+        # adding +1 to num_resources since we filter out existing resource.id
+        search = search.extra(size=num_resources + 1)
     search = search.filter("term", resource_type=COURSE_TYPE)
     search = search.query(
         MoreLikeThis(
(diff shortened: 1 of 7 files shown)


learning_resources_search/api.py
```diff
@@ -870,3 +870,51 @@ def get_similar_topics(
...
+            minimum times a term needs to show up in docs
+
+    Returns:
+        list of str:
+            list of topic values
+    """
+    indexes = relevant_indexes([COURSE_TYPE], [], endpoint=LEARNING_RESOURCE)
+    search = Search(index=",".join(indexes))
+    if num_resources:
+        search = search.extra(size=num_resources)
```
abeglova, line 898: This should be num_resources+1  since the original resource is included in the results and later filtered out. Currently you get limit-1 resources in the response

abeglova (approved):
LGTM - one minor fix required

learning_resources_search/api.py
shanbady: good catch! fixed

PR #1751, opened 2024-10-25
move facets first in tab order

ChristopherChudzicki:
Closes https://github.com/mitodl/hq/issues/5419

Puts facets after search results in tab order (and DOM order)

### How can this be tested?
1. There should be no visual change.
2. Tab though the search page—either at `/search` or on the channel pages. Starting from the search text field, you should see:
    1. Text field
    2. Tabs (All | Courses | Programs) *Only the active tab is tabable. The other tabs are navigated to via arrow keys.)
    3. Sort dropdown
    4. Search results x 20 (or whatever)
    5. THEN facets

Note: This change makes the facets a bit harder to get to, but the search results themselves are easier to get to.

Facets are still—somewhat—easily navigated to with a screenreader via the "Search FIlters" heading.


jonkafton (approved):
Even after my short time spent VoiceOver testing https://github.com/mitodl/mit-learn/pull/1778, I am very grateful for this change!

PR #1769, opened 2024-10-28
Image and video optimizations

jonkafton:

Closes https://github.com/mitodl/hq/issues/5857

Upgrades to Next.js v15

- Replaces `` tags with `<Image>` components where it makes sense to.
  - There are places where it doesn't make sense to, such as where we have containers that would prevent layout shift and where we are already able to provide the image as a static import (Next.js set the `src` string to its image optimization endpoint), e.g. the PersonalizeSection image. 

  - Static imports for images in repo (import src is a path to the Next.js image optimization endpoint). This is used for both `` tags and CSS background-image.

  - Provides a utility that produces a CSS `image-set()` declaration for a statically imported image. These are links for the image from the optimization endpoint at resolutions for 1x and 2x screen pixel densities, see https://developer.mozilla.org/en-US/docs/Web/CSS/image/image-set. Applied to the homepage personalize section background. These can also be used on `` tags ([srcset](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/img#srcset) attribute). We should consider applying elsewhere, particularly for large images, though also should use with caution as it may not be desirable to render high resolution images for example for backgrounds that contain little detail.

  - The resource drawer images are now optimized by Next.js with noticeable improvement to quality:
    - Before:
      image

    - After:
      image

- Resource drawer YouTube videos are displayed using a simple iframe (YouTube's embed html is similar). Non-YouTube videos fall back to Embedly, though all of our videos are YouTube at present. This also removes some padding added by Embedly and sizes the videos to match our aspect ratios and applies the border radius.

    - Before: 
       image

    - After:
       image

  - Updates ol-components/Logo to support unit logos and distinguish between units and platforms so it can be used on the unit channel pages. We can't import from the `./public/images` folder to use image optimization as we have the rule that ol-components are leaf imports, though there is no value for SVGs.
  
  - Clean up repetitions of the Embedly key / move into EmbedlyCard.

### How can this be tested?

Check that images and logos display correctly at mobile and desktop. Particular attention to:

- Resource drawer images and video.
- Homepage personalize section background.
- Nav drawer icons.
- About page main image.
- Topics listing page banner background.
- Channel pages banner background.
- Units listing page card logos.


frontends/main/src/app-pages/ChannelPage/UnitChannelTemplate.tsx
```diff
@@ -126,7 +135,20 @@ const UnitChannelTemplate: React.FC<UnitChannelTemplateProps> = ({
...
+                        unitCode={name as OfferedByEnum}
+                        height={50}
+                      />
+                    </DesktopOnly>
+                    <MobileOnly>
+                      <PlatformLogoInverted
+                        unitCode={name as OfferedByEnum}
+                        height={40}
+                      />
+                    </MobileOnly>
```
jonkafton, line 150: Here we produce for each side of the width breakpoint and use the css media query to hide one. This is wordy, though needed for varying the height with only CSS. Where the images are in the repo we can generally pull them in as static imports, so don't need to specify the dimensions, though in this case the ol-components PlatformLogo is not allowed to import from outside its workspace. Further, the width and height on <Image> components do not size the image itself - it is just to block space to prevent layout shift, which can be done with CSS. Here though the logos are of different sizes so we need to hardcode the dimensions (here providing fixed height with width calculated in PlatformLogo).

frontends/main/src/app-pages/DepartmentListingPage/DepartmentListingPage.tsx
```diff
@@ -211,6 +210,7 @@ const DepartmentListingPage: React.FC = () => {
             current="Departments"
           />
         }
+        backgroundUrl={backgroundSteps.src}
```
jonkafton, line 213: The `<Banner>` component has a default image path (background_steps.jpg), but as this is just a path string, the image is loaded directly as opposed to from the image optimization endpoint. Static imports get the image optimzation endpoint, but as a rule the `<Banner>` component cannot import from outside ol-components.

frontends/main/src/app-pages/HomePage/HomePage.tsx
```diff
@@ -43,24 +43,28 @@ const HomePage: React.FC = () => {
     <>
       <LearningResourceDrawer />
       <FullWidthBackground>
-        <Container>
+        <Container style={{ overflow: "hidden" }}>
```
jonkafton, line 46: Fixes this issue where the hero image pushes past the container causing horizontal scroll:

image

jonkafton: Ignore, addressed in https://github.com/mitodl/mit-learn/pull/1777

frontends/main/src/app-pages/HomePage/HomePage.tsx
```diff
@@ -41,26 +41,30 @@ const MediaCarousel = styled(ResourceCarousel)(({ theme }) => ({
...
+      <Container>
+        <LearningResourceDrawer />
+      </Container>
       <FullWidthBackground>
-        <Container>
-          <HeroSearch />
-          <section>
+        <HeroSearch />
+        <section>
+          <Suspense>
```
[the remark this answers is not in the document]
jonkafton: Addressed separately in https://github.com/mitodl/mit-learn/pull/1776

ChristopherChudzicki (changes requested):
This is working great. I left a few comments, some of which we should probably make sense to do now (remove comments, maybe use backgroundSrcSetCSS more, add fallbak to backgroundSrcSetCSS?).

frontends/ol-utilities/src/images/backgroundImages.ts
```diff
@@ -0,0 +1,27 @@
...
+
+  const imageSet = props.srcSet
+    ?.split(", ")
+    .map((str) => {
+      const [url, dpi] = str.split(" ")
+      return `url("${url}") ${dpi}`
+    })
+    .join(", ")
+
+  return imageSet ? `image-set(${imageSet})` : ""
```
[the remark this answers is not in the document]
jonkafton: `props.srcSet` is typed `string | undefined`, though I'm not sure how it could be. Either way the backround image would be broken, so have removed the ternary.

frontends/main/src/app-pages/TopicsListingPage/TopicsListingPage.tsx
```diff
@@ -275,6 +275,7 @@ const TopicsListingPage: React.FC = () => {
         }
         title="Browse by Topic"
         header="Select a topic below to explore relevant learning resources across all Academic and Professional units."
+        backgroundUrl={backgroundSteps.src}
```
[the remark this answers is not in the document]
jonkafton: For this one the `getImageProps()` only produces the single 1x image, so using srcset has no impact.

Also, the `backgroundUrl` prop is used in the `<Banner>` css like `linear-gradient(rgba(0 0 0 / ${backgroundDim}%), rgba(0 0 0 / ${backgroundDim}%)), url('${backgroundUrl}')`, so we needed some extra gymnastics:

```
    const backgroundUrlFn = backgroundUrl.startsWith("image-set(")
      ? backgroundUrl
      : `url('${backgroundUrl}')`
```

Works fine though.

ChristopherChudzicki (approved):
👍

ChristopherChudzicki (approved):
👍

Issue #1772, opened 2024-10-29
Flaky test learning_resources/views_test.py::test_similar_resources_endpoint_only_returns_published


shanbady:
Flaky [test failure](https://github.com/mitodl/mit-learn/actions/runs/11573783309/job/32216615661#step:11:440) `learning_resources/views_test.py::test_similar_resources_endpoint_only_returns_published - assert 0 == 1`

### Current Behavior

This test appears to fail when the endpoint in the test returns an empty set. I think this happens when only published resource in the test is the same as the resource we pass into the similar resources endpoint.

PR #1773, opened 2024-10-29
fix intermittent failure for flaky test

shanbady:

Closes https://github.com/mitodl/mit-learn/issues/1772

Fixes the intermittent test failure in test_similar_resources_endpoint_only_returns_published

### How can this be tested?
1. checkout **main**
2. decorate learning_resources/views_test.py::test_similar_resources_endpoint_only_returns_published  with `@pytest.mark.parametrize('execution_number', range(100))` and run it via `pytest learning_resources/views_test.py::test_similar_resources_endpoint_only_returns_published -s -vvv -x` to see it fail
3. checkout this branch and try the same and note it succeeds every time

### DIFF
--- a/learning_resources/views_test.py
+++ b/learning_resources/views_test.py
@@ -1007,6 +1007,7 @@ def test_similar_resources_endpoint_only_returns_published(mocker, client):
     resources = LearningResourceFactory.create_batch(5)
 
     resource_ids = [learning_resource.id for learning_resource in resources]
+    similar_for = resource_ids.pop()
     mocker.patch.object(Search, "execute")
     response_resources = LearningResource.objects.for_search_serialization().filter(
         id__in=resource_ids
@@ -1031,7 +1032,6 @@ def test_similar_resources_endpoint_only_returns_published(mocker, client):
             "took": 123,
         },
     ).hits
-    similar_for = resource_ids[0]
     resp = client.get(
         reverse("lr:v1:learning_resources_api-similar", args=[similar_for])
     )


ChristopherChudzicki (approved):
Approving because this works, but:

1. What was causing the failures before?
2. Mentioning: I see that `get_similar_resources` does
    ```
            return LearningResource.objects.for_search_serialization().filter(
            id__in=[
                resource.id
                for resource in response.hits
                if resource.id != value_doc["id"] and resource.published
            ]
        )
    ```
    So I guess this tests the last part of that filter. I think we also exclude unpublished stuff from the search indexes (do we?) . But if I'm correct, no harm in doing it here, too.

shanbady:
> Approving because this works, but:
> 
> 1. What was causing the failures before?
> 2. Mentioning: I see that `get_similar_resources` does
>    ```
>            return LearningResource.objects.for_search_serialization().filter(
>            id__in=[
>                resource.id
>                for resource in response.hits
>                if resource.id != value_doc["id"] and resource.published
>            ]
>        )
>    ```
>    
>    
>        
>          
>        
>    
>          
>        
>    
>        
>      
>    So I guess this tests the last part of that filter. I think we also exclude unpublished stuff from the search indexes (do we?) . But if I'm correct, no harm in doing it here, too.

The initial issue happened when the document id we pass into the view happens to be one published=True resource we expect in the response. The fix was to ensure that the document id we pass in is not in the list of mocked resources for the response (so that the one published=True resource is not the same as the id passed in)

PR #1774, opened 2024-10-29
Search Vector Embeddings

shanbady:
Part of https://github.com/mitodl/hq/issues/5865

This PR is the first part of https://github.com/mitodl/hq/issues/5865 which adds a management command that sets up opensearch with vector embeddings.

**Note:** The follow up to this work is to add a new and adjacent similar resources endpoint that uses embeddings instead of "More Like This" queries so we can compare results.

This PR adds the following:
- some new settings in [settings.py](https://github.com/mitodl/mit-learn/blob/shanbady/similar-embeddings-endpoint/main/settings.py#L790) for specifying embedding and vector search models:
    - EMBEDDING_PIPELINE_NAME
    - EMBEDDING_MODEL_NAME
    - EMBEDDING_MODEL_VERSION
    - EMBEDDING_MODEL_FORMAT
    - SEARCH_TOKENIZER_MODEL_NAME
    - SEARCH_TOKENIZER_MODEL_VERSION
    - SEARCH_TOKENIZER_MODEL_FORMAT
- A django management command that sets up and enables the embedding ingestion pipeline `./manage.py setup_embedding_pipeline`
- Some tweaks to the opensearch single node compose file so it works locally (changes to avoid opensearch circuit breaker issues)

### How can this be tested?
The best way to test this end to end is to start with a fresh opensearch instance
1. Checkout this branch
2. do a `docker compose down`
3. remove the persisted elastic search volume via `docker volume rm  mit-open_opensearch-data1`
4. do a docker compose up
5.  run the management command to install the embedding pipeline: `./manage.py setup_embedding_pipeline`
6. run management commands to load resources into opensearch. Here is a convenient one-liner 
  ```python manage.py update_offered_by && python manage.py update_platforms && python manage.py update_departments_schools && python manage.py update_course_number_departments && python manage.py backpopulate_mitxonline_data && python manage.py backpopulate_micromasters_data && python manage.py backpopulate_prolearn_data && python manage.py backpopulate_oll_data && python manage.py backpopulate_sloan_data && python manage.py backpopulate_resource_channels --overwrite --all && python manage.py recreate_index --all```
7. visit the opensearch indexes endpoint (http://0.0.0.0:9100/_cat/indices) and find the name of the course index
8. visit the _search endpoint for the courses endpoint (example: http://0.0.0.0:9100/_cat/mitlearn_course_78ce08364d744fbf80c13908f4dc9fd0/_search)
9. note that there are new fields for each of the results ("title_embeddings" and "description_embeddings") that contain sparse embeddings

### Additional Context
Some considerations:
- this PR allows us to enable this pipeline via a django command - there are two main reasons for this:
  - to allow us to toggle this manually in live/other environments while testing
  - I was running into some performance/memory issues in codespaces (and at times locally) so i think it makes sense to disable this by default until it is battle tested. 
  - Locally it uses a smaller mini model for embeddings - presumably we would use a larger one in production.
 

What we still need to figure out:
- What the production setup looks like. Locally this all runs on a single node. Opensearch docs suggest having a separate ML node in production. I will need input from the team on how/what needs to get setup
- should we be using gpu enabled instances for the ml node in prod? (or can we get away with cpu)


shanbady:
closing in favor of qdrant based approach

PR #1777, opened 2024-10-29
Fix issue with hero image causing horizontal scroll

jonkafton:

N/A

Issue in production where the Hero image causes overflow to the right of its container causing the page to be horizontally scrollable. To replicate, reduce your screen width to <1365px and scroll left:

image

<br/>
<br/>
<br/>
<br/>

By design the image is translated right by 48px so is cropped right on the page. With this fix:

image

<br/>
<br/>
<br/>
<br/>

We could add an `overflow: hidden` to its container, though the issue is that the image scales and at larger screen widths we want to see the whole image, otherwise it looks like this: 

image

<br/>
<br/>
<br/>
<br/>

This fix add a media breakpoint so that the image displays in full at larger widths, but has overflow hidden where it would interfere with the page margins.

Open to a better solution with the css, or an update to the design where the image is not translated right (not sure how intentional this was), which looks like:

image

- [ ] Desktop screenshots

### How can this be tested?

The homepage should not be scrollable at all screen widths and the hero image should display correctly.

### DIFF
--- a/frontends/main/src/app-pages/HomePage/HomePage.tsx
+++ b/frontends/main/src/app-pages/HomePage/HomePage.tsx
@@ -38,12 +38,18 @@ const MediaCarousel = styled(ResourceCarousel)(({ theme }) => ({
   },
 }))
 
+const StyledContainer = styled(Container)({
+  "@media (max-width: 1365px)": {
+    overflow: "hidden",
+  },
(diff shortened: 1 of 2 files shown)


shanbady (approved):
looks good 👍

PR #1778, opened 2024-10-30
Card Accessibility Improvements

ChristopherChudzicki:
Closes https://github.com/mitodl/hq/issues/5538

### Description  (What does it do?)
This PR makes some accessibility enhancements to cards and learning resource cards in particular. 

**Card Title is Link:** Previously the entire card was a link. Now just the title is a link, leading to much better link names. *The whole card remains clickable.*

**LearningResourceCards are `article`s**: Learning resource cards now use `article` as a root element with `[resource.type]: [resource.title]` as the article name. Consequently, cards are landmarks that appear in landmark navigation

**Simplified Positioning**: Cards no longer use absolute positioning for buttons. 

### Screenshots

landmarks_rotor
links_rotor

Card aesthetics should be unchanged:

Screenshot 2024-10-30 at 7 57 45 AM
Screenshot 2024-10-30 at 7 57 55 AM
Screenshot 2024-10-30 at 7 58 04 AM
Screenshot 2024-10-30 at 7 58 12 AM

### How can this be tested?
1. Inspect cards and check that their aesthetic is unchanged. Cards include:
    - Carousel resource cards (LearningResourceCard)
    - Search result cards (LearningResourceListCard)
    - Userlist cards at `/dashboard/my-lists/` 
    - Draggable resource cards at `/dashboard/my-lists/{list_id}` when "Reordering"
    - News & Events cards on the homepage 
2.Where cards were clickable as links before, they should be now.
    - You should still be able to ctrl+click to open in a new tab
3. Interactive children in cards should still work, and should NOT navigate the whole link.  Examples are
   - "AddToListDialog" (learning path and userlist versions) 
   - The "..." menu on learning path cards.

### DIFF
--- a/frontends/ol-components/src/components/Card/Card.tsx
+++ b/frontends/ol-components/src/components/Card/Card.tsx
@@ -5,6 +5,7 @@ import React, {
   isValidElement,
   CSSProperties,
   useCallback,
+  AriaAttributes,
 } from "react"
 import styled from "@emotion/styled"
 import { theme } from "../ThemeProvider/ThemeProvider"
@@ -18,7 +19,7 @@ type LinkableProps = {
   href?: string
   children?: ReactNode
   className?: string
-}
+} & Pick<AriaAttributes, "aria-label">
 /**
  * Render a NextJS link if href is provided, otherwise a span.
  * Does not scroll if the href is a query string.
@@ -27,10 +28,16 @@ export const Linkable: React.FC<LinkableProps> = ({
   href,
   children,
   className,
+  "aria-label": ariaLabel,
 }) => {
   if (href) {
     return (
-      <Link className={className} href={href} scroll={!href.startsWith("?")}>
+      <Link
+        aria-label={ariaLabel}
+        className={className}
+        href={href}
+        scroll={!href.startsWith("?")}
+      >
         {children}
       </Link>
     )
@@ -206,7 +213,8 @@ type TitleProps = {
   children?: ReactNode
   lines?: number
   style?: CSSProperties
-}
+} & Pick<AriaAttributes, "aria-label">
+
 type SlotProps = { children?: ReactNode; style?: CSSProperties }
 
 type Card = FC<CardProps> & {
--- a/frontends/ol-components/src/components/Card/Card.tsx
+++ b/frontends/ol-components/src/components/Card/Card.tsx
@@ -19,7 +19,7 @@ type LinkableProps = {
   href?: string
   children?: ReactNode
   className?: string
-} & Pick<AriaAttributes, "aria-label">
+}
 /**
  * Render a NextJS link if href is provided, otherwise a span.
  * Does not scroll if the href is a query string.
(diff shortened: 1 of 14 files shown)


frontends/ol-components/src/components/Card/Card.tsx
```diff
@@ -130,17 +139,72 @@ const Bottom = styled.div`
...
+   * href link as well.
+   *
+   * NOTES:
+   *  - If using Card.Content to customize, you must ensure the content includes
+   *  an anchor with the card's href.
+   *  - Clicks will NOT be forwarded if it is (or is a child of):
+   *    - an anchor or button element
+   *    - OR an element with data-card-action
+   */
+  forwardClicksToLink?: boolean
```
ChristopherChudzicki, line 204: Originally, I made this the default behavior. However, since reliably NOT triggering the link when clicking interactive children requires some effort from consumer (adding `data-card-action` if the interactive child is not a anchor or button) I decided to make it opt-in to increase visibility.

frontends/ol-components/src/components/Card/Card.tsx
```diff
@@ -130,17 +139,72 @@ const Bottom = styled.div`
 const Actions = styled.div`
   display: flex;
   gap: 8px;
-  position: absolute;
-  bottom: 16px;
-  right: 16px;
```
ChristopherChudzicki, line 135: Previously, we used absolute positioning because:
- The "main" card element was an anchor
- anchors can't have interactive children (invalid HTML)
- so we made anchor + buttons children of a wrapper

Since the main card element is no longer an anchor, this is no longer necessary.

frontends/ol-components/src/components/Card/Card.tsx
```diff
@@ -196,54 +267,64 @@ const Card: Card = ({ children, className, size, href }) => {
...
+      {...others}
+      className={allClassNames}
+      size={size}
+      onClick={handleClick}
+    >
+      {image && (
+        // alt text will be checked on Card.Image
+        // eslint-disable-next-line styled-components-a11y/alt-text
+        <Image
+          className="MitCard-image"
```
jonkafton, line 304: Not one for this PR as this hasn't changed here, but can we use `data-testid`s if these are just for selection in the tests (or ideally avoid altogether). `.MitCard-root` and `.MitCard-info` are the only ones in use for CSS selectors.

jonkafton (approved):
Looks good. Just the one comment with repeated content for the small resources cards, though not sure if this is controlloable or a VoiceOver artifact as the HTML looks the same.

frontends/ol-components/src/components/Card/Card.tsx
ChristopherChudzicki: We do use `MitCard-root` and `MitCard-info` for styling at least once, which is why they were added. When I added them, I added for other parts (like `MitCard-title`) that made sense but weren't actually used (as of yet).

PR #1780, opened 2024-10-31
Clear resource_type filter when leaving Learning Materials search tab

mbertrand:
Closes https://github.com/mitodl/hq/issues/5333

If you are on the "Learning Materials" tab of the search page with resource_type filters selected ("Video" for example), then switch to another tab, the `resource_type` filter should be cleared,  but not other filters like topic.

### How can this be tested?
- Go to http://open.odl.local:8062/search?resource_category=learning_material&resource_type=video&topic=Biology, the "Video" option should be checked for "Resource Type" facet and "Biology" for the Topic facet.  Course and Program tab headers should show counts of 0.  The "All" tab header should show the same count as the current tab.
- Switch to the "Courses" tab.  You should get results back with a count > 0, the biology topic should still be selected, and there should be no `resource_type` parameter in the url : http://open.odl.local:8062/search?topic=Biology&resource_category=course
- Repeat the previous 2 steps, but this time switch to the "All" tab instead of the "Course" tab.  The behavior should be the same, with a url of http://open.odl.local:8062/search?topic=Biology

### DIFF
--- a/frontends/main/src/app-pages/SearchPage/SearchPage.test.tsx
+++ b/frontends/main/src/app-pages/SearchPage/SearchPage.test.tsx
@@ -491,6 +491,32 @@ describe("Search Page Tabs", () => {
     expect(params2.get("department")).toBe("8") // should preserve other params
   })
 
+  test("Switching from learning materials tab clears resource type only", async () => {
+    setMockApiResponses({
+      search: {
+        count: 1000,
+        metadata: {
+          aggregations: {
+            resource_type: [{ key: "video", doc_count: 100 }],
+          },
+          suggestions: [],
+        },
+      },
+    })
+    const { location } = renderWithProviders(<SearchPage />, {
+      url: "?resource_category=learning_material&resource_type=video&topic=Biology",
+    })
+    const tabLM = screen.getByRole("tab", { name: /Learning Materials/ })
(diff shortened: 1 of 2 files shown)


abeglova (changes requested):
This works great. Should the counts for the tabs also ignore the resource_type filter setting? That way the count will be the number of resources the user will see if they click on the tab:

If so, you should replace the following lines with
https://github.com/mitodl/mit-learn/blob/main/learning_resources_search/api.py#L489-L491 with:

```
            ignored_keys = ["resource_type", 'resource_category'] if aggregation == "resource_category " else [aggregation]
            other_filters = [
                filter_clauses[key] for key in filter_clauses if not key in ignored_keys
            ]
```

mbertrand:
@abeglova thanks, I tried that suggestion but when I select a resource type facet like "Podcast", the course/program tab counts remain the same (which is good) but so does the learning material tab count (not good?):

podcast count: 22, tab count 1154

Screenshot 2024-11-01 154936

abeglova:
Oh yeah - you are right . So maybe what you have is better

abeglova (approved):

abeglova:
I guess we can make two seperate aggregations and combine them but maybe that's also confusing. What you have is probably the least bad solution

PR #1784, opened 2024-11-01
Shanbady qdrant integration for vector based similar resources endpoint

shanbady:

Closes https://github.com/mitodl/hq/issues/5865

This PR introduces the following
- a new endpoint to show similar resources using vector embeddings via [QDrant](https://qdrant.tech/)
- a "generate_embeddings" management command to embed learning resources and contentfiles

### How can this be tested?
1. Checkout this branch
2. You will need to rebuild your images to install dependencies. (`docker compose down` `docker compose build` )
3. load up learning resources via management command
4. run the embedding management command `./manage.py generate_embeddings --all --skip-content-files` (skip contentfiles for now - more on this in the additional context section)
5. once embeddings have been generated, visit the [qdrant dashboard](http://localhost:6333/dashboard) and see that the resources are populated.
6. find a course id and visit the new similar resources endpoint http://open.odl.local:8063/api/v1/learning_resources/**<<resource_id>>**/vector_similar/ and validate the results

### Additional Context
1. The results do not currently at contentfiles (just learning resources). Contentfiles are also stored in a separate collection (unlike in opensearch where they sit with the index of the associated learning resource). Once this goes out and we see this works get a sense of performance, it should be easy to follow up with the addition of contentfile data . Either keep it in a separate collection and make the relation to its parent resource in code or embed learning resources with contentfile vectors in a single collection.
2. The "generate_embeddings" management command is currently not based into the celery crontab - Once this is deployed and manually validated, I'll follow up with a pr that adds this into the celery schedule.
3. Contentfiles (atleast locally) take a long time but i dont think much longer than opensearch indexing. 
4. I was running into some uwsgi exception locally which appears to be resolved by setting the harakiri timeout in uwsgi.ini. Not sure if we expect this to happen when deployed but we can set it there if we need do.

### DIFF
--- a/learning_resources/views.py
+++ b/learning_resources/views.py
@@ -225,7 +225,48 @@ class LearningResourceViewSet(
             id=pk
         )
         resource_data = serialize_learning_resource_for_update(learning_resource)
-        similar = get_similar_resources(resource_data, limit, 2, 3)
+        similar = get_similar_resources(
+            resource_data, limit, 2, 3, use_embeddings=False
+        )
+        return Response(LearningResourceSerializer(list(similar), many=True).data)
+
+    @extend_schema(
+        summary="Get similar resources using vector embeddings",
+        parameters=[
+            OpenApiParameter(name="id", type=int, location=OpenApiParameter.PATH),
+            OpenApiParameter(name="limit", type=int, location=OpenApiParameter.QUERY),
+        ],
+        responses=LearningResourceSerializer(many=True),
+    )
+    @action(
+        detail=True,
+        methods=["GET"],
+        name="Fetch similar resources using embeddings for a resource",
+    )
+    @method_decorator(
+        cache_page_for_all_users(
+            settings.SEARCH_PAGE_CACHE_DURATION, cache="redis", key_prefix="search"
+        )
+    )
+    def vector_similar(self, request, *_, **kwargs):
+        """
+        Fetch similar learning resources
+
+        Args:
+        id (integer): The id of the learning resource
+
+        Returns:
+        QuerySet of similar LearningResource for the resource matching the id parameter
+        """
+        limit = request.GET.get("limit", 10)
+        pk = int(kwargs.get("id"))
+
+        learning_resource = get_object_or_404(LearningResource, id=pk)
+        learning_resource = LearningResource.objects.for_search_serialization().get(
+            id=pk
+        )
+        resource_data = serialize_learning_resource_for_update(learning_resource)
+        similar = get_similar_resources(resource_data, limit, 2, 3, use_embeddings=True)
         return Response(LearningResourceSerializer(list(similar), many=True).data)
 
 
--- a/learning_resources_search/api.py
+++ b/learning_resources_search/api.py
@@ -906,22 +906,19 @@ def get_similar_resources(
 
 
 def _qdrant_similar_results(doc, num_resources):
-    from sentence_transformers import SentenceTransformer
-
     from learning_resources_search.indexing_api import qdrant_client
 
     client = qdrant_client()
-    encoder = SentenceTransformer("all-MiniLM-L6-v2")
     return [
-        hit.payload
-        for hit in client.query_points(
-            collection_name="resource_embeddings.resources",
-            query=encoder.encode(
+        hit.metadata
+        for hit in client.query(
+            collection_name=f"{settings.QDRANT_BASE_COLLECTION_NAME}.resources",
+            query_text=(
                 f'{doc.get("title")} {doc.get("description")} '
                 f'{doc.get("full_description")} {doc.get("content")}'
-            ).tolist(),
+            ),
             limit=num_resources,
-        ).points
+        )
     ]
 
 
--- a/learning_resources_search/api.py
+++ b/learning_resources_search/api.py
@@ -873,7 +873,11 @@ def get_similar_topics(
 
 
 def get_similar_resources(
-    value_doc: dict, num_resources: int, min_term_freq: int, min_doc_freq: int
+    value_doc: dict,
+    num_resources: int,
+    min_term_freq: int,
+    min_doc_freq: int,
+    use_embeddings,
 ) -> list[str]:
     """
     Get a list of similar resources based on another resource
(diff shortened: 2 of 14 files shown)


learning_resources_search/api.py
```diff
@@ -887,11 +891,81 @@ def get_similar_resources(
...
+            a document representing the data fields we want to search with
+        num_topics (int):
+            number of resources to return
+
+    Returns:
+        list of str:
+            list of learning resources
+    """
+    hits = _qdrant_similar_results(value_doc, num_resources)
+    return LearningResource.objects.for_search_serialization().filter(
```
mbertrand, line 940: This is resulting in numerous n+1 queries. Adjusting the `for_search_serialization` function to this seems to get rid of them:

```python
    def for_search_serialization(self):
        return self.for_serialization().annotate(in_featured_lists=Count("parents__parent__channel"))
```

learning_resources/views.py
```diff
@@ -225,7 +225,48 @@ def similar(self, request, *_, **kwargs):
...
+        Args:
+        id (integer): The id of the learning resource
+
+        Returns:
+        QuerySet of similar LearningResource for the resource matching the id parameter
+        """
+        limit = request.GET.get("limit", 10)
+        pk = int(kwargs.get("id"))
+
+        learning_resource = get_object_or_404(LearningResource, id=pk)
```
mbertrand, line 264: This variable is immediately reassigned in the next line.  Maybe could do something like this instead:

```python
        try:
            learning_resource = LearningResource.objects.for_search_serialization().get(
                id=pk
            )
        except LearningResource.DoesNotExist:
            raise Http404("No LearningResource matches the given query.")
```

mbertrand (changes requested):
Results look good.  However the response time for me using vector search was really slow, about 7-8 seconds per request, compared to 1-1.5 seconds for the regular similar endpoint.

Also noticed it may take awhile to regenerate embeddings for all resources.  I have ~10K and I aborted the mgmt command after 30 minutes.  Generating embeddings for just programs/courses (545) took ~100 seconds.

learning_resources/views.py
shanbady: good catch!

shanbady:
@mbertrand this is ready for another look.

I followed the "increase throughput" section from [qdrants performance guide](https://qdrant.tech/documentation/guides/optimize/) and I got it to index 8148 resources in 702 seconds by reducing the number of segments to just 2.

learning_resources_search/tasks.py
```diff
@@ -540,6 +540,35 @@ def wrap_retry_exception(*exception_classes):
...
+    """
+    try:
+        with wrap_retry_exception(*SEARCH_CONN_EXCEPTIONS):
+            api.embed_learning_resources(ids, resource_type)
+    except (RetryError, Ignore):
+        raise
+    except SystemExit as err:
+        raise RetryError(SystemExit.__name__) from err
+    except:  # noqa: E722
+        error = "index_courses threw an error"
```
mbertrand, line 567:
```python
error = "generate_embeddings threw an error"
```

learning_resources_search/tasks.py
```diff
@@ -660,6 +689,81 @@ def start_recreate_index(self, indexes, remove_existing_reindexing_tags):
     )
 
 
+@app.task(bind=True)
+def start_embed_resources(self, indexes, skip_content_files):
+    index_tasks = []
```
mbertrand, line 694: Might be worth checking here to see if the required `QDRANT_*` settings have values, and if not return immediately.

mbertrand (commented):
Just a couple more suggestions, looking great.  The `vector_similar` endpoint is even slower for me now, taking 20+ seconds and sometimes timing out, returning a 502 error.  Do you expect a hosted version of qdrant to be significantly faster?

learning_resources_search/tasks.py
```diff
@@ -660,6 +689,81 @@ def start_recreate_index(self, indexes, remove_existing_reindexing_tags):
     )
 
 
+@app.task(bind=True)
+def start_embed_resources(self, indexes, skip_content_files):
```
mbertrand, line 693: Add a unit test?

uwsgi.ini
```diff
@@ -18,6 +18,7 @@ offload-threads = 2
 thunder-lock =
 if-env = DEV_ENV
 python-autoreload = 1
+harakiri = 30
```
mbertrand, line 21: Might be worth checking with Nathan to see if this change is good for production.

shanbady:
@mbertrand I'm going to try something locally and see if I can resolve that issue (which I think could happen once deployed as well) - I think it might be due to the segment size optimization but I will have to confirm.

learning_resources_search/tasks.py
shanbady: Added a check for this

learning_resources_search/tasks.py
shanbady: added test for the task and for exiting out of embedding without the settings specified

learning_resources_search/tasks.py
shanbady: good call. fixed

shanbady:
@mbertrand I had modified the code to use a hybrid search approach as opposed to the default which is to use fast-embed (requires far less resources). Using fast embed is probably the right thing for now. After reverting to fast embed indexing ~8k resources took only 248 seconds and the results loaded much faster too (without the gateway errors)

learning_resources_search/api.py
```diff
@@ -887,11 +891,81 @@ def get_similar_resources(
...
     """
+    if use_embeddings:
+        return get_similar_resources_qdrant(value_doc, num_resources)
+    return get_similar_resources_opensearch(
+        value_doc, num_resources, min_term_freq, min_doc_freq
+    )
+
+
+def _qdrant_similar_results(doc, num_resources):
+    from learning_resources_search.indexing_api import qdrant_client
```
mbertrand, line 909: add docstring

mbertrand (approved):
LGTM, response time is much better than before though still takes 5-8 seconds, maybe it will be faster on a more powerful qdrant instance.  Just one missing docstring to add in.

PR #1792, opened 2024-11-05
upgrade stories & events grid to grid2

ChristopherChudzicki:
Closes https://github.com/mitodl/hq/issues/6007

This PR fixes a bug that was introduced when MUI was updated from v5 to v6 as part of #1776.

MUI v5 had `Grid` and `Unstable_Grid2` components. In v6, Grid2 was stabilized Grid was deprecated. Additionally, v6 seems to have introduced a bug in `Grid`. The bug only affects Grid usage that has string-values for `columnSpacing`. We had exactly one instance of string-valued `columnSpacing`. So I've upgraded that particular usage to `Grid2` and everything looks fine now.

See https://github.com/mui/material-ui/issues/44321 for MUI bug.

### How can this be tested?
1. On this branch, compare styles of Stories & Events section of the homepage with that on https://learn.mit.edu/. (The bug has not yet made it to production). When comparing styles, I recommend:
    - Put local branch in browser Tab 1, production in Tab 2
    - Ensure zoom level is same in both tabs
    - Ensure both tabs have same widths (e.g., close dev tools on both tabs)
    - swap between Tab 1 and Tab 2.
2. Suggest doing additional style comparison of other pages, e.g. `/departments`, `/topics`, `/units`, `/search`, `/dashboard` + its tabs, and channel pages.
    - when comparing, ensure both tabs have scrollbar or both tabs do not have scrollbar. E.g., if viewing `/dashboard/my-lists`, if one tab has 1 list and the other has 10 lists, scrollbar might affect positioning in the longer list.

### DIFF
--- a/frontends/main/src/app-pages/HomePage/NewsEventsSection.tsx
+++ b/frontends/main/src/app-pages/HomePage/NewsEventsSection.tsx
@@ -4,7 +4,7 @@ import {
   styled,
   theme,
   Typography,
-  Grid,
+  Grid2,
   Card,
   TypographyProps,
 } from "ol-components"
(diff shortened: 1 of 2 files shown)


jonkafton (approved):
Looks good / fixed. Instances of Grid v1 also good.

PR #1793, opened 2024-11-05
Card bug fixes

ChristopherChudzicki:
- Follow up to #1778
- Fixes  https://github.com/mitodl/hq/issues/6001

See comments, but roughly:
- Alters how cards' `forwardClicksToLink` target is found/marked
- Fixes card links on `/units` page
- Makes whole card clickable on `/dashboard/my-lists`

### How can this be tested?
1. Cards on `/units` page should be clickable (whole card)
    - Additionally, cards on the unit page should show the same description on desktop vs mobile (See https://github.com/mitodl/hq/issues/6001)
2. Whole card (not just title) on `/dashbaord/my-lists` should be clickable.
    - Check re-ordering these, too. 
3. Other cards should work as before: whole card clickable. (Carousels, search results, Stories + Events)

### DIFF
--- a/frontends/main/src/app-pages/UnitsListingPage/UnitsListingPage.test.tsx
+++ b/frontends/main/src/app-pages/UnitsListingPage/UnitsListingPage.test.tsx
@@ -1,29 +1,8 @@
 import React from "react"
 import { renderWithProviders, screen, waitFor, within } from "@/test-utils"
-import type { LearningResourcesSearchResponse } from "api"
 import UnitsListingPage from "./UnitsListingPage"
 import { factories, setMockResponse, urls } from "api/test-utils"
 import { assertHeadings } from "ol-test-utilities"
-
-const makeSearchResponse = (
-  aggregations: Record<string, number>,
-): LearningResourcesSearchResponse => {
-  return {
-    metadata: {
-      suggestions: [],
-      aggregations: {
-        topic: Object.entries(aggregations).map(([key, docCount]) => ({
-          key,
-          doc_count: docCount,
-        })),
-      },
-    },
-    count: 0,
-    results: [],
-    next: null,
-    previous: null,
-  }
-}
 
 describe("DepartmentListingPage", () => {
   const setupApis = () => {
@@ -59,12 +38,6 @@ describe("DepartmentListingPage", () => {
       value_prop: "Professional Unit 2 value prop",
       professional: true,
     })
-    const professionalUnit3 = make.offeror({
-      code: "professionalUnit3",
-      name: "Professional Unit 3",
-      value_prop: "Professional Unit 3 value prop",
-      professional: true,
-    })
 
     const units = [
       academicUnit1,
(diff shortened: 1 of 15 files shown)


frontends/ol-components/src/components/Card/Card.tsx
```diff
@@ -180,27 +192,22 @@ export const useClickChildHref = (
...
-  /**
-   * If provided, the card will render its title as a link.
-   *
-   * Clicks on the entire card can be forwarded to the link via `forwardClicksToLink`.
-   */
-  href?: string
   /**
    * Defaults to `false`. If `true`, clicking the whole card will click the
-   * href link as well.
+   * child anchor with data-card-link="true".
```
ChristopherChudzicki, line 205: Previously, if `forwardClicksToLink` was true, clicking the whole card would click the child anchor with the card's `href`.

This turned out to be a bit fragile with NextJS. For example, on the `/units` listing page, the backend currently returns `channel_url` values like `http://localhost:8062/c/unit/mitx/`. But NextJS standardizes the href to NOT have a trailing slashing: `http://localhost:8062/c/unit/mitx`. 

We could try to trim the trailing slash, but that seems complicated for things like `/some/path/?foo=bar` vs `/some/path?foo=bar`. So instead, let's just find the relevant anchor by explicitly marking it with `data-card-link`.

Consequently, the `Card` no longer needs the href, just the `Card.Title`.

frontends/main/src/app-pages/UnitsListingPage/UnitsListingPage.test.tsx
```diff
@@ -138,49 +99,43 @@ describe("DepartmentListingPage", () => {
...
+
     renderWithProviders(<UnitsListingPage />)
+
+    const academicSection = screen.getByTestId("UnitSection-academic")
+    const professionalSection = screen.getByTestId("UnitSection-professional")
+
     await waitFor(() => {
-      const academicSection = screen.getByTestId("UnitSection-academic")
-      const professionalSection = screen.getByTestId("UnitSection-professional")
-      units.forEach(async (unit) => {
```
ChristopherChudzicki, line 145: This PR is fixing a bug where `/units` page links were broken. This test should have caught that, except it wasn't actually doing anything because `forEach` can't be awaited (it always returns `undefined`) So the test effectively ended on line 144.

frontends/ol-components/src/components/Card/Card.tsx
gumaerc: Something I'm noticing is that previously, the unit cards were wrapped in an `a` tag when this was rendered in the browser. Now, following the link seems to be handled by Javascript. It doesn't make logical sense to me either that an `href` property would be set on the title, as the whole card is supposed to be clickable?

gumaerc (approved):
I'm okay with approving this based on our discussion of why the link forwarding functionality needs to work the way it does. For historical context we discussed how wrapping entire cards in an `a` tag causes all of the text inside of that card to become the link's `textContent` property, which makes screen readers read everything inside the card. In order to prevent this from happening, the actual anchor tag is being placed on the title, or on a child element you manually specify using the `data-card-link` property and then if you set `forwardClicksToLink` on the parent `Card` element, clicks on the whole card will go to the nested `a` tag. We discussed how I think this could be potentially somewhat confusing for new developers, but ultimately decided it was the best solution. We did discuss doing whatever is possible to indicate to developers how this works though, including docstings for Intellisense.

PR #1799, opened 2024-11-06
identify users in posthog via user id

ChristopherChudzicki:
https://github.com/mitodl/hq/issues/5977

This PR enables identifying users in posthog via MIT Learn user ids. This lets us use, e.g., specific feature flags.

### How can this be tested?
Here's one way you could test this.
1. Connect a personal PostHog account to MIT Learn by setting `POSTHOG_PROJECT_API_KEY` in `env/shared.local.env`
2. Create two users in your instance of Learn
    - *I find it convenient to use separate Chrome profiles or separate browsers for different users.*
3. In PostHog, create a feature flag named `lr_drawer_v2`. Set it to apply to 100% of users with a specific id. Something like:
    Screenshot 2024-11-06 at 12 25 52 PM
4. Check that:
    - When logged in as that user, you see the new learning resource drawer (wider drawer—same as RC)
    - Now log out, still in same browser window. You should see old drawer (narrower drawer—same as Prod)
    - When logged in as a different user, you should see old drawer (narrower drawer, same as Prod)


jkachel (approved):
Tested and started seeing events come through with my user ID.

LGTM 👍

PR #1801, opened 2024-11-07
Remove unnecessary percolate calls

shanbady:
This PR removes calling percolate on newly inserted documents.During the initial implementation of the subscription feature we had a plugin hook that percolated new documents on creation which is currently unused (we switched to a celery task that manually does this when needed). I think the plugin hook and parameter percolate=False is still a useful placeholder for when we do need something to trigger off of this down the road.

### How can this be tested?
I noticed that there was a stream of percolate operations when pulling down lots of new items via etl pipelines. 

To see this issue:
1. checkout main
2. make sure you have your settings in place (copy it from rc) to fetch content from youtube (YOUTUBE_DEVELOPER_KEY, YOUTUBE_CONFIG_URL) 
3. run `python manage.py backpopulate_youtube_data fetch`
4. sometime after this completes notice a stream of percolate operations in the celery logs that follow
5. check out this branch and restart containers and note that this does not happen

### DIFF
--- a/learning_resources/etl/loaders.py
+++ b/learning_resources/etl/loaders.py
@@ -73,7 +73,7 @@ def update_index(learning_resource, newly_created):
     if not newly_created and not learning_resource.published:
         resource_unpublished_actions(learning_resource)
     elif learning_resource.published:
-        resource_upserted_actions(learning_resource, percolate=newly_created)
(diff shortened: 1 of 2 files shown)


gumaerc (approved):
👍

PR #1805, opened 2024-11-07
Make Heroku API deployments conditional

rhysyngsun:

N/A

We're having issues getting backend API deployments to heroku to work and I suspect the github action for that isn't behaving correct. I'm adding some conditionals here to only deploy the code if an environment variable is set to `true`

### How can this be tested?

We'll have to test this by running a release candidate.


shanbady (approved):
not sure how to test locally but changes look ok.

ChristopherChudzicki:
Closing this since we're off heroku now.

PR #1808, opened 2024-11-08
Endpoints for userlist/learningpath memberships

mbertrand:
Closes https://github.com/mitodl/hq/issues/5159

- Creates an endpoint to retrieve all learning path relationships
- Creates an endpoint to retrieve all user list relationships

### How can this be tested?
- Log in as an admin or a learning path editor and go to http://api.open.odl.local:8063/api/v1/learningpaths/membership/
- You should get a list of all the relationships (id, parent, child attributes) from all the learning paths. 
- Create a couple user lists if you haven't already.
- Go to http://api.open.odl.local:8063/api/v1/userlists/membership/
- You should get a list of all the relationships from all of your userlists
- Log in as another user, with no or different user lists, and who is not an admin or learning path editor, then go back to http://api.open.odl.local:8063/api/v1/userlists/membership/
- You should get a list of all the relationships (if any) from all of this user's userlists
- Go to http://api.open.odl.local:8063/api/v1/learningpaths/membership/, you should get a 403
- Log out
- Go back to both urls, you should get a 403


gumaerc (approved):
👍 LGTM!

PR #1809, opened 2024-11-08
show more button for v2 drawer dates

gumaerc:
Closes https://github.com/mitodl/hq/issues/5874

This PR adds a "show more" link to the start dates section of the info displayed in the v2 learning resource drawer. This expands and collapses the start dates for the course.

image
image
image
image

### How can this be tested?
 - Ensure you have sufficient data backpopulated into your local instance, especially Sloan courses
 - Search for a course with identical data but with multiple start dates, like "Artificial Intelligence in Pharma and Biotech" for example
 - Confirm that you can click "Show more" to expand the rest of the start dates, that the link text changes to "Show less" after it expands, and that you can click "Show less" to collapse back down
 - Search for a course with different data and multiple start dates, such as "Leading the AI-Driven Organization"
 - Confirm that all of the start dates are shown without the show more button

### DIFF
--- a/frontends/ol-components/src/components/LearningResourceExpanded/InfoSectionV2.tsx
+++ b/frontends/ol-components/src/components/LearningResourceExpanded/InfoSectionV2.tsx
@@ -89,7 +89,7 @@ const InfoValue = styled.div({
   ...theme.typography.body3,
 })
 
-const DateWrapper = styled.span({
+const NoWrap = styled.span({
   whiteSpace: "nowrap",
 })
 
@@ -170,38 +170,40 @@ const RunDates: React.FC<{ resource: LearningResource }> = ({ resource }) => {
   const showMore = allRunsAreIdentical(resource) && sortedDates.length > 2
   if (showMore) {
     const showMoreLink = (
-      <ShowMoreLink
-        color="red"
-        size="small"
-        onClick={() => setShowingMore(!showingMore)}
-      >
-        {showingMore ? "Show less" : "Show more"}
-      </ShowMoreLink>
+      <NoWrap>
+        <ShowMoreLink
+          color="red"
+          size="small"
+          onClick={() => setShowingMore(!showingMore)}
+        >
+          {showingMore ? "Show less" : "Show more"}
+        </ShowMoreLink>
+      </NoWrap>
     )
     return (
       <span data-testid="drawer-run-dates">
         {sortedDates.slice(0, 2).map((runDate, index) => {
           return (
-            <DateWrapper key={`run-${index}`}>
+            <NoWrap key={`run-${index}`}>
               <InfoItemValue
                 label={runDate}
                 index={index}
                 total={showingMore ? 3 : 2}
               />
-            </DateWrapper>
+            </NoWrap>
           )
         })}
         {!showingMore && showMoreLink}
         {showingMore &&
           sortedDates.slice(2).map((runDate, index) => {
             return (
-              <DateWrapper key={`run-${index + 2}`}>
+              <NoWrap key={`run-${index + 2}`}>
                 <InfoItemValue
                   label={runDate}
                   index={index}
                   total={sortedDates.length - 2}
                 />
-              </DateWrapper>
+              </NoWrap>
             )
           })}
         {showingMore && showMoreLink}
@@ -211,13 +213,13 @@ const RunDates: React.FC<{ resource: LearningResource }> = ({ resource }) => {
     const runDates =
       sortedDates.map((runDate, index) => {
         return (
-          <DateWrapper key={`run-${index}`}>
+          <NoWrap key={`run-${index}`}>
             <InfoItemValue
               label={runDate}
               index={index}
               total={sortedDates.length}
             />
-          </DateWrapper>
+          </NoWrap>
         )
       }) ?? []
     return <span data-testid="drawer-run-dates">{runDates}</span>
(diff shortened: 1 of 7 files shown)


frontends/ol-components/src/components/LearningResourceExpanded/InfoSectionV2.tsx
```diff
@@ -145,6 +155,77 @@ const InfoItemValue: React.FC<InfoItemValueProps> = ({
...
+        return (
+          <NoWrap key={`run-${index}`}>
+            <InfoItemValue
+              label={runDate}
+              index={index}
+              total={sortedDates.length}
+            />
+          </NoWrap>
+        )
+      }) ?? []
```
jonkafton, line 224: Left hand side will never be null or undefined, so we can remove `?? []`, although it does appear `sortedDates` can possibly be undefined from L163.

frontends/ol-utilities/src/learning-resources/learning-resources.ts
```diff
@@ -116,12 +120,59 @@ const formatRunDate = (
...
+  if (resource.runs.length <= 1) {
+    return true
+  }
+  const prices: LearningResourcePrice[] = []
+  const deliveryMethods = []
+  const locations = []
+  for (const run of resource.runs) {
+    if (run.resource_prices) {
+      run.resource_prices.forEach((price) => {
+        if (price.amount !== "0") {
```
jonkafton, line 142: Is it intentional that several resource runs all with `"0"` price are considered different?

frontends/ol-utilities/src/learning-resources/learning-resources.ts
```diff
@@ -116,12 +120,59 @@ const formatRunDate = (
...
+      })
+    }
+    if (run.delivery) {
+      deliveryMethods.push(run.delivery)
+    }
+    if (run.location) {
+      locations.push(run.location)
+    }
+  }
+  const distinctPrices = [...new Set(prices.map((p) => p.amount).flat())]
```
jonkafton, line 154: These work fine, but shorter to use Set throughout instead of converting between arrays, ie.

```
136:    const prices = new Set<LearningResourcePrice>()

144:    run.resource_prices?.forEach(price => {
            if (price.amount !== "0") prices.add(price.amount);
        });

161:    prices.size === 1

File: frontends/ol-components/src/components/LearningResourceExpanded/InfoSectionV2.tsx [hunk 4]
```diff
@@ -145,6 +155,77 @@ const InfoItemValue: React.FC<InfoItemValueProps> = ({
...
+  const sortedDates =
+    resource.runs
+      ?.sort((a, b) => {
+        if (a?.start_date && b?.start_date) {
+          return Date.parse(a.start_date) - Date.parse(b.start_date)
+        }
+        return 0
+      })
+      .map((run) => formatRunDate(run, asTaughtIn)) ?? []
+  const showMore = allRunsAreIdentical(resource) && sortedDates.length > 2
```
jonkafton, line 170: Just to check - it's possible that a single run date ("Start Date:" info item) can be displayed in addition to the differing runs table (several runs, same date, dates repeated in differing runs table and here)?

jonkafton (commented):
A couple of questions that may or may not lead to changes - looks good.

frontends/ol-utilities/src/learning-resources/learning-resources.ts
gumaerc: I tweaked this logic a bit. The idea here is that a free course that offers a certificate is always going to have the "0" price along with one other, so the course doesn't have just "one" price. I could just check `resource.free` in the conditional at the bottom, but instead I'm checking it here and just ignoring the zero price if the resource is free. That way, if there are any different prices for the certificate they will be the only ones in the set.

frontends/ol-components/src/components/LearningResourceExpanded/InfoSectionV2.tsx
gumaerc: I checked with @sovsey on this, and I had a misinterpretation here. In https://github.com/mitodl/hq/issues/5681, it says "Only courses that have multiple runs with DIFFERENT data display data in the table format" but doesn't say anything about hiding the start dates info item when that is the case. I've changed up the logic here so that the start dates info item is *not* shown if there are unique runs.

gumaerc:
@jonkafton This is ready for another look. I also noticed another issue with the CTA image and couldn't help myself but to fix it. Remember how you discovered that issue where it was floating *above* the sticky header when scrolling, and I put in a fix to move it below with `z-index: -1`? Well, that's still necessary, but I noticed that although the issue had been fixed visually I was unable to click the X to close the drawer when the image was sitting under it, because the `NextImage` is `position: absolute` by default. The existing code was scaling the image using the `aspect` property by calculating a padding offset. Since `aspect-ratio` is usable in all browsers, I fixed it by applying this style to `NextImage`:

```ts
const Image = styled(NextImage)<{ aspect: number }>`
  position: relative !important;
  border-radius: 8px;
  width: 100%;
  aspect-ratio: ${({ aspect }) => aspect};
  object-fit: cover;
  z-index: -1;
`
```

jonkafton:
> Since aspect-ratio is usable in all browsers, I fixed it by applying this style to NextImage

Better, plus we use elsewhere. The % padding does feel like a trick.

jonkafton (approved):

PR #1821, opened 2024-11-12
Add data-ph- elements to CTA buttons

jkachel:

https://github.com/mitodl/hq/issues/5989

We need a clear way to identify when users click on the CTA button in the Learning Resource drawer - this adds that, in the form of `data-ph-` attributes. 

The attributes that are added are:
- `data-ph-action` - set to `click-cta` (this is the standard PostHog attribute we decided on)
- `data-ph-offered-by` - set to the offeror code
- `data-ph-resource-type` - set to the resource type
- `data-ph-resource-id` - set to the resource ID

These are added to both the V1 and V2 LearningResourceExpanded components since the v2 one isn't rolled out yet.

### How can this be tested?

This does not handle video resources - videos will require more work (due to the CTA being the embedded video itself). So, choose resources that aren't videos. These will be handled in a future PR.

The easiest way to test this is to create an Action in PostHog's Data Management tab. Set up an Action that has one Match Group. The Match Group should be set to Autocapture and the "Element matches HTML selector" should be set to `[data-ph-action="click-cta"]`. 

Then, load some resources and click the CTA buttons in the drawers. You should see events match in the Action after some delay.

### DIFF
--- a/frontends/ol-components/src/components/LearningResourceExpanded/LearningResourceExpandedV2.tsx
+++ b/frontends/ol-components/src/components/LearningResourceExpanded/LearningResourceExpandedV2.tsx
@@ -285,7 +285,7 @@ const getCallToActionText = (resource: LearningResource): string => {
   const accessCourseMaterials = "Access Course Materials"
   const watchOnYouTube = "Watch on YouTube"
   const listenToPodcast = "Listen to Podcast"
-  const learnMore = "Learn More About"
+  const learnMore = "Learn More"
   const callsToAction = {
     [ResourceTypeEnum.Course]: learnMore,
     [ResourceTypeEnum.Program]: learnMore,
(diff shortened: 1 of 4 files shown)


frontends/ol-components/src/components/LearningResourceExpanded/LearningResourceExpandedV2.tsx
```diff
@@ -285,7 +285,7 @@ const getCallToActionText = (resource: LearningResource): string => {
   const accessCourseMaterials = "Access Course Materials"
   const watchOnYouTube = "Watch on YouTube"
   const listenToPodcast = "Listen to Podcast"
-  const learnMore = "Learn More"
+  const learnMore = "Learn More About"
```
shanbady, line 288: I'm assuming V2 isnt displayed anywhere yet. was this copy change intentional?

shanbady (changes requested):

frontends/ol-components/src/components/LearningResourceExpanded/LearningResourceExpandedV2.tsx
jkachel: it was not, fixed in https://github.com/mitodl/mit-learn/pull/1821/commits/bb7bd021c24482556056f1238d6480fecc166db0

shanbady (approved):

PR #1828, opened 2024-11-15
separate starts and as taught in, show anytime availability

gumaerc:
Closes https://github.com/mitodl/hq/issues/6075

This PR revises the logic surrounding dates. Previously, we were showing *either* the "Starts" or "As taught in" fields, based on the output of `shouldShowAnytime`. This function takes a `LearningResource` and returns `true` if its `availability` property is set to "anytime" and its `resource_type` is "course" or "program." Now, if `availability` is "anytime" then the "Starts" field will show "Anytime," otherwise it will show the start dates. The "As taught in" field also appears if `availability` is "anytime," but if no dates are returned from `totalRunsWithDates`, then it will be hidden.

image
image

### How can this be tested?
 - Ensure you have sufficient data backpopulated into your local instance, including some OCW and Sloan courses
 - Go to the search page at http://localhost:8062/search and filter on OCW courses
 - Browse through the courses, clicking them to bring up the drawer
 - All OCW courses should have "Starts: Anytime" and if there is a year and semester available for the course, it should show under "As taught in"
 - Look for a Sloan course with "anytime" `availability`, such as "Generative AI Business Sprint"
 - Ensure that "Starts: Anytime" is shown, but "As taught in" does not

### DIFF
--- a/frontends/ol-components/src/components/LearningResourceCard/testUtils.ts
+++ b/frontends/ol-components/src/components/LearningResourceCard/testUtils.ts
@@ -105,7 +105,12 @@ const courses = {
     }),
     anytime: makeResource({
       resource_type: ResourceTypeEnum.Course,
-      runs: [factories.learningResources.run()],
+      runs: [
+        factories.learningResources.run({
+          year: 2022,
+          semester: "Spring",
+        }),
+      ],
       free: true,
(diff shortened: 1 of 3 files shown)


shanbady (approved):
LGTM

PR #1831, opened 2024-11-18
LocalDate and NoSSR components to render localized dates only on client

jonkafton:

Fixes https://github.com/mitodl/hq/issues/6083

Provides a \<NoSSR\> utility component that renders children only after it has first rendered (ie. always on the client).

Provides \<LocalDate\> component that for displaying dates adjusted to the user's timezone with the \<NoSSR\> wrapper, that may otherwise cause a hydration error due a mismatch between the server and client rendered display dates.

Wraps all uses of `moment().format()` with \<LocalDate\> or \<NoSSR\>.

Sets `TZ=UTC` on the Next.js server environment to that hydration errors will be triggered during development (for anyone not on GMT+0).

### How can this be tested?

Find a course that has a start time between midnight and midnight minus your UTC offset (ie. 5am in EST), e.g. "Professional Certificate in Data Engineering".

Ensure that the start date renders correctly without any hydration error in the console. The above course has a start date of 2024-11-27T00:00:00Z, so should display "November 26, 2024" in EST, though would otherwise render on the UTC server as "November 27, 2024", causing the hydration error.

### DIFF
--- a/frontends/ol-utilities/src/ssr/NoSSR.tsx
+++ b/frontends/ol-utilities/src/ssr/NoSSR.tsx
@@ -5,12 +5,12 @@ type NoSSRProps = {
   onSSR?: ReactNode
 }
 
-export const NoSSR: React.FC<NoSSRProps> = ({ children, onSSR = <></> }) => {
+export const NoSSR: React.FC<NoSSRProps> = ({ children, onSSR = null }) => {
   const [isClient, setClient] = useState(false)
 
   useEffect(() => {
     setClient(true)
   }, [])
 
-  return <>{isClient ? children : onSSR}</>
+  return isClient ? children : onSSR
 }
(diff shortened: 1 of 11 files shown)


frontends/ol-utilities/src/ssr/NoSSR.tsx
```diff
@@ -0,0 +1,16 @@
...
+}
+
+export const NoSSR: React.FC<NoSSRProps> = ({ children, onSSR = <></> }) => {
+  const [isClient, setClient] = useState(false)
+
+  useEffect(() => {
+    setClient(true)
+  }, [])
+
+  return <>{isClient ? children : onSSR}</>
```
ChristopherChudzicki, line 15: Any need for the fragment wrapper?

And maybe just `null` for `onSSR` default?

ChristopherChudzicki (approved):

PR #1839, opened 2024-11-20
Consistent qdrant point ids

shanbady:

Closes https://github.com/mitodl/hq/issues/6094

This PR generates reproducible uuids off of the resource readable id for vector points stored in Qdrant. What this lets us do, is directly reference and check for existing embeddings in Qdrant if we have a learning resource or content file. Currently for vector similarity, the endpoint unnecessarily re-embeds the referenced document even though the embeddings for that already exist in qdrant (causes a slight delay when loading /api/v1/learning_resources/181/vector_similar/) - this is resolved in this PR since we can re-use the existing embedding

### How can this be tested?

1. Checkout main and make sure you have learning resources locally
2. clear existing collections and generate the embeddings via `python manage.py generate_embeddings --all --skip-contentfiles`
3. find some learning resource id and load the vector similarity endpoint `/api/v1/learning_resources/{resource id}/vector_similar/`- note the delay in loading
4. Checkout this branch
5. make sure you have learning resources locally
6. clear existing collections and generate the embeddings via `python manage.py generate_embeddings --all --skip-contentfiles`
7. find some learning resource and load the vector similarity endpoint `/api/v1/learning_resources/{resource id}/vector_similar/` -  note how much faster it loads

### Additional Context
- We generate the uuid off of the resource "readable_id" instead of the "id" so that if we had some "master embeddings" snapshot - it can be instantly re-used in any environment.

### DIFF
--- a/learning_resources/models.py
+++ b/learning_resources/models.py
@@ -474,6 +474,9 @@ class LearningResource(TimestampedModel):
     )
     location = models.CharField(max_length=256, blank=True)
 
+    def vector_point_id(self):
+        return uuid.uuid5(uuid.NAMESPACE_DNS, self.readable_id)
+
     @property
     def audience(self) -> str | None:
         """Returns the audience for the learning resource"""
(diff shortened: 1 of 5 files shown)


abeglova (approved):
lgtm

PR #1840, opened 2024-11-20
Refactor date/time parsing for mitpe events

mbertrand:
Closes https://github.com/mitodl/hq/issues/6114

Refactors the functions for mitpe event date/time parsing so they better handle the formats in the mitpe events api.

### How can this be tested?
- Run `./manage.py  backpopulate_news_events --pipelines mitpe_events_etl` on the main branch, you will see lots of errors about the date/time being unparseable.
- Switch to this branch and try again you should see very few or no errors.  If any errors do occur, they should be for very badly formatted time_range values from the mitpe api.

### DIFF
--- a/news_events/etl/utils.py
+++ b/news_events/etl/utils.py
@@ -134,18 +134,40 @@ def parse_date(text_date: str) -> datetime:
     return dt_utc
 
 
+def convert_to_utc(dt: datetime, known_tz: str) -> datetime:
+    """
+    Convert a datetime object to UTC timezone. If its
+    orignal timezone is not known, assume it is in US/Eastern.
+
+    Args:
+        dt (datetime): The datetime object to convert
+        known_tz (str): The timezone string if known
+
+    Returns:
+        datetime: The datetime object in UTC timezone
+    """
+    if not dt:
+        return None
+    if not known_tz:
+        # Assume it is in US/Eastern where MIT is
+        dt = dt.replace(tzinfo=ZoneInfo("US/Eastern"))
+    return dt.astimezone(UTC)
+
+
 def parse_date_time_range(
-    start_date_str, end_date_str, time_range_str: str
+    start_date_str: str, end_date_str: str, time_range_str: str
 ) -> tuple[datetime, datetime]:
     """
     Attempt to parse the time range from the MITPE events API.
     The field might not actually contain a time or range.
 
     Args:
+        start_date_str (str): start date string
+        end_date_str (str): end date string
         time_range (str): time range string
 
     Returns:
-        tuple: start and end times as strings
+        tuple: start and end datetimes
 
     """
     start_time, start_ampm, end_time, end_ampm, tz = "", "", "", "", ""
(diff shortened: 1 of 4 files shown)


news_events/etl/utils.py
```diff
@@ -131,3 +132,69 @@ def parse_date(text_date: str) -> datetime:
...
+    """
+    if not dt:
+        return None
+    if not known_tz:
+        # Assume it is in US/Eastern where MIT is
+        dt = dt.replace(tzinfo=ZoneInfo("US/Eastern"))
+    return dt.astimezone(UTC)
+
+
+def parse_date_time_range(
```
abeglova, line 157: This does not seem to get the correct time 100% of the time. For example

start_date_str="2020-01-28"
end_date_str="2020-01-28"
time_range_str="2-3:00pm"
returns start_date = datetime.datetime(2020, 1, 28, 2, 0)
end_date = datetime.datetime(2020, 1, 28, 15, 0) 

before the conversion to utc

Also some events just have a start date like "1pm" and the time isn't parsed at all in that case

news_events/etl/utils.py
```diff
@@ -131,3 +132,69 @@ def parse_date(text_date: str) -> datetime:
...
+        end_date_str (str): end date string
+        time_range (str): time range string
+
+    Returns:
+        tuple: start and end datetimes
+
+    """
+    start_time, start_ampm, end_time, end_ampm, tz = "", "", "", "", ""
+    time_regex = re.compile(
+        r"(\d{1,2})(:\d{2})?\s*(am|pm)?\s*-?\s*(\d{1,2})(:?\d{2})?\s*(am|pm)?\s*([A-Za-z]{2,3})?",
```
abeglova, line 175: The regex is really complicated. Could it be broken up into multiple steps? Like one step to break the time string into two in the case where there is both a start and end date and a seperate step to extract pm/am. As it is i think it's hard to troubleshoot or understand what the regex is doing

news_events/etl/utils.py
abeglova: Some events also have time strings that are going to be really hard to parse like "8:00pm; doors open at 7:30pm" I wonder if it makes sense to give up, store event dates then store event time as a string so it's displayed the way it was in the original listing

mbertrand: Maybe I was trying too hard to get the times right; what's most important is really the start date because that's the only thing that's actually shown on the frontend.  Next most important is end_date, because both are used to determine if the event should be published (start date or end date should be < now).  If the professional ed api is sending back dates like "1 pm", that's just bad data and on them to fix.

How about I make sure the dates are correctly parsed (assuming the data is good), then for times, if it can't figure it out, it will just default to noon?  That way when the pipeline runs in the early AM, it will keep events happening that day published for display.

news_events/etl/utils.py
mbertrand: Split up the regex and also split up the function.  Added more inline comments and more unit test cases

abeglova (approved):
lgtm

Issue #1843, opened 2024-11-21
codespace config: Update minimum host requirements and add qdrant service


shanbady:
### Description/Context
The minimum requirements for a working codespace have changed and now require atleast a 4cpu 16gb memory instance. 

We need to update the codespace config so that this is the machine type launched by default (it will otherwise cause confusion when people auto-launch and the instances hangs/freezes).

We also need to add qdrant as a service in the devcontainer.json

### Plan/Design
update the devcontainer.json config with minimum host requirements and include qdrant service in the runServices section

PR #1845, opened 2024-11-21
set min cpu and memory. also add qdrant container

shanbady:
Closes https://github.com/mitodl/mit-learn/issues/1843

updates the default configuration so that codespaces will launch a 4cpu and 16gb memory instance by default. This Pr also updates runServices in devcontainer.json to include the qdrant service.
### How can this be tested?
1. Click this button to open the default options for the **main branch** 
  [Open in GitHub Codespaces](https://codespaces.new/mitodl/mit-learn/tree/main)
2. observe that "2-code" is pre-selected as the machine type
3. Open the launch page for this branch
  [Open Main branch in GitHub Codespaces](https://codespaces.new/mitodl/mit-learn/tree/shanbady%2Fcodespace-set-host-config)
4. see that the "4-core" option is now the default
4. use the link above and actually launch the codespace for this branch.
5. once it finishes building and you can access the terminal, confirm the qdrant container is running by curling the qdrant dashboard page `curl http://qdrant:6333/dashboard`


shanbady:
> ```json
> containerEnv
> ```

Good idea. I exposed the qdrant port.

re: secrets in the container - interesting I'm going to have to look a bit more into that because right now there are certain secrets that have defaults set via docker-compose.codespaces.yml which pulls in from envs/codespaces.env (which I think do get properly set during build time) in addition to secrets added via [github settings](https://docs.github.com/en/codespaces/managing-your-codespaces/managing-your-account-specific-secrets-for-github-codespaces#adding-a-secret). Its possible that some of those are not properly getting forwarded into the container

PR #1849, opened 2024-11-22
returning all resource types for similar resources endpoint

shanbady:

Closes https://github.com/mitodl/hq/issues/6132

Fixes the similar resources endpoint so that it returns more than just courses

### How can this be tested?
1. checkout this branch
2. load up resources locally. ensure you have a variety of resource types (videos, courses, podcasts etc)
3. go to the search page, find a learning resource and open up the drawer. in the "similar resources" carousel you should see a mixture of resources types and not just courses.


mbertrand (approved):
👍

PR #1859, opened 2024-11-25
Randomize hero image index on the server

jonkafton:

Closes: https://github.com/mitodl/hq/issues/5921

Selects the random Hero image index on the server so it can be server rendered.

### How can this be tested?

In development watch mode, refresh the homepage to see a the a randomized image selection and ensure that it is server rendered (ie. hero-_n_.png) appears in the HTML source.

### Additional context

We're caching for 30 mins on the CDNs - each image selection will be shown for that duration.


shanbady (approved):
👍 works as intended

PR #1863, opened 2024-11-26
Filter out content type on similar resources endpoint

shanbady:
Fixes https://github.com/mitodl/mit-learn/issues/1862
This PR fixes a bug with the similar resources endpoint that happens when a opensearch result is missing the published attribute (happens on unpublished contentfiles).

### How can this be tested?
I have included a test that will fail on main and pass on this branch. 
You can also visit the similar resources endpoint `http://open.odl.local:8063/api/v1/learning_resources/181/similar/` and validate that only learning resources are returned.

### Additional Context
Restricting this to just learning resources should fix the issue but I have also moved the check for "published" to the django filter instead of the opensearch response for added measure since we cant rely on that attribute in the index.


mbertrand (approved):
👍

PR #1867, opened 2024-12-03
fix edx file imports

abeglova:
closes https://github.com/mitodl/hq/issues/6031

This pr removes duplicate edx content files and cleans up content file content for subtitle files.  We were creating duplicates of edx content files because both the code that was intended to find existing documents to update them instead of creating new ones and the code that unpublished and deleted documents that were no longer in the course folder had bugs.

It also populates the file path and file extension for edx content files and does not attempt to extract content with tilka for non-text files such as csv files or various code file formats.

### How can this be tested?
If you don't already have some xpro files in your database run ` docker-compose run web ./manage.py backpopulate_xpro_data`,  `docker-compose run web ./manage.py backpopulate_xpro_files in the main branch

On this branch

Run `docker-compose run web ./manage.py backpopulate_xpro_data` , `docker-compose run web ./manage.py backpopulate_xpro_files --overwrite`
Go to http://api.open.odl.local:8063/api/v1/contentfiles/?platform=xpro and verify that you have 10315 files
From the shell verify that
 ContentFile.objects.filter(run__learning_resource__platform__code='xpro' ).filter(file_extension__isnull=True).count() is 0

Find a srt file useing

ContentFile.objects.filter(run__learning_resource__platform__code='xpro').filter(file_extension='.srt').first()
Verify that the content field does not have timestamps


learning_resources/etl/utils.py
```diff
@@ -376,17 +319,11 @@ def documents_from_olx(
     Yields:
         tuple: A list of (bytes of content, metadata)
     """
-    try:
-        yield from get_xbundle_docs(olx_path)
```
abeglova, line 380: verticals are just duplicates of files that are imported in the next step

learning_resources/etl/utils.py
```diff
@@ -395,14 +332,50 @@ def documents_from_olx(
...
+
+    Returns:
+        str: The content as a string without timestamps
+    """
+    # Remove timestamps
+    content = re.sub(
+        r"\d{2}:\d{2}:\d{2},\d{3} --> \d{2}:\d{2}:\d{2},\d{3}\n", "", content
+    )
+    # Remove sequence numbers
+    content = re.sub(r"\d+\n", "", content)
```
shanbady, line 360: it seems like there are a lot of srt files that still have timestamps - I think this is because of the '\n' at the end of the regex. I havent combed through all the cases that were missed but the ones i spotted ended with the timestamp which you should be able to fix by adjusting for end of match `r"\d{2}:\d{2}:\d{2},\d{3} --> \d{2}:\d{2}:\d{2},\d{3}(\n|$)"`

shanbady: also - should lock this down with some tests (or adjust existing tests)

learning_resources/etl/utils.py
```diff
@@ -395,14 +332,50 @@ def documents_from_olx(
...
+    # Remove empty lines
+    return re.sub(r"\n\s*\n", " ", content)
+
+
+def clean_sjson_content(content: str):
+    """
+    Remove timestamps  from sjson content
```
shanbady, line 367: may want to clarify the docstring since it seems like this method returns the text content rather than removing anything. (minor nitpick - seems like some extra spaces in the docstring)

shanbady (changes requested):
Left some notes on the srt filter. One other thing I noticed is that it seems there are about 3089 xpro contentfiles that have empty content. is this something we should be filtering out as well?

abeglova:
I didn't filter the files with empty content because for ocw we want files with non text content to show up in the search and be searchable by title.  Since we don't have plans for a edx content search, I will update the code to just not save content files without content for edx.

learning_resources/etl/utils.py
abeglova: Thanks for the suggestion, I will make that change. Since for now we will just be using the content for vector search, a few stray timestamps are ok, we just don't want the entirety of the content to be interspersed by timestamps.

shanbady (approved):
👍

PR #1868, opened 2024-12-03
App env vars for use with local vault agent

shaidar:

Part of https://github.com/mitodl/ol-infrastructure/issues/1816

This shell script is required for local dev environments when used in conjunction a locally running vault agent to render secrets so that there would be no need to copy/paste from Heroku apps. Step-by-step instructions on how that file is to be used with vault can be found [here](https://github.com/mitodl/platform-engineering-site/blob/main/docs/how_to/mitol_dev_access_to_vault.md)


env/app_vars.sh
```diff
@@ -0,0 +1,112 @@
...
+echo "OPENSEARCH_INDEXING_CHUNK_SIZE=75" >>.env
+echo "QDRANT_COLLECTION_NAME=mitlearn-qa" >>.env
+echo "PROLEARN_CATALOG_API_URL=https://prolearn.mit.edu/graphql" >>.env
+echo "SEE_API_URL=https://mit-unified-portal-prod-78eeds.43d8q2.usa-e2.cloudhub.io/api/" >>.env
+echo "SEE_API_ACCESS_TOKEN_URL=https://mit-unified-portal-prod-78eeds.43d8q2.usa-e2.cloudhub.io/oauth/token" >>.env
+echo "SECURE_CROSS_ORIGIN_OPENER_POLICY=None" >>.env
+echo "SEE_BASE_URL=https://executive.mit.edu/" >>.env
+echo "SOCIAL_AUTH_OL_OIDC_KEY=ol-mitlearn-client" >>.env
+echo "USE_X_FORWARDED_HOST=True" >>.env
+echo "USE_X_FORWARDED_PORT=True" >>.env
```
rhysyngsun, line 72: I don't think we want these as we aren't behind a reverse proxy locally.

env/app_vars.sh
```diff
@@ -0,0 +1,112 @@
...
+echo "MITOL_FRONTPAGE_DIGEST_MAX_POSTS=10" >>.env
+echo "MITOL_USE_S3=True" >>.env
+echo "MITOL_NOTIFICATION_EMAIL_BACKEND=anymail.backends.mailgun.EmailBackend" >>.env
+echo "MITPE_BASE_URL=https://professional.mit.edu/" >>.env
+echo "MITX_ONLINE_BASE_URL=https://mitxonline.mit.edu/" >>.env
+echo "MITX_ONLINE_COURSES_API_URL=https://mitxonline.mit.edu/api/v2/courses/" >>.env
+echo "MITX_ONLINE_LEARNING_COURSE_BUCKET_NAME=mitx-etl-mitxonline-production" >>.env
+echo "MITX_ONLINE_PROGRAMS_API_URL=https://mitxonline.mit.edu/api/v2/programs/" >>.env
+echo "NEW_RELIC_LOG=stdout" >>.env
+echo "NODE_MODULES_CACHE=False" >>.env
```
rhysyngsun, line 51: I'm not sure about the exact implications of this but I suspect it's not something we want locally.

env/app_vars.sh
```diff
@@ -0,0 +1,112 @@
...
+echo "MITOL_FROM_EMAIL=MITOpen <mitopen-support@mit.edu>" >>.env
+echo "MITOL_FRONTPAGE_DIGEST_MAX_POSTS=10" >>.env
+echo "MITOL_USE_S3=True" >>.env
+echo "MITOL_NOTIFICATION_EMAIL_BACKEND=anymail.backends.mailgun.EmailBackend" >>.env
+echo "MITPE_BASE_URL=https://professional.mit.edu/" >>.env
+echo "MITX_ONLINE_BASE_URL=https://mitxonline.mit.edu/" >>.env
+echo "MITX_ONLINE_COURSES_API_URL=https://mitxonline.mit.edu/api/v2/courses/" >>.env
+echo "MITX_ONLINE_LEARNING_COURSE_BUCKET_NAME=mitx-etl-mitxonline-production" >>.env
+echo "MITX_ONLINE_PROGRAMS_API_URL=https://mitxonline.mit.edu/api/v2/programs/" >>.env
+echo "NEW_RELIC_LOG=stdout" >>.env
```
rhysyngsun, line 50: Shouldn't need this as we don't use it locally.

env/app_vars.sh
```diff
@@ -0,0 +1,112 @@
...
+echo "CKEDITOR_UPLOAD_URL=${CKEDITOR_UPLOAD_URL}" >>.env
+echo "EDX_API_CLIENT_ID=${EDX_API_CLIENT_ID}" >>.env
+echo "EDX_API_CLIENT_SECRET=${EDX_API_CLIENT_SECRET}" >>.env
+echo "EMBEDLY_KEY=${EMBEDLY_KEY}" >>.env
+echo "MAILGUN_KEY=${MAILGUN_KEY}" >>.env
+echo "OCW_NEXT_SEARCH_WEBHOOK_KEY=${OCW_NEXT_SEARCH_WEBHOOK_KEY}" >>.env
+echo "OCW_WEBHOOK_KEY=${OCW_WEBHOOK_KEY}" >>.env
+echo "OLL_API_CLIENT_ID=${OLL_API_CLIENT_ID}" >>.env
+echo "OPENSEARCH_HTTP_AUTH=${OPENSEARCH_HTTP_AUTH}" >>.env
+echo "SECRET_KEY=${SECRET_KEY}" >>.env
```
rhysyngsun, line 15: We shouldn't need the value from a deployed environment it could just be defined as a static string since it's only for local so commiting it shouldn't be an issue.

env/app_vars.sh
```diff
@@ -0,0 +1,112 @@
...
+echo "SENTRY_DSN=${SENTRY_DSN}" >>.env
```
rhysyngsun, line 16: Should also statically set `SENTRY_ENV=dev`

env/app_vars.sh
```diff
@@ -0,0 +1,112 @@
...
+echo "ACCESS_TOKEN_URL=https://sso-qa.ol.mit.edu/realms/olapps/protocol/openid-connect/token" >>.env
+echo "AUTHORIZATION_URL=https://sso-qa.ol.mit.edu/realms/olapps/protocol/openid-connect/auth" >>.env
+echo 'CORS_ALLOWED_ORIGINS=["https://mitopen-rc.odl.mit.edu", "https://ocw-next.netlify.app", "https://ol-devops-ci.odl.mit.edu", "https://draft-qa.ocw.mit.edu", "https://live-qa.ocw.mit.edu"]' >>.env
+echo 'CSRF_TRUSTED_ORIGINS=["https://api.mitopen-rc.odl.mit.edu", "https://mitopen-rc.odl.mit.edu"]' >>.env
+echo "KEYCLOAK_BASE_URL=https://sso-qa.ol.mit.edu/" >>.env
+echo "MAILGUN_FROM_EMAIL=MIT Open <no-reply@discussions-mail.odl.mit.edu" >>.env
+echo "MAILGUN_SENDER_DOMAIN=discussions-mail.odl.mit.edu" >>.env
+echo "MAILGUN_URL=https://api.mailgun.net/v3/discussions-mail.odl.mit.edu" >>.env
+echo "OIDC_ENDPOINT=https://sso-qa.ol.mit.edu/realms/olapps" >>.env
+echo "SESSION_COOKIE_DOMAIN=.rc.learn.mit.edu" >>.env
```
rhysyngsun, line 93: I think this probably needs to be `.odl.local` instead, but someone would need to test to verify.

env/app_vars.sh
```diff
@@ -0,0 +1,112 @@
...
+echo 'SOCIAL_AUTH_ALLOWED_REDIRECT_HOSTS=["live-qa.ocw.mit.edu", "mitopen-rc.odl.mit.edu"]' >>.env
+echo "SOCIAL_AUTH_OL_OIDC_OIDC_ENDPOINT=https://sso-qa.ol.mit.edu/realms/olapps" >>.env
+echo "USERINFO_URL=https://sso-qa.ol.mit.edu/realms/olapps/protocol/openid-connect/userinfo" >>.env
+echo >>.env
+echo "# Heroku Vars" >>.env
+echo "CELERY_WORKER_MAX_MEMORY_PER_CHILD=125000" >>.env
```
rhysyngsun, line 99: I don't think we need this because it's heroku-specific performance tuning.

env/app_vars.sh
```diff
@@ -0,0 +1,112 @@
...
+echo "OLL_LEARNING_COURSE_BUCKET_PREFIX=open-learning-library/courses/" >>.env
+echo "OPENSEARCH_DEFAULT_TIMEOUT=30" >>.env
+echo "OPENSEARCH_INDEXING_CHUNK_SIZE=75" >>.env
+echo "QDRANT_COLLECTION_NAME=mitlearn-qa" >>.env
+echo "PROLEARN_CATALOG_API_URL=https://prolearn.mit.edu/graphql" >>.env
+echo "SEE_API_URL=https://mit-unified-portal-prod-78eeds.43d8q2.usa-e2.cloudhub.io/api/" >>.env
+echo "SEE_API_ACCESS_TOKEN_URL=https://mit-unified-portal-prod-78eeds.43d8q2.usa-e2.cloudhub.io/oauth/token" >>.env
+echo "SECURE_CROSS_ORIGIN_OPENER_POLICY=None" >>.env
+echo "SEE_BASE_URL=https://executive.mit.edu/" >>.env
+echo "SOCIAL_AUTH_OL_OIDC_KEY=ol-mitlearn-client" >>.env
```
rhysyngsun, line 70: This client will need to be set up with the local hostname redirect urls configured, are we ok with the same client for RC being setup like that? There's also some cases where keycloak will redirect to the client's main url, which will end the user up on RC instead of their local setup.

rhysyngsun (changes requested):

env/app_vars.sh
shaidar: I think we have two use cases around this:
- Dev wanting to go through the flow and login to QA/RC which this would cover
- Dev wanting to login through Keycloak to their own instance of learn. I think in this case, I'll create a separate client. The trick is the redirect as that's dynamic, however, we can setup a wildcard (not ideal) however between requiring the `client_secret` and enabling PKCE, we should be on the safer side.

rhysyngsun (approved):

PR #1869, opened 2024-12-03
OpenAI embeddings for vector search

shanbady:

Closes https://github.com/mitodl/hq/issues/6152

This PR:
- allows us to use OpenAI (or any other provider we configure) for embeddings.
- introduces a daily task to embed new learning resources
- When developing locally it defaults to using fastembed on the cpu
- production/staging/live environments will be configured to use the openai api

The configuration for this is in main/settings.py:
- QDRANT_ENCODER
    - this settings specifies which encoder/embeddings class to use. vector_search/encoders contains classes for each provider. If we need to add a new one we can subclass the BaseEncoder and implement the required methods. Locally this is set to the FastEmbedEncoder and production/staging will be set to use OpenAI. The encoder classes allow us to encapsulate settings and functionality specific to each provider.
- QDRANT_DENSE_MODEL
    - specifies the model to use. Allows us to switch between models for a given provider.
    - locally for fastembed this is BAAI/bge-small-en-v1.5 and for openai will be "text-embedding-3-small" or "text-embedding-3-large" (more expensive and which we currently dont have access to)

There is also now a new "embed_new_learning_resources" celery task which runs daily and embeds just new learning resources. If we ever need to backfill we can manually run the management command (initial embeddings will be generated this way)

#### using fastembed
1. Checkout this branch
2. rebuild your containers `docker compose build celery web` (and restart) `docker compose down web celery && docker compose up -d`
3. populate a bunch of learning resources locally
4. run the management command to generate embeddings locally ` python manage.py generate_embeddings --all --recreate-collections --skip-contentfiles`
5. make sure it completes. go to the [qdrant dashboard](http://localhost:6333/dashboard) and note the model used for the collection and that the collection is populated Screenshot 2024-12-03 at 2 10 05 PM
6. find a learning resource id and visit the vector based similar resources endpoint http://open.odl.local:8063/api/v1/learning_resources/181/vector_similar/ and validate it works.
7. visit the vector search endpoint and validate it also works http://open.odl.local:8063/api/v0/learning_resources_vector_search/

#### using openai
1. in settings.py set QDRANT_ENCODER to the openai class "vector_search.encoders.openai.OpenAIEncoder". (optional: set QDRANT_DENSE_MODEL to "text-embedding-3-small" - it will default to this anyways if None)
2. set the the value for OPENAI_EMBEDDINGS_API_KEY from the [rc settings in heroku](https://dashboard.heroku.com/apps/mitopen-rc/settings)
3. restart celery so the new settings stick 
4. re-run through steps 4-7 from "using fastembed" steps above. Both vector search and similarity should work. When visiting the qdrant dashboard should see the embeddings that are stored are openai embeddings. you can dig further into the collection details to see and compare the dimensions, etc.

### Additional Context
1. openai generation time locally for 10148 resources took 234.973707 seconds and cost $0.09 usinf the text3-small model
2. vector search endpoint was much faster than cpu fastembed version
3. If we want to switch or add providers for embeddings, we can simply add an encoder in the [encoders folder](https://github.com/mitodl/mit-learn/tree/499aa279a1fa856a945a58a76167996d8d94bfd7/vector_search/encoders)
4. Future enhancement: since the embeddings now sit under their own named key in the qdrant collection,  we have the ability to let them coexist with other embeddings (for example a single record can contain embeddings for fastembed, openai text-embed3, image embeddings etc). We would just need to add some logic to upsert the new embedding key rather than recreate the record from scratch.


Ferdi:
https://docs.litellm.ai/docs/embedding/supported_embedding

vector_search/tasks.py
```diff
@@ -148,3 +150,30 @@ def start_embed_resources(self, indexes, skip_content_files):
...
+@app.task(bind=True)
+def embed_new_learning_resources(self, period="daily"):
+    """
+    Embed new resources in period by type
+    """
+    log.info("Running %s embedding task", period)
+    delta = datetime.timedelta(days=1)
+    if period == "weekly":
+        delta = datetime.timedelta(days=7)
+    since = now_in_utc() - delta
```
mbertrand, line 164: Do you think there's a chance some resources might occasionally fall through the cracks?  For example maybe this task gets occasionally backed up behind a bunch of other tasks and runs late.  Wondering if we should buffer this a bit by making the timedelta something like 25 hours for daily and 7*24+1 hours for weekly?  Or instead of using `now_in_utc() - timedelta`, filter by `created_on__gt=<today's date at a specific time(like 12:00 AM) - 1 day/week>`?

vector_search/utils.py
```diff
@@ -123,12 +161,11 @@ def embed_learning_resources(ids, resource_type):
...
-    client.add(
-        collection_name=collection_name,
-        ids=ids,
-        documents=docs,
-        metadata=metadata,
-    )
+    encoder = dense_encoder()
+    embeddings = encoder.encode_batch(docs)
+    vector_name = encoder.model_short_name()
+    points = points_generator(ids, metadata, embeddings, vector_name)
```
mbertrand, line 167: Would be good to have a unit test to make sure this works as expected (for `create_qdrand_collections` and/or for `points_generator`)

vector_search/tasks.py
mbertrand: I kicked off this task in a django shell `embed_new_learning_resources.delay()`, it frequently overwhelmed my celery workers, though they do start up again, not sure if it will complete successfully, it is still trying.

Screenshot 2024-12-04 at 11 37 28 AM

vector_search/encoders/dummy.py
```diff
@@ -0,0 +1,18 @@
+import numpy as np
+
+from vector_search.encoders.base import BaseEncoder
+
+
+class DummyEmbedEncoder(BaseEncoder):
```
mbertrand, line 6: This only seems to be used for unit tests, maybe it could could go in a `vector_search/conftest.py` file instead?

mbertrand (changes requested):
Overall looks great, needs some additional docstrings, and another unit test or two.

vector_search/encoders/openai.py
```diff
@@ -0,0 +1,26 @@
+from django.conf import settings
+from openai import OpenAI
+
+from vector_search.encoders.base import BaseEncoder
+
+
+class OpenAIEncoder(BaseEncoder):
```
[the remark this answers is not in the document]
shanbady: I've used langchain the most but there were certain edgecase issues I ran into with the composable pattern it forces you to use so I'm open to alternatives

vector_search/tasks.py
mbertrand: PS Did not have this problem after switching the new env variables over to openai

main/settings.py
```diff
@@ -795,13 +795,15 @@ def get_all_config_keys():
...
-)
+QDRANT_DENSE_MODEL = get_string(name="QDRANT_DENSE_MODEL", default=None)
 QDRANT_SPARSE_MODEL = get_string(
     name="QDRANT_SPARSE_MODEL", default="prithivida/Splade_PP_en_v1"
 )
+QDRANT_ENCODER = get_string(
+    name="QDRANT_ENCODER", default="vector_search.encoders.fastembed.FastEmbedEncoder"
+)
+OPENAI_EMBEDDINGS_API_KEY = get_string(
+    name="OPENAI_EMBEDDINGS_API_KEY",
```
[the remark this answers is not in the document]
shanbady: This may be a good discussion point. I currently have the embedding api key scoped for permissions to just embeddings. Do we want to have a catch all key or should we scope them?

vector_search/tasks.py
shanbady: The error you were seeing could be due to constrained docker resources when running embedding via fastembed  - you might have to increase cpu allocation for docker. I have mine set to the max and it still seems to be overflowing  (but not crashing) during the local embedding process

Screenshot 2024-12-04 at 3 23 47 PM
Screenshot 2024-12-04 at 3 24 44 PM

main/settings.py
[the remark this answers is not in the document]
shanbady: I swapped out the openai encoder for a litellm encoder which means we will have to use OPENAI_API_KEY as the key name and have to set that as an environment variable so that litellm picks it up. to test the new litellm encoder set the OPENAI_API_KEY environment variable and settings.QDRANT_ENCODER to the litellm encoder.

vector_search/encoders/openai.py
```diff
@@ -0,0 +1,26 @@
...
+
+    def encode(self, text: str) -> list:
+        return (
+            self.client.embeddings.create(model=self.model_name, input=[text])
+            .data[0]
+            .embedding
+        )
+
+    def encode_batch(self, texts: list[str]) -> list[list[float]]:
+        return [
```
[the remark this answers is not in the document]
shanbady: modified this to have encode call encode batch (other way around introduced an api call for each text)

vector_search/utils.py
shanbady: 👍 tests added for both

shanbady:
@mbertrand this is ready for another look. set of notable new changes (besides docstring and prev comments):
- swapped out the OpenAIEncoder for a litellm encoder - this requires that the OPENAI_API_KEY **env var** is set locally (not via settings.py)
- created a conftest.py and moved all default mocks, dummy encoder and settings there.
- tests in place for create_qdrand_collections and points_generator
- added a check for existing qdrant embeddings and increased frequency of new resource embedding task to every 30 minutes. I removed the period argument from the task. its not needed since we just look back 1 day and embed whatever is not akleady embedded every half hour.

pyproject.toml
```diff
@@ -81,6 +81,9 @@ uwsgitop = "^0.12"
 pytest-lazy-fixtures = "^1.1.1"
 pycountry = "^24.6.1"
 qdrant-client = {extras = ["fastembed"], version = "^1.12.0"}
+onnxruntime = "1.19.2"
+openai = "^1.55.3"
+litellm = "^1.53.5"
```
mbertrand, line 86: Is it important to have this exact onnxruntime version?  `1.20.0` is already included in the main branch.

If not I think you can remove this, and openai too, just add `litellm` - litellm will bring in openai anyway (version 1.57.0)

I tried it out by resetting both pyproject.toml and poetry.lock to the `main` branch versions and adding only litellm, then regenerated embeddings.  Everything still seemed work.

mbertrand (approved):
Everything works, but I think it's possible to get rid of a couple additions to the pyproject.toml file.

pyproject.toml
shanbady: I had hit merge just before catching this. Yes there was a reason for that specific pin. I can't find the exact thread but using the latest onnxruntime had issues with certain models. I was seeing it locally where the process consistently got killed and the only way around it was to downgrade (similarly suggested here https://github.com/microsoft/onnxruntime/issues/10039) -  my suggestion is to leave atleast the onnyxruntime pin for now unless there is a compelling reason to upgrade. ill see if I can put together a repro case

PR #1871, opened 2024-12-03
fix card line clamping

gumaerc:
Follow-up for https://github.com/mitodl/mit-learn/pull/1850

In the above PR, we adjusted the line clamping rules in the `Card` component. Unfortunately, it didn't work in every scenario and this adjustment is needed to ensure that all children of the `Title` within `Card` have the line clamp rule applied.

image
image
image
image

### How can this be tested?
 - Ensure you have sufficient data backpopulated into your local instance, including some OCW and Sloan courses
 - Visit the home page at http://localhost:8062/
 - Scroll down to the news section and verify that 3 lines of the title are revealed, with an ellipsis included if line clamping is happening
 - Click on some results until you get one with a recommended course card with a long title, and verify that it is being clamped at 3 lines

### DIFF
--- a/frontends/ol-components/src/components/Card/Card.tsx
+++ b/frontends/ol-components/src/components/Card/Card.tsx
@@ -125,8 +125,9 @@ const Title = styled(Linkable, titleOpts)<{ size?: Size }>`
     -webkit-box-orient: vertical;
     -webkit-line-clamp: 3;
 
-    > * {
-      -webkit-line-clamp: 3;
+    /* Any and all children of the card need to be clamped */
+    * {
+      -webkit-line-clamp: 3 !important;
     }
   }
 `
--- a/frontends/ol-components/src/components/Card/Card.tsx
+++ b/frontends/ol-components/src/components/Card/Card.tsx
@@ -12,6 +12,7 @@ import { theme } from "../ThemeProvider/ThemeProvider"
 import { pxToRem } from "../ThemeProvider/typography"
 import Link from "next/link"
 import { default as NextImage, ImageProps as NextImageProps } from "next/image"
+import { TruncateText } from "../TruncateText/TruncateText"
 
 export type Size = "small" | "medium"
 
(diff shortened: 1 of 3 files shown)


jonkafton:
Currently on `main` the News section stories are clamped at 3 lines - the Title link has no children aside from its text, so the change doesn't seem to have any effect here:

image

<br /><br />
Separately but related, there was a previous issue that needed the News section cards to be bottom aligned. It's a design question and I'd assume we favor readability, but do we need to make an adjustment somewhere else to put them back in alignment?...

<br />

image

The resource cards do have a nested element, the `EllipsisTitle` in LearningResourceCard.tsx, which uses `TruncateText` to apply the line clamp, which is being overridden by the change in Card.tsx. If all cards are clamped at 3 lines we can remove the EllipsisTitle; thinking though it would be good to build `TruncateText` into Card.Title in Card.tsx for reuse.

frontends/ol-components/src/components/Card/Card.tsx
```diff
@@ -103,13 +104,14 @@ const Info = styled.div<{ size?: Size }>`
 `
 
 const titleOpts = {
-  shouldForwardProp: (prop: string) => prop !== "size",
+  shouldForwardProp: (prop: string) => prop !== "lines" && prop !== "size",
 }
-const Title = styled(Linkable, titleOpts)<{ size?: Size }>`
+const Title = styled(Linkable, titleOpts)<{ size?: Size; lines?: number }>`
```
ChristopherChudzicki, line 109: **Suggestion:** use the `truncateText` css helper instead of putting an extra div inside Title: 
```tsx
const Title = styled(
  Linkable,
  titleOpts,
)<{ size?: Size; lines?: number }>(({ theme, size, lines = 3 }) => {
  return [
    {
      display: "flex",
      textOverflow: "ellipsis",
      overflow: "hidden",
      margin: 0,
      ...truncateText(lines),
    },
    size === "small"
      ? {
          ...theme.typography.subtitle2,
          height: theme.typography.pxToRem(18 * lines),
        }
      : {
          ...theme.typography.subtitle1,
          height: theme.typography.pxToRem(20 * lines),
        },
  ]
})
```
then just 
```tsx
        <Title
          data-card-link={!!title.href}
          className="MitCard-title"
          size={size}
          {...title}
        />
```
below. (`lines` and `children` will come from `...title`)


**Additionally:** You could do
```
height: `calc(${lines} * ${theme.typography.subtitle1.lineHeight})` 
```
if you want. TS will tell you `lineHeight` might be undefined. We actually do define it for every typography style, and I can see how using it (as here) is useful, so we could make the theme aware of the fact that it is always defined. Happy to help if you wanna do that.

**Aside:** Apparently there are line-height CSS units, `lh` and `rlh` (see https://developer.mozilla.org/en-US/docs/Learn/CSS/Building_blocks/Values_and_units#numbers_lengths_and_percentages) but their browser support is only ~91% . So sad.

ChristopherChudzicki (approved):
👍

PR #1874, opened 2024-12-04
fix drawer sticking

ChristopherChudzicki:
Fixes https://github.com/mitodl/hq/issues/6150

Fixes a bug where the drawer was sticking sometimes—getting stuck in an openable state.

### How can this be tested?
1. On `main` homepage: Click a card to open the drawer, then click outside the drawer to close it, and quickly click a different card.
    - you'll be stuck in a state where `?resource=1234` is in the URL, but drawer is not open.
2. Try on this branch. Drawer will not stick.

### DIFF
--- a/frontends/ol-components/src/components/RoutedDrawer/RoutedDrawer.tsx
+++ b/frontends/ol-components/src/components/RoutedDrawer/RoutedDrawer.tsx
@@ -1,4 +1,4 @@
-import React, { useCallback, useEffect, useMemo } from "react"
+import React, { useCallback, useEffect, useMemo, useRef } from "react"
 import Drawer from "@mui/material/Drawer"
 import styled from "@emotion/styled"
 import type { DrawerProps } from "@mui/material/Drawer"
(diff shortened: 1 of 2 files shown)


frontends/ol-components/src/components/RoutedDrawer/RoutedDrawer.tsx
```diff
@@ -91,25 +79,21 @@ const RoutedDrawer = <K extends string, R extends K = K>(
     }
   }, [router, searchParams, params])
 
+  const childParams = isOpen ? filteredParams : lastValidParams.current
   return (
-    <Drawer
-      open={open}
-      onTransitionExited={removeUrlParams}
-      onClose={setOpen.off}
```
ChristopherChudzicki, line 98: In old version, sometimes `onTransitionExited` was not firing, which meant that the query param never got removed from the URL. Consequently, the effect with dependencies `[requiredArePresent, setOpen, requiredParams]` would stop firing, and drawer would stay stuck.

ChristopherChudzicki:
I'm going to close this for now. I'm fairly sure that https://github.com/mitodl/mit-learn/pull/1872 is going to resolve the original issue.

PR #1876, opened 2024-12-05
fix price comparison in differing runs table

gumaerc:
Closes https://github.com/mitodl/hq/issues/6169

This PR fixes an issue with the function `allRunsAreIdentical`, which determines if all of the runs of a given course are identical or not for the purposes of displaying the "differing runs table" in the new learning resource drawer. Previously, a string comparison was being made against `price.amount` and a string of `"0"`. Some runs have their price stored as `"0.00"` which was making the comparison fail. To be sure it won't fail in the future, this value is being converted to a `number` before comparison, which should negate any issues with string comparison.

### How can this be tested?
 - Ensure you have Sloan courses backpopulated in your database
 - Search for the course "Supply Chain Fundamentals"
 - Click the top search result and verify that the differing runs table does not appear
 - Search for "Communication and Persuasion in the Digital Age"
 - Click the top search result and verify that the differing runs table still appears for this course

### DIFF
--- a/frontends/ol-utilities/src/learning-resources/learning-resources.ts
+++ b/frontends/ol-utilities/src/learning-resources/learning-resources.ts
@@ -136,7 +136,7 @@ const allRunsAreIdentical = (resource: LearningResource) => {
   for (const run of resource.runs) {
     if (run.resource_prices) {
       run.resource_prices.forEach((price) => {
-        if (!(resource.free && price.amount === "0")) {
+        if (!(resource.free && Number(price.amount) === 0)) {
           amounts.add(price.amount)
           currencies.add(price.currency)
         }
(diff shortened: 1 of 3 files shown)


frontends/ol-utilities/src/learning-resources/learning-resources.ts
```diff
@@ -136,7 +136,7 @@ const allRunsAreIdentical = (resource: LearningResource) => {
   for (const run of resource.runs) {
     if (run.resource_prices) {
       run.resource_prices.forEach((price) => {
-        if (!(resource.free && price.amount === "0")) {
+        if (!(resource.free && Number(price.amount) === 0)) {
```
shanbady, line 139: We should include a test case for this.

shanbady (changes requested):
Change looks good and works. just missing a test case to hold it in place

gumaerc:
@shanbady Thanks for your review. I changed up the logic and added a new test to check that values are being compared numerically. Basically, if a course is free but offers certification we should expect the zero price as well as the certificate price and all other courses should only have one unique price if all runs are identical.

shanbady (approved):
👍

PR #1877, opened 2024-12-05
add a trailing slash to non contentfile OCW resources

gumaerc:
Closes https://github.com/mitodl/hq/issues/6187

This PR alters the OCW ETL code to add a trailing slash to the end of OCW URLs that reference pages or whole sites. OCW sites are generated with the Hugo static site generator and are configured to always generate a folder with an `index.html` file inside of it for each page. Although this PR is about indexing in OER commons and duplicates appearing, this is important to consider because not all OCW content are pages. Some of the content files are raw images, pdf's, etc. These URLs have not been modified, because a slash at the end of them would be incorrect.

### How can this be tested?
 - Ensure that you have proper AWS credentials in your environment and that you have set `OCW_LIVE_BUCKET=ocw-content-live-production`
 - Spin up your containers
 - Run `docker compose exec web ./manage.py backpopulate_ocw_data --skip-contentfiles --overwrite`
 - After the backpopulate has completed, enter a Django shell with `docker compose exec web ./manage.py shell`
 - Run the following commands in the shell:
 ```py
LearningResource.objects.filter(resource_type="course", offered_by_id="ocw")[0].url
 ```
 - You should see an OCW URL with a trailing slash like `'http://ocw.mit.edu/courses/4-463-building-technology-iii-building-structural-systems-fall-2004/'`


learning_resources/migrations/0079_ocw_pages_trailing_slash.py
```diff
@@ -0,0 +1,26 @@
+# Generated by Django 4.2.16 on 2024-12-06 22:02
+
+from django.db import migrations
+
+
+def ensure_trailing_slashes(apps, schema_editor):
+    LearningResource = apps.get_model("learning_resources", "LearningResource")
+
+    for resource in LearningResource.objects.filter(
+        resource_type="course", offered_by_id="ocw"
```
mbertrand, line 10: Could also filter by `etl_source`  alone, I think the end result will be the same:
```python
for resource in LearningResource.objects.filter(etl_source="ocw"):
```

mbertrand (approved):
LGTM, the migration's query filter could be simplified a bit to just `filter(etl_source="ocw")` but works fine as is.

PR #1878, opened 2024-12-06
specify non-free course for differing prices test

gumaerc:
Follow-up for https://github.com/mitodl/mit-learn/pull/1876

This PR fixes a flaky test introduced in the above PR

### How can this be tested?
 - In `frontends/ol-utilities/src/learning-resources/learning-resources.test.ts`, on line 125, modify it to be `test.each(new Array(1000).fill(null))("returns false if prices differ", () => {` to repeat the test 1000 times
 - Spin up `mit-learn`
 - Run `docker compose exec watch yarn test learning-resources`
 - Verify that the tests pass every time


jonkafton (approved):
Changes make sense for the `expectedPrices` counts and not seeing flakiness after running 15 or so times locally.

PR #1889, opened 2024-12-09
Filter params for vector search

shanbady:

Closes https://github.com/mitodl/hq/issues/6188

This PR enables filter parameters on the vector search endpoint /api/v0/learning_resources_vector_search/

available filters:
- readable_id
- resource_type
- certification
- certification_type
- professional
- free
- course_feature
- content_feature_type
- run_id
- resource_id
- topic
- ocw_topic
- level
- department
- platform
- offered_by
- delivery
- resource_category

### How can this be tested?
1. checkout this branch
2. make sure you have learning resources loaded up locally
3. make sure you have embeddings generated otherwise regenerate them via ```python manage.py generate_embeddings --all --skip-contentfiles --recreate-collections```
4. once embeddings have been generated, visit the vector search endpoint /api/v0/learning_resources_vector_search/
5. try different filter parameters from the list above and validate the responses
6. play around with the limit and offset parameters and validate the results


mbertrand (changes requested):
Overall works great, but noticed a couple of issues with certain parameters:

`?run_id=course-v1:MITxT+10.50.CH03x+1T2023`: Doesn't seem to do any filtering.
`?resource_id=1001001`: ditto
`readable_id`: works for some values (`e4e59bf6-3ad6-49b2-ac6f-c2e62875cba3`, `feeds.soundcloud.com/users/soundcloud:users:561805800/sounds.rss`) but not others (`program-v1:xPRO+QCR`, `course-v1:MITxT+3.012Sx`) -maybe the `+` is the cause?.

Also, the `count` shown is always 10000, regardless of how many matches are actually found, even when none are found.

shanbady:
> Overall works great, but noticed a couple of issues with certain parameters:
> 
> `?run_id=course-v1:MITxT+10.50.CH03x+1T2023`: Doesn't seem to do any filtering. `?resource_id=1001001`: ditto `readable_id`: works for some values (`e4e59bf6-3ad6-49b2-ac6f-c2e62875cba3`, `feeds.soundcloud.com/users/soundcloud:users:561805800/sounds.rss`) but not others (`program-v1:xPRO+QCR`, `course-v1:MITxT+3.012Sx`) -maybe the `+` is the cause?.
> 
> Also, the `count` shown is always 10000, regardless of how many matches are actually found, even when none are found.

Looking into the encoding issue for the values but regarding the counts - qdrant for some reason [doesn't return counts](https://github.com/qdrant/qdrant/issues/4882) with the search response. It requires a [separate call](https://api.qdrant.tech/v-1-12-x/api-reference/points/count-points) which I was initially planning on handling separately but i can add it to this pr.

shanbady:
@mbertrand added the counts to the results - as far as the 

> @shanbady are these filters the same as what we support with ES endpoint ?

Yes with the exception of the ES specific/admin filters like slop, min_score, max_incompleteness_penalty etc.

shanbady:
> @shanbady is there a parameter for search ? The equivalent of "q" in ES?

Yes that is already there and live https://api.learn.mit.edu/api/v0/learning_resources_vector_search/?q=biology

vector_search/utils.py
```diff
@@ -4,13 +4,35 @@
...
+    "readable_id": "readable_id",
+    "resource_type": "resource_type",
+    "certification": "certification",
+    "certification_type": "certification_type.code",
+    "professional": "professional",
+    "free": "free",
+    "course_feature": "course_feature",
+    "content_feature_type": "content_feature_type",
+    "run_id": "run_id",
+    "resource_id": "resource_id",
```
mbertrand, line 25: These two filters still aren't working, I believe because they have not been added to `LearningResourcesVectorSearchRequestSerializer`.  They are not available to the regular search filter either though, so I think it might be okay to either remove them here, or add them to the serializer if filtering by these might be useful.

mbertrand (changes requested):

vector_search/utils.py
shanbady: 👍  removed them

mbertrand (approved):
👍

PR #1891, opened 2024-12-09
Populate OCW file extensions

abeglova:
https://github.com/mitodl/hq/issues/6222 - for now this just populates the file extension and does not do additional cleaning pending this discussion https://github.com/mitodl/hq/discussions/6230

After  https://github.com/mitodl/mit-learn/pull/1867 we populate file extensions for edx files. This populates them for ocw

### How can this be tested?
run ./manage.py
backpopulate_ocw_data 6-100l-introduction-to-cs-and-programming-using-python-fall-2022 --overwrite

From the shell run
```
run = LearningResource.objects.filter(url__contains='6-100l-introduction-to-cs-and-programming-using-python-fall-2022').runs.first()
ContentFile.objects.filter(run=run).filter(content_type='file').filter(file_extension__isnull=True).count() 
```
should return zero

Spot check some content files from `ContentFile.objects.filter(run=run).filter(content_type='file')` to check that that the file extension makes sense

### DIFF
--- a/learning_resources/etl/ocw.py
+++ b/learning_resources/etl/ocw.py
@@ -254,10 +254,12 @@ def transform_contentfile(
         content_type = CONTENT_TYPE_VIDEO
         file_s3_path = contentfile_data.get("transcript_file")
         image_src = contentfile_data.get("thumbnail_file")
+        file_extension = Path(contentfile_data.get("file")).suffix
     else:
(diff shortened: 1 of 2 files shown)


shanbady (approved):
change seems good. One thing i will suggest is improving the docstring for the test_transform_content_files test

PR #1893, opened 2024-12-09
drawer v2 - share resource

gumaerc:
Closes https://github.com/mitodl/hq/issues/5876

This PR adds the "share" section of the new learning resource drawer and adjusts the layout of the call to section. 

image
image

### How can this be tested?
 - Ensure you have sufficient data backpopulated into your local instance to test a variety of resources
 - Visit http://localhost:8062/search
 - Do *not* log in yet
 - Click any of the search results to bring up the learning resource drawer
 - Verify that you see two buttons below the unit logo, Bookmark and Share
 - Click the Share button
 - Verify that the share section appears and that the URL in the textbox looks correct
 - Click the social media share icons and verify that you are taken to each given social media site and their respective interfaces are displayed for sharing a post with a link to the resource
 - Click the Copy Link button and verify that the link is properly copied to your clipboard
 - Log in as a superuser
 - Go back to the search page and click on any learning resource to bring the drawer back up
 - Verify that you now see 3 buttons; Add to list, Bookmark and Share

### DIFF
--- a/frontends/ol-components/src/components/LearningResourceExpanded/LearningResourceExpandedV2.tsx
+++ b/frontends/ol-components/src/components/LearningResourceExpanded/LearningResourceExpandedV2.tsx
@@ -276,6 +276,9 @@ const ShareLink = styled(Link)({
 
 const CopyLinkButton = styled(StyledButton)({
   flexGrow: 0,
+  flexBasis: "112px",
+  padding: "12px 16px",
+  whiteSpace: "nowrap",
   "span:first-of-type": {
     color: theme.custom.colors.red,
   },
--- a/frontends/ol-components/src/components/LearningResourceExpanded/LearningResourceExpandedV2.tsx
+++ b/frontends/ol-components/src/components/LearningResourceExpanded/LearningResourceExpandedV2.tsx
@@ -270,7 +270,6 @@ const ShareLink = styled(Link)({
 const CopyLinkButton = styled(NoWrapButton)({
   flexGrow: 0,
   flexBasis: "112px",
-  // padding: "12px 16px",
   "span:first-of-type": {
     color: theme.custom.colors.red,
   },
--- a/frontends/ol-components/src/components/LearningResourceExpanded/LearningResourceExpandedV2.tsx
+++ b/frontends/ol-components/src/components/LearningResourceExpanded/LearningResourceExpandedV2.tsx
@@ -1,7 +1,6 @@
 import React, { useCallback, useRef, useState } from "react"
 import styled from "@emotion/styled"
 import Skeleton from "@mui/material/Skeleton"
-import Typography from "@mui/material/Typography"
 import { default as NextImage } from "next/image"
 import { ActionButton, Button, ButtonLink, ButtonProps } from "../Button/Button"
 import type { LearningResource } from "api"
@@ -12,8 +11,12 @@ import {
   RiBookmarkLine,
   RiCloseLargeLine,
   RiExternalLinkLine,
+  RiFacebookFill,
+  RiLink,
+  RiLinkedinFill,
   RiMenuAddLine,
   RiShareLine,
+  RiTwitterXLine,
 } from "@remixicon/react"
 import type { ImageConfig } from "../../constants/imgConfigs"
 import { theme } from "../ThemeProvider/ThemeProvider"
@@ -23,6 +26,8 @@ import type { User } from "api/hooks/user"
 import { LearningResourceCardProps } from "../LearningResourceCard/LearningResourceCard"
 import VideoFrame from "./VideoFrame"
 import { Link } from "../Link/Link"
+import { Input } from "../Input/Input"
+import { Typography } from "../.."
 
 const DRAWER_WIDTH = "900px"
 
@@ -225,6 +230,49 @@ const StyledButton = styled(Button)<{ filled?: number }>((props) => {
   }
 })
 
+const ShareContainer = styled.div({
+  display: "flex",
+  flexDirection: "column",
+  alignItems: "center",
+  alignSelf: "stretch",
+  padding: "16px 0 8px 0",
+  gap: "12px",
+})
+
+const ShareLabel = styled(Typography)({
+  ...theme.typography.body3,
+  color: theme.custom.colors.darkGray2,
+})
+
+const ShareInput = styled(Input)({
+  display: "flex",
+  flexDirection: "column",
+  justifyContent: "center",
+  alignItems: "flexStart",
+  alignSelf: "stretch",
+})
+
+const ShareButtonContainer = styled.div({
+  display: "flex",
+  justifyContent: "center",
+  alignItems: "center",
+  alignSelf: "stretch",
+  gap: "16px",
+})
+
+const ShareLink = styled(Link)({
+  color: theme.custom.colors.silverGrayDark,
+})
+
+const CopyLinkButton = styled(StyledButton)({
+  "span:first-of-type": {
+    color: theme.custom.colors.red,
+    "&:hover": {
+      color: theme.custom.colors.white,
+    },
+  },
+})
+
 const CarouselContainer = styled.div({
   display: "flex",
   flexDirection: "column",
@@ -247,6 +295,7 @@ const CarouselContainer = styled.div({
 type LearningResourceExpandedV2Props = {
   resource?: LearningResource
   user?: User
+  location: string
   imgConfig: ImageConfig
   carousels?: React.ReactNode[]
   inLearningPath?: boolean
@@ -407,6 +456,7 @@ const CallToActionSection = ({
   resource,
   hide,
   user,
+  location,
   inUserList,
   inLearningPath,
   onAddToLearningPathClick,
@@ -416,6 +466,7 @@ const CallToActionSection = ({
   resource?: LearningResource
   hide?: boolean
   user?: User
+  location: string
   inUserList?: boolean
   inLearningPath?: boolean
   onAddToLearningPathClick?: LearningResourceCardProps["onAddToLearningPathClick"]
@@ -444,6 +495,7 @@ const CallToActionSection = ({
   const addToLearningPathLabel = "Add to list"
   const bookmarkLabel = "Bookmark"
   const shareLabel = "Share"
+  const copyLinkLabel = "Copy Link"
   return (
     <CallToAction data-testid="drawer-cta">
       <ImageSection resource={resource} config={imgConfig} />
@@ -498,6 +550,29 @@ const CallToActionSection = ({
           {shareLabel}
         </CallToActionButton>
       </ButtonContainer>
+      <ShareContainer>
+        <ShareLabel>Share a link to this Resource</ShareLabel>
+        <ShareInput value={location} />
+        <ShareButtonContainer>
+          <ShareLink>
+            <RiFacebookFill />
+          </ShareLink>
+          <ShareLink>
+            <RiTwitterXLine />
+          </ShareLink>
+          <ShareLink>
+            <RiLinkedinFill />
+          </ShareLink>
+          <CopyLinkButton
+            size="small"
+            edge="circular"
+            startIcon={<RiLink />}
+            aria-label={copyLinkLabel}
+          >
+            {copyLinkLabel}
+          </CopyLinkButton>
+        </ShareButtonContainer>
+      </ShareContainer>
     </CallToAction>
   )
 }
@@ -557,6 +632,7 @@ const LearningResourceExpandedV2: React.FC<LearningResourceExpandedV2Props> = ({
   resource,
   imgConfig,
   user,
+  location,
   carousels,
   inUserList,
   inLearningPath,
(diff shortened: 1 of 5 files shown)


gumaerc:
@sovsey @mbilalmughal I had a thought while working on the "Copy Link" button. Any site I've seen that uses one of these gives you some kind of feedback that the link was copied to your clipboard. Either a notification, changing the text of the button to "Copied" (at least temporarily) or some other way to at least let the user know it has been done and the text is in their clipboard. Some sites even highlight the link and copy it to your clipboard immediately upon clicking the textbox. Should we do that as well?

mbilalmughal:
@gumaerc 
I like the suggestion to change the button text to "Link copied." This way, the CTA text will update, and the user will clearly know that the link has been copied.
CC: @steven-hatch @sovsey

mbilalmughal:
This is even better, but we don’t have this kind of floating notification anywhere. I wish we could add this—it looks really good, IMO.
Changing the button text will work as well. I’m fine with either option.
image

CC: @steven-hatch

frontends/ol-components/src/components/LearningResourceExpanded/LearningResourceExpandedV2.tsx
```diff
@@ -194,11 +207,86 @@ const OnPlatform = styled.span({
...
+const ButtonContainer = styled.div({
   display: "flex",
+  width: "100%",
   gap: "8px",
   flexGrow: 1,
-  justifyContent: "flex-end",
+  justifyContent: "center",
+})
+
+const StyledButton = styled(Button)<{ filled?: number }>((props) => {
```
ChristopherChudzicki, line 218: @gumaerc `Button` has a lot of styles built in, `StyledButton` overrides them, and then those are further overidden by `CopyLinkButton`. I'm a little worried this will become hard to keep track of.

In Figma, this looks like a new button variant `"bordered"`:

Screenshot 2024-12-12 at 10 20 50 AM

Figma also shows the button having a "selected" state:
Screenshot 2024-12-12 at 10 20 58 AM

I don't see "selected" state details on the [main button page ](https://www.figma.com/design/Eux3guSenAFVvNHGi1Y9Wm/MIT-Design-System?node-id=13735-67429&node-type=symbol&m=dev) and I'm not sure what the selected state should look like for other variants. So, **my suggestion is:**
- Add `"bordered"` as a variant for the main button component
- Add a `selected` prop locally in `LearningResourceExpanded2.tsx`.

I took a quick stab and came up with:
```diff
diff --git a/frontends/ol-components/src/components/Button/Button.tsx b/frontends/ol-components/src/components/Button/Button.tsx
index d30d7c6ac..1b21e0efc 100644
--- a/frontends/ol-components/src/components/Button/Button.tsx
+++ b/frontends/ol-components/src/components/Button/Button.tsx
@@ -10,6 +10,7 @@ type ButtonVariant =
   | "primary"
   | "secondary"
   | "tertiary"
+  | "bordered"
   | "text"
   | "noBorder"
   | "inverted"
@@ -98,7 +99,7 @@ const buildStyles = (props: ButtonStyleProps & { theme: Theme }) => {
     ...props,
   }
   const { colors } = theme.custom
-  const hasBorder = variant === "secondary"
+  const hasBorder = variant === "secondary" || variant === "bordered"
   return css([
     {
       color: theme.palette.text.primary,
@@ -142,11 +143,6 @@ const buildStyles = (props: ButtonStyleProps & { theme: Th
me }) => {
         boxShadow: "none",
       },
     },
-    hasBorder && {
-      backgroundColor: "transparent",
-      borderColor: "currentcolor",
-      borderStyle: "solid",
-    },
     variant === "success" && {
       backgroundColor: colors.darkGreen,
       color: colors.white,
@@ -163,13 +159,10 @@ const buildStyles = (props: ButtonStyleProps & { theme: Theme }) => {
         boxShadow: "none",
       },
     },
-    hasBorder && {
+    variant === "secondary" && {
       backgroundColor: "transparent",
       borderColor: "currentcolor",
       borderStyle: "solid",
-    },
-    variant === "secondary" && {
-      color: colors.red,
       ":hover:not(:disabled)": {
         backgroundColor: tinycolor(colors.brightRed).setAlpha(0.06).toString(),
       },
@@ -217,6 +210,20 @@ const buildStyles = (props: ButtonStyleProps & { theme: Theme }) => {
       borderColor: colors.mitRed,
       borderStyle: "solid",
     },
+    variant === "bordered" && {
+      borderStyle: "solid",
+      borderColor: colors.silverGrayLight,
+      color: colors.silverGrayDark,
+      backgroundColor: "transparent",
+      ":hover:not(:disabled)": {
+        color: colors.darkGray2,
+        backgroundColor: colors.lightGray2,
+      },
+      ":disabled": {
+        border: colors.lightGray2,
+        backgroundColor: colors.lightGray2,
+      },
+    },
     // edge
     edge === "rounded" && {
       borderRadius: "4px",
```

```tsx
// In LearningResourceExpanded

const SelectableButton = styled(Button)<{
  selected?: boolean
  variant: "bordered"
}>(({ selected, theme, variant }) => [
  variant === "bordered" &&
    selected && {
      backgroundColor: theme.custom.colors.red,
      color: theme.custom.colors.white,
      borderColor: theme.custom.colors.red,
    },
])
```

gumaerc: Yea, this makes sense and I kind of figured it was the case. However, @sovsey @mbilalmughal, it would be nice if you could directly notify us when you are adding new button variants so we can create a separate issue to add them to the codebase first, rather than having to shoehorn them into unrelated PR's such as this one.

ChristopherChudzicki:
> This is even better, but we don’t have this kind of floating notification anywhere. I wish we could add this—it looks really good, IMO.

This is available if we want it: https://mui.com/material-ui/react-snackbar/

frontends/ol-components/src/components/LearningResourceExpanded/LearningResourceExpandedV2.tsx
```diff
@@ -413,55 +509,112 @@ const CallToActionSection = ({
...
+            <ShareInput value={shareUrl} />
+            <ShareButtonContainer>
+              <ShareLink
+                href={`${facebookShareBaseUrl}?u=${encodeURIComponent(shareUrl)}`}
+                target="_blank"
+              >
+                <RiFacebookFill size={socialIconSize} />
+              </ShareLink>
+              <ShareLink
+                href={`${twitterShareBaseUrl}?text=${encodeURIComponent(resource.title)}&url=${encodeURIComponent(shareUrl)}`}
```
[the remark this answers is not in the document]
gumaerc: @sovsey This is actually something I meant to ask about. I put `text` in here because otherwise the tweet is just the URL in the text field, but we probably want something other than the resource title:

image

It's a bit different to the interface on other platforms like LinkedIn which automatically detect that you're sharing a link and exclude it from the text content of the post:

image

So, for Twitter at least, we probably want to frontload some text as that's more traditionally how a tweet is written anyway. However, when you actually go to post a tweet the URL is obfuscated and replaced with the typcial image based social media preview. So, we could leave it as-is but I feel like most people add some text there like "Check out this post on MIT Learn!"

gumaerc:
@sovsey @mbilalmughal I'm aware of right clicking and viewing the component. What we're talking about here is a process issue. If new variants are added in Figma, there should be an accompanying Github issue and we should add the variant as its own PR. That way, if there's an issue with the variant, it's easily searchable in the Github history and we can see the code that was used without other unrelated code like the share button. It also makes it easier to revert if there's a problem with it.

Anytime you find yourself adding a new variant for something, just create a Github issue for it and let us know which component and what the variant is called so we can get it in the codebase before work gets done on issues like this one that actually use it.

frontends/ol-components/src/components/LearningResourceExpanded/LearningResourceExpandedV2.tsx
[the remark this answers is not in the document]
sovsey: Let's have just the title and URL (as you have now) for Twitter/X. Was browsing a few new sites, and that seems to be typical at least.

ChristopherChudzicki:
@mbilalmughal I'm still not sure what the style of the "selected" state should be when hovered:

Screenshot 2024-12-12 at 4 59 21 PM

gumaerc:
@ChristopherChudzicki This is all set for another look. I do think we should discuss the -4px, 8px margins that get set on `Button` elements with icons with @mbilalmughal. Figma shows a 4px separation between the icon and the text:

image

But, The button component sets a -4px margin on the left and an 8px margin on the right to "compensate for padding on the button itself" but there still seems to be 8 pixels in between the icon and the text:

image

Compared to changing it to 0px on the left, 4px on the right:

image

cc @sovsey

mbilalmughal:
> @mbilalmughal I'm still not sure what the style of the "selected" state should be when hovered:
> 
> Screenshot 2024-12-12 at 4 59 21 PM

@ChristopherChudzicki We should avoid adding a hover state to the selected state. It should function as follows:

1. As we know, the Default state already has the Hover
2. If click - Selected State - Red
3. If clicked again - revert to the Default  state
CC: @steven-hatch

gumaerc:
@mbilalmughal No worries. You can see the changes in action here: https://learn.dev.c4103.com/search?resource=12638

mbilalmughal:
> @mbilalmughal No worries. You can see the changes in action here: https://learn.dev.c4103.com/search?resource=12638

Why, carey WHY?

image

gumaerc:
For what it's worth, what @mbilalmughal is seeing above with the V1 drawer is because he had an ad blocker enabled. Ad blockers will prevent the site from talking to Posthog and determining that the feature flag is enabled.

frontends/ol-components/src/components/LearningResourceExpanded/LearningResourceExpandedV2.tsx
```diff
@@ -413,55 +502,132 @@ const CallToActionSection = ({
...
+          >
+            {shareLabel}
+          </CallToActionButton>
+        </ButtonContainer>
+        {shareExpanded && shareUrl && (
+          <ShareContainer data-testid="drawer-share">
+            <ShareLabel>Share a link to this Resource</ShareLabel>
+            <Input
+              fullWidth
+              ref={inputRef}
```
ChristopherChudzicki, line 579:
```suggestion
              inputRef={inputRef}
```
would simplify the stuff below.

Actually, though: You don't need a ref at all. The click handler could be
```ts
(event) => {
  const insputEl = event.currentTarget.querySelector("input")
  if (!inputEl) return
  input.select()
}

File: frontends/ol-components/src/components/LearningResourceExpanded/LearningResourceExpandedV2.tsx [hunk 4]
```diff
@@ -194,11 +207,68 @@ const OnPlatform = styled.span({
...
+const ButtonContainer = styled.div({
   display: "flex",
+  width: "100%",
   gap: "8px",
   flexGrow: 1,
-  justifyContent: "flex-end",
+  justifyContent: "center",
+})
+
+const SelectableButton = styled(Button)<{ selected?: number }>((props) => [
```
ChristopherChudzicki, line 218: Could you make `selected` a boolean?

ChristopherChudzicki (changes requested):
Looks good. Noticed A few minor things.

mbilalmughal:
@gumaerc few spacing adjustments

1-The gap between the description and the metadata is 24px, but when the description is shorter, the gap increases too much. Can we do something about this?
image

Browser 
image


2-I suggest using a gap of 12px instead of 16px, as the browser's rendering of the metadata row height is 24px. In Figma, it's shown as 20px, which increases the overall gap between each line. Please check the following image.

image

3-The Bookmark and Share buttons should stretch all the way. Even if the Add to List button appears in some cases, it should adjust to fit in a single row.

image

4- The gap between Icon & text is 12px right now its 16px (12px is also acceptable value in our spacing rule)
image

5-Is there any way we could remove this gap? Only this one—the icon and label gap is fine, but this one is unnecessary.
I think it would fit nicely if we could remove that gap. Just to try it, I added margin-left: -12px, which was used as a gap, to see how it looks. I hope we can find a better fix, but this works fine as well.
image

6- In the course material drawer, when we don't have the provider name, the space should be smaller, as shown in Figma.
image

7- A small text cut is visible when the 'Show More' button appears.
image

image


@steven-hatch I have reservations about this empty space because we are showing the bookmark icon there, but it's wasting a lot of space and taking up a lot of vertical space."

image

frontends/ol-components/src/components/LearningResourceExpanded/LearningResourceExpandedV2.tsx
gumaerc: I tried this previously and it wasn't working, but seems to be working now. Previously, it was complaining about it being a boolean because React was writing the props as HTML attributes. This was probably because I was passing `selected` by itself, rather than with property expansion.

gumaerc:
@ChristopherChudzicki Thanks again for your feedback. This is ready for another look! I addressed the button width issue @mbilalmughal pointed out, but as we discussed the other feedback will go into the QA doc for triaging.

ChristopherChudzicki (approved):
👍 Looks nice.

PR #1897, opened 2024-12-11
add nprogress loading bar

ChristopherChudzicki:
Closes https://github.com/mitodl/hq/issues/6267

Adds a loading bar for internal navigations. 

This only triggers between different nextjs routes—not for client-side stuff, which we generally handle via skeleton uis.

Me navigating from homepage to `/departments`

Screenshot 2024-12-11 at 6 26 03 PM

### How can this be tested?
1. Try navigating between paths via links (e.g., in the left-hand nav drawer)
2. Try navigating between paths via `useRouter` (e.g., the homepage search box)
3. Try adding a new `useRouter` import from `next/navigation`; it should error.

### DIFF
--- a/frontends/.eslintrc.js
+++ b/frontends/.eslintrc.js
@@ -36,6 +36,11 @@ module.exports = {
     ...restrictedImports({
       patterns: [
         {
+          group: ["next/navigation"],
+          importNames: ["useRouter"],
+          message: "Please use `useRouter` from `next-nprogress-bar` instead.",
+        },
+        {
           group: ["@mui/material*", "@mui/lab/*"],
           message:
             "Please use 'ol-components' isInterfaceDeclaration; Direct use of @mui/material is limited to ol-components.",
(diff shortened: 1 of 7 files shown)


steven-hatch:
Looks good to me (based on yesterdays convo and screenshot). Approved, thanks Chris

jonkafton (approved):
Very effective and a nice simple solution - perfect!

Could we hook this into course-search-utils navigations? That's using `history.pushState()`, so maybe not so simple, though the search pages would really benefit from similar treatment. A separate issue perhaps - something like fading the outgoing search results could work.

Would be good to add for next/link navigations that set the search params without changing the route (e.g. select Courses from the nav drawer while already on the search page). That may have been the intention with `shallowRouting` (I'd expect that to enable it for shallow routing, not disable), though we'd need to remove / set to its default false.

ChristopherChudzicki:
> though the search pages would really benefit from similar treatment. A separate issue perhaps - something like fading the outgoing search results could work.

Better loading states for search page is on radar — i know we got some feedback about that recently.

PR #1900, opened 2024-12-12
add bordered button variant

gumaerc:
Closes https://github.com/mitodl/hq/issues/6277

This PR adds the `bordered` variant to the `Button` component and updates the Storybook stories to demonstrate its usage.

image
image

### How can this be tested?
 - Visit Storybook at http://localhost:6006
 - Verify that the `bordered` button variant behaves as expected in normal, hover and disabled states according to the designs linked in the issue

### DIFF
--- a/frontends/ol-components/src/components/Button/Button.tsx
+++ b/frontends/ol-components/src/components/Button/Button.tsx
@@ -10,11 +10,11 @@ type ButtonVariant =
   | "primary"
   | "secondary"
   | "tertiary"
-  | "text"
   | "bordered"
   | "noBorder"
   | "inverted"
   | "success"
+  | "text"
 type ButtonSize = "small" | "medium" | "large"
 type ButtonEdge = "circular" | "rounded" | "none"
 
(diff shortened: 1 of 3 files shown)


frontends/ol-components/src/components/Button/Button.tsx
```diff
@@ -188,6 +182,19 @@ const buildStyles = (props: ButtonStyleProps & { theme: Theme }) => {
...
+      color: colors.silverGrayDark,
+      border: `1px solid ${colors.silverGrayLight}`,
+      ":hover:not(:disabled)": {
+        backgroundColor: colors.lightGray1,
+      },
+      ":disabled": {
+        backgroundColor: colors.lightGray2,
+        border: `1px solid ${colors.silverGrayDark}`,
+        color: colors.silverGrayDark,
+      },
```
ChristopherChudzicki, line 196: The color darkens on hover, and the border when disabled should be same color as background. (You're right that Figma says the border color is silverGrayDark, but figma also says the border width is 0. I think setting the width to zero will mess up the height. I don't trust figma css :p) 
```suggestion
      ":hover:not(:disabled)": {
        backgroundColor: colors.lightGray1,
        color: colors.darkGray2,
      },
      ":disabled": {
        backgroundColor: colors.lightGray2,
        borderColor: colors.lightGray2
        color: colors.silverGrayDark,
      },
```

ChristopherChudzicki (approved):

PR #1905, opened 2024-12-16
Contentfile chunk embeddings

shanbady:

Closes https://github.com/mitodl/hq/issues/6271

This PR does the following:
- ensures that contentfiles are chunked before embedding
- introduces a flag that allows us to pass in a specific set of resources (and their associated contentfiles) to embed

### How can this be tested?
1. checkout this branch
2. restart celery
3. ensure that you have a good amount of contentfiles loaded locally
4. get a list of ids for resources with contentifiles and run the management command to embed them via ```python manage.py generate_embeddings --all --recreate-collections --resource-ids 1,2,3,4``` note the ids are comma separated
5. visit the [quadrant dashboard](http://localhost:6333/dashboard#/collections/resource_embeddings.content_files) and view the contentfiles collection. each record is a contentfile chunk.

### Additional Context
When generating the embeddings we conditionally load a the text splitter (RecursiveCharacterTextSplitter or TokenTextSplitter). Under the hood TokenTextSplitter uses the tiktoken library to ensure the chunks are within openai's token limits.

This pr also has a change that takes the embeddings from the contentfile chunks and associates them with their parent learning resource under a separate named vector as [multivectors](https://qdrant.tech/documentation/concepts/vectors/#multivectors). This sets us up to be able to take contentfiles embeddings into account when doing  learning resource vector searches (will need a tweak to the search endpoint to do this) 

related infra pr to switch us to using the large embedding model https://github.com/mitodl/ol-infrastructure/pull/2854

note: This pr will be followed up separately with a pr that exposes an endpoint to search contentfile chunks


gumaerc (approved):
I was able to limit `ContentFile` embedding generation using the `resource-ids` argument as described, and verified that each chunk is limited to 384, so everything here seems to work correctly.

Issue #1910, opened 2024-12-18
create initial qdrant collection on init


shanbady:
### Description/Context
The initial qdrant collections should get created by default in local development when bringing up containers.

### Plan/Design
run the management command ```python manage.py generate_embeddings --recreate-collections --all``` to populate the qdrant collections in scripts/run-django-dev.sh

PR #1911, opened 2024-12-18
create default collections on local dev initialization

shanbady:

Fixes https://github.com/mitodl/mit-learn/issues/1910

This PR introduces a new management command that gets called when initializing local environments to create qdrant collections if they dont exist.

### How can this be tested?
1. checkout this branch
2. run ```docker compose down web qdrant```
3. remove the existing qdrant volume (which removes all qdrant data) ```docker volume rm $(docker volume ls | grep qdrant )```
4. rebuild the web container so the changes in the init script are there ```docker compose build web```
5. bring up just the qdrant container ```docker compose up -d qdrant``` - visit [the dashboard](http://localhost:6333/dashboard#/collections) and confirm there are no  collections
6. bring up the web container and let it finish initializing ```docker compose up -d``` - visit the qdrant dashboard again and confirm collections exist

### Additional Context
I also fixed an issue with the prune subscription command in the init script

### DIFF
--- a/docker-compose.apps.yml
+++ b/docker-compose.apps.yml
@@ -20,6 +20,8 @@ services:
         condition: service_healthy
       redis:
         condition: service_healthy
+      qdrant:
+        condition: service_healthy
     volumes:
       - .:/src
       - django_media:/var/media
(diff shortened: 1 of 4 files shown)


gumaerc (approved):
:+1:

PR #1912, opened 2024-12-18
add topics carousels to v2 drawer

gumaerc:
Closes https://github.com/mitodl/hq/issues/5879

This PR adds a carousel to the new learning resource drawer for every `topic` related to the given resource being displayed, excluding the resource itself.

image
image

### How can this be tested?
 - Ensure that you have backpopulated some courses into your local database for testing
 - Click different types of courses and other resources, bringing up the new learning resource drawer
 - Verify that a carousel is displayed for each topic relevant to the resource below the similar resources carousel

### DIFF
--- a/frontends/main/src/page-components/LearningResourceDrawer/LearningResourceDrawerV2.test.tsx
+++ b/frontends/main/src/page-components/LearningResourceDrawer/LearningResourceDrawerV2.test.tsx
@@ -10,7 +10,8 @@ import LearningResourceDrawerV2 from "./LearningResourceDrawerV2"
 import { urls, factories, setMockResponse } from "api/test-utils"
 import { LearningResourceExpandedV2 } from "ol-components"
 import { RESOURCE_DRAWER_QUERY_PARAM } from "@/common/urls"
-import { ResourceTypeEnum } from "api"
+import { CourseResource, LearningResource, ResourceTypeEnum } from "api"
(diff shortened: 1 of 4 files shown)


shanbady (changes requested):
Works pretty well however I noticed that for certain topics there is a lot of cross-over and the carousels were not very different. for example:
Screenshot 2024-12-18 at 2 37 09 PM

Screenshot 2024-12-18 at 2 32 53 PM

if these are cases we care about you might be able to fix it by changing the excludeResourceId to an array fetching a few extra items to remove duplicates and fill the gaps

gumaerc:
@shanbady I discussed this with @sovsey and we are okay with duplicates for now. There isn't really a great way to exclude them, and the criteria we would use to do so is something to solve in and of itself. The current sort for topics carousels are by views descending, so it's going to show the most popular resources in each topic. If we start removing results from one topic and excluding from the ones below it, then the recommendations on the next carousel would be less relevant. Recommendation carousels on Amazon also have duplicates between carousels that sit right next to each other. In a way, this messages to the user that that particular resource is extra relevant because it is popular and relevant to all the subtopics it shows up in. What we will do is merge this as-is for now (unless you have other recommendations) and then evaluate what the carousels look like on RC and see if we want to make changes.

shanbady (approved):
👍

PR #1914, opened 2024-12-18
Gives LR Drawer role "dialog" and a name.

ChristopherChudzicki:
Closes https://github.com/mitodl/hq/issues/5418

This PR:
1. Gives the RoutedDrawer and hence learning resource drawer role "dialog" with an accessible name.
2. Improves the loading skeleton for the drawer title a little bit.

### How can this be tested?
I suggest comparing screenreader behavior on RC (or prod) behavior with this branch. I used VoiceOver.

With a screenreader activated:
1. Open the Learning resource drawer on RC or prod.
    - When the drawer first opens, it will read something about the carousel slides.
2. Now on this branch:
   - screenreader should read something like "Dialog, [ResourceType] ResourceTitle]".
3. If you have https://chromewebstore.google.com/detail/axe-devtools-web-accessib/lhdoppojpmngadmnindnejefpokejbdd or equivalent installed, you could check that the drawer no longer triggers "All page content must be included in a landmark role".


frontends/ol-components/src/components/RoutedDrawer/RoutedDrawer.tsx
```diff
@@ -80,11 +81,22 @@ const RoutedDrawer = <K extends string, R extends K = K>(
...
+      "[Dev-only Warning]: RoutedDrawer requires an aria-label or aria-labelledby prop."
+    invariant(props["aria-label"] || props["aria-labelledby"], msg)
+  }
+
   return (
     <Drawer
       open={open}
       onTransitionExited={removeUrlParams}
       onClose={setOpen.off}
+      role="dialog"
```
ChristopherChudzicki, line 98: So, I dunno why MUI doesn't put role=dialog on drawer by default. 

Bootstrap does, github does it on its drawers, and it resolves the accessibility checker warning because 
- https://github.com/dequelabs/axe-core/issues/991

Additionally, the drawer title is now read aloud when users open it, which is a big improvement.

gumaerc (approved):
These changes make sense, and if you've tested them on Voiceover and it works well then I think we should merge it. I tested this with NVDA, the free screen reader for Windows, When you click a card to bring up the drawer, it announces "dialog" like you talk about here, but then after that it still announces the close button first, then the content of the recommended resources carousel. It goes on and on announcing all of the UI components in the carousel, repeating itself about "groupings" over and over. I'm not sure how we would fix... maybe some aria attributes on the carousel itself?

ChristopherChudzicki:
> These changes make sense, and if you've tested them on Voiceover and it works well then I think we should merge it. I tested this with NVDA, the free screen reader for Windows, When you click a card to bring up the drawer, it announces "dialog" like you talk about here, but then after that it still announces the close button first, then the content of the recommended resources carousel. It goes on and on announcing all of the UI components in the carousel, repeating itself about "groupings" over and over. I'm not sure how we would fix... maybe some aria attributes on the carousel itself?

Discussed this a bit with Carey, and thanks @gumaerc for helping test this on Windows screenreaders. Conclusions:
- On Mac with VoiceOver, the drawer was also behaving a little different when "first opened" vs "re-opened".
- This, and the title quirks carey noted, seem to be because the drawer doesn't initially have all of its content. (Since the `learning_resources/:id` api call is still resolving).

I wish I understood better the right way to deal with this. Especially the fact that the drawer **title** doesn't necessary exist until the API call resolves. I looked at a few possible solutions, including:
- hiding the drawer via `aria-hidden={isLoading}` until the drawer content finishes loading
- fallback titles while drawer is loading
- using `role="region"` instead

None of these worked super well. Since the fundamental problem is that the drawer data isn't available until the API call finishes resolving...I decided to just prepopulate the query cache. In all cases, we have the resource data somewhere, so we generally don't need to load it again. See other comment.

frontends/main/src/page-components/ResourceCard/ResourceCard.tsx
```diff
@@ -54,7 +55,10 @@ const useResourceCard = (resource?: LearningResource | null) => {
       }
     }, [user])
 
+  const onClick = useLearningResourceDetailSetCache(resource)
```
ChristopherChudzicki, line 58: I considered various possible ways to pre-populate the `useLearningResourceDetail` data, e.g., 
- inside `useLearningResourceDetail` hook, search the query cache for the resource
    - hard because the query cache stores lots of different shapes of data, and we generally don't know during search whether something is a resource
- prepopulate the details cache whenever we make a listing call
    - hard because there are so many different listing calls

In the end, setting it here seemed easiest. We have many different versions of "listings" (featured courses, various other carousels, search results, userlist resources...) but generally it's pretty rare that we use a details. And currently the only time we do, it is triggered via the resource cards.

ChristopherChudzicki: bonus side effect... drawer loads way faster for everyone!

frontends/ol-components/src/components/LearningResourceExpanded/InfoSectionV2.tsx
```diff
@@ -49,7 +49,7 @@ const SeparatorContainer = styled.span({
  * breaks after the pipe but not before it.
  */
 const Separator: React.FC = () => (
-  <SeparatorContainer>|&#8203;</SeparatorContainer>
+  <SeparatorContainer aria-label=", ">|&#8203;</SeparatorContainer>
```
ChristopherChudzicki, line 52: as-was, this was reading "vertical line" a lot.

gumaerc (approved):
Tested again with Windows Narrator and NVDA with positive results; every time I opened a card (including the first time) the title and description was properly read.

PR #1916, opened 2024-12-18
remove unncessary wrapper

ChristopherChudzicki:
Closes https://github.com/mitodl/hq/issues/6325

This PR removes an unnecessary wrapper around an icon. Previously the wrapping div had an aria-label; this is technically disalllowed and causes some accessibility checkers to issue a warning.

Now the aria-label is on an svg, which is 👍 .

### Screenshots
Small cards with a certificate are "affected", but there is no visual change:

Screenshot 2024-12-18 at 3 28 52 PM

### How can this be tested?
1. View a small card with a certificate on RC. These can be found (possibly) in your dashboard, or in the "Similar Resources" carousels inside the learning resource drawer. Check the certificate icon dimensions:
    - svg is 12px by 12px
    - so is the div wrapping it
2. Do same locally; svg should still be 12px by 12px. The div wrapping it is gone. The SVG itself now has aria-label.
3. Additionally / Optionally, you could check that an accessibility checker like [axe devtools ](https://chromewebstore.google.com/detail/axe-devtools-web-accessib/lhdoppojpmngadmnindnejefpokejbdd) no longer flags the aria-label as invalid. (Run it with the drawer open).

### DIFF
--- a/frontends/ol-components/src/components/LearningResourceCard/LearningResourceCard.tsx
+++ b/frontends/ol-components/src/components/LearningResourceCard/LearningResourceCard.tsx
@@ -79,9 +79,7 @@ const Info = ({
           <Certificate>
             {size === "small" ? (
               <Tooltip title="Certificate">
-                <CertificateIconContainer>
-                  <RiAwardFill />
-                </CertificateIconContainer>
+                <RiAwardFill />
               </Tooltip>
             ) : (
               <RiAwardFill />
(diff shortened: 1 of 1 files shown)


ChristopherChudzicki:
The JS tests will pass after
- https://github.com/mitodl/mit-learn/pull/1917
is merged.

The Python test failure is flakiness, i guess.

shanbady (approved):
👍

PR #1917, opened 2024-12-18
Update version of react used in tests to same as production

ChristopherChudzicki:
None

_Related to https://github.com/mitodl/mit-learn/pull/1916, which is failing artificially due to incorrect version of react._

This PR updates the version of React to ^19.

**This is the version of react used in our JSDOM tests, not production.** NextJS bundles react internally, and we're on NextJS 15. It already uses react 19. [^1]

[^1]: Our version of NextJS actually uses a canary react 19, namely `19.0.0-rc-02c0e824-20241028`, which you can see if you add a `console.log(React)` into the application somewhere. Projects like NextJS using canary react is expected by the react developers. See https://react.dev/blog/2023/05/03/react-canaries. We should update to a newer version of NextJS 15, which will use a non-canary react, but let's do that separately.

### How can this be tested?
CI should pass. (This change does not affect production.)


frontends/main/src/test-utils/index.tsx
```diff
@@ -124,7 +124,7 @@ const expectLastProps = (
 ) => {
   expect(fc).toHaveBeenLastCalledWith(
     expect.objectContaining(partialProps),
-    expect.anything(),
+    expect.toBeOneOf([expect.anything(), undefined]),
```
ChristopherChudzicki, line 127: Evidently how react calls functional components changed a little between react 18 -> 19.

It used to call them as `func(props, something_i_didnt_understand)`, now it calls them as `func(props, undefined)`.  This is irrelevant and is is already what's happening in MIT Learn on RC/Prod.

jonkafton (approved):
:+1:

PR #1918, opened 2024-12-19
Demo Course Recommendation Chatbot (not production-ready, for RC only)

mbertrand:
Related to https://github.com/mitodl/hq/issues/6126

Implements a rudimentary chatbot that recommends learning resources based on the user's input.
Hidden behind a posthog feature flag that is enabled for RC only.
Lacking lots of unit tests, and the final implementation will be very different, but this needs to be available as a demo ASAP.

- [ ] Desktop screenshots

### How can this be tested?
- Use the same `POSTHOG_*` env settings as RC, otherwise the functionality will be disabled.  Turn off adblockers, they don't like posthog.
- Set the `OPENAI_API_KEY` to a value (I can send one to you)
- Go to http://open.odl.local:8062/chat
- Try it out, it should work, there will be debug info at the bottom.
- Comment out your posthog env settings, then reload the page.  It should be blank.

### Additional Context
Demo purposes only, not intended for production


abeglova (approved):
LGTM

PR #1920, opened 2024-12-20
Cg/fix podcast cta

gumaerc:
Closes https://github.com/mitodl/hq/issues/6264

This PR adds an `audio_url` field to the `PodcastEpisode` model and explicitly fills it with `rss_data.enclosure["url"]` during ETL. The `url` property of the `LearningResource` associated with a `PodcastEpisode` now defaults to `rss_data.link.text` (aka `episode_link`) and falls back to `audio_url` if that's not available. The frontend was set up to waterfall in the order of `episode_link` > `audio_url` > `url`.

### How can this be tested?
 - Run `docker compose exec web ./manage.py backpopulate_podcast_data` and wait for it to finish
 - Go to the search page at http://localhost:8062/ and search for "J. Mijin Cha, "A Just Transition for All: Workers and Communities for a Carbon-Free Future" (MIT Press, 2024)"
 - Click on the search result to bring up the drawer
 - Click on "Listen to Podcast" and verify that it links directly to the MP3 file for the podcast episode


frontends/ol-components/src/components/LearningResourceExpanded/LearningResourceExpandedV2.tsx
```diff
@@ -410,7 +410,11 @@ const ImageSection: React.FC<{
...
   switch (resource.resource_type) {
     case ResourceTypeEnum.PodcastEpisode:
-      return resource.podcast_episode?.episode_link
+      return (
+        resource.podcast_episode?.episode_link ||
+        resource.podcast_episode?.audio_url ||
+        resource.url
+      )
     default:
       return resource.url
```
ChristopherChudzicki, line 419: Now that podcasts use `"url": episode_link or audio_url, the switch statement seems unnecessary to me.

In fact, could we just get rid of `getCallToActionUrl` and do `href={resource.url}` in the component?

learning_resources/models.py
```diff
@@ -980,6 +980,7 @@ class PodcastEpisode(LearningResourceDetailModel):
     )
 
     transcript = models.TextField(blank=True, default="")
+    audio_url = models.URLField(null=True, max_length=2048)  # noqa: DJ001
```
ChristopherChudzicki, line 983: Could we make this `null=False`?

Both of these things indicate it's required:
- https://github.com/Podcast-Standards-Project/PSP-1-Podcast-RSS-Specification?tab=readme-ov-file#required-item-elements
- https://podcasters.apple.com/support/823-podcast-requirements 
    > Each episode must have a unique [<enclosure>](https://help.apple.com/itc/podcasts_connect/#/itcb54353390) tag, along with its three required components — URL, length, and type ...

ChristopherChudzicki (changes requested):
Two small requests

ChristopherChudzicki (approved):
👍

PR #1936, opened 2025-01-02
Import AI agent within view post code

mbertrand:
N/A

Should improve api performance

### How can this be tested?
Hit multiple pages of `http://api.open.odl.local/api/v1/learning_resources` and also detail urls.  The response should almost always be < 1 second.   If you try the same on the main branch, the response will almost always be 4-5 seconds.

Also check that `http://open.odl.local:8062/chat` still works (you'll need to copy the POSTHOG and OPENAI_API_KEY values from RC)

### DIFF
--- a/ai_chat/apps.py
+++ b/ai_chat/apps.py
@@ -1,4 +1,4 @@
-"""learning_resources app config"""
+"""ai_chat app config"""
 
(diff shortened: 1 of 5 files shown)


rhysyngsun (approved):
LGTM

PR #1945, opened 2025-01-07
dedupe edx content files further

abeglova:
closes https://github.com/mitodl/hq/issues/6343

The edx course folders contain duplicates of some files. This PR updates the file etl to dedupe files that have the same run and content 

### How can this be tested?
If you don't have xpro files already populated, run 
Run `docker-compose run web ./manage.py backpopulate_xpro_data` , `docker-compose run web ./manage.py backpopulate_xpro_files --overwrite`
in the main branch

From the shell run
```
ContentFile.objects.filter(run__learning_resource__platform_id='xpro').count()
ContentFile.objects.filter(checksum='c701f34e5593021eec27db37561c45de').count()
```

You should have about 7187 xpro content files and 33 content files with checksum c701f34e5593021eec27db37561c45de

In this branch run 
 `docker-compose run web ./manage.py backpopulate_xpro_files --overwrite`

From the shell run
```
ContentFile.objects.filter(run__learning_resource__platform_id='xpro').count()
ContentFile.objects.filter(checksum='c701f34e5593021eec27db37561c45de').count()
```

You will have about 5555 xpro content files and 1 file with checksum c701f34e5593021eec27db37561c45de

### DIFF
--- a/learning_resources/etl/utils.py
+++ b/learning_resources/etl/utils.py
@@ -330,16 +330,17 @@ def documents_from_olx(
                     filebytes = f.read()
 
                 mimetype = mimetypes.types_map.get(extension_lower)
+                checksum = md5(filebytes).hexdigest()  # noqa: S324
 
                 yield (
                     filebytes,
                     {
-                        "key": f"{path}/{filename}",
+                        "key": checksum,
                         "content_type": CONTENT_TYPE_FILE,
                         "mime_type": mimetype,
-                        "checksum": md5(filebytes).hexdigest(),  # noqa: S324
+                        "checksum": checksum,
                         "file_extension": extension_lower,
-                        "source_path": path,
+                        "source_path": f"{path}/{filename}",
                     },
                 )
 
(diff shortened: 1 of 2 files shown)


learning_resources/etl/utils.py
```diff
@@ -330,16 +330,17 @@ def documents_from_olx(
                     filebytes = f.read()
 
                 mimetype = mimetypes.types_map.get(extension_lower)
+                checksum = md5(filebytes).hexdigest()  # noqa: S324
 
                 yield (
                     filebytes,
                     {
-                        "key": f"{path}/{filename}",
+                        "key": checksum,
```
shanbady, line 338: curious if there are unexpected cases here since it seems like the "key" is unique with the associated run. are there cases where we could have 2 separate files in the same run that have the same content?

shanbady (changes requested):
testing it seemed to work as intended. left minor note about potential edge case

learning_resources/etl/utils.py
abeglova: The scenario you are describing is exactly what this pr is fixing - we found that there are duplicate files with different file names and the same content in lots of edx course folders. This pr fixes that issue.

There is an edge case where files with different content that isn't parsed fully by tikka for whatever reason might end up looking like the same (almost blank) file. But since those files don't have useful content anyway, that's ok

shanbady (approved):
LGTM

PR #1949, opened 2025-01-09
Dashboard Tab Keyboard Navigation

ChristopherChudzicki:
Closes https://github.com/mitodl/hq/issues/6405

Makes dashboard navigation tabs keyboard navigable.

Also:
- updates some `@testing-library` dependencies. I was getting a bunch of console errors during tests, and the updates fixed it.

There should be no visible difference.

**Before / After**
Screenshot 2025-01-09 at 2 39 07 PM Screenshot 2025-01-09 at 2 38 23 PM

Screenshot 2025-01-09 at 2 39 23 PM Screenshot 2025-01-09 at 2 39 44 PM

### How can this be tested?
Check that the dashboard tabs are keyboard navigable. You should be able to tab to them, and navigate between them with left/right arrows.


frontends/main/src/app-pages/DashboardPage/Dashboard.test.tsx
```diff
@@ -226,41 +223,28 @@ describe("DashboardPage", () => {
...
+    const desktopTabs = within(desktopTabList).getAllByRole("tab")
+    const mobileTabs = within(mobileTabList).getAllByRole("tab")
+
+    // Check URLs
+    expect(desktopTabs.map((el) => el.getAttribute("href"))).toEqual(urls)
+    expect(mobileTabs.map((el) => el.getAttribute("href"))).toEqual(urls)
+
+    // Check labels
+    expect(desktopTabs.map((el) => el.textContent)).toEqual(labels)
+    expect(mobileTabs.map((el) => el.textContent)).toEqual(labels)
```
ChristopherChudzicki, line 247: We used to test the href and label separately, but 90% of the test was the same, so might as well do it together.

frontends/main/src/app-pages/DashboardPage/DashboardPage.tsx
```diff
@@ -277,38 +277,65 @@ const TabLabels = {
...
+const DesktopTabLabel: React.FC<{
   icon: React.ReactNode
-  text: string
-  tabKey: string
-  value: string
-  currentValue: string
-  onClick?: () => void
-}
-
-const UserMenuTab: React.FC<UserMenuTabProps> = (props) => {
```
ChristopherChudzicki, line 289: The main issue was that, apparently, if you wrap the MUI tab component in your own custom component, you need to spread props, so something like:
```tsx
const UserMenuTab = ({ icon, text, tabKey, value, currentValue, onClick , ...others }) => {
    /* What we have now */
    return ( <Tab /*What we have now*/ {...others} />)
} 
```
would've worked.

But I just removed our custom wrapper and did a little refactoring.

See https://github.com/mui/material-ui/issues/27947#issuecomment-905318861

gumaerc (approved):
👍

PR #1951, opened 2025-01-09
fix search text

abeglova:
closes https://github.com/mitodl/hq/issues/6403

Currently hitting the back button changes search text in the url but not the text box. This fixes the issue.

### How can this be tested?
1) Go to http://open.odl.local:8062/search
2) Make a text search
3) Search for some different text
4) Click the back button

Your search page should show the text from step 2 and the appropriate results

Go to a channel page (such as http://open.odl.local:8062/c/unit/mitx) and repeat the above

### DIFF
--- a/frontends/main/src/app-pages/ChannelPage/ChannelSearch.tsx
+++ b/frontends/main/src/app-pages/ChannelPage/ChannelSearch.tsx
@@ -1,4 +1,4 @@
-import React, { useCallback, useMemo } from "react"
+import React, { useCallback, useMemo, useEffect } from "react"
 import { ChannelTypeEnum } from "api/v0"
 import { useOfferorsList } from "api/hooks/learningResources"
 import { useResourceSearchParams } from "@mitodl/course-search-utils"
@@ -94,6 +94,10 @@ const ChannelSearch: React.FC<ChannelSearchProps> = ({
     onFacetsChange,
   })
   const page = +(searchParams.get("page") ?? "1")
+
+  useEffect(() => {
+    setCurrentText(params.q ?? "")
+  }, [params, setCurrentText])
 
   return (
     <section>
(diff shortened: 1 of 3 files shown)


frontends/main/src/app-pages/ChannelPage/ChannelSearch.tsx
```diff
@@ -95,6 +95,10 @@ const ChannelSearch: React.FC<ChannelSearchProps> = ({
   })
   const page = +(searchParams.get("page") ?? "1")
 
+  useMemo(() => {
+    setCurrentText(searchParams.get("q") ?? "")
+  }, [searchParams])
```
ChristopherChudzicki, line 100: **Suggestion:** `useMemo` for calculating values, `useEffect` for side effects.

So I think here we want:
```ts
  useEffect(() => {
    setCurrentText(params.q ?? "")
  }, [params.q, setCurrentText])
```
- `params.q` is the the q param, the actual search query
- current text is what user has typed, maybe not yet submitted
- when submitted query changes (e.g., back navigation), make current text match it.

ChristopherChudzicki (changes requested):
Works well 👍  but could you add a small test to `SearchPage.test.tsx` that fails on main and succeeds here? I think something like
```tsx
  test.only("routing to new query sets currentText", () => {
    setMockApiResponses({})
    renderWithProviders(<SearchPage />)
    // get input, assert empty
    act(() => {
      // Navigate to a new query:
      window.history.pushState({}, "", "?q=meow")
    })
    // Check input has value "meow"
    // might need `await waitFor...`, I' not sure
    expect(textInput).toBe("meow")
  })
```

ChristopherChudzicki (approved):

ChristopherChudzicki (approved):

PR #1953, opened 2025-01-10
Fix Posthog multiple inits

jonkafton:

Relates to https://github.com/mitodl/hq/issues/6290

Whilst looking into initial page load performance and script blocking, Posthog showed up as significantly adding to the Total Blocking Time.

During build, we see:

```
[PostHog.js] was already loaded elsewhere. This may cause issues.
[PostHog.js] was already loaded elsewhere. This may cause issues.
[PostHog.js] was already loaded elsewhere. This may cause issues.
[PostHog.js] was already loaded elsewhere. This may cause issues.
[PostHog.js] was already loaded elsewhere. This may cause issues.
[PostHog.js] was already loaded elsewhere. This may cause issues.
[PostHog.js] was already loaded elsewhere. This may cause issues.
[PostHog.js] was already loaded elsewhere. This may cause issues.
[PostHog.js] was already loaded elsewhere. This may cause issues.
[PostHog.js] was already loaded elsewhere. This may cause issues.
[PostHog.js] was already loaded elsewhere. This may cause issues.
[PostHog.js] was already loaded elsewhere. This may cause issues.
```

Again we see the log line at runtime on each page load. It is indeed causing issues - running the profiler in Chrome Dev Tools, the flame chart shows this long blocking task with many calls into [rrweb](https://github.com/rrweb-io/rrweb), a library used by PostHog for session replay, here serializing the DOM: https://github.com/rrweb-io/rrweb/blob/52842ba09a975436c7748161887fde0efd875068/packages/rrweb-snapshot/src/snapshot.ts#L900. This profile was run at 20x CPU slowdown:

<br /><br />

image

<br /><br />

With this change to initialize Posthog once, which is in line with [Posthog's documentation for Next.js](https://posthog.com/docs/libraries/next-js), the long blocking task that shows up between the DOMContentLoaded and Onload events is gone entirely:

<br /><br />

image

### How can this be tested?

- Posthog should function as previously with feature flags applied and events reported.

- The long blocking task should not appear in the profiler flame chart (Performance tab in Chrome Dev Tools) - this shows up prominently on the main branch when the profile is set to 20x CPU slowdown. Next.js must be run in production mode (`yarn build; yarn start;`) and the site should be profiled in a private window to eliminate any browser extensions.

image

<br /><br />

This also shaves a second or two off the Onload timing (running at 20x CPU slowdown). Be mindful though of the initial server response time when testing as this is a significant factor in marker timings (here we can measure between First Paint and Onload).

### DIFF
--- a/frontends/main/src/page-components/ConfiguredPostHogProvider/ConfiguredPostHogProvider.tsx
+++ b/frontends/main/src/page-components/ConfiguredPostHogProvider/ConfiguredPostHogProvider.tsx
@@ -1,9 +1,9 @@
 import React, { useEffect } from "react"
+import posthog from "posthog-js"
 import { PostHogProvider, usePostHog } from "posthog-js/react"
 import { useUserMe } from "api/hooks/user"
-import type { PostHogConfig } from "posthog-js"
 
-const PosthogIdenifier = () => {
+const PosthogIdentifier = () => {
   const { data: user } = useUserMe()
   const posthog = usePostHog()
   /**
@@ -28,30 +28,37 @@ const PosthogIdenifier = () => {
   return null
 }
 
+const POSTHOG_API_KEY = process.env.NEXT_PUBLIC_POSTHOG_API_KEY
+const POSTHOG_API_HOST = process.env.NEXT_PUBLIC_POSTHOG_API_HOST
+
 const ConfiguredPostHogProvider: React.FC<{ children: React.ReactNode }> = ({
   children,
 }) => {
-  const apiKey = process.env.NEXT_PUBLIC_POSTHOG_API_KEY || ""
-  const apiHost =
-    process.env.NEXT_PUBLIC_POSTHOG_API_HOST || "https://us.i.posthog.com"
-  const featureFlags = JSON.parse(process.env.FEATURE_FLAGS || "")
+  useEffect(() => {
+    const featureFlags = JSON.parse(process.env.FEATURE_FLAGS || "")
 
-  const postHogOptions: Partial<PostHogConfig> = {
-    api_host: apiHost,
-    bootstrap: {
-      featureFlags: featureFlags,
-    },
+    if (POSTHOG_API_KEY) {
+      console.info("Initializing PostHog")
+      posthog.init(POSTHOG_API_KEY, {
+        api_host: POSTHOG_API_HOST,
+        bootstrap: {
+          featureFlags,
+        },
+      })
+    }
+  }, [])
+
+  if (!POSTHOG_API_KEY) {
+    return children
   }
 
-  return apiKey ? (
-    <PostHogProvider apiKey={apiKey} options={postHogOptions}>
-      <PosthogIdenifier />
+  return (
+    <PostHogProvider client={posthog}>
+      <PosthogIdentifier />
       {children}
     </PostHogProvider>
-  ) : (
-    children
   )
 }
 
 export default ConfiguredPostHogProvider
-export { PosthogIdenifier }
+export { PosthogIdentifier }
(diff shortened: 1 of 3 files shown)


frontends/main/src/page-components/ConfiguredPostHogProvider/ConfiguredPostHogProvider.tsx
```diff
@@ -28,30 +28,36 @@ const PosthogIdenifier = () => {
...
   children,
 }) => {
-  const apiKey = process.env.NEXT_PUBLIC_POSTHOG_API_KEY || ""
-  const apiHost =
-    process.env.NEXT_PUBLIC_POSTHOG_API_HOST || "https://us.i.posthog.com"
-  const featureFlags = JSON.parse(process.env.FEATURE_FLAGS || "")
+  useEffect(() => {
+    const featureFlags = JSON.parse(process.env.FEATURE_FLAGS || "")
+
+    if (POSTHOG_API_KEY) {
```
ChristopherChudzicki, line 40: Could we initialize the client outside the component rather than in the effect?

Then it's 100% obvious it only runs once, even in strict mode.

ChristopherChudzicki: **Request:** Unless there's a reason not to, I think we should initialize posthog in the outer scope, not in a useEffect. 

I just double checked and if you add a:
```
const posthog = usePosthog()
useEffect(() => {
  console.log(`Child says ${posthog.__loaded}`)
},[posthog])
```
to a child of `PostHogProvider`, you'll see `"Child says false"` logged. Since child effects run before parent effects, the child effect runs with un-initialized posthog.

ChristopherChudzicki (changes requested):
**Comment:** I guess the reason that this was double-initializing was... PosthogProvider initializes [inside](https://github.com/PostHog/posthog-js/blob/38dfb96357e25b22f11cdde5dc85a1c1c4bf6e38/react/src/context/PostHogProvider.tsx#L18) a `useMemo`, and `useMemo` isn't guaranteed to run more often than it needs to.

I remember reading that `useMemo` might run more often than it needs to, but I don't remember actually ever observing that. But you can see it here if you put `useMemo(()=>console.log('Provider Memo'), [])` in the provider. It runs several times. Weirdly if you put it lower in the tree, you'll only see it log once.

So, yeah, I guess we should listen when React says not to count it on for single-running.

frontends/main/src/page-components/ConfiguredPostHogProvider/ConfiguredPostHogProvider.tsx
jonkafton: Concern realized - once the `posthog.init()` is moved out to the module scope, we again see the long blocking task. It appears that when initialized in `useEffect()`, the DOM is not serialized for session replay, or at least not during the initial page load. The Posthog docs do instruct us to initialize inside `useEffect()`. I'm not 100% on whether initializing it here is broken and we don't get the session replay feature and hence don't see the blocking task, or whether it should actually be initialized in the `useEffect()` where session replay works fine but is handled without blocking the main thread. You're observation with the `posthog.__loaded` does indicate that either their docs are incorrect or that PostHog not being loaded when components first mount is the correct behavior.

I'm hesitant to conclude that their docs are wrong, though testing on a personal PostHog instance, I only see session replay when initialized in the outer scope.

We can disable session recording in the config to remove the feature and associated performance hit. I'll add an issue to set by environment variable, as discussed. These can also be enabled by feature flag: https://posthog.com/docs/session-replay/how-to-control-which-sessions-you-record#with-feature-flags

ChristopherChudzicki (approved):
👍 

I didn't realize their docs showed it in useEffect—happy to revert to that if you wanna try it on RC. I don't think it should matter for their session replay, and i strongly expect they de-prioritize recordings for free accounts

PR #1957, opened 2025-01-13
make useSearchParams mock return stable reference

ChristopherChudzicki:


frontends/ol-test-utilities/src/mocks/nextNavigation.ts
```diff
@@ -73,7 +74,14 @@ export const nextNavigationMocks = {
...
     const url = new URL(router.asPath, "http://localhost")
-    return url.searchParams
+    // Ensure the reference to the URLSearchParams object is stable
+    // across renders.
+    // NextJS guarantees this.
+    const ref = useRef<URLSearchParams>(url.searchParams)
+    if (url.searchParams.toString() !== ref.current.toString()) {
+      ref.current = new URLSearchParams(url.searchParams)
+    }
+    return ref.current
```
ChristopherChudzicki, line 84: @abeglova in NextJS (and react router) `useSearchParams` returns a "stable" object reference:
```ts
const s1 = useSearchParams()
const s2 = useSearchParams()
```
`s1 === s2`, and the object reference will only change if the search URL actually updates.

the mock router we were using didn't behave this way, which caused your artificial test failures.

PR #1959, opened 2025-01-14
drawer v2 design review fixes

gumaerc:
Closes https://github.com/mitodl/hq/issues/6294

This PR corrects a number of small spacing problems noted in the above issue, discovered during QA testing of the V2 drawer.

image
image
image

### How can this be tested?
 - Ensure that you have backpopulated some courses and resources like podcasts into your local database for testing
 - Bring up the issue above and go through each example, verifying that it has been addressed

### DIFF
--- a/frontends/main/src/page-components/LearningResourceDrawer/LearningResourceDrawerV2.tsx
+++ b/frontends/main/src/page-components/LearningResourceDrawer/LearningResourceDrawerV2.tsx
@@ -168,7 +168,7 @@ const PAPER_PROPS: RoutedDrawerProps["PaperProps"] = {
     }),
     minWidth: (theme) => ({
       [theme.breakpoints.down("md")]: {
-        minWidth: "100%",
+        maxWidth: "100%",
       },
     }),
   },
--- a/frontends/ol-components/src/components/Carousel/Carousel.test.tsx
+++ b/frontends/ol-components/src/components/Carousel/Carousel.test.tsx
@@ -142,8 +142,8 @@ describe("Carousel", () => {
   })
 
   test.each([
-    { initialIndex: 2, finalIndex: 5, nextDisabled: false },
-    { initialIndex: 4, finalIndex: 7, nextDisabled: true },
+    { initialIndex: 1, finalIndex: 5, nextDisabled: false },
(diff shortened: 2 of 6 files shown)


gumaerc:
@shanbady IMO this is normal behavior when there aren't enough results to show, but I can understand why we would only want to show complete slides here. We could limit it to 8 instead of 12, but the same problem would happen if the endpoint returned less than 8 results. The original problem here was that clicking the next slide arrow was only skipping forward 3 items, showing you the last item from the first slide first on the second slide and so on. This leaves an empty space at the end even if you are displaying 4 cards at a time and have 12 total, because you are skipping forward 3 at a time. It keeps skipping forward 3 (one less than the cards per slide) to show you that context. I don't think that the "context" here is particularly helpful since it's only a couple slides of results. However, as part of this PR I have made a change to the `Carousel` component globally so that `slickNext` advances by the amount of items per slide evenly. It's worth discussing if we want it to work this way globally.

shanbady (approved):
LGTM

gumaerc:
@shanbady I looked into this further, and the issue with the Carousel is a bit different than I thought. The `Carousel` component was deciding to only skip forward 3 cards at a time because of this code:

```ts
  const currentRect = current.getBoundingClientRect()
  const adjRect = adjacent.getBoundingClientRect()
  const containerRect = container.getBoundingClientRect()
  const itemWidth = currentRect.width
  if (!itemWidth) return { currentIndex, slidesPerPage: 1 }
  const gap = Math.abs(adjRect.x - currentRect.x) - itemWidth
  const fractional =
    Math.round(containerRect.width + gap) / Math.round(itemWidth + gap)
```

With the width of the drawer and the surrounding padding, the right card gets cut off by 4 pixels:
image

The `fractional` variable here in debugging ended up being 3.91, which is reduced to 3 by `Math.min` and `Math.floor` when calculating `slidesPerPage.

@mbilalmughal  - In order to fix this issue we need to either reduce the padding on the drawer as a whole or make the drawer slightly wider to accommodate.

mbilalmughal:
@gumaerc 

> horizontal paddings to 28px instead of 32px fixes it.

I think this should work. Increasing the drawer width might be risky since we would need to review all the content again, as the rest of the content would need to adjust to an additional 8px overall.

PR #1960, opened 2025-01-14
Onboarding Accessibility Improvements

ChristopherChudzicki:
Closes https://github.com/mitodl/hq/issues/6435

This makes screenreaders announce step changes during onboarding flow.

### How can this be tested?
With a screenreader enabled (I used VoiceOver on MacOS), as an authenticated user:
1. Visit http://localhost:8062/onboarding
2. Change the pages, (Forward and back) you should hear:
  - "Step X of Y"
  - The step title should be read aloud.
3. Throughout all, there should be no visible changes.

### Additional Context
MUI's stepper screenreader behavior isn't great. See https://github.com/mui/material-ui/issues/32826 for a good discussion. Unfortunately most of that is focused on MUI's  "nonlinear" stepper where you can click any step to navigate to it.

### DIFF
--- a/frontends/main/src/app-pages/OnboardingPage/OnboardingPage.test.tsx
+++ b/frontends/main/src/app-pages/OnboardingPage/OnboardingPage.test.tsx
@@ -52,7 +52,10 @@ const STEP_TITLES = [
   "Are you seeking a certificate?",
   "What is your current level of education?",
   "What course format are you interested in?",
-]
+].map((title, index) => ({
+  title,
+  step: index,
+}))
 
 const PROFILES_FOR_STEPS = times(STEPS_DATA.length, profileForStep)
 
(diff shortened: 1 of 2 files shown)


frontends/main/src/app-pages/OnboardingPage/OnboardingPage.test.tsx
```diff
@@ -87,166 +90,66 @@ const queryBackButton = () => screen.queryByRole("button", { name: "Back" })
 const queryFinishButton = () => screen.queryByRole("button", { name: "Finish" })
 
 describe("OnboardingPage", () => {
```
ChristopherChudzicki, line 92: This used to be organized as
```ts
describe("OnboardingPage", () => {
  describe("Page 1", () => {
    /* tests */
  })
  describe("Page 2", () => {
    /* tests */
  })
  // etc
})
```

The tests were really only checking navigation.

I moved it into `each` blocks so that if we add/remove pages it's easier to update.

abeglova (approved):
lgtm

PR #1966, opened 2025-01-16
Remove prolearn ETL code

mbertrand:
Closes https://github.com/mitodl/hq/issues/6272

Removes prolearn ETL code and management command.  The only reference left should be in the `ETLSource` enum (because there will still be some old unpublished resources in the database that came from this pipeline).

### How can this be tested?
The following should still work fine, as long as `MITPE_BASE_URL` and `SEE_*` env settings are the same as on RC:

```
./manage.py backpopulate_mitpe_data
```


abeglova (changes requested):
Why are we keeping the prolearn ETLSource and unpublished resources instead of just deleting them?  Other than that this looks good

abeglova (approved):
lgtm

PR #1967, opened 2025-01-16
fix line clamping of HTML description

gumaerc:
Closes https://github.com/mitodl/hq/issues/6512

This PR fixes some issues with line clamping of the resource description in the V2 learning resource drawer. What's happening here is similar to the issue we had with some descriptions having extraneous padding at the top; nested paragraph elements in the description HTML in the database. Previously, I had simply set the first child top padding and last child bottom padding to zero, but it seems now that we've added the "show more" line clamping functionality we need a bit more sophisticated of a solution. The main issue is that paragraphs (`p` elements) get a default amount of padding applied to them, both top and bottom. When we set up line clamping, we used the `-webkit-line-clamp` rule, which is supported by most browsers, but we have a fallback in there for unsupported browsers that sets:
```
maxHeight: `calc(${theme.typography.body2.lineHeight} * 5)`
```
This rule simply takes the font's line height value and multiplies it by 5 to cap the height of the description container at 5 lines. The `body2` font in use here has a line height of `1.125rem`, which is `18px` at `16px` base font size. Since the `p` elements being inserted inside the description have a padding of 14 pixels, the 5th "line" of text gets pushed down by that much, but only the very top of it is exposed. This is because the `max-height` rule being set only accounts for the literal pixel height of 5 lines of text at `1.125rem` line height.

image
image

### How can this be tested?
 - Ensure that you have backpopulated some courses and resources that have HTML descriptions in your database
 - Spin up `mit-learn` on the `main` branch
 - Search for the Program "Architecture and Systems Engineering: Models and Methods to Manage Complex Systems"
 - Verify that you see the issue as described; a cut off line of text at the bottom of the clamped description
 - Switch to this PR branch
 - Refresh the page and verify that you now do not see the cut off line of text

### DIFF
--- a/frontends/ol-components/src/components/LearningResourceExpanded/LearningResourceExpandedV2.tsx
+++ b/frontends/ol-components/src/components/LearningResourceExpanded/LearningResourceExpandedV2.tsx
@@ -174,20 +174,23 @@ const Description = styled.p({
   ...theme.typography.body2,
   color: theme.custom.colors.black,
   margin: 0,
-  whiteSpace: "pre-wrap",
   wordBreak: "break-word",
-  "p:first-child": {
-    marginTop: 0,
-  },
-  "p:last-child": {
-    marginBottom: 0,
+  "> *": {
+    ":first-child": {
+      marginTop: 0,
+    },
+    ":last-child": {
+      marginBottom: 0,
+    },
+    ":empty": {
+      display: "none",
+    },
   },
 })
 
 const DescriptionCollapsed = styled(Description)({
   display: "-webkit-box",
   overflow: "hidden",
-  maxHeight: `calc(${theme.typography.body2.lineHeight} * 5)`,
   "@supports (-webkit-line-clamp: 5)": {
     WebkitLineClamp: 5,
     WebkitBoxOrient: "vertical",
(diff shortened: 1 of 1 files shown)


frontends/ol-components/src/components/LearningResourceExpanded/LearningResourceExpandedV2.tsx
```diff
@@ -174,20 +174,23 @@ const Description = styled.p({
...
+    ":empty": {
+      display: "none",
+    },
   },
 })
 
 const DescriptionCollapsed = styled(Description)({
   display: "-webkit-box",
   overflow: "hidden",
-  maxHeight: `calc(${theme.typography.body2.lineHeight} * 5)`,
```
ChristopherChudzicki, line 190: The browser support for line-clamp is so good I don't think it matters, but if you really want to keep the fallback, you could do
```
  maxHeight: `calc(${theme.typography.body2.lineHeight} * 5)`,
  "@supports (-webkit-line-clamp: 5)": {
    maxHeight: "unset",                     <-- turn off the fallback on supported browsers
    WebkitLineClamp: 5,
    WebkitBoxOrient: "vertical",
  },
```

ChristopherChudzicki (approved):

PR #1980, opened 2025-01-21
variable contentfile chunk sizes

shanbady:

Closes https://github.com/mitodl/hq/issues/6531

This PR consists of 2 simple changes that allow us to quickly test different content file chunk sizes.
1. a "CONTENT_FILE_EMBEDDING_CHUNK_SIZE_OVERRIDE" setting. if this is set, it forces the use of the TokenTextSplitter with an explicit chunk size.
2. Addition of a "collection_name" parameter on the vector content file endpoint that lets use specify which collection to query.

The plan is to:
1. embed a specific set of contentfiles (see https://github.com/mitodl/hq/issues/6419) locally using the openai large embedding model using  2 new chunk sizes( 512 and 1024)
2.  snapshot and upload these as separate collections to rc and prod (content_files_512, content_files_1024 etc)
3. The LLM can then use the  collection_name parameter to query the collections of a specific chunk size

### How can this be tested?
1. Checkout this branch. make sure you have some resources and contentfiles locally
2. in settings.py set CONTENT_FILE_EMBEDDING_CHUNK_SIZE_OVERRIDE to some small value like 100 - set CONTENT_FILE_EMBEDDING_CHUNK_OVERLAP to 20
3. restart celery
4. delete existing qdrant collections
5. pick a few resources that have contentfiles associated with them. embed them via `python manage.py generate_embeddings --resource-ids 111,222,333`
6. visit the contentfiles collection in the [qdrant dashboard](http://localhost:6333/dashboard#/collections/resource_embeddings.content_files) note that the "chunk_contents" should be much smaller.

after performing the previous steps, we can test the  "collection_name" param:
1. visit the [qdrant dashboard](http://localhost:6333/dashboard#/collections/)
2. create a snapshot of the contentfiles collection and download it
Screenshot 2025-01-21 at 4 26 46 PM

3. re-upload the snapshot and specify some other name (but keep the "resource_embedding.*" prefix )
4. visit the [vector contentfile endpoint](http://open.odl.local:8063/api/v0/vector_content_files_search/?limit=10&collection_name=testing) and specify the collection name (without the "resource_embedding.*" prefix) - results are pulled in from the new collection. delete the new collection and see that the api fails with a "collection not found error"


abeglova (approved):
👍

PR #1984, opened 2025-01-23
drawer topic carousels should not be filtered on just courses

gumaerc:
https://github.com/mitodl/hq/issues/6565

When adding the topic carousels to the new learning resource drawer, the `TopicCarouselConfig` class was reused from the Dashboard. This config was set up by default to filter on only courses. The carousels in the drawer are not supposed to be this way. By default the `TopicCarouselConfig` does not filter on courses now, and that is only set in the Dashboard.

### How can this be tested?
 - Ensure you have a variety of resources of different types populated in your database
 - Visit the search page at http://localhost:8062/search/
 - Click on some search results and verify that the topic carousels pull in more than just courses
 - You should test several resources as some topics may show all courses just by chance

### DIFF
--- a/frontends/main/src/common/carousels.ts
+++ b/frontends/main/src/common/carousels.ts
@@ -1,5 +1,11 @@
-import type { ResourceCarouselProps } from "@/page-components/ResourceCarousel/ResourceCarousel"
-import { LearningResourcesSearchRetrieveDeliveryEnum } from "api"
+import type {
+  ResourceCarouselProps,
+  TabConfig,
+} from "@/page-components/ResourceCarousel/ResourceCarousel"
+import {
+  LearningResourcesSearchRetrieveDeliveryEnum,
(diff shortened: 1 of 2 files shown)


frontends/main/src/common/carousels.ts
```diff
@@ -38,26 +44,29 @@ const TopPicksCarouselConfig: TopPicksCarouselConfigProps = (
 
 type TopicCarouselConfigProps = (
   topic: string | undefined,
+  resourceType?: LearningResourcesSearchRetrieveResourceTypeEnum[] | undefined,
 ) => ResourceCarouselProps["config"]
 
 const TopicCarouselConfig: TopicCarouselConfigProps = (
   topic: string | undefined,
+  resourceType?: LearningResourcesSearchRetrieveResourceTypeEnum[] | undefined,
```
ChristopherChudzicki, line 52: `resourceTypes` ? or maybe `filteredTypes` ?

PR #1987, opened 2025-01-24
move LearningResourceExpanded to main

ChristopherChudzicki:
For https://github.com/mitodl/hq/issues/6554

This moves LearningResourceExpanded out of ol-components and into `main`. The immediate goal is to make it simpler to use Posthog within the Learning Resource drawer, but not clear this ever belonged inside ol-components IMO. It's not particularly generic.

As part of https://github.com/mitodl/hq/issues/6554 I'd like to use posthog within the drawer itself. Making this PR sepearately because I am worried that with lots of other changes to the drawer, git won't do a good job detecting the simple move.

### How can this be tested?
Drawer should still open fine.


frontends/ol-components/src/components/LearningResourceExpanded/LearningResourceExpanded.stories.tsx
ChristopherChudzicki, line 1: this was written for the v1 drawer, no point in keeping it.

gumaerc (approved):
👍 

In the future, I think we should also move some of the things that are being passed as arguments into `LearningResourceExpanded` directly into it. The learning path and user list hooks and event handlers as well as the share URL I think can be moved in, as well as the hook to get the resource itself. The carousel slots should stay as arguments, although since they just accept an arbitrary list of `ReactNode` they should probably just be renamed to something like `upperContent` and `lowerContent`.

PR #1988, opened 2025-01-24
remove react router

ChristopherChudzicki:
Removes react-router.

We were using it a few places that it was irrelevant, and we shouldn't have used it at all since migrating to NextJS.

The only uses were in test files or types, so won't affect production deployments.

### How can this be tested?
Tests should pass.


shanbady (approved):
LGTM 👍

PR #1998, opened 2025-01-27
Enable Posthog session replay by env var

jonkafton:

https://github.com/mitodl/hq/issues/6477

Provides an env var, `NEXT_PUBLIC_POSTHOG_ENABLE_SESSION_RECORDING`, to enable Posthog session recording when set to `"True"` or `"true"`. In the CI release workflows this is mapped from `POSTHOG_ENABLE_SESSION_RECORDING` (the `NEXT_PUBLIC_` prefix is a requirement in Next.js for all env vars that arrive at the browser).

The Posthog session replay feature was identified as a performance issue contributing to Total Blocking Time during page load.

### How can this be tested?

The full test is to build the frontend with `NEXT_PUBLIC_POSTHOG_ENABLE_SESSION_RECORDING=true yarn build` and confirm that you see session recordings in the Posthog dashboard from your local. The inverse should be true with `NEXT_PUBLIC_POSTHOG_ENABLE_SESSION_RECORDING` unset. 
<br/><br/>
image
<br/><br/>

This requires setting up a Posthog project for your local, however session replays are sampled so the absence of a recording does not guarantee that it is actually disabled. The intention is to eliminate blocking on the main thread, so it's sufficient to run the Profiler "Record and reload (⌘ ⇧ E)" with CPU at 20x slowdown to amplify the result.

With Posthog session recording enabled, you should see blocking between the DOMContentLoaded and Onload events plus `serializeNodeWithID` calls as Posthog traverses the DOM:

<br/><br/>
image
<br/><br/>

Disabled you should no longer see the thread blocking task and the DCL and L events should fire together:
<br/><br/>
image

### Additional Context

The aim is retest in PageSpeed and weigh up performance improvement against desirability. Follow up PR in ol-infrastructure needed to set the env var on the GitHub Actions runners if we decide to enable.


abeglova (approved):
lgtm

PR #2001, opened 2025-01-29
disable daily embed task in development

shanbady:

Closes https://github.com/mitodl/hq/issues/6615

This pr disables the daily embedding task for local environments by conditionally including the task based on the environment.

### How can this be tested?
1. checkout main
2. go into the celery container and go into celery shell via `celery -A main shell`. then list periodic tasks via `celery.conf.beat_schedule` - note that the 'daily_embed_new_learning_resources' task in the list of periodic tasks
3. checkout this branch
4. restart the celery container `docker compose restart celery`
5. repeat step 2. note that the 'daily_embed_new_learning_resources' is no longer in the list of periodic tasks

### DIFF
--- a/main/settings_celery.py
+++ b/main/settings_celery.py
@@ -6,7 +6,7 @@ from celery.schedules import crontab
 
 from main.envs import get_bool, get_int, get_string
 
-ENVIRONMENT = get_string("MITOL_ENVIRONMENT", "dev")
+DEV_ENV = get_bool("DEV_ENV", False)  # noqa: FBT003
 USE_CELERY = True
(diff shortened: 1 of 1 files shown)


main/settings_celery.py
```diff
@@ -6,6 +6,7 @@
 
 from main.envs import get_bool, get_int, get_string
 
+DEV_ENV = get_bool("DEV_ENV", False)  # noqa: FBT003
```
mbertrand, line 9: Could use `settings.ENVIRONMENT` instead of creating this new one

Default value locally is "dev"

```python
    if settings.ENVIRONMENT == "dev":
        // do something
```

mbertrand (approved):
Works well, only suggestion is to use `settings.ENVRONMENT` instead of adding `settings.DEV_ENV`

PR #2006, opened 2025-01-30
Better React Testing Library Errors

ChristopherChudzicki:
None

This PR makes the errors from `@testing-library/react` unfound matchers slightly better. In particular:

**Before:**
Screenshot 2025-01-30 at 11 53 19 AM

**After**
Screenshot 2025-01-30 at 11 50 41 AM

### How can this be tested?
Tests should pass.

If you want to see the behavior change, you can add a fake error to a test file, e.g.,
```ts
// ResourceCard.test.tsx
  test.only("Applies className to the resource card", () => {
    const { view } = setup({ user: {}, props: { className: "test-class" } })
    expect(view.container.firstChild).toHaveClass("test-class")
    screen.getByRole("button", { name: "Fake Element" })
  })
```
and run `yarn test frontends/main/src/page-components/ResourceCard/ResourceCard.test.tsx` on this branch vs `main`.

### DIFF
--- a/frontends/jest-shared-setup.ts
+++ b/frontends/jest-shared-setup.ts
@@ -85,7 +85,7 @@ afterEach(() => {
  *
  * However, method calls occurring in beforeEach hooks earlier or afterEach hooks
  * after this line will not error.
- * - beforeEach hooks declared earlier than this all
+ * - beforeEach hooks declared earlier than this call
  * - afterEach hooks declared later than this call
  */
 failOnConsole()
--- a/frontends/jest-shared-setup.ts
+++ b/frontends/jest-shared-setup.ts
@@ -1,7 +1,6 @@
 import failOnConsole from "jest-fail-on-console"
 import "@testing-library/jest-dom"
 import "cross-fetch/polyfill"
-import { configure } from "@testing-library/react"
 import { resetAllWhenMocks } from "jest-when"
 import * as matchers from "jest-extended"
 import { mockRouter } from "ol-test-utilities/mocks/nextNavigation"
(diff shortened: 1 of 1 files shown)


frontends/jest-shared-setup.ts
```diff
@@ -59,32 +58,17 @@ const polyfillResizeObserver = () => {
...
-   *
-   * This does change the stacktrace a bit: The line causing the error is still
-   * there, but the line where the error is generated (below) is most visible.
-   */
-  getElementError(message, _container) {
-    const error = new Error(message ?? "")
-    error.name = "TestingLibraryElementError"
-    return error
-  },
-})
```
ChristopherChudzicki, line 79: The comment here said:
> By default, testing-library prints much too much of the DOM.

This doesn't seem to change how much of the DOM gets printed, and the DOM elements it does print are helpful anyway, IMO. (I guess my opinion has changed since 22 months ago.)

frontends/jest-shared-setup.ts
```diff
@@ -59,32 +58,17 @@ const polyfillResizeObserver = () => {
...
   return {
     ...jest.requireActual("ol-test-utilities/mocks/nextNavigation")
       .nextNavigationMocks,
   }
 })
 
+beforeEach(() => {
+  mockRouter.setCurrentUrl("/")
+  window.history.replaceState({}, "", "/")
+})
```
ChristopherChudzicki, line 71: **tldr:** This resets some state. It can happen before or after, but "after" is more error prone.

-----------

Simplify removing the `configure` call above produced a bunch of `act` console errors. What was happening was:

1. We had an `afterEach` hook that reset the `mockRouter` state.
    - *If react components are still mounted, this will cause them to update.*
2. `@testing-library/react` registers some hooks that automatically unmount components after each test. 

If (1) happens before (2), then we get some `act` console errors.

When we were calling `configure(...)`, RTL hooks ended up getting registered first, so everything was fine.  When `configure` was removed, the RTL hooks were registered afterwards, and so console errors started showing up[^1]. Reseting the state before the tests instead of after avoids the issue. (Since react components won't be mounted before the test runs).

[^1]: But didn't cause our tests to fail, for the reason noted in comment on `failOnConsole`.

shanbady (approved):
looks good. the output  is far more useful

PR #2012, opened 2025-02-05
Remove imports for builtin django User model

rhysyngsun:

- Adds a lint to block direct importing of `django.contrib.auth.models.User`
- Replaces all such imports with `get_user_model()` or `settings.AUTH_USER_MODEL`

### How can this be tested?

You should be able to run the app without issue.

You can also test the lint rule by adding an import to `django.contrib.auth.models.User` in any file and run `pre-commit run` and it should error.

You can observe that the initial commit for this PR that added this check failed for all of the existing usages: https://results.pre-commit.ci/run/github/672068771/1738783559.KuMD_PjETSqF69WSAV5Ufg

### DIFF
--- a/learning_resources/models.py
+++ b/learning_resources/models.py
@@ -3,9 +3,9 @@
 import uuid
 from abc import abstractmethod
 from functools import cached_property
-from typing import Optional
+from typing import TYPE_CHECKING, Optional
 
-from django.contrib.auth.models import User
+from django.conf import settings
 from django.contrib.postgres.fields import ArrayField
 from django.db import models
 from django.db.models import CharField, Count, JSONField, OuterRef, Prefetch, Q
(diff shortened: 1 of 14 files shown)


learning_resources/models.py
```diff
@@ -25,6 +25,11 @@
 )
 from main.models import TimestampedModel, TimestampedModelQuerySet
 
+if TYPE_CHECKING:
+    from django.contrib.auth import get_user_model
+
```
shanbady, line 30: what does this code do? It doesnt seem like User is referenced anywhere in this file. Might be worth leaving a comment

shanbady (changes requested):
worked in catching the User import. had a question about the `if TYPE_CHECKING:`  condition

learning_resources/models.py
rhysyngsun: It's referenced in the types of some of the function arguments: https://github.com/mitodl/mit-learn/pull/2012/files/6a35cb4f70d5a4364f2eb25905f46b76d27b788a#diff-3e244d0725c2dbf2dfd444546f36377feb3dcb0bfee9838f50a971237def1e72R336

shanbady (approved):
LGTM

PR #2013, opened 2025-02-05
Add initial users app

rhysyngsun:

Note: this is based on the PR https://github.com/mitodl/mit-learn/pull/2012.

This adds an initial users app with an empty migration, per the steps outlined in the issue above.

There are also a few changes regarding SSL certificates and database backups I wanted to commit to avoid an extra PR just for those.

### How can this be tested?

The app should still function, the main point of this PR is to get an entry for `0001_initial.py` into the `django_migrations` table.

**NOTE: you don't need to reverse the migration after using this branch, in fact you'll have to apply it later anyway so you might as well leave it**


shanbady (approved):
pr looks good. i did have some lingering questions about what we end up doing with the profiles app if it makes sense to consolidate the custom user object into that.

PR #2015, opened 2025-02-05
Add custom user model

rhysyngsun:
NOTE: https://github.com/mitodl/mit-learn/pull/2013 should be reviewed first, as changes from it are included here.

### What are the relevant tickets?

Part of https://github.com/mitodl/hq/issues/6679

### Description (What does it do?)

This adds a custom user model and points the `AUTH_USER_MODEL` to it.
It updates the implementation of the `0001_initial.py` model by having deleted this file and then recreating it after updating `AUTH_USER_MODEL`.

### How can this be tested?

- Ensure you have run the migrations from [this PR](https://github.com/mitodl/mit-learn/pull/2013) before proceeding so that `0001_initial.py` gets recorded as having been run. If it's been merged, you can just go ahead and run the migration off `main`.
- Checkout this branch and test that `./manage.py showmigrations` still lists `0001_initial.py` under the users app as having been applied.
- Run `./manage.py makemigrations --dry` and verify there are no changes.
- Start the app, verify the migrations run, and the app is functional.

**When you are done with testing, be sure to run `./manage.py migrate users 0001` to revert not break things on other branches.**

PR #2016, opened 2025-02-05
Rename user table from `auth_user` to `users_user`

rhysyngsun:
NOTE: https://github.com/mitodl/mit-learn/pull/2015 should be reviewed first

This renames the table as described. Additionally a Postgres view is created so the table is still accessible under the previous name so we don't see errors during deployment.

### How can this be tested?

- Before switching to this branch, you should probably backup your db: 
  `docker compose exec db pg_dump --file=/mnt/backups/dump.sql -U postgres -W -d postgres`
- Start the app, the migrations should succeed and the app should run.
- Check the db schema, the following should be true:
  - The `auth_user` table should be renamed to `users_user`.
  - In the db, `SELECT * from django_content_type WHERE model = 'user';` should show that the `app_label` has been updated to `"users"`

**Be sure to `./manage.py migrate users 0001` before switching back to other branches**


shanbady (approved):
👍

PR #2017, opened 2025-02-06
Skip existing embeddings

shanbady:

Closes https://github.com/mitodl/hq/issues/6670

This PR makes it so that the embedding generation for both learning resources and contentfiles skips over existing embeddings by default. We can force re-embed by passing the --overwrite flag to the generate_embeddngs command.

### How can this be tested?
1. checkout this branch
2. restart celery
3.get the id of learning resource that has associated contentfiles and pass it into the generate_embeddings command `python manage.py generate_embeddings --resource-ids <the id>`
4. observe the celery container's output - there should be some output indicating that embeddings are being generated (different output depending on if fastembed or [ollama via litellm](https://github.com/mitodl/mit-learn/blob/main/docs/how-to/embeddings.md#ollama-for-hardware-accelerated-embedding) is used).
5. Once that completes, verify there are embeddings on the [qdrant dashboard](http://localhost:6333/dashboard#/collections)
6. while keeping an eye on the celery output, re-run the embedding command
7. note that embeddings are not skipped
8. re-run the command with the overwrite flag `python manage.py generate_embeddings --resource-ids <the id> --overwrite` and note that once again embeddings are getting generated

### DIFF
--- a/vector_search/constants.py
+++ b/vector_search/constants.py
@@ -1,8 +1,6 @@
 from django.conf import settings
-from django.contrib.auth.models import User
 from qdrant_client import models
 
-user = User()
 RESOURCES_COLLECTION_NAME = f"{settings.QDRANT_BASE_COLLECTION_NAME}.resources"
 CONTENT_FILES_COLLECTION_NAME = f"{settings.QDRANT_BASE_COLLECTION_NAME}.content_files"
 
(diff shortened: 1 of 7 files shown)


mbertrand (approved):
👍  works great

PR #2020, opened 2025-02-07
Update actions/setup-python to use pyproject.toml

rhysyngsun:

N/A

This updates the github actions usage of `actions/setup-python` to pull the version from  `pyproject.toml` instead of a hardcoded value. This will streamline renovate python upgrades because it doesn't update github actions.

### How can this be tested?

Tests should pass.


ChristopherChudzicki (approved):
Awesome.

PR #2040, opened 2025-02-11
Remove `@lukemorales/query-key-factory`, unblock react-query v5 upgrade

ChristopherChudzicki:
Prerequisite for https://github.com/mitodl/mit-learn/pull/1940

This PR removes `@lukemorales/query-key-factory`, which we had been using to create query configurations for `@tanstack/react-query`. However:
-  `@tanstack/react-query` was released about 1.5 years ago and has a number of improvements. A long open issi
- `@lukemorales/query-key-factory` does not support v5, and the open issue https://github.com/lukemorales/query-key-factory/issues/73 to add support shows no progress. A PR to add support was closed. I am not optimistic support will be added any time soon.
    - `query-key-factory` more or less works OK with v5 despite official support. However, it does lead to some TS issues that are very hard to resolve.

### How can this be tested?
The site should function exactly as before.

What I have checked is:
1. basic functionality of pages (home, search, dashboard, department/unit/topic listing, etc)
2. query invalidations:
   - Add / remove a resource to userlist. The bookmark should fill / unfill appropriately.
   - At `/dashboard/my-lists/`, create / delete / edit  userlist. The listing page should update appropriately.


frontends/api/src/hooks/articles/queries.ts
```diff
@@ -0,0 +1,27 @@
...
+import { articlesApi } from "../../clients"
+import type { ArticlesApiArticlesListRequest as ArticleListRequest } from "../../generated/v1"
+
+const articleKeys = {
+  root: ["articles"],
+  listRoot: () => [...articleKeys.root, "list"],
+  list: (params: ArticleListRequest) => [...articleKeys.listRoot(), params],
+  detailRoot: () => [...articleKeys.root, "detail"],
+  detail: (id: number) => [...articleKeys.detailRoot(), id],
+}
```
ChristopherChudzicki, line 11: The structure above (and in other queries files) is based on https://tkdodo.eu/blog/effective-react-query-keys#use-query-key-factories (blog by one of the maintainers).

Still a "key factory", but a little less automated than it was with `@lukemorales/query-key-factory`.

jonkafton (approved):
Looks good and is actually a cleaner API that we had previously (loses `._ctx` and `._def`).

I'd consider an alternative along the lines of below, though I appreciate it may be better to alway call a function for the keys regardless of whether it accepts params.

```
const ROOT_KEY: RootKey = "articles"

const articleKeys = {
  root: [ROOT_KEY],
  listRoot: [ROOT_KEY, "list"],
  list: (params: ArticleListRequest) => [ROOT_KEY, "list", params],
  detailRoot: [ROOT_KEY, "detail"],
  detail: (id: number) => [ROOT_KEY, "detail", id],
};
```

I'm not massively seeing the value of a key factory layer inside of the query hooks, which could directly create their keys and call the APIs. Convenient and a safer way for retrieving keys for clearing cache across files perhaps, though perhaps when we upgrade react-query we can find a pattern that better encapsulates this logic without exporting keys.

All works great.

PR #2041, opened 2025-02-11
Embeddings - Stay within openai request size limits

shanbady:
Fixes https://github.com/mitodl/hq/issues/6724

This PR makes it so that when we stay under openai's per-request rate limits when sending up chunks of contentfiles to embed. This currently affects a small number of resources on prod:
**tokens per request:** 600,000
**max array size:** 2048

### How can this be tested?
1. Checkout this branch
2. Make sure you have some contentfiles locally
3. run embeddings for some number of resources with contentfiles associated with them  `python manage.py generate_embeddings  --resource-ids <resource id> --overwrite` - it should succeed and function as normal without issues - this change only affects usage of openai in that we stay under their limits by breaking up requests

### DIFF
--- a/vector_search/utils.py
+++ b/vector_search/utils.py
@@ -262,7 +262,19 @@ def _process_content_embeddings(serialized_content):
             )
             for md in split_metadatas
         ]
-        split_embeddings = list(encoder.embed_documents(split_texts))
+        split_embeddings = []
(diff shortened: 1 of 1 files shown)


mbertrand (approved):
Works great 👍

PR #2050, opened 2025-02-14
embed course metadata as contentfile

shanbady:

Fixes (part 1 of) https://github.com/mitodl/hq/issues/6725

This PR adds embeddings for general course info (what shows up in the resource panel) in the contentfiles collection so that the chat agent can get that info from the contentfile vector endpoint directly.

**How it works**
Anytime we embed a new resource, we also generate an "about this course" document with all the course info and put that in the contentfiles collection. We can follow this same pattern for whatever else we might need to enrich the chat agent's response

### How can this be tested?
1. checkout this branch
2. make sure you have some contentfiles locally
3. find a learning resource id for a resource that has contentfiles
4. generate embeddings for that resource via `python manage.py generate_embeddings --resource-ids <id>`
5. hit the contentfile vector search endpoint - set the resource_readable_id parameter to the readable_id (make the readable_id is url encoded - this has get me a few times). 
http://open.odl.local:8063/api/v0/vector_content_files_search/?limit=10&resource_readable_id=course-v1%3AMITxT%2B14.73x&q=who%20offers%20this%20course?
6. try seting the "q=" parameter to questions about the course (info that shows up in the resource panel) for example "?q=who is teaching this course?" or "q=how much does it cost? will I earn a certificate?". The results should surface a chunk that starts with "Information about this course" and has a "file_extension" of ".md"

### Additional Context
- The "course info" document won't always be surfaced as the top result - its heavily dependent on other contentfiles for the resource - in lots of cases the same info is found in other chunks. This does however fill the gap where that info is absent in contentfiles.
- There will be a follow on ticket about pulling in full marketing site content into the chunk collection - related https://github.com/mitodl/hq/issues/6699


vector_search/utils.py
```diff
@@ -214,7 +196,146 @@ def _chunk_documents(encoder, texts, metadatas):
...
+        )
+    if len(docs) > 0:
+        embeddings = encoder.embed_documents(docs)
+        return points_generator(ids, metadata, embeddings, vector_name)
+    return None
+
+
+def generate_metadata_document(serialized_resource):
+    """
+    Generate a plaint-text info document to embed in the contentfile collection
```
abeglova, line 220: Can this be replaced with a serializer that takes a learning resource? Something like ContentFileLearningResourceMetadataSerializer

vector_search/utils.py
```diff
@@ -214,7 +196,146 @@ def _chunk_documents(encoder, texts, metadatas):
...
+    vector_name = encoder.model_short_name()
+    metadata = []
+    ids = []
+    docs = []
+    for doc in serialized_resources:
+        readable_id = doc["readable_id"]
+        resource_vector_point_id = str(vector_point_id(readable_id))
+        ids.append(resource_vector_point_id)
+        course_info_document = generate_metadata_document(doc)
+        metadata.append(
```
abeglova, line 311: Seems like other than resource_vector_point_id, most of this can be added to generate_metadata_document (or the serializer that i think you should replace it with)

abeglova (changes requested):
Back to you with some comments!

vector_search/utils.py
shanbady: 👍 That might be a cleaner way to do this. will have a look

abeglova (approved):
lgtm

PR #2058, opened 2025-02-18
Recommendation bot styling and text updates

jonkafton:

Item 2. of https://github.com/mitodl/hq/issues/6747

- Updates title to "What do you want to learn from MIT?"
- Removes strap lines "Need assistance getting started?" and "Let me know how I can help.".
- Adds a drawer close button to the entry screen; reuses for the chat screen (replacing chat built-in).
- Removes the "I am curious about AI applications for business..." starter prompt.
- Also makes some updates to the mobile layout so that it holds in advance of designs.

image

image

### How can this be tested?

- Assign `recommendationBotEnabled` to `true` in HeroSearch.tsx to bypass the feature flag and show the AskTIM button on the homepage.
- Hit the AskTim button to open the drawer. Confirm that updates have been made according to item 2 in https://github.com/mitodl/hq/issues/6747 and that the layout and behaviors are as expected. Layout should match https://www.figma.com/design/Eux3guSenAFVvNHGi1Y9Wm/MIT-Design-System?node-id=17143-54099&m=dev (though one starter prompt is removed.)

### DIFF
--- a/frontends/main/src/page-components/AiRecommendationBot/AiRecommendationBotDrawer.tsx
+++ b/frontends/main/src/page-components/AiRecommendationBot/AiRecommendationBotDrawer.tsx
@@ -101,7 +101,7 @@ const Starter = styled.button(({ theme }) => ({
   color: theme.custom.colors.darkGray2,
   backgroundColor: "transparent",
   textAlign: "left",
-  [theme.breakpoints.down("md")]: {
+  [theme.breakpoints.down("sm")]: {
     textAlign: "center",
     padding: "12px 36px",
   },
--- a/frontends/main/src/page-components/AiRecommendationBot/AiRecommendationBotDrawer.tsx
+++ b/frontends/main/src/page-components/AiRecommendationBot/AiRecommendationBotDrawer.tsx
@@ -18,8 +18,10 @@ const EntryScreen = styled.div(({ theme }) => ({
   justifyContent: "center",
   gap: "16px",
   padding: "136px 40px 24px 40px",
+  width: "900px",
   [theme.breakpoints.down("md")]: {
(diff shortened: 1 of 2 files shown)


frontends/main/src/page-components/AiRecommendationBot/AiRecommendationBotDrawer.tsx
```diff
@@ -1,20 +1,43 @@
...
-})
+  padding: "136px 40px 24px 40px",
+  width: "900px",
+  [theme.breakpoints.down("md")]: {
+    padding: "136px 24px 24px 24px",
+    width: "100%",
+  },
+}))
+
+const CloseButton = styled(ActionButton)(({ theme }) => ({
```
ChristopherChudzicki, line 28: I believe Figma has a component `Close-Modal` that is used both here and in the resource drawer.

Ideally, I think this should be a button variant coordinated with Steve and Bilal. But in the meantime, might be worth sticking this in `/main/src/components/CloseButton` and using in both places :shrug:

ChristopherChudzicki (approved):
👍 

Unrelated to the changes in this PR, I am noticing that the hero area "AskTim" button causes some layout shift. If we're going to enable that on prod soon, might be worth reserving some space for it on the server-rendered page.

frontends/main/src/page-components/AiRecommendationBot/AiRecommendationBotDrawer.tsx
jonkafton: Yes - it now appears here, in the resource drawer and in the recommendation bot drawer (which turns off this one to reuse between screens).

PR #2061, opened 2025-02-19
APISIX integration

mbertrand:
Closes https://github.com/mitodl/hq/issues/6756

Related infrastructure PR: https://github.com/mitodl/ol-infrastructure/pull/2983

- Makes mit-learn compatible with APISIX, using the code from unified-ecommerce and learn-ai as a template.
- Adds two optional docker containers to start up local instances of Keycloak and APISIX.
- Removes LiteLLM proxy service code (that is now in learn-ai)

### How can this be tested?
Follow instructions in the new [README](https://github.com/mitodl/mit-learn/blob/98bf39baf18a4bc0e7b781c4e8377b2ab6286ec7/README-keycloak.md).  You should be able to log in and log out.


main/settings.py
```diff
@@ -309,45 +310,6 @@
...
-    # Populate the extra_data field in the social record with the values
-    # specified by settings (and the default ones like access_token, etc).
-    "social_core.pipeline.social_auth.load_extra_data",
-    # Update the user record with any changed info from the auth service.
-    "social_core.pipeline.user.user_details",
-    # Create a favorites list for new users
-    "authentication.pipeline.user.user_created_actions",
-    # redirect new users to onboarding
-    "authentication.pipeline.user.user_onboarding",
-)
```
mbertrand, line 350: This didn't seem to be getting used even without keycloak and apisix

mbertrand: PS the final pipeline step is supposed to redirect the user to "/onboarding" but that isn't currently happening on RC with new users I created there, even though `MITOL_NEW_USER_LOGIN_URL=https://rc.learn.mit.edu/onboarding` - should it be?  Haven't tried it on production yet.

jkachel (approved):
LGTM 👍 was able to log in successfully via Keycloak. It also attached the keycloak user to the `admin@odl.local` superuser I created via the management command.

docker-compose.services.yml
```diff
@@ -88,9 +88,58 @@ services:
...
+      - ./config/keycloak/realms:/opt/keycloak/data/import
+      - ./config/keycloak/providers:/opt/keycloak/providers
+      - ./config/keycloak/themes:/opt/jboss/keycloak/themes
+
+  apigateway:
+    profiles:
+      - keycloak
+    image: apache/apisix:latest
+    depends_on:
+      - keycloak
```
rhysyngsun, line 121: I would remove `profiles` and `depends_on` here, because otherwise I'm forced to spend resources running a duplicate keycloak instance I won't use if I'm not running the keycloak container defined above.

mbertrand: Might it be useful to still keep the `profiles` part, but change it to its own value so that apisix (and keycloak) don't start automatically, only when explicitly requested? So:

`COMPOSE_PROFILES=backend,frontend,keycloak,apisix` starts everything
`COMPOSE_PROFILES=backend,frontend,apisix` starts apisix but not keycloak
etc

mbertrand: I just made that change (removed `depends_on`,  gave keycloak and apisix containers their own separate profiles)

mbertrand: Forgot you can just call them by name in this case, so updated again

main/middleware/apisix_user.py
```diff
@@ -0,0 +1,53 @@
...
+
+    header = "HTTP_X_USERINFO"
+
+    def process_request(self, request):
+        """
+        Modify the header to contaiin username, pass off to RemoteUserMiddleware
+        """
+        if request.META.get(self.header):
+            new_header = decode_apisix_headers(request, self.header)
+            log.error("FOUND APISIX HEADER: %s", new_header)
```
jkachel, line 50: nit: this is logging as error

jkachel (approved):
Simpler workflow seems to work pretty well. I have one nit regarding logging.

I had some issues getting to the onboarding until I figured out that the onboarding URL is configurable. May be a good idea to add that to the docs (or update the pipeline or default to use the app base URL) - it defaults to just `/onboarding` which ends up sending the user to the API service and results in an error.

main/middleware/apisix_user.py
mbertrand: I got rid of that stray test logging statement and added info to the README on setting the correct onboarding url.  Also added `credentials: "include"` to the frontend AI tsx pages since I think that may be necessary with or without keycloak/apisix

PR #2063, opened 2025-02-20
Full width slide down syllabus chat for the resource drawer

jonkafton:

Closes https://github.com/mitodl/hq/issues/6747

- Implements 1. in the linked issue.
- Refactors out the various LearningResourceExpanded components (the file was becoming unwieldy).
- Refactors out our chat with entry screen (now an easy move to Smoot Design).

- [ ] Desktop screenshots

image

<br/><br/>

image

<br/><br/>

image

### How can this be tested?

- Set `chatEnabled` to `true` in LearningResourceExpanded.tsx and HeroSearch.tsx.
- Click a resource card to open the drawer. The "AskTIM about this course" opener should display.
  - Click the opener to slide down the chat entry screen
  - Confirm the entry screen text input and starter buttons take you to the chat.
  - Confirm the chat functions and scrolls correctly.
    - The Smoot Design AiChat component has no padding, with the expectation that it would expand to a container provided by the calling component. This is not ideal is its scrollbar interferes with the messages. Ideally it would behave as the resource content, scrolling at full height with the sticky title and content masked by the opener button and horizontal line, though that has its challenges:

image

  - Check above at mobile widths.
  - On closing, the drawer content should be displayed with scroll preserved.
- The chat with entry is reused for the syllabus and recommendation bots.
  - Click the AskTIM button in the Hero search. Confirm


.github/workflows/production.yml
```diff
@@ -52,7 +52,7 @@ jobs:
           POSTHOG_API_HOST: https://app.posthog.com
           POSTHOG_PROJECT_ID: ${{ secrets.POSTHOG_PROJECT_ID_PROD }}
           POSTHOG_API_KEY: ${{ secrets.POSTHOG_PROJECT_API_KEY_PROD }}
-          POSTHOG_ENABLE_SESSION_RECORDING: ${{ secrets.POSTHOG_PROJECT_ENABLE_SESSION_RECORDING_PROD }}
+          POSTHOG_ENABLE_SESSION_RECORDING: ${{ secrets.POSTHOG_ENABLE_SESSION_RECORDING_PROD }}
```
jonkafton, line 55: Slipped this in - it was just badly named. https://github.com/mitodl/ol-infrastructure/pull/2978 sets the env vars.

ChristopherChudzicki (commented):
Just starting to look at code. Generally worked well, but noticed a few issues:

1. Drawer's scrollbar is covered up by title
    - no obvious suggestion... Some z-index issue?
3. Title close button shifts annoyingly when chat is toggled (see video)
    - IMO, we should add `scrollbar-gutter: stable` to the scrolling container so that it reserves space for the scrollbar. (I believe that this is based on the user agent, so if the user agent uses overlay scrolling, no space is reserved.) 
3. **Accessibility:** When chat is visible, users can still tab to the hidden content, and vice versa. 
    - I think that the easiest solution would be to conditionally apply [inert](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/inert) to the chat vs regular content. `inert` doesn't have fantastic browser support, but if you're missing it, the page is still pretty functional.
    - Altenratively, could use transitionend events, I feel like there may be some caveats about when exactly those fire. (Caveats might not be relevant to our case; but e.g., if the transition css is removed so the transition starts but then finishes abruptly, i believe the event never fires.)

ChristopherChudzicki:
https://github.com/user-attachments/assets/62acae51-1006-435b-b325-7e98b0e9bd79

jonkafton:
Posting as a backup option if we see cross platform scrollbar issues (@ChristopherChudzicki has a solution, commit below) - A workaround is to only scroll the content section when chat is enabled. Although scrollbars at the right of the window don't look good if they're not full height, the horizontal line on the "AskTIM..." opener affords us the option:

https://github.com/user-attachments/assets/2ae866c5-4e58-4c4e-a0f7-53935287920c

ChristopherChudzicki (changes requested):
All the scrolling issues and shifting seem resolved to me, 👍 👍 

I'm still seeing one accessibility issue.

- When Chat is open, I can't tab to content ✅  (regular content is inert)
- When Chat is closed, I can still tab into it. ❌ 
    - The chat section needs inert when not open. 

Left a few other extremely minor comments.

frontends/main/src/page-components/LearningResourceExpanded/AiChatSyllabusSlideDown.tsx
```diff
@@ -0,0 +1,174 @@
...
+            apiUrl: process.env.NEXT_PUBLIC_LEARN_AI_SYLLABUS_ENDPOINT!,
+            fetchOpts: {
+              headers: {
+                "X-CSRFToken": getCsrfToken(),
+              },
+            },
+            transformBody: (messages) => ({
+              collection_name: "content_files",
+              message: messages[messages.length - 1].content,
+              course_id: resource?.readable_id,
```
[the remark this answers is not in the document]
jonkafton: Good spot.

frontends/main/src/page-components/LearningResourceExpanded/LearningResourceExpanded.tsx
```diff
@@ -49,7 +49,7 @@ const ChatLayer = styled("div")<{ top: number; chatExpanded: boolean }>(
     right: 0,
     pointerEvents: chatExpanded ? "auto" : "none",
     overflow: "hidden",
-    scrollbarGutter: "stable",
+    scrollbarGutter: chatExpanded ? "auto" : "stable",
```
ChristopherChudzicki, line 52: As discussed in slack, this has the downside that users with classic scrollbars see a small layout shift when expanding/collapsing.

Another way to get the divider to be full-width w/o the layout shift would be to keep this stable, and set the divider's width to 100vw when it's open. Then it will overflow to the edge of the screen. (It needs to be 100% when not open, though, or it overflows into the scrollbar undesirably.)

That said, my suggest seems a little hacky. Up to you.

ChristopherChudzicki (approved):
👍

frontends/main/src/page-components/LearningResourceExpanded/LearningResourceExpanded.tsx
jonkafton: I think the shift is ok in this case as the scrollbars appear occupying space, and the button shifts to remain centered. As the shift coincides with the scrollbars appearing, it's inoffensive. Let's get feedback though and can revisit if needed.

PR #2064, opened 2025-02-20
Handle "next" query string param in CustomLogoutView

gumaerc:
Closes https://github.com/mitodl/hq/issues/6789

This PR sets up the `CustomLogoutView`'s `_keycloak_logout_url` function to properly pass the `next` query string param to Keycloak as `post_logout_redirect_uri` if it is passed along in the initial request to `/logout`. In this case, it would be sent instead of the value of `settings.LOGOUT_REDIRECT_URL`.

### How can this be tested?
 - Spin up `mit-learn` locally on this branch
 - Log in to the site at http://localhost:8062/ in your usual way
 - Click the User Menu in the upper right, then right click on "Log Out" and click "Copy link address"
 - Open a new tab and paste in the URL, but don't go to it yet
 - At the end of the URL, add: `?next=https://www.google.com
 - Verify that when you hit enter, you are brought to Google
 - Navigate back to your instance of MIT Learn and verify that you are logged out

### DIFF
--- a/authentication/views.py
+++ b/authentication/views.py
@@ -5,6 +5,7 @@ from urllib.parse import urlencode
 from django.conf import settings
 from django.contrib.auth import views
 from django.shortcuts import redirect
+from social_core.utils import sanitize_redirect
 from social_django.utils import load_strategy
 
 from authentication.backends.ol_open_id_connect import OlOpenIdConnectAuth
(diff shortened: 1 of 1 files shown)


authentication/views.py
```diff
@@ -33,12 +33,13 @@ def _keycloak_logout_url(self, user):
             user, provider=OlOpenIdConnectAuth.name
         ).first()
         id_token = user_social_auth_record.extra_data.get("id_token")
+        qs_next = self.request.GET.get("next")
```
rhysyngsun, line 36: You can use social auth's redirect sanitization for this: https://github.com/python-social-auth/social-core/blob/master/social_core/actions.py#L27-L31. You'll need to pull the value of allowed redirects more directly from the settings though.

ChristopherChudzicki (approved):
👍 

**Note:** To test this, I rebased against `mb/apisix` since I don't have keycloak set up locally, and that branch (soon to be merged) makes running local keycloak easy.

Also had to add `SOCIAL_AUTH_ALLOWED_REDIRECT_HOSTS=["google.com"]` to `backend.local.env`.

PR #2074, opened 2025-02-26
Make embedding generation task use correct run

shanbady:

Closes https://github.com/mitodl/hq/issues/6832

This Pr makes it so that the embedding process for contentfiles pulls the contentfiles from the "next run" of the learning resource rather than the latest run.

### How can this be tested?
1. Checkout this branch.
2. embedding contentfiles should work as expected ```python manage.py generate_embeddings --all```

### DIFF
--- a/vector_search/tasks.py
+++ b/vector_search/tasks.py
@@ -104,7 +104,11 @@ def start_embed_resources(self, indexes, skip_content_files, overwrite):
                     .exclude(readable_id=blocklisted_ids)
                     .order_by("id")
                 ):
-                    run = course.next_run
+                    run = (
+                        course.next_run
+                        if course.next_run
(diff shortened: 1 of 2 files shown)


mbertrand (changes requested):
Turns out this won't work with OLL courses (they are all archived, so there is no "next" run).  Looking at the `sync_edx_course_files` function, it does this:

```python
        resource = run.learning_resource
        if run != (
            resource.next_run
            or resource.runs.filter(published=True).order_by("-start_date").first()
        ): 
            continue
 ```
 
So the code should probably make sure that resource.next_run isn't false, and if it is then use the run with the latest start date.

mbertrand (approved):
👍

PR #2075, opened 2025-02-26
Add all Contentfile metadata to chunk responses

shanbady:

Closes https://github.com/mitodl/hq/issues/6829

Currently, when retrieving contentfile chunks via the [vector content files endpoint](https://api.rc.learn.mit.edu/api/v0/vector_content_files_search/), we return the payload that was initially stored with the embedding document - however, for learning resources, since there is a 1-1 relation with what is in qdrant, we pull the serialized resources from the database and render that.

This PR makes it so that the vector contentfile endpoint does the same. What prompted this was the need for having the edx_block_id in the response.

### How can this be tested?
1. Checkout this branch
2. make sure you have some contentfiles locally otherwise populate them `python manage.py backpopulate_mitxonline_files`
3. generate embeddings for contentfiles via `python manage.py generate_embeddings --all` or `python manage.py generate_embeddings --resource-ids <comma separated resource ids>` 
4. visit your [local contentfiles vector](http://open.odl.local:8063/api/v0/vector_content_files_search/?limit=23) endpoint and inspect the response. compare this to [the production response](https://api.rc.learn.mit.edu/api/v0/vector_content_files_search/?limit=23) - see that there is more info including the edx_block_id and database specific ids.


abeglova (approved):
lgtm

PR #2076, opened 2025-02-27
Tie chatbots to URL parameters

ChristopherChudzicki:
Closes https://github.com/mitodl/hq/issues/6809

This PR:
- Changes the smoot-design theme's `LinkAdapter` to support shallow routing 
- Recommendation bot drawer is now tied to a URL paramter, `?recommender` 
- Syllabus chat expanded/collapsed state now tied to URL parameter, `?syllabus`

### How can this be tested?
1. Enable the chatbots via `home-page-recommendation-bot` and `lr-drawer-chatbot` if not already enabled.
2. On the homepage, click "Ask Tim" button. Expect:
    - drawer opens
    - URL updates to `http://learn.odl.local:8062/?recommender`
3. Close drawer; `?recommender` should be removed from url.
4. In URL bar, navigate to http://learn.odl.local:8062/?recommender ... The drawer should be open. 
5. Now test Syllabus Chat. Click on any course card to open the resource drawer.
6. Click "AskTim". Expect a URL like `?resource=123&syllabus=`
7. Refresh the page. Syllabus chat should still be expanded.
8. Close the drawer. URL params should be removed.

### DIFF
--- a/frontends/main/src/page-components/LearningResourceExpanded/LearningResourceExpanded.tsx
+++ b/frontends/main/src/page-components/LearningResourceExpanded/LearningResourceExpanded.tsx
@@ -4,7 +4,6 @@ import { theme } from "ol-components"
 import type { ImageConfig, LearningResourceCardProps } from "ol-components"
 import { ResourceTypeEnum } from "api"
 import type { LearningResource } from "api"
-import { useToggle } from "ol-utilities"
 import InfoSection from "./InfoSection"
 import type { User } from "api/hooks/user"
 import TitleSection from "./TitleSection"
@@ -13,6 +12,7 @@ import ResourceDescription from "./ResourceDescription"
 import { FeatureFlags } from "@/common/feature_flags"
 import { useFeatureFlagEnabled } from "posthog-js/react"
 import AiSyllabusBotSlideDown from "./AiChatSyllabusSlideDown"
+import { RESOURCE_DRAWER_PARAMS } from "@/common/urls"
 
 const DRAWER_WIDTH = "900px"
 
@@ -112,6 +112,7 @@ const TopCarouselContainer = styled.div({
 
 type LearningResourceExpandedProps = {
   resourceId: number
+  chatExpanded: boolean
   titleId?: string
   resource?: LearningResource
   user?: User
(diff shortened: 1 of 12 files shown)


pdpinch:
How/might this work with the Tutor and VideoGPT bots? I'm not sure this is a meaningful question since I don't know if they use smoot design.

frontends/main/src/page-components/LearningResourceExpanded/LearningResourceExpanded.tsx
```diff
@@ -126,6 +127,25 @@ type LearningResourceExpandedProps = {
...
 
+const closeChat = () => {
+  const params = new URLSearchParams(window.location.search)
+  params.delete(RESOURCE_DRAWER_PARAMS.syllabus)
+  window.history.replaceState({}, "", `?${params.toString()}`)
+}
+const openChat = () => {
+  const params = new URLSearchParams(window.location.search)
+  params.set(RESOURCE_DRAWER_PARAMS.syllabus, "")
+  window.history.replaceState({}, "", `?${params.toString()}`)
```
jonkafton, line 138: I'd be surprised if this causes anyone much bother - small request though to remove the trailing = so we have `/?resource=16993&syllabus` (not `/?resource=16993&syllabus=`).

We prob need to do it manually:

```
  params.append(RESOURCE_DRAWER_PARAMS.syllabus, "")
  window.history.replaceState({}, "", `?${params.toString().replace(/=$/g, "")}`)
```

jonkafton (approved):
Works great. Added a non blocking change request for very small thing.

ChristopherChudzicki:
> How/might this work with the Tutor and VideoGPT bots? I'm not sure this is a meaningful question since I don't know if they use smoot design.

@pdpinch It won't affect them at all. No code from this repo touches the tutor / video bot; it's all in https://github.com/mitodl/smoot-design.

It's a bit less obvious to me how to do something similar for the tutor/video bots. When you click "AskTim" in edx,  the xblock iframe sends a bunch of configuration information (block id, title initial questions, chatbot api endpoint—tutor vs video) to the parent MFE, which then opens the drawer. So I think we'd have to either:
1. encode all the data that the XBlock passes to the chat drawer (title, block id, initial messge, ...) as part of the URL. 
2. or jsut put the xblock id in the URL, and add some code for MFE to send a message to problem iframe to retrieve the relevant information.
    - This might be simpler if the iframe's can access the parent page URL. Generally I don't think they can, but the iframe can access the iframe's document.referrer, which might be the same as the parent URL.

frontends/main/src/page-components/LearningResourceExpanded/LearningResourceExpanded.tsx
ChristopherChudzicki: Going to live with this potential extra, inconsequential `=`, rather than manual altering it.

PR #2069, opened 2025-02-24
fix: opensearch container on ARM64 based architecture

arslanashraf7:
### What are the relevant tickets?
None, Noticed it while running the project containers on Apple Silicon(M4) machine.

### Description (What does it do?)
I couldn't run the `Opensearch` container while setting up this project on Macbook Pro m4. While running the opensearch container using the current configuration breaks with the JDK version. This seems to be an incompatible JDK version error.

This seems to be a known issue in JDK with ARM architecture reported
- https://bugs.openjdk.org/browse/JDK-8345296
- https://github.com/opensearch-project/opensearch-build/issues/5217

I've applied the suggested solution based on the above reported issues.

**The error log:**

```
2025-02-24 16:00:45 Disabling OpenSearch Security Plugin
2025-02-24 16:00:45 Enabling execution of OPENSEARCH_HOME/bin/opensearch-performance-analyzer/performance-analyzer-agent-cli for OpenSearch Performance Analyzer Plugin
2025-02-24 16:00:45 #
2025-02-24 16:00:45 # A fatal error has been detected by the Java Runtime Environment:
2025-02-24 16:00:45 #
2025-02-24 16:00:45 #  SIGILL (0x4) at pc=0x0000ffff8fd3fc5c, pid=33, tid=34
2025-02-24 16:00:45 #
2025-02-24 16:00:45 # JRE version:  (21.0.6+7) (build )
2025-02-24 16:00:45 # Java VM: OpenJDK 64-Bit Server VM (21.0.6+7-LTS, mixed mode, sharing, tiered, compressed oops, compressed class ptrs, g1 gc, linux-aarch64)
2025-02-24 16:00:45 # Problematic frame:
2025-02-24 16:00:45 # j  java.lang.System.registerNatives()V+0 java.base@21.0.6
2025-02-24 16:00:45 #
2025-02-24 16:00:45 # No core dump will be written. Core dumps have been disabled. To enable core dumping, try "ulimit -c unlimited" before starting Java again
2025-02-24 16:00:45 #
2025-02-24 16:00:45 # An error report file with more information is saved as:
2025-02-24 16:00:45 # /usr/share/opensearch/hs_err_pid33.log
2025-02-24 16:00:45 [0.024s][warning][os] Loading hsdis library failed
2025-02-24 16:00:45 #
2025-02-24 16:00:45 # The crash happened outside the Java Virtual Machine in native code.
2025-02-24 16:00:45 # See problematic frame for where to report the bug.
[...]

PR #2082, opened 2025-02-28
fix: env based _JAVA_OPTIONS for opensearch container

arslanashraf7:
None, More flexible fix for https://github.com/mitodl/mit-learn/pull/2069

- Makes the `_JAVA_OPTIONS` setting in opensearch container configurable through env vars. Earlier, this was hardcoded to `-XX:UseSVE=0` but the x86 systems pick it and don't understand it hence stopping the container right after starting it. 

- Anyone needing `-XX:UseSVE=0` would now set it in their `backend.local.env` file

- [ ] Desktop screenshots

### How can this be tested?
We might want multiple people to test it by having different machines based on x86, ARM architecture.

- Switch to this branch and try to start your OpenSearch container 
- Ideal, Only, if you could destroy your existing Opensearch container and rebuild and start it
- Make sure that your container starts normally.

### DIFF
--- a/env/backend.env
+++ b/env/backend.env
@@ -33,7 +33,3 @@ UWSGI_THREADS=35
 
 TIKA_SERVER_ENDPOINT=http://tika:9998/
 TIKA_CLIENT_ONLY=True
-
-# _JAVA_OPTIONS for Opensearch container are not consistent between different CPU architectures e.g ARM and x86.
-# Here, anyone can set options like -XX:UseSVE=0 based on their CPU.
-JAVA_OPTIONS=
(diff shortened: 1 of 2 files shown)


env/backend.env
```diff
@@ -33,3 +33,7 @@ UWSGI_THREADS=35
 
 TIKA_SERVER_ENDPOINT=http://tika:9998/
 TIKA_CLIENT_ONLY=True
+
+# _JAVA_OPTIONS for Opensearch container are not consistent between different CPU architectures e.g ARM and x86.
+# Here, anyone can set options like -XX:UseSVE=0 based on their CPU.
+JAVA_OPTIONS=
```
mbertrand, line 39: Since it's null by default, could move this to `backend.local.example.env` instead, and anyone who needs it can add  `JAVA_OPTIONS=-XX:UseSVE=0` to their `backend.local.env` file.

mbertrand (approved):
Works great on my old mac!  Just have a minor suggestion. 👍

env/backend.env
mbertrand: Looks good

mbertrand (approved):
👍

rhysyngsun (approved):
LGTM, works again for me

shanbady (approved):
👍

PR #2107, opened 2025-03-03
Increase nginx header size limit to 12k

rhysyngsun:

Closes https://github.com/mitodl/hq/issues/6858

Increases the nginx header size limit to 12k

### How can this be tested?

`docker compose up` and verify the site is reachable and no errors from nginx.


mbertrand (approved):
👍 assuming tests pass

PR #2110, opened 2025-03-04
Chatbot UI issues

jonkafton:

Closes: https://github.com/mitodl/hq/issues/6815

- The bulk of the change is for the Syllabus bot to share the parent drawer scroll container within the chat slide down (see acceptance in test checks below).

<br />

- AskTIM hero button

image

image

<br />

- Entry screen submit prompt (also no longer disabled when empty):

image

image

<br />

- Alignment with prompt suggestions on mobile:

image

- Offset "AskTIM about this course" opener for visual centering:

image

image

### How can this be tested?

Syllabus bot drawer scrolling:
- Set `chatEnabled` to true in LearningResourceExpanded, open a resource drawer and find the "AskTIM about this course" opener.
  - The chat entry screen should slide down smoothly.
  - The background resource content should not shift as the chat slides down.
  - The background resource content should remain at the same scroll position when the chat is closed.
  - After submitting a prompt on the entry screen, enter further prompts until the message window is taller than the available space.
      - The scrollbar should occupy the full height of the window.
      - The prompt input and disclaimer should be pinned to the bottom.
      - The drawer title and AskTIM opener should remain fixed.
      - The messages should scroll to bottom when a new prompt is submitted.
      - The messages scroll should snap to the bottom when scrolled near the bottom (threshold is set to 200px).
      - The messages scroll should not snap to the bottom when not scrolled near the bottom.
      - The resource drawer content should not be focusable by tab or visible to screen readers.
- Above should be true and layout preserved with always-visible scrollbar gutters (on macOS, settings => Scroll bar behavior -> Show scroll bars -> Always).
  - Chat should not be scrollable horizontally.
- Above should be true and layout preserved at mobile screen widths.


ChristopherChudzicki (changes requested):
A few issues I noticed while reviewing this:

1. Padding below disclaimer too small
    Screenshot 2025-03-05 at 11 18 37 AM
    Screenshot 2025-03-05 at 11 18 19 AM
2. Scrollsnap issues:
     -  If I scroll up while streaming, scroll is pulled back to bottom. _This might not be a big deal, but most streaming chats I've seen respect that if you scroll up during streaming, you're trying to read something above, and so don't pull you back down._ 
3. ⚠️  **Header not fixed, chat visible after close** See video below 


https://github.com/user-attachments/assets/32a00ddc-96a1-4e43-997e-03c6f2512945

4. If I go directly to the chat url, e.g., http://learn.odl.local:8062/?resource=2326&syllabus=, then the splash screen has too much height. 

#### Regarding (3)

I think this is caused by the root div of `AiSyllabusBotSlideDown` — its height is constrained by something (screen height?) and it has absolute positioning. One of its children has absolute positioning, too.

I think we have two options:
1. Stop using absolutely positioned divs for stuff, let divs have their normal height. This makes the "slide" transitions hard. But "collapse" transitions are much easier. I took a quick stab at this on jk/6815-chatbot-ui-issues-cc-testing using MUI's `Collapse` component. (I also used `Fade` for the other drawer content, mostly to handle the accessibility stuff for us.)
    - IMO this generally worked very well, and using Collapse (and Fade) was very easy..
    - The only piece I know I didn't get quite right is the min-height of the (non-splash) chat component.
2. Or don't use root of drawer as scroll container, but something else. (Possibly irrespective of chat opened/closed).

jonkafton:
> 1. Padding below disclaimer too small

I see a 14px padding. I haven't fixed intentionally, so please let me know if you still see the issue.

image

<br />

> 2. If I scroll up while streaming, scroll is pulled back to bottom. This might not be a big deal, but most streaming chats I've seen respect that if you scroll up during streaming

If you scroll up you should quickly be outside the snap threshold, so I don't see that as an issue, though let me know if I have missed the point. I set this to 200px which felt about right. The smoot-design ScollSnap defaults to only 2px.

> 3. ⚠️ Header not fixed, chat visible after close 

That's fixed.

> If I go directly to the chat url, e.g., http://learn.odl.local:8062/?resource=2326&syllabus=, then the splash screen has too much height.

Also fixed.
 
Other notes.

> Stop using absolutely positioned divs for stuff, let divs have their normal height. 

I've tried a lot of options here - we have a unique combination of factors, some of them:

- We need a slide transition (container must be positioned).
- We want to have a scroll on a parent element (the drawer paper), which contains nested fixed positioned elements (title, opener), that nest content that grows in height.
- Although these elements are fixed position (in terms of the design), we cannot use fixed positioning as a/ it's within a fixed positioned drawer, so we must control the width (not too hard), but b/ we are abutting the right of the viewport and must accommodate classic and floating scrollbars - there is no way to hook into the gutter presence. The fallback is to use sticky positioning, though we must then control the height of containers (or we end up with 3. ⚠️ Header not fixed, or the chat input not properly pinned to the bottom, etc).
- We want the scrollbar to overlay positioned elements that are upstream from the scrolling content (limited options for scrolling close to the content). Related to "2. Or don't use root of drawer as scroll container"

frontends/main/src/page-components/LearningResourceExpanded/AiChatSyllabusSlideDown.tsx
```diff
@@ -113,60 +140,99 @@ const getInitialMessage = (
...
+  contentTopPosition: number
+  chatTransitionState: ChatTransitionState
 }) => {
   const user = useUserMe()
+  const ref = useRef<HTMLDivElement>(null)
+
+  useEffect(() => {
+    const element = ref.current
+    const _onTransitionEnd = (event: TransitionEvent) => {
+      if (event.target === element) {
```
ChristopherChudzicki, line 192: i guess you could also check propertyName https://developer.mozilla.org/en-US/docs/Web/API/TransitionEvent/propertyName

ChristopherChudzicki (approved):
Working well 👍

PR #2114, opened 2025-03-06
Always show chat entry screen when navigating courses

jonkafton:

Follow on for https://github.com/mitodl/hq/issues/6836

The previous PR https://github.com/mitodl/mit-learn/pull/2111 ensure the course details are updated in the chat, but also changed the entry screen to only show once within a drawer (ie. selecting a new course from within the drawer and opening another chat would take you straight to the chat screen). This change keys on the resource readable ID to ensure the chat component with entry screen loads fresh for each new course.


ChristopherChudzicki (approved):
Approving this, but let's confirm with @sovsey before merging.

As mentioned in slack:

> there's a small quirk if you go from course X -> Y -> X (back to same course).
> See video... Second time visiting X, splash screen shows, but so does the previous chat history.

https://github.com/user-attachments/assets/b711831d-6e28-4b02-93ba-cd3cdd0fc766


IMO, fixing this would be straightforward if we refactor `AiChat` to be `AiChatProvider + AiChatDisplay + `useAiChat()` hook. Then the splash screen can use `useAiChat` to determine if it has already been shown. _The splash screen could also be moved to smoot, but not required._

PR #2119, opened 2025-03-10
Create/login/logout users via apisix middleware only

mbertrand:
- https://github.com/mitodl/hq/issues/6885

Related ol-infrastructure PR is https://github.com/mitodl/ol-infrastructure/pull/3031

- Removes social auth pipeline and other social auth code
- Adjusts the apisix middleware to handle user registration and login/logout
- Got rid of the logout suffix code since the default `logout` url works fine

### How can this be tested?
- If you don't have any users yet, create them on the main branch:
  - Run `COMPOSE_PROFILES=backend,frontend,keycloak,apisix docker compose up` and wait for everything to finish starting up.
  - Log in and log out 2 different users: `student@odl.local` and `admin@odl.local` password is username minus `@odl.local`
- Stop docker containers, switch to this branch, and start up docker containers again.
- Check that `user.global_id` values for the 2 users are the same as in the `social_auth_usersocialauth` table.
- Log in and out as one of those users
- Delete the two users: `User.objects.all().delete()`
- Log in/out again as those 2 users again, they should be recreated with non-null global ids.

### Additional Context
I did not yet remove `social_django`  from `INSTALLED_APPS` or `pyproject.toml` because it is needed for one of the migrations here.  I will remove it in a subsequent PR.

Much of the apisix middleware code that was originally copied over from learn-ai was brought back in this PR (handling user/profile creation, login, logout).

One thing I initially tried was to have the apisix middleware handle user/profile creation only if necessary then pass off login/logout duties to `RemoteUserMiddleware` but that didn't seem to work.


main/middleware/apisix_user.py
```diff
@@ -33,7 +49,77 @@ def decode_apisix_headers(request, header):
...
+        User object
+
+    """
+
+    User = get_user_model()
+
+    if not decoded_headers:
+        return None
+
+    global_id = decoded_headers.get("global_id", None)
```
rhysyngsun, line 78: If the global id is `None`, we should log an error and return as if there were no headers, otherwise we'll end up creating a user per request.

main/middleware/apisix_user.py
```diff
@@ -33,7 +49,77 @@ def decode_apisix_headers(request, header):
...
+
+    user, created = User.objects.filter(global_id=global_id).get_or_create(
+        defaults={
+            "global_id": global_id,
```
rhysyngsun, line 82: Should be this, the other version might work in simple cases but direct use of `get_or_create` without the filter is probably safer against edge cases.:

```suggestion
    user, created = User.objects.get_or_create(
        global_id=global_id,
        defaults={
```

rhysyngsun (changes requested):

users/migrations/0006_user_global_id.py
```diff
@@ -0,0 +1,36 @@
...
+
+    users_with_uid = User.objects.annotate(
+        social_uid=models.Subquery(
+            UserSocialAuth.objects.filter(
+                user=models.OuterRef("pk"),
+            ).values("uid")[:1]
+        )
+    )
+
+    users_with_uid.update(global_id=models.F("social_uid"))
```
rhysyngsun, line 19: Should be careful here, this could be an issue with large numbers of users because it takes a row-level lock on the rows effected by this query, plus the whole migration is wrapped in a transaction. So this is effectively a full table lock which could either fail or cause errors on anything else trying to update the row (something as trivial as login which updates `last_login` could hit this). You probably want to do this in batches like I did here: https://github.com/mitodl/mit-learn/blob/main/users/migrations/0005_set_user_scim_id.py#L19

Note: it needs to be done in a separate migration that has `atomic = False` so that it doesn't wrap the migration in a transaction

users/migrations/0006_user_global_id.py
```diff
@@ -1,22 +1,31 @@
...
+
 
 def transfer_global_id(apps, schema_editor):
     """Transfer the social uid to the user global_id field"""
     UserSocialAuth = apps.get_model("social_django", "UserSocialAuth")
     User = apps.get_model("users", "User")
 
-    users_with_uid = User.objects.annotate(
+    users_with_uid = User.objects.filter(
+        id__in=User.objects.filter(global_id__isnull=True).only("id")[:BATCH_SIZE]
```
rhysyngsun, line 18: Needs an additional `global_id__isnull=False` to avoid an infinite loop - using the `while` syntax depends on the `update()` call updating the rows for a given batch so that they won't match on the next call since you're always fetching the first 10k rows.

main/middleware/apisix_user.py
```diff
@@ -33,7 +49,77 @@ def decode_apisix_headers(request, header):
...
+        log.error("No global_id found in APISIX headers")
+        return None
+
+    user, created = User.objects.update_or_create(
+        global_id=global_id,
+        defaults={
+            "username": decoded_headers.get("username", ""),
+            "email": decoded_headers.get("email", ""),
+            "first_name": decoded_headers.get("first_name", ""),
+            "last_name": decoded_headers.get("last_name", ""),
```
rhysyngsun, line 89: We need to also update `profile.name`, because otherwise new users don't have their name show up. Does APISIX pass this?

mbertrand: updated to get it from fullName, which should now be sent from the local keycloak instance

rhysyngsun (commented):
Everything is looking good, I pushed some fixes for the following:
- Removed references to `/login/ol-oidc` in the frontend
- Removed a usage of `/django_logout` in `settings.py`
- Updated realm settings to enable registration and password reset

Someone else should review these and the Copilot comments about redirect url validation should be addressed too.

authentication/views.py
```diff
@@ -61,18 +45,25 @@ def get(
...
+            logout(request)
+        return redirect(get_redirect_url(request))
 
-        The logout redirect path the user follows is:
 
-        - api.example.com/logout (this view)
-        - keycloak.example.com/realms/REALM/protocol/openid-connect/logout
-        - api.example.com/app (see main/urls.py)
-        - app.example.com
+class CustomLoginView(views.LogoutView):
```
jkachel, line 55: Should this be `views.LoginView`? Not sure this is actually a thing that needs to be addressed but it just sort of stuck out that the login view subclassed the logout view.

jkachel (approved):
I think this looks OK. Log in/out worked as expected and I got users with SSO IDs as I expected, I got onboarding for new users and user lists were created. I am working from a clean slate deployment so using the supplied Keycloak realm data.

I did flag something but not sure it's an actual issue so not considering it a blocker for approval.

jkachel (changes requested):
Sorry, was reviewing prior review comments and noticed something that I'd missed in the global ID migration. Should be a quick fix though.

users/migrations/0007_user_global_id_assign.py
```diff
@@ -0,0 +1,44 @@
...
+            )
+        )
+        .filter(social_uid__isnull=False)
+    )
+
+    while num_updates := users_with_uid.update(global_id=models.F("social_uid")):
+        log.info("Updated %s user records", num_updates)
+
+
+class Migration(migrations.Migration):
```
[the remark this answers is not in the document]
mbertrand: added

jkachel (approved):
LGTM 👍

rhysyngsun (approved):

PR #2124, opened 2025-03-13
Migrated AiChat entry screen

jonkafton:

https://github.com/mitodl/hq/issues/6902

Upgrades to the latest Smoot Design with the migrated chat entry screen. The external scroll container is now handled in the AiChat component, so we're removing styles that enabled it - such as the unconstrained message container height and fixed title and prompt input.

### How can this be tested?

- Assign `recommendationBotEnabled` and `chatEnabled` to true in HeroSearch.tsx and LearningResourceExpanded.tsx.
- Open the chatbots via the AskTIM button on the homepage and the slide down in course drawers.
- There should be no functional change or changes to the designs with the exception that the padding around the chat disclaimer has been adjusted to match [designs](https://www.figma.com/design/Eux3guSenAFVvNHGi1Y9Wm/MIT-Design-System?node-id=16842-209507&m=dev).


ChristopherChudzicki (approved):
👍  Working well!

Nice to have this moved to smoot.

PR #2135, opened 2025-03-17
Fix contentfile embeddings

shanbady:

Closes https://github.com/mitodl/hq/issues/6931

This PR resolves:
- an [initial issue](https://github.com/mitodl/hq/issues/6931) where the non-opensearch contentfile endpoint errors out if a contentfile's run is null (all the marketing page contentfiles we generate)
- a separate issue in which the embedding task does not pick up on contentfiles that don't have runs.This pr adjust the serializers so that they work with "run-less" contentfiles.

### How can this be tested?
1. checkout this branch
2. restart celery
3. run any etl pipeline for courses `python manage.py backpopulate_mitxonline_data`
4. upon completion, inspect the database and see that there are ContentFile objects with a "file_type" of "marketing_page"
```python
from learning_resources.models import ContentFile, LearningResource
ContentFile.objects.filter(file_type="marketing_page").all()
```
5. get the resouurce id of one of these marketing page contentfiles 
```python 
# print resource ids
print(ContentFile.objects.filter(file_type="marketing_page").values_list('learning_resource',flat=True))
```
6. generate embeddings for that resource:
```bash
python manage.py generate_embeddings --resource-ids <resource id> --overwrite
```
7. once complete visit your [local qdrant dashboard](http://localhost:6333/dashboard#/collections/)
8. search the contentfile collection for "file_type:marketing_page" and see that there is a result(s) Screenshot 2025-03-17 at 9 56 58 AM

9. browse around the regular [contentfiles endpoint](http://open.odl.local:8063/api/v1/contentfiles/?limit=10&offset=25) - it should not throw exceptions when encountering a contentfile without a run

### DIFF
--- a/learning_resources/serializers.py
+++ b/learning_resources/serializers.py
@@ -773,12 +773,10 @@ class ContentFileSerializer(serializers.ModelSerializer):
         # prefetch related run and learning resource
         queryset = models.ContentFile.objects.prefetch_related(
             "learning_resource__course",
-            "content_tags",
             "learning_resource__platform",
-            Prefetch(
-                "run",
-                queryset=models.LearningResourceRun.objects.for_serialization(),
-            ),
+            "run__learning_resource__course",
+            "run__learning_resource__platform",
+            "content_tags",
             Prefetch(
                 "learning_resource__topics",
                 queryset=models.LearningResourceTopic.objects.for_serialization(),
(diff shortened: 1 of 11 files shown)


learning_resources/serializers.py
```diff
@@ -754,38 +754,61 @@ class ContentFileSerializer(serializers.ModelSerializer):
...
+
+    @extend_schema_field({"type": "string"})
+    def get_resource_id(self, instance):
+        return str(self.get_learning_resource(instance).id)
+
+    @extend_schema_field(LearningResourceTopicSerializer(many=True))
+    def get_topics(self, instance):
+        return LearningResourceTopicSerializer(
+            self.get_learning_resource(instance).topics, many=True
+        ).data
```
mbertrand, line 805: Getting lots of n+1 queries now when querying the opensearch contentfiles (and probably when indexing them too), probably due to the above changes, so needs to be refactored.

mbertrand (changes requested):
Have to go offline for a bit, here's one preliminary comment.

learning_resources/serializers.py
shanbady: Resolved the n+1 queries

learning_resources/serializers.py
```diff
@@ -769,6 +769,40 @@ class ContentFileSerializer(serializers.ModelSerializer):
...
+            ),
+            Prefetch(
+                "content_tags",
+            ),
+            Prefetch(
+                "learning_resource__departments",
+                queryset=models.LearningResourceDepartment.objects.for_serialization().select_related(
+                    "school"
+                ),
+            ),
```
mbertrand, line 795: Still seeing some:

```
Potential n+1 query detected on `LearningResourceDepartment.school
Potential n+1 query detected on `LearningResourceTopic.channel_topic_details
Potential n+1 query detected on `LearningResourceRun.learning_resource
```

mbertrand (approved):
Looks good! 👍

PR #2141, opened 2025-03-18
global_id migrations based on scim id

mbertrand:
Related to https://github.com/mitodl/mit-learn/pull/2119

Assigns global_id based on scim external id

### How can this be tested?
Follow instructions for https://github.com/mitodl/mit-learn/pull/2119


main/middleware/apisix_user_test.py
```diff
@@ -62,6 +62,45 @@ def test_get_request(mocker, mock_login):
...
+        META={
+            "HTTP_X_USERINFO": b64encode(json.dumps(apisix_user_info).encode()),
+        },
+        user=AnonymousUser(),
+    )
+    apisix_middleware = ApisixUserMiddleware(mocker.Mock())
+    apisix_middleware.process_request(mock_request)
+    mock_login.assert_called_once()
+    user.refresh_from_db()
+    assert user.username == user.username
```
rhysyngsun, line 80: This and other `assert user.FIELD == user.FIELD` assertions will always be true, I think we want to compare these values against what's in `apisix_user_info`

main/middleware/apisix_user_test.py
```diff
@@ -62,6 +62,45 @@ def test_get_request(mocker, mock_login):
...
+    assert user.email == apisix_user_info["email"]
+    assert user.global_id == apisix_user_info["sub"]
+
+
+@pytest.mark.django_db(transaction=True)
+def test_get_request_existing_user_with_globalid(mocker, mock_login):
+    """Test that a valid request creates a new user."""
+    close_old_connections()
+    user = UserFactory.create(email=apisix_user_info["email"], global_id="abc123")
```
rhysyngsun, line 89: I'm not sure I follow this test because I don't think this will match on the `filter()` because it doesn't match on `global_id` or null `global_id` + `email`. I would've expected this to fail because it'll try to create a user and conflict.

rhysyngsun (changes requested):

main/middleware/apisix_user_test.py
rhysyngsun: Probably worth having this test, but it should wrap `process_request()` in a `pytest.raises(IntegrityError)` and then you don't need any other asserts.-

main/middleware/apisix_user_test.py
```diff
@@ -62,6 +62,65 @@ def test_get_request(mocker, mock_login):
...
+    apisix_middleware = ApisixUserMiddleware(mocker.Mock())
+    apisix_middleware.process_request(mock_request)
+    user.refresh_from_db()
+    assert user.username == old_username
+    assert user.global_id == "abc123"
+    assert user.email == apisix_user_info["email"]
+
+
+@pytest.mark.django_db(transaction=True)
+def test_get_request_existing_user_with_email(mocker, mock_login):
```
rhysyngsun, line 106: I think this is the same as the first test. We do need a test though where the `global_id` matches and the email should be updated though.

mbertrand:
Ready for another look, though javascript tests have failed twice now, have not looked into why yet but it shouldn't be related to this PR.

jkachel (approved):
I think this looks OK - I did not have any issues with login or logout. 👍

PR #2144, opened 2025-03-21
qdrant performance tweaks

shanbady:

Closes https://github.com/mitodl/hq/issues/6921

This Pr does the following:
- fix naming of indexed fields and data types in qdrant (some were missing or incorrectly named)
- adds "indexed_only" as a search param which tells qdrant to only look at indexed items (so it wont hang during large re-index operations)
- fixes a flaky test (test_learning_resources_display_info_list_view) https://github.com/mitodl/hq/issues/6975

### How can this be tested?
1. checkout this branch
2. recreate all the qdrant collections via `./manage.py generate_embeddings --recreate-collections --all```
3. it should embed without issues and search should work fine
4. go to the info page for each collection ([example](http://localhost:6333/dashboard#/collections/resource_embeddings.content_files#info)) and note that chunk_number is now an indexed field  

### Additional Context
Other changes that have been made on prod related to this:
- created 2 new nodes (this lets us do rolling updates without freezing the entire cluster)
- sharded both collections across nodes (replication factor 3)
- enabled binary quantization (performance boost and lower mem usage)
- updated qdrant instance to the latest version


vector_search/utils.py
```diff
@@ -311,27 +309,9 @@ def _process_content_embeddings(serialized_content):
...
-        try:
-            # sometimes we can't update the multi-vector if max size is exceeded
-
-            client.update_vectors(
-                collection_name=RESOURCES_COLLECTION_NAME,
-                points=resource_points,
-            )
-        except Exception as e:  # noqa: BLE001
-            msg = f"Exceeded multi-vector max size: {e}"
-            logger.warning(msg)
```
mbertrand, line 334: Just curious why these 2 chunks of code are no longer necessary

mbertrand (commented):
Waiting on embeddings to finish generating (taking awhile on ~550 contentfiles even with `QDRANT_CHUNK_SIZE=20`) but code looks good, just had a question about the deleted lines of code.

mbertrand (approved):
I deleted all but a handful of contentfiles to get the embedding generation to finish quickly, and everything seems to work great.  👍 assuming the deleted blocks of code aren't needed anymore.

PR #2145, opened 2025-03-21
Enable redirects on logout

mbertrand:
Closes https://github.com/mitodl/hq/issues/6923

Creates a new `/next_logout/?next=<url>` endpoint that will save the value of the `next` parameter to a short-lived redis key, then redirect to the usual `/logout/oidc` url.  
When keycloak/apisix redirects back to the `/logout/` view, the cached key is retrieved and the user is redirected to that url if set to anything, else it goes to the default `/app`
The authentication CustomLoginView and CustomLogoutView classes were also both changed to inherit from the standard django `View` class because a) the tests were throwing errors due to the imminent deprecation of allowing such views to accept GET requests, and b) because they weren't actually doing any login/logout functions (the apisix middleware is).

### How can this be tested?
- Set `MITOL_API_BASE_URL=http://api.open.odl.local:8065` in your `backend.local.env` file
- Run `COMPOSE_PROFILES=backend,frontend,keycloak,apisix docker compose up`
- Log in and log out as usual to make sure that still works normally.
- Log in
- Go to the url `http://api.open.odl.local:8065/next_logout/?next=https://ocw.mit.edu`
- You should be redirected to OCW
- Go back to http://open.odl.local:8062, you should be logged out
- Log in and out again

### Additional Context

~~The original idea was to set a cookie, but cookies can only be set in responses and not requests, which didn't mesh well with the logout flow.  So a temporary redis cache key/value pair was used instead.~~
Figured out that I can set a cookie from the apisix middleware for this.


gumaerc:
@mbertrand I tested this and everything worked perfectly, code looks good too. Good tests, and good call on using the Redis cache. It does look like you have some upstream conflicts to resolve before merging though. Just let me know when that's done and I'll approve the PR

mbertrand:
@gumaerc just did a fresh rebase,  ready for another look.

authentication/views.py
```diff
@@ -34,6 +34,16 @@ def get_redirect_url(request):
...
 
 
+def next_cache_key(username):
+    """
+    Get the cache key for the next URL for a user.
+
+    Args:
+        request: Django request object
+    """
+    return f"{username}_next_logout"
```
gumaerc, line 44: Is this cache key appropriate? I don't see this function being used anywhere, is it called automatically by Next for cache invalidation or something? Maybe it should be `f"{session_key}_auth"` or something along those lines?

gumaerc (approved):
Everything works great, just a question above about the cache key function.

authentication/views.py
mbertrand: That's left over from the redis implementation, deleted it.

PR #2152, opened 2025-03-25
typeerror nonetype error

shanbady:

Closes https://github.com/mitodl/hq/issues/6986

This PR fixes a 500 being thrown at the [learning resource display info](http://api.rc.learn.mit.edu/api/v1/learning_resource_display_info/) endpoint when a run has None for the prices attribute.

### How can this be tested?
1. checkout main 
2. visit the learning resource display info endpoint and make note of a run in the response
3. set the prices for that run to None and save it
4. go back to the endpoint - it should throw an exception (you may need to cache break by passing some random parameter) 
5. checkout this branch and see no exceptions


gumaerc (approved):
:+1:

PR #2142, opened 2025-03-18
Cc/card style updates

ChristopherChudzicki:
### Screenshots (if appropriate):
Screenshot 2025-03-18 at 4 44 38 PM

PR #2154, opened 2025-03-25
design refactor: make link black style bolder

ChristopherChudzicki:
None, but came across this discrepancy during https://github.com/mitodl/mit-learn/pull/2142

This PR updates the `Link` component to match designs:
- the change is `color="black"` now has a bolder typography.
- **There are no visual changes.** In all cases where we were using `color="black"` (previously also the default value of `color`, but that prop is now required) we were always overriding the styling to match figma on a case-by-case basis.

### How can this be tested?
Check that things look the same. Some things worth checking:
- the social media share icons on drawer
- the homepage "Non Degree Learning" link in hero section
- Card titles

### DIFF
--- a/frontends/ol-components/src/components/Link/Link.tsx
+++ b/frontends/ol-components/src/components/Link/Link.tsx
@@ -5,15 +5,14 @@ import { theme } from "../ThemeProvider/ThemeProvider"
 import { LinkAdapter } from "../LinkAdapter/LinkAdapter"
 
 type LinkStyleProps = {
+  color: "black" | "white" | "red"
   size?: "small" | "medium" | "large"
-  color?: "black" | "white" | "red"
   hovercolor?: "black" | "white" | "red"
   nohover?: boolean
 }
 
-const DEFAULT_PROPS: Required<LinkStyleProps> = {
+const DEFAULT_PROPS: Required<Omit<LinkStyleProps, "color">> = {
   size: "medium",
-  color: "black",
   hovercolor: "red",
   nohover: false,
 }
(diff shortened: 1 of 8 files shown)


frontends/ol-components/src/components/Link/Link.tsx
```diff
@@ -33,16 +32,22 @@ const NO_FORWARD = Object.keys({
  */
 const linkStyles = (props: LinkStyleProps) => {
```
ChristopherChudzicki, line 33: IMO it's a bit weird that the typography style depends on color, but that's what how our [designs](https://www.figma.com/design/Eux3guSenAFVvNHGi1Y9Wm/MIT-Design-System?node-id=470-11729&m=dev) do it, so worth matching that rather than applying the bolder black font on a case-by-case basis.

frontends/ol-components/src/components/Link/Link.tsx
```diff
@@ -5,15 +5,14 @@ import { theme } from "../ThemeProvider/ThemeProvider"
 import { LinkAdapter } from "../LinkAdapter/LinkAdapter"
 
 type LinkStyleProps = {
+  color: "black" | "white" | "red"
```
ChristopherChudzicki, line 8: I made `color` required because:
- it helped me check that this had no unintential changes
- in contrast to things like `nohover` and `size`, I don't think what the default  value of `color` should be is obvious.

frontends/ol-components/src/components/Link/Link.tsx
```diff
@@ -33,16 +32,22 @@ const NO_FORWARD = Object.keys({
...
+          ? { ...theme.typography.subtitle3 }
+          : { ...theme.typography.body3 },
+      medium:
+        color === "black"
+          ? { ...theme.typography.subtitle2 }
+          : { ...theme.typography.body2 },
+      large:
+        color === "black"
+          ? { ...theme.typography.subtitle1 }
+          : { ...theme.typography.body1 },
```
jonkafton, line 49: A little uneasy about this - we're saying that the same link with a different color adopts styles defined for specific typographic elements. They may match those styles, but if we change the `body1` styles, should a non-black link change or is it incidental? I do see that `color` in Figma is effectively a variant and not a style, so the issue is really in the naming there.

jonkafton (approved):
Looks good. It would be good though to review the relationship between our naming in components in code and Figma and guidelines around the degree to which Figma is canonical and should be synced to.

We have e.g. size: h5 and and color: BreadcrumbDark in the Link component in Figma - we're selective and it's not always clear what forms part of the _definitive_ design system.

frontends/ol-components/src/components/Link/Link.tsx
ChristopherChudzicki: I agree the naming is confusing. Best to match Figma, though, imo.

PR #2156, opened 2025-03-26
Make /users/me/ api endpoint accessible for anonymous users

mbertrand:
Related to https://github.com/mitodl/mit-learn/pull/2119

- Makes the `api/v0/users/me` endpoint accessibe to anonymous users
- Adjusts the `UserSerializer`, making `profile` optional (anon users don't have one) and adding an `is_authenticated` boolean field.

### How can this be tested?
- You should still see the `Login` button at the top if not authenticated and your name with a dropdown menu if authenticated.  Editing your profile at https://open.odl.local:8062/dashboard/profile should still work. Tests should pass.

### Additional Context
This is being done to see if a redirect to keycloak can be avoided on RC/prod when initially loading this url > 30 minutes after logging in.

### DIFF
--- a/frontends/api/src/generated/v0/api.ts
+++ b/frontends/api/src/generated/v0/api.ts
@@ -3391,6 +3391,12 @@ export interface PatchedUserRequest {
    * @memberof PatchedUserRequest
    */
   email?: string
+  /**
+   *
+   * @type {boolean}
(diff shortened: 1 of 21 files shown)


ChristopherChudzicki (approved):
👍 Glad the frontend can get rid of the 403 handling / extra (non-api) user property now.

PR #2158, opened 2025-03-27
Fix for potential missing file attribute in contentfile json

mbertrand:
Closes https://github.com/mitodl/hq/issues/6939

Handles missing/empty file attribute in contentfile JSON

### How can this be tested?
 - Make sure the following env variables are set in your backend.local.env file:
    ```
   OCW_LIVE_BUCKET=ocw-content-live-production
   AWS_ACCESS_KEY_ID
   AWS_SECRET_ACCESS_KEY
   ``` 
   

 - Run the following command, you should not see any logging errors like "TypeError: argument should be a str or an os.PathLike object"
 ```bash
 ./manage.py backpopulate_ocw_data --course-name 21m-342-composing-for-jazz-orchestra-fall-2008
 ```

I also changed a percolate info statement to a debug statement because it was filling up the console when running (@shanbady is this okay or are those log info statements necessary?)

### DIFF
--- a/learning_resources/etl/ocw.py
+++ b/learning_resources/etl/ocw.py
@@ -254,10 +254,15 @@ def transform_contentfile(
         content_type = CONTENT_TYPE_VIDEO
         file_s3_path = contentfile_data.get("transcript_file")
         image_src = contentfile_data.get("thumbnail_file")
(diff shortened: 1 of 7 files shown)


abeglova (approved):
👍

PR #2159, opened 2025-03-27
Make gemini optional by request

rhysyngsun:

N/A

This makes Gemini optional but folks can still request it manually to cut down on unwanted PR review noise.

### DIFF
--- a/.gemini/config.yaml
+++ b/.gemini/config.yaml
@@ -1,0 +1,10 @@
+---
+have_fun: false
+code_review:
+  disable: false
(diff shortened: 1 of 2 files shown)


.gemini/config.yaml
```diff
@@ -0,0 +1,10 @@
+---
+have_fun: false
```
ChristopherChudzicki, line 2: 🙏 thank you

ChristopherChudzicki (approved):

PR #2162, opened 2025-03-28
Decrease lookback window for embedding new resources and contentfiles

shanbady:
Closes https://github.com/mitodl/hq/issues/7008

We have celery tasks that pick up new learning resources and contentfiles from within the last day and run embeddings for them. This typically is not an issue unless we run etl pipelines manually in which case if there are several tens of thousands of new contentfiles, the periodic task (which runs every 30 minutes) may not complete and a new one gets spawned.

This PR reduces the lookback window to "embed new contentfiles/learning resources" from 40 minutes ago. 

### How can this be tested?
Its a faily straightforward adjustment to the timedelta for the lookback window. Tests have been adjusted for it and should pass.

To manually test:
1. checkout this branch.
2. find a contentfiles and learning resources and set their "created_on" attribute to within the last 40 minutes
```python
from learning_resources.models import LearningResource,ContentFile
import datetime
for i in range(10):
    lr = LearningResource.objects.all()[i]
    lr.created_on =now_in_utc() - datetime.timedelta(minutes=20)
    lr.save()
for i in range(10):
    cf = ContentFile.objects.all()[i]
    cf.created_on =now_in_utc() - datetime.timedelta(minutes=30)
    cf.save()
```
3. manually run the embeddings tasks to see that those resources are picked up:
```python
from vector_search.tasks import embed_new_content_files,embed_new_learning_resources
 embed_new_content_files.apply()
embed_new_learning_resources.apply()
```
when the task succeeds the number of "None" elements outputted should match the number of resources you would have expected to see get picked up
<code>succeeded in 0.00019108298874925822s: **[None, None, None, None, None, None, None, None, None, None]**</code>

### DIFF
--- a/main/settings_celery.py
+++ b/main/settings_celery.py
@@ -6,6 +6,16 @@ from celery.schedules import crontab
 
 from main.envs import get_bool, get_int, get_string
 
+"""
+the lookback window (in minutes) for the embeddings task
+the frequency of the embeddings tasks will be 1/4 of this for example
+lookback = 2 hours; task frequency = 30 minutes
+"""
+QDRANT_EMBEDDINGS_TASK_LOOKBACK_WINDOW = get_int(
+    name="QDRANT_EMBEDDINGS_TASK_LOOKBACK_WINDOW", default=60 * 2
+)
+EMBEDDING_SCHEDULE_MINUTES = int(QDRANT_EMBEDDINGS_TASK_LOOKBACK_WINDOW / 4)
+
 DEV_ENV = get_bool("DEV_ENV", False)  # noqa: FBT003
 USE_CELERY = True
 CELERY_BROKER_URL = get_string("CELERY_BROKER_URL", get_string("REDISCLOUD_URL", None))
@@ -147,13 +157,13 @@ if not DEV_ENV:
     CELERY_BEAT_SCHEDULE["daily_embed_new_learning_resources"] = {
         "task": "vector_search.tasks.embed_new_learning_resources",
         "schedule": get_int(
-            "EMBED_NEW_RESOURCES_SCHEDULE_SECONDS", 60 * 30
+            "EMBED_NEW_RESOURCES_SCHEDULE_SECONDS", 60 * EMBEDDING_SCHEDULE_MINUTES
         ),  # default is every 30 minutes
     }
     CELERY_BEAT_SCHEDULE["daily_embed_new_content_files"] = {
         "task": "vector_search.tasks.embed_new_content_files",
         "schedule": get_int(
-            "EMBED_NEW_CONTENT_FILES_SCHEDULE_SECONDS", 60 * 30
+            "EMBED_NEW_CONTENT_FILES_SCHEDULE_SECONDS", 60 * EMBEDDING_SCHEDULE_MINUTES
         ),  # default is every 30 minutes
     }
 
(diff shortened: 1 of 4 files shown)


vector_search/tasks.py
```diff
@@ -271,10 +271,10 @@ def embed_new_learning_resources(self):
 @app.task(bind=True)
 def embed_new_content_files(self):
     """
-    Embed new content files from the last day
+    Embed new content files from 40 minutes ago
     """
     log.info("Running content file embedding task")
-    delta = datetime.timedelta(days=1)
+    delta = datetime.timedelta(minutes=40)
```
abeglova, line 277: This works but it's not great because we need to remember to change it again if the celery beat schedule changes . The celery beat task is called "daily_embed_new_content_files" so i think this task being updated to run every 30 minutes instead of daily is how this error occurred in the first place (things moved around enough that i'm having trouble confirming from git blame). Can you create some setting variables for  CONTENT_FILE_EMBED_FREQUNCY_MINUETES and LEARNING_RESOURCES_EMBED_FREQUENCY_MINUERES and set the value here to CONTENT_FILE_EMBED_FREQUNCY_MINUETES + 10?

abeglova (changes requested):
Left a comment. Since the beat schedule and the lookback timedelta are dependent on each other they should read from one constant

abeglova:
Additionally, things sit in the celery queue for more than 10 minutes pretty frequently, I think this change will result in a lot of missed files

vector_search/tasks.py
```diff
@@ -237,10 +237,10 @@ def embed_learning_resources_by_id(self, ids, skip_content_files, overwrite):
 @app.task(bind=True)
 def embed_new_learning_resources(self):
     """
-    Embed new resources from the last day
+    Embed new resources from 40 minutes ago
```
abeglova, line 240: update the docstring

main/settings_celery.py
```diff
@@ -6,6 +6,16 @@
...
 
+"""
+the lookback window (in minutes) for the embeddings task
+the frequency of the embeddings tasks will be 1/4 of this for example
+lookback = 2 hours; task frequency = 30 minutes
+"""
+QDRANT_EMBEDDINGS_TASK_LOOKBACK_WINDOW = get_int(
+    name="QDRANT_EMBEDDINGS_TASK_LOOKBACK_WINDOW", default=60 * 2
+)
+EMBEDDING_SCHEDULE_MINUTES = int(QDRANT_EMBEDDINGS_TASK_LOOKBACK_WINDOW / 4)
```
abeglova, line 17: So this is a nitpick but probably a fixed difference between QDRANT_EMBEDDINGS_TASK_LOOKBACK_WINDOW 
 and EMBEDDING_SCHEDULE_MINUTES  makes more sense than a ratio since the concern is how long celery takes to pick up tasks and that is the same whether the task runs every 30 minutes or every 3 hours. Also when we later decide that we are comfortable with a smaller difference between QDRANT_EMBEDDINGS_TASK_LOOKBACK_WINDOW and EMBEDDING_SCHEDULE_MINUTES because we have a task that picks up any missed embeddings  it makes more sense to think of that as a number of minutes than a ratio as well

abeglova (changes requested):

main/settings_celery.py
shanbady: pushed a change to make the lookback window a fixed 60 minutes more than the schedule frequency.

abeglova (approved):
the two docstrings in tasks.py still need to be updated. Otherwise lgtm

PR #2171, opened 2025-04-01
reorganize dashboard tabs

ChristopherChudzicki:
Toward https://github.com/mitodl/hq/issues/7004

This PR breaks apart the large `DashboardPage` file (and its associated tests). Now:
- `DashboardLayout.tsx` controls the overall layout (sidepanel / content area). This is used in `app/dashboard/layout.tsx`, a NextJS "layout" file
- The individual tab contents are inserted into DashboardLayout via NextJS's file-based routing.

### How can this be tested?
There should be no functional changes. This was just a reorganization to make the dashboard files a bit more independent  / smaller / easier to deal with before I add more.

I suggest testing:
1. Open https://rc.learn.mit.edu/dashboard as first tab in your browser window
2. Open http://learn.odl.local:8062/dashboard as second tab in browser window
3. ensure both are at 100% zoom (cmd + 0)
4. Alternative between your two tabs and make sure there are no visual changes in any dashboard page 
     - Other than possibly the feature-flagged content on Home tab, which is different between `main` and RC` right now.


frontends/main/src/app-pages/DashboardPage/HomeContent.tsx
```diff
@@ -93,7 +94,6 @@ const HomeContent: React.FC = () => {
           title="Top picks for you"
           isLoading={isLoadingProfile}
```
ChristopherChudzicki, line 95: On `main`, `isLoading={true}` here shows a loading UI, but the underlying carousel still makes an API call with its given configuration. However, while `isLoading` is true, the config isn't really complete, so no API calls should be made.

Originally I believe `ResourceCarousel` did NOT make any API calls when `isLoading` was true. I think that behavior was introduced in https://github.com/mitodl/mit-learn/pull/1195/files#diff-fc04870f8cbdc1285ac40a4fe30850eff6668ee699c299d18c1de9c4f3430689L104 when ResoureCarousel stopped using DataPanel. (Which was a very good change, aside from this effect).

**Why am I mentioning this?**: In this PR I disabled API calls for `ResourceCarousel` when `isLoading` is true because they were causing erroneous API calls in test files (and rc/prod). The erroneous API calls during tests resulted in some excessive API mocking in the dashboard tests.

frontends/main/src/page-components/ResourceCarousel/ResourceCarousel.tsx
```diff
@@ -200,41 +224,15 @@ const ResourceCarousel: React.FC<ResourceCarouselProps> = ({
...
-        case "lr_similar":
-          return learningResourceQueries.similar(
-            tab.data.params.id,
-          ) as CarouselQuery
-        case "lr_vector_similar":
-          return learningResourceQueries.vectorSimilar(
-            tab.data.params.id,
-          ) as CarouselQuery
-      }
+      return { ...getTabQuery(tab), enabled: !isLoading }
```
ChristopherChudzicki, line 235: See https://github.com/mitodl/mit-learn/pull/2171/files#r2021973542

frontends/main/src/app-pages/ChannelPage/ChannelPage.test.tsx
```diff
@@ -422,14 +422,6 @@ describe("Channel Pages, Unit only", () => {
...
       )
     })
-
-    await waitFor(() => {
-      expect(makeRequest).toHaveBeenCalledWith(
-        "get",
-        urls.learningResources.featured({ limit: 12 }),
-        undefined,
-      )
-    })
```
ChristopherChudzicki, line 432: This API call was being made before, and is not being made now.

But... I'm not sure why we were testing for it. This was another case of the carousel making an API call while its configuration was loading. (Difference, btw, from the carousel loading its children AFTER the config is supplied.)

gumaerc (approved):
:+1:

I wasn't able to see any differences between my local and RC when comparing them with a fine toothed comb. All my tests also pass locally as well. I looked through all the code changes and there seems to be a significant amount of changes to the Dashboard tests besides just reorganization, but everything still seems to be covered and passing.

PR #2174, opened 2025-04-02
backpopulate files by resource id

abeglova:

- [ ] Desktop screenshots


shanbady (changes requested):
Not sure exactly whats happening buy when i try running  ``` python manage.py backpopulate_mitxonline_files  --overwrite --learning_resource_id 8077``` it does not appear to be updating the content after I changed it. i also tried deleting the contentfile object altogether to see if it would come back but it didnt. the celery task did seem to only process one item though just looking at the runtime

shanbady (approved):
👍

PR #2178, opened 2025-04-03
Fix type error on main

ChristopherChudzicki:
Non

There is currently a typescript error on `main` (see, e.g., https://github.com/mitodl/mit-learn/actions/runs/14247995768/job/39933709546?pr=2177)

- Starting in https://github.com/mitodl/mit-learn/pull/2156/files?show-viewed-files=true&file-filters%5B%5D=#diff-83282fef07409ca22ec9db563d88e9059520e6da5071c3c54411042687adbe29, `user.profile` is no longer a required property on `User` object.
- prior to merging ☝️ some code was added that assumed `user.profile` was non-null.
- I **think** we merged the above PR without rebasing, which is why the type error wasn't picked up until merges to main.

### How can this be tested?
Tests should pass. There are no changes here except to two test files. No runtime behavior change.


rhysyngsun (approved):
LGTM

PR #2180, opened 2025-04-03
Scrape marketing sites with dynamic content

shanbady:

Closes  https://github.com/mitodl/hq/issues/6982

This PR:
- enables the use of a chrome webdriver to scrape marketing sites that have js loaded/dynamic content.
- moves the scraper out of etl process and into its own periodic celery task

### How can this be tested?
1. checkout main
2. run the sloan etl pipeline ```python manage.py backpopulate_sloan_data```
3. Find a marketing page contentfile for one of the sloan courses and inspect the content:
```
from learning_resources import models
models.ContentFile.objects.get(file_type="marketing_page", learning_resource__id=<sloan resource id>).content
```
4. visit the marketing page for the sloan course (ex: https://executive.mit.edu/course/frontiers-of-generative-ai-in-business/a05U1000002xz4jIAA.html) and click on the "faculty" tab. note that info in the faculty section does not appear in the contentfile.
5. delete all existing marketing pages ```models.ContentFile.objects.filter(file_type="marketing_page").delete()```
6. Checkout this branch
7. rebuild your web and celery container so webdriver dependencies are installed 
8. in settings.py enable the webdriver by setting EMBEDDINGS_EXTERNAL_FETCH_USE_WEBDRIVER to True
9. manually trigger the celery task to fetch marketing site content:
```python
from vector_search.tasks import scrape_marketing_pages
scrape_marketing_pages.run()
```
10. Find a marketing page contentfile for one of the sloan courses and inspect the content:
```python
from learning_resources import models
models.ContentFile.objects.get(file_type="marketing_page", learning_resource__id=<sloan resource id>).content
```
11. Note that all contents from all tabs are there (and other potentially js loaded content)

### Additional Context
- There are Dockerfile changes that install the webdriver - heroku has a buildpack to install this but using it as-is  will put us over the slug size limit (500mb). I have [a fork ](https://github.com/shanbady/heroku-buildpack-chrome-for-testing/) that swaps out chrome for chrome-headless-shell which gets us under the hard limit for slug size.
- moving this into a celery task also sets us up to be able to pickup on any live updates on the marketing site without having to re-run the etl.
- This is the first step in being able to scrape dynamic data and there is currently special case logic to deal with sloan pages which has info on the tabs that are only loaded in once a user clicks them 
    - The idea is that this custom logic will go away once we need to customize this to handle another page type - I was thinking of more declarative ways to specify actions to take depending on the url match - possibly defining them as [scrapy spiders](https://docs.scrapy.org/en/latest/topics/spiders.html#scrapy.spiders.Spider) and putting the custom logic there


main/settings_celery.py
```diff
@@ -148,6 +148,12 @@
         "task": "learning_resources_search.tasks.update_featured_rank",
         "schedule": crontab(minute=30, hour=7),  # 3:30am EST
     },
+    "scrape-marketing-pages-every-1-hour": {
+        "task": "vector_search.tasks.scrape_marketing_pages",
```
mbertrand, line 152: Every hour seems like a lot, I doubt the marketing pages will change that often.  The main sloan task only runs once per day.

vector_search/tasks.py
```diff
@@ -296,3 +301,58 @@ def embed_new_content_files(self):
...
+            content_file, _ = ContentFile.objects.update_or_create(
+                learning_resource=learning_resource,
+                file_type=MARKETING_PAGE_FILE_TYPE,
+                defaults={
+                    "file_extension": ".md",
+                },
+            )
+            content_file.key = marketing_page_url
+            content_file.content = html_to_markdown(page_content)
+            content_file.save()
```
mbertrand, line 358: Seems like it would be cleaner to have ContentFile creation for marketing pages kept in the `learning_resources` app where all the other contentfile creation methods are, instead of `vector_search`, there doesn't seem to be anything specific to vector search code happening here.

mbertrand (commented):
Works well, a couple minor nitpicks

vector_search/tasks.py
shanbady: Went back and forth on this but I agree. learning_resources seems like the best place considering lack of vector based functionality.

learning_resources/tasks.py
```diff
@@ -430,3 +431,58 @@ def summarize_unprocessed_content(
...
+            published=True, resource_type__in=["course", "program"]
+        ).values_list("id", flat=True)
+    )
+
+    existing_page_resource_ids = set(
+        ContentFile.objects.filter(file_type="marketing_page").values_list(
+            "learning_resource_id", flat=True
+        )
+    )
+    missing_pages = list(resource_ids.difference(existing_page_resource_ids))
```
mbertrand, line 454: One issue with this is that existing pages are never updated, so they could get out of date if the instructors change from one run to another for example.  Generally the ETL pipelines update everything even if something has changed or not.  That could be improved on at some point by comparing previous attribute values vs freshly retrieved values, but that seems like work for another PR.

main/settings_celery.py
```diff
@@ -148,6 +148,12 @@
         "task": "learning_resources_search.tasks.update_featured_rank",
         "schedule": crontab(minute=30, hour=7),  # 3:30am EST
     },
+    "scrape-marketing-pages-every-1-hour": {
+        "task": "learning_resources.tasks.scrape_marketing_pages",
+        "schedule": get_int(
+            "SCRAPE_MARKETING_PAGES_SCHEDULE_SECONDS", 60 * 60 * 12
+        ),  # default is every hour
```
mbertrand, line 155: \# default is every 12 hours

mbertrand (commented):
Looks good, just one concern about existing pages never getting updated.  Might be worth discussing with Simone/Ferdi/Peter what they would expect for that, but updating existing page content on every task run is more in line with the ETL pipelines.

main/settings_celery.py
shanbady: fixed comment and also fixed name of scheduled task

mbertrand (approved):
👍

PR #2184, opened 2025-04-07
Workflow updates

Ardiea:
https://github.com/mitodl/ol-infrastructure/issues/3073

- Created new workflow for deploying the CI frontend to S3
- Updated names on existing workflows to be more accurate about what they are actually doing (only deploying the application frontend...) 

### How can this be tested?
???

### DIFF
--- a/.github/workflows/ci-frontend-deploy.yaml
+++ b/.github/workflows/ci-frontend-deploy.yaml
@@ -1,0 +1,106 @@
+name: CI Frontend Deploy
+
+on:
+  workflow_run:
+    workflows: [CI]
+    types: [completed]
+    branches: [main]
+  workflow_dispatch: # manual trigger
+
+jobs:
+  # runs if CI workflow was successful OR if this was manually triggered
+  on-success:
+    runs-on: ubuntu-22.04
+    if: >
+      github.event_name == 'workflow_dispatch' ||
+      github.event.workflow_run.conclusion == 'success'
+    steps:
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4
+        with:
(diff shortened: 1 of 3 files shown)


.github/workflows/ci-frontend-deploy.yaml
```diff
@@ -0,0 +1,106 @@
...
+          heroku_app_name: mitopen-rc
+          heroku_email: ${{ secrets.HEROKU_EMAIL }}
+          branch: main
+
+      - name: Write commit SHA to file
+        run: echo $GITHUB_SHA > frontends/main/public/hash.txt
+
+      - name: Build and push the NextJS Docker image
+        env:
+          HEROKU_APP_NAME: mitopen-rc-nextjs
```
blarghmatey, line 43: This seems like the wrong app name. Is it worth setting up that app in Heroku? Or should we do the work to just run that in K8s as well?

.github/workflows/ci-frontend-deploy.yaml
```diff
@@ -0,0 +1,106 @@
...
+          HEROKU_API_KEY: ${{ secrets.HEROKU_API_KEY }}
+          ORIGIN: https://rc.learn.mit.edu
```
blarghmatey, line 45: This should be `ci.learn.mit.edu`

.github/workflows/ci-frontend-deploy.yaml
Ardiea: oofff, I did not think of this.

Ardiea: I would rather not go through getting this to run in k8s at the moment so I will take care of heroku today.

PR #2187, opened 2025-04-08
enrollment dashboard mobile layout

gumaerc:
https://github.com/mitodl/hq/issues/7042

This PR implements mobile designs on the enrollments display in the dashboard home page.

image
image
image

### How can this be tested?
- Ensure you have Posthog connected to your local instance of `mit-learn`
- In  your connected Posthog project, make sure you have added and enabled the `enrollment-dashboard` feature flag
- Visit http://open.odl.local:8062/dashboard and log in (if you are not already logged in)
- Ensure nothing has changed with the desktop view of the enrollments display
- View the site using your browser's mobile simulator and ensure that it looks and behaves properly

### DIFF
--- a/frontends/main/src/app-pages/DashboardPage/CoursewareDisplay/DashboardCard.tsx
+++ b/frontends/main/src/app-pages/DashboardPage/CoursewareDisplay/DashboardCard.tsx
@@ -34,6 +34,7 @@ const MobileOnly = styled.div(({ theme }) => ({
 }))
 
 const CardRoot = styled.div(({ theme }) => ({
+  position: "relative",
   border: `1px solid ${theme.custom.colors.lightGray2}`,
   borderRadius: "8px",
   backgroundColor: theme.custom.colors.white,
@@ -53,7 +54,9 @@ const CardRoot = styled.div(({ theme }) => ({
 }))
 
 const MenuButton = styled(ActionButton)({
-  marginLeft: "-8px",
+  position: "absolute",
+  top: "0",
+  right: "0",
 })
 
 const getCoursewareText = (endDate?: string | null) => {
--- a/frontends/main/src/app-pages/DashboardPage/CoursewareDisplay/DashboardCard.tsx
+++ b/frontends/main/src/app-pages/DashboardPage/CoursewareDisplay/DashboardCard.tsx
@@ -24,6 +24,11 @@ const CardRoot = styled.div(({ theme }) => ({
   display: "flex",
   gap: "8px",
   alignItems: "center",
+  [theme.breakpoints.down("md")]: {
+    borderTop: "none",
+    borderRadius: "0px",
+    boxShadow: "none",
+  },
 }))
 
 const MenuButton = styled(ActionButton)({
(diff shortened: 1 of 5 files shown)


frontends/main/src/app-pages/DashboardPage/CoursewareDisplay/DashboardCard.tsx
```diff
@@ -246,21 +292,77 @@ const DashboardCard: React.FC<DashboardCardProps> = ({
...
         </Stack>
         {run.startDate ? (
           <CourseStartCountdown startDate={run.startDate} />
         ) : null}
       </Stack>
     </CardRoot>
   )
+
+  const mobileLayout = (
+    <CardRoot data-testid="enrollment-card">
```
ChristopherChudzicki, line 305: **Request:** I think we should use separate testids for the desktop/mobile versions, and one of them (doesn't matter which) should just be `enrolment-card`. (Or I guess maybe we should change it to `dashboard-card`).

The current structure has a few disadvantages, imo:
- makes the `DashboardCard.test.tsx` file awkward. You had change a bunch of things like `courseLink` to `courseLinks`, and it's not clear how many `courseLinks` there are.
- The tests for a component that renders a list of dashboard cards now feels weird, too. If the test renders 7 resources, you get 14 cards.

If we use separate testids, then both these problems go away:
- the `DashboardCard.test.tsx` can make all the assertions it wants about both variations,and it's clear there is only one `courseLink` (per variation).
- Other tests see 1 `enrollment-card` per resource. (And also 1 `enrollment-card-mobile` per resource, but other tests can ignore that...the equivalence of desktop/mobile versions in terms of logic / meaning is thoroughly tested in `Dashboard.test.tsx`

I took a quick stab at this in https://github.com/mitodl/mit-learn/compare/cg/enrollment-dashboard-mobile...cg/enrollment-dashboard-mobile-cc-edits?expand=1 (The DashboardCard.test.tsx file passes; some other test file fails because it is expecting 2 cards per resource. I didn't change that one.)

ChristopherChudzicki: If you look at https://github.com/mitodl/mit-learn/compare/cg/enrollment-dashboard-mobile...cg/enrollment-dashboard-mobile-cc-edits?expand=1 you can see my old man bias against mobile... Could have called them `enrollment-card` and `enrollment-card-desktop` 😛

frontends/main/src/app-pages/DashboardPage/CoursewareDisplay/DashboardCard.tsx
```diff
@@ -246,21 +292,77 @@ const DashboardCard: React.FC<DashboardCardProps> = ({
...
+      <Stack
+        direction="row"
+        justifyContent="space-between"
+        alignItems="stretch"
+        flex={1}
+        width="100%"
+      >
+        <Stack direction="column" gap="8px">
+          <Link size="medium" color="black" href={marketingUrl}>
```
ChristopherChudzicki, line 314: **Suggestion:** Might be worth styling the Link so that it has a max width slightly less than 100% (Maybe something like `calc(100% - 4px)`.

Else this happens: 
Screenshot 2025-04-09 at 9 02 48 AM

ChristopherChudzicki (changes requested):
Looks great. One minor suggestion, and one request about tests.

gumaerc:
@ChristopherChudzicki Thanks, this is ready for another look.

ChristopherChudzicki (approved):
👍

PR #2188, opened 2025-04-08
Changes to use posthog proxy

shaidar:

https://github.com/mitodl/hq/issues/6690

Changes required for app to use the correct url for the managed posthog proxy.


ChristopherChudzicki (approved):
👍 

I didn't realize our `ActionsSecret` thing supported variables in addition to secrets. None of the frontend stuff is secret, so nice.

PR #2204, opened 2025-04-15
redirect new users to onboarding

shanbady:

Closes https://github.com/mitodl/hq/issues/6857

This PR adds a "completed_onboarding" flag to the user profile that gets set to True once the user has interacted with the onboarding page after they first login.

### How can this be tested?
1. Checkout this branch
2. run migrations
3. Assuming you have an existing user locally, manually set user.profile.completed_onboarding to False (migrations automatically set to True for all existing users)
4. login and note that you are automatically redirected to the onboarding flow
5. upon completing the onboarding flow you will be redirected to the "next" parameter.
6. manually set your user's completed_onboarding flag to False
7. navigate to some other (non homepage) page and login from there. You should land back on the page you started on once done with onboarding.

test the "skip_onboarding" get parameter which skips the onboarding. 
1. perform step 3 from above
2. visit the login with ?skip_onboarding=1 (http://api.open.odl.local:8065/login?next=http%3A%2F%2Fopen.odl.local%3A8062%2F&skip_onboarding=1) You should not be forwarded to the onboarding page.


gumaerc:
I tested this with the OCW integration, setting a login URL of:

```
const loginUrl = new URL(
    `/login?next=${encodedLocation}&skip_onboarding=1`,
    apiBaseUrl
  ).toString()
```

Onboarding is indeed skipped when I log in from the button in OCW, but it is *also* skipped if I don't set `skip_onboarding` at all. With the auth setup we are using now with APISIX, `next` seems to be taking precedence either way. When we were using `django-social-auth`, the custom Logout view would override the redirect URL regardless, but that doesn't seem to be the case here for some reason. I sanity checked this by resetting the onboarding status and logging in directly to learn, where I was prompted to complete onboarding if I didn't set the flag.

shanbady:
adding note from slack discussion regarding ocw redirects "Since django-social-auth is no longer part of the picture, Keycloak never redirects back to Learn when next is specified."

authentication/views.py
```diff
@@ -73,4 +73,15 @@ def get(
         """
         GET endpoint for logging a user in.
         """
-        return redirect(get_redirect_url(request))
+        redirect_url = get_redirect_url(request)
+        if not request.user.is_anonymous:
+            profile = request.user.profile
+            if (
+                not profile.completed_onboarding
+                and request.GET.get("skip_onboarding", "0") != "1"
```
rhysyngsun, line 81: Nit because non-negated comparisons are easier to read, plus it works then for any value that's not a 0, just in case:
```suggestion
                and request.GET.get("skip_onboarding", "0") == "0"
```

authentication/views.py
```diff
@@ -73,4 +73,15 @@ def get(
...
+            ):
+                params = urlencode({"next": redirect_url})
+                redirect_url = f"{settings.MITOL_NEW_USER_LOGIN_URL}?{params}"
+            profile.completed_onboarding = True
+            profile.save()
```
rhysyngsun, line 86: I think we only want to mark them as having completed onboarding if the user was redirected there. Otherwise if a request specifies `skip_onboarding` (presumably because we have a redirect url we want them to go straight to), they'll at least be prompted for the onboarding once when they return to Learn directly.

rhysyngsun (changes requested):
Things are generally looking good, a few small things.

authentication/views.py
shanbady: Updated the logic such that the flag will only get set at point of redirect to the onboarding page

rhysyngsun (approved):
LGTM

PR #2211, opened 2025-04-22
Attempt to reduce nextjs memory usage

rhysyngsun:

N/A

This attempts to reduce nextjs memory usage so it doesn't get oom-killed.

### How can this be tested?

Run `docker compose up`, it should run without being oom-killed. The build may be slower but it should use less memory.

### DIFF
--- a/frontends/main/next.config.js
+++ b/frontends/main/next.config.js
@@ -100,6 +100,19 @@ const nextConfig = {
   env: {
     FEATURE_FLAGS: JSON.stringify(processFeatureFlags()),
   },
+
+  experimental: { webpackMemoryOptimizations: true },
+
+  webpack: (config, { dev }) => {
+    if (dev) {
+      config.cache = Object.freeze({
+        type: "filesystem",
+        allowCollectingMemory: true,
+      })
+    }
+    // Important: return the modified config
+    return config
+  },
 }
 
 // Injected content via Sentry wizard below
--- a/docker-compose.apps.yml
+++ b/docker-compose.apps.yml
@@ -36,7 +36,6 @@ services:
     command:
       - |
         yarn install --immutable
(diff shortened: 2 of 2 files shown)


frontends/main/next.config.js
```diff
@@ -100,6 +100,17 @@ const nextConfig = {
...
+  experimental: { webpackMemoryOptimizations: true },
+
+  webpack: (config) => {
+    config.cache = Object.freeze({
+      type: "filesystem",
+      allowCollectingMemory: true,
+    })
+    // Important: return the modified config
+    return config
+  },
```
ChristopherChudzicki, line 113: **Suggestion:** Only make this change in development mode.

The webpack docs say 
> Cache the generated webpack modules and chunks to improve build speed. cache is set to type: 'memory' in [development mode](https://webpack.js.org/configuration/mode/#mode-development) **and disabled in [production mode](https://webpack.js.org/configuration/mode/#mode-production)**.

so i'm not sure if the dev check here is necessary.

```suggestion
  webpack: (config, { dev } ) => {
    if (dev) {
      config.cache = Object.freeze({
        type: "filesystem",
        allowCollectingMemory: true,
      })
    }
    // Important: return the modified config
    return config
  },
```

ChristopherChudzicki: Also, any particular reason for the freeze?

ChristopherChudzicki (changes requested):
Could we limit this to dev mode?

**Also:** In `docker-compose.apps.yml`, it might be worth removing the storybook command.

frontends/main/next.config.js
rhysyngsun: It's what I saw in code samples.

ChristopherChudzicki (approved):

PR #2217, opened 2025-04-29
embedding content updates and removals

shanbady:

Closes https://github.com/mitodl/hq/issues/6754

This PR introduces updates/upserts and removals for embedding vectors. 

The main changes:

**For contentfiles**
1. the original "checksum" attribute on contentfiles has been renamed to "archive_checksum" to maintain functionality in how the ETL uses it to check tar file checksums.
2. the _new_ "checksum" attribute contentfiles is a checksum of the content attribute which gets generated on each save
3. When generating embeddings for contentfiles, the checksum is used to determine if the contentfile has changed. If so all existing chunks for that contentfile are removed. The file is then re-chunked and re-embedded. 
4. If the content checksum is the same - regeneration of the embedding is skipped but the "metadata"/attributes are upserted - this accounts for cases where the content has not changed but there may have been updates to some other attributes.

**For resources**
1. The embedding context is just the "{title} {description} {full description}" which we check against to see if it needs updating. 

#### testing the etl hooks
1. checkout this branch
2. in settings.py set QDRANT_ENABLE_INDEXING_PLUGIN_HOOKS to True
3. restart celery
4. (optional step - makes it easier to see something happened) visit your local [qdrant dashboard](http://localhost:6333/dashboard) and delete both collections
5. run any backpopulate management command like `python manage.py backpopulate_micromasters_data`
6. you should see the qdrant records getting populated with learning resources that have been pulled down, updated or removed.

For reference, here is a list of things that this PR (and functionality before it) should ensure:
- New Learning resources are embedded. New contentfiles are chunked and embedded as new records. 
- When contentfile contents are changed, old points are removed and the contentfile is re-chunked and embedded
- regardless of changes to embedding content (for both contentfiles and learning resources) updates to indexed metadata fields are still applied.
- when contentfiles or learning resources are unpublished/deleted, the associated records are removed from qdrant


learning_resources_search/plugins.py
```diff
@@ -133,11 +146,14 @@ def bulk_resources_unpublished(self, resource_ids, resource_type):
...
         ):
-            try_with_retry_as_task(
-                tasks.bulk_deindex_learning_resources,
-                ids,
-                resource_type,
-            )
+            unpublished_tasks = [
+                tasks.bulk_deindex_learning_resources.si(ids, resource_type),
+            ]
+            if django_settings.QDRANT_ENABLE_INDEXING_PLUGIN_HOOKS:
```
abeglova, line 152: I think the elasticsearch tasks should be chained before the vector embedding tasks since the vector embedding tasks can potentially take a long time. It's better to make the elasticsearch index up to date sooner

vector_search/utils.py
```diff
@@ -199,25 +206,121 @@ def _chunk_documents(encoder, texts, metadatas):
...
+def update_payload(
+    serialized_document,
+    resource_type,
+):
+    """
+    Update the payload of a document in Qdrant
+    based on the resource_type (learning resource vs content file)
+    """
+    client = qdrant_client()
+    if resource_type != CONTENT_FILE_TYPE:
```
abeglova, line 252: Does it make sense to have a seperate version of update_payload(), should_generate_embeddings(),  _embedding_context() etc for learning resources and content files instead of having all the if statements?you should be able to know which version of the function to call in the task

vector_search/utils.py
```diff
@@ -199,25 +206,121 @@ def _chunk_documents(encoder, texts, metadatas):
...
+    if not all([points, payload]):
+        return
+    client.set_payload(
+        collection_name=collection_name,
+        payload=payload,
+        points=points,
+    )
+
+
+def should_generate_embeddings(serialized_document, resource_type):
```
abeglova, line 284: I think it makes sense to have seperate content file and learning resource verisons of this function instead of all the if statements. 
You can consider separate content file or resource specific versions of  update_payload(), and  _embedding_context() too

abeglova (changes requested):
This seems to work well although i haven't tested it will a delete or update for existing vectors. I will figure out how to test that out. In the meantime, here are some comments

learning_resources/migrations/0089_contentfile_archive_checksum.py
```diff
@@ -0,0 +1,17 @@
...
+
+class Migration(migrations.Migration):
+    dependencies = [
+        ("learning_resources", "0088_add_content_summarizer_config"),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name="contentfile",
+            name="archive_checksum",
```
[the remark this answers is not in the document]
shanbady: good call.

vector_search/utils.py
shanbady: good idea. I updated the code so they each have their own _embedding_context, update_payload and should_generate_embeddings methods

abeglova (approved):
lgtm

PR #2219, opened 2025-04-29
update coursenumber weight in search

abeglova:
https://github.com/mitodl/hq/issues/7211

This updates the weight of coursenumber in search results similar to instructor names

### How can this be tested?
Run `docker-compose run -u root --rm web ./manage.py backpopulate_ocw_data --course-name 18-01-single-variable-calculus-fall-2005`
Go to `http://open.odl.local:8062/search?q=18.01`
OCW single variable calculus should be the first result. This is a bit hard to test locally since that course would probably come up first anyway if you don't have many learning resources . We can test more on rc


mbertrand (approved):
👍

PR #2222, opened 2025-04-30
Force https for urls in scraper

shanbady:

Closes https://github.com/mitodl/hq/issues/7216

This PR makes it so that http start urls are forced to https before beginning the scraping process. Some of the apis our etls hit return http urls and the redirect to https is causing unexpected behavior.

### How can this be tested?
1. Make sure EMBEDDINGS_EXTERNAL_FETCH_USE_WEBDRIVER is True (even as you switch branches to test)
2. On main (or any other branch but this one) - scrape an "http" url using the scraper and observe the content size:
```python
from learning_resources.site_scrapers.utils import scraper_for_site
scraper = scraper_for_site("http://micromasters.mit.edu/scm/")
content = scraper.scrape()
print(len(content))
```
3. checkout this branch.
4. re-run the script and observe the content length - it should be significantly more since it scraped multiple pages


learning_resources/site_scrapers/utils.py
```diff
@@ -5,6 +5,7 @@
 
 
 def scraper_for_site(url):
+    url = url.replace("http://", "https://")
```
ChristopherChudzicki, line 8: BTW `http://` is legal in query parameters, so theoretically this could change more than the protocol. (E.g., our login URL is `https://api.rc.learn.mit.edu/learn/login?next=https://rc.learn.mit.edu/`)

Very likely those sorts of URLs don't matter for this scraping (and maybe changing http to https in query params would be desirable, too).

ChristopherChudzicki (commented):
I'm getting 70853 on both main and your branch after restarting all containers when running
```python
from learning_resources.site_scrapers.utils import scraper_for_site
scraper = scraper_for_site("http://micromasters.mit.edu/scm/")
content = scraper.scrape()
print(len(content))
```

But this change makes sense to me, and seems plausible that it's not always reproducible, so 👍 

Edit: Making sure `EMBEDDINGS_EXTERNAL_FETCH_USE_WEBDRIVER=True`, now I get expected difference. 👍

ChristopherChudzicki (approved):
oops, meant to approve!

PR #2230, opened 2025-05-06
Retry requests on connection errors and additional behavior

jonkafton:

Closes https://github.com/mitodl/hq/issues/7205

- Fixes the core issue that errors thrown while fetching learning resource content to build page metadata (for pages loaded with the drawer open) were thrown up to the wide global error boundary causing the fallback error page to be shown in place of content. These are now caught, reported to Sentry and fall back to the default page metadata.

- Retries requests that have failed due to network issues (status 0 or undefined).

- Replaces our whitelist of status codes to retry and extends to all statuses with the exception of some that we wouldn't expect to succeed on retry, `[400, 401, 403, 404, 405, 409, 422]`.

- Also applies the above logic so that requests a retried during SSR prefetching. To note - we were retrying any requests that might have slipped through our prefetching on the server according to the previous rules (the `QueryClientProvider` provided query client), whereas the QueryClient created during the prefetch phase had no rules specified.

- During SSR prefetching and for any requests made from on the server not prefetched (none expected), fixes the retry delay to 1s.

### How can this be tested?

It's tricky to test error responses / network errors - I've committed a small proxy to an otherwise identical branch for testing: https://github.com/mitodl/mit-learn/tree/jk/7205-query-retry-proxy/sandbox. This closes requests sockets with `ECONNRESET` and includes counts to test for retries. It will need be restarted after any test dependent on the counts:

```bash
cd ./sandbox
yarn install
yarn proxy
```

The Next.js server should be run in production mode as it will behave differently in dev mode:

```bash
NEXT_PUBLIC_MITOL_API_BASE_URL=http://api.open.odl.local:8063 \
NEXT_PUBLIC_CSRF_COOKIE_NAME=any \
NEXT_PUBLIC_MITOL_SUPPORT_EMAIL=any \
NEXT_PUBLIC_SITE_NAME=any \
NEXT_PUBLIC_ORIGIN=http://learn.odl.local:8062 \
yarn build

PORT=8062 yarn start
```

Navigate to the homepage with a drawer open, http://learn.odl.local:8062/?resource=17261 (or update proxy.ts for a different resource)
- Confirm that the error fetching page metadata for the open drawer does not produce the "Something went wrong." error page.
- The error should be logged in the Next.js console.
- The page should have default title "MIT Learn" and description "Learn with MIT".

The proxy is set to error on all requests to /api/v1/featured (homepage carousels) from the server and the first 3 from the client (per querystring / carousel).

Navigate to the homepage, http://learn.odl.local:8062/
- Observe that failed prefetch requests are made on the the server with 3 retries and a 1s delay.
- The homepage should load with skeleton carousels.
- The browser should retry requests unless success on the third.

Apply the proxy request counts from the server to confirm that it retries and prefetches successfully if < 3.

Note that the Next.js server should be restarted between tests as responses are cached, as should the proxy to reset counts.

### DIFF
--- a/frontends/api/src/ssr/prefetch.ts
+++ b/frontends/api/src/ssr/prefetch.ts
@@ -2,14 +2,11 @@ import { QueryClient, dehydrate } from "@tanstack/react-query"
 import type { Query } from "@tanstack/react-query"
 
 const MAX_RETRIES = 3
-const NO_RETRY_CODES: (number | undefined)[] = [404, 403, 401]
+const NO_RETRY_CODES = [400, 401, 403, 404, 405, 409, 422]
 
 type MaybeHasStatus = {
   response?: {
     status?: number
-  }
-  config?: {
-    url?: string
   }
 }
 
@@ -21,39 +18,31 @@ export const prefetch = async (
   queries: (Query | unknown)[],
   queryClient?: QueryClient,
 ) => {
+  /**
+   * The SSR QueryClient is transient - it is created only for prefetch
+   * while API requests are made to server render the page and discarded
+   * as the dehydrated state is produced and sent to the client.
+   *
+   * Create a new query client if one is not provided.
+   */
   queryClient =
     queryClient ||
     new QueryClient({
       defaultOptions: {
         queries: {
+          /**
+           * React Query's default retry logic is only active in the browser.
+           * Here we explicitly configure it to retry MAX_RETRIES times on
+           * the server, with an exclusion list of statuses that we expect not
+           * to succeed on retry.
+           *
+           * Includes status undefined as we want to retry on network errors
+           */
           retry: (failureCount, error) => {
             const status = (error as MaybeHasStatus)?.response?.status
             const isNetworkError = status === undefined || status === 0
-            const id = (error as MaybeHasStatus)?.config?.url
-            const isBrowser = typeof window !== "undefined"
-            const isOffline =
-              isBrowser && typeof navigator !== "undefined" && !navigator.onLine
-            console.log(
-              "retry",
-              failureCount,
-              isBrowser ? "client" : "server",
-              status,
-              id,
-              "network error:",
-              isNetworkError,
-              "id in browser:",
-              isBrowser ? id : undefined,
-              "offline:",
-              isOffline,
-            )
-            /**
-             * React Query's default behavior is to retry all failed queries 3
-             * times. Many things (e.g., 403, 404) are not worth retrying. Let's
-             * just retry some explicit whitelist of status codes.
-             *
-             * Includes status undefined as we want to retry on network errors
-             */
-            if (status === undefined || !NO_RETRY_CODES.includes(status)) {
+
+            if (isNetworkError || !NO_RETRY_CODES.includes(status)) {
               return failureCount < MAX_RETRIES
             }
             return false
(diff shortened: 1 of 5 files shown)


frontends/api/src/ssr/prefetch.ts
```diff
@@ -9,7 +18,47 @@ export const prefetch = async (
...
 ) => {
-  queryClient = queryClient || new QueryClient()
+  /**
+   * The SSR QueryClient is transient - it is created only for prefetch
+   * while API requests are made to server render the page and discarded
+   * as the dehydrated state is produced and sent to the client.
+   *
+   * Create a new query client if one is not provided.
+   */
+  queryClient =
```
ChristopherChudzicki, line 28: **Minor Suggestion:** For readability, maybe move this to a function and default parameter?

```ts
export const prefetch = async (
  queries: (Query | unknown)[],
  queryClient = getServerQueryClient(),
) => {....}
```

ChristopherChudzicki (approved):
This looks good and makes sense. I think this will probably solve the problem.

I do want to suggest one small additional change.

One thing that I found odd in the sentry logs around the time of these errors was:
```
TypeError
Cannot read properties of undefined (reading 'results')
```
([Example](https://mit-office-of-digital-learning.sentry.io/issues/6340553017/?environment=production&project=4506260054540288&query=is%3Aunresolved%20issue.priority%3A%5Bhigh%2C%20medium%5D&referrer=issue-stream&stream_index=1))

I think this error was arising because of a somewhat erroneous non-null assertion in `frontends/main/src/app/c/[channelType]/[name]/page.tsx`:

```diff
--- a/frontends/main/src/app/c/[channelType]/[name]/page.tsx
+++ b/frontends/main/src/app/c/[channelType]/[name]/page.tsx
@@ -74,29 +74,28 @@ const Page: React.FC = async ({
   const offerors = queryClient
     .getQueryData<PaginatedLearningResourceOfferorDetailList>(
       offerorQueries.list({}).queryKey,
-    )!
-    .results.reduce(
+    )?.results.reduce(
       (memo, offeror) => ({
         ...memo,
         [offeror.code]: offeror,
       }),
-      [],
-    )
+      {} as Record<string, LearningResourceOfferorDetail>,
+    ) ?? {}
```

I say "somewhat erroneous" because the only way for the results to be null is if the prefetch failed, in which case `getQueryData` will return undefined. Accounting for the possibly-undefined data results in it trying to be refetched on the client.

This likely is much less or not an issue anymore. Previously the typeerror would occur with a single failure on `offerorQueries.list({})`, now it must fail three consecutive times.

(An alternative to `?.results.reduce` would be to call `invariant(...)` on the offerors data. Then at least we'd get a more intelligible error on the server.)

jonkafton:
> I say "somewhat erroneous" because the only way for the results to be null is if the prefetch failed, in which case getQueryData will return undefined. Accounting for the possibly-undefined data results in it trying to be refetched on the client.

> This likely is much less or not an issue anymore. Previously the typeerror would occur with a single failure on offerorQueries.list({}), now it must fail three consecutive times.

The `throwOnError` rules in our query client did get me thinking about error handling in components generally. We don't normally check for errors from our useQuery() hooks, if ever, but instead just fail gracefully if there is no data or often assume that we are still waiting for the response and return a null for the component. If we know a request has errored and retries are exhausted we'd ideally offer some feedback instead of displaying nothing or load skeletons indefinitely.

PR #2232, opened 2025-05-07
Fix ordering of modules in UAI dashboard

gumaerc:
https://github.com/mitodl/hq/issues/7255

This PR adds a transform function for properly sorting the modules listed in the UAI dashboard, sorting them by Enrolled, Completed, and then the rest sorted by the order defined within the Program.

image
image

### How can this be tested?
- Ensure that you have a Posthog project configured and have run the basic setup for `mit-learn`
- Enable both the `enrollment-dashboard` and `mitlearn-organization-dashboard` feature flags
- Visit the dashboard at http://open.odl.local:8062/dashboard (or wherever your site is hosted), logging in if necessary
- The "My Learning" dashboard should be unchanged by this PR
- Click either "Organization X" or Organization Y" to open the UAI dashboard
- Ensure that the order matches what's described in the issue: first Enrolled courses, then Completed courses, then the order matching the one defined in the program

### DIFF
--- a/frontends/api/src/mitxonline/test-utils/factories/index.ts
+++ b/frontends/api/src/mitxonline/test-utils/factories/index.ts
@@ -1,44 +1,5 @@
 import * as mitx from "./enrollment"
 import * as programs from "./programs"
 import * as courses from "./courses"
-import * as u from "api/test-utils"
-import { urls } from "api/mitxonline-test-utils"
-import { mockOrgData } from "../../hooks/enrollment"
-import { setMockResponse } from "../../../test-utils"
 
-const setupProgramsAndCourses = () => {
-  const user = u.factories.user.user()
-  setMockResponse.get(u.urls.userMe.get(), user)
-
-  const orgId = mockOrgData.orgX.id
-  const coursesA = courses.courses({ count: 4 })
-  const coursesB = courses.courses({ count: 3 })
-  const programA = programs.program({
-    courses: coursesA.results.map((c) => c.id),
-  })
-  const programB = programs.program({
-    courses: coursesB.results.map((c) => c.id),
-  })
-
-  setMockResponse.get(
-    urls.programs.programsList({ orgId: mockOrgData.orgX.id }),
-    { results: [programA, programB] },
-  )
-  setMockResponse.get(urls.courses.coursesList({ id: programA.courses }), {
-    results: coursesA.results,
-  })
-  setMockResponse.get(urls.courses.coursesList({ id: programB.courses }), {
-    results: coursesB.results,
-  })
-
-  return {
-    orgId,
-    user,
-    programA,
-    programB,
-    coursesA: coursesA.results,
-    coursesB: coursesB.results,
-  }
-}
-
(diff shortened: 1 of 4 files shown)


frontends/api/src/mitxonline/test-utils/factories/index.ts
```diff
@@ -1,5 +1,44 @@
 import * as mitx from "./enrollment"
 import * as programs from "./programs"
 import * as courses from "./courses"
+import * as u from "api/test-utils"
+import { urls } from "api/mitxonline-test-utils"
+import { mockOrgData } from "../../hooks/enrollment"
+import { setMockResponse } from "../../../test-utils"
 
-export { mitx as enrollment, programs, courses }
+const setupProgramsAndCourses = () => {
```
ChristopherChudzicki, line 9: **Request:** IMO a better spot for this would be `frontends/main/src/app-pages/DashboardPage/CoursewareDisplay/test-utils.ts`. (Or a sinilka

Factory files should be data-creators, not API mockers.

Also, I think it would be good to add a docstring describing what test scenario it sets up:
```ts
/**
 *  blah blah
 */
const setupProgramsAndCourses = () => {}
```

ChristopherChudzicki (changes requested):
Makes sense 👍 , but two small requests.

frontends/main/src/app-pages/DashboardPage/CoursewareDisplay/transform.ts
```diff
@@ -105,4 +105,43 @@ const mitxonlineProgram = (raw: V2Program): DashboardProgram => {
...
+    if (!aEnrolled && bEnrolled) {
+      return 1
+    }
+    if (aCompleted && !bCompleted) {
+      return -1
+    }
+    if (!aCompleted && bCompleted) {
+      return 1
+    }
+    return (
```
ChristopherChudzicki, line 129: **Request:** The goal here is to keep original order, right? If yes, just return 0.

From MDN:
>compareFn(a, b) return value | sort order
> -- | --
> \> 0 | sort a after b, e.g., [b, a]
> \< 0 | sort a before b, e.g., [a, b]
> === 0 | keep original order of a and b
>
> – https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort#description

Also... copy pasting from MDN gave me legit markdown. That's cool.

gumaerc: For some reason I thought in our conversation about this that we had thought the proper order wasn't being returned from the API, but this wouldn't be the way to deal with that anyway. Good suggestion.

ChristopherChudzicki: I believe that the API does return a meaningful order, it's just that we want to use that order as a fallback only.

gumaerc:
@ChristopherChudzicki Thanks, it's ready for another look.

ChristopherChudzicki (approved):

PR #2233, opened 2025-05-08
Remove param based onboarding navigation  

jonkafton:

Closes https://github.com/mitodl/hq/issues/7223

Removes the onboarding finish navigation configurable by a `next` param on the querystring as we always want to go to the dashboard screen.

Previously the backend /login and onboarding handling looked like:

- A user on e.g. the topic page hits "Log in" and requests /learn/login?next=https://learn.mit.edu/topics
- The login handler determines whether they have completed onboarding
- If yes:
   - User redirected to https://learn.mit.edu/topics
- If no:
   - User redirected to https://learn.mit.edu/onboarding?next=https://learn.mit.edu/topics
   - User completes onboarding
   - User is redirected in the frontend to https://learn.mit.edu/topics when they hit "Finish" at the end of onboarding.
 
We instead want the user to always go to the dashboard when they complete the onboarding, irrespective of their starting point prior to login.

### How can this be tested?

The "Log In" button always links to the login endpoint with the current address as set on a `next` param. Starting from any page as a user that has not completed onboarding (manually set `user.profile.completed_onboarding` to False):
- Log in.
- Step through onboarding and hit "Finish".
- Confirm that you are on the dashboard page at /dashboard.

### DIFF
--- a/authentication/views.py
+++ b/authentication/views.py
@@ -4,7 +4,7 @@ import logging
 
 from django.contrib.auth import logout
 from django.shortcuts import redirect
-from django.utils.http import url_has_allowed_host_and_scheme, urlencode
+from django.utils.http import url_has_allowed_host_and_scheme
 from django.views import View
 
 from main import settings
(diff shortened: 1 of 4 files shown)


shanbady (approved):
works as intended 👍 . only comment is that it feels like we are removing functionality that may need to come back at some point but we can address that when it does

jonkafton:
> works as intended 👍 . only comment is that it feels like we are removing functionality that may need to come back at some point but we can address that when it does

I agree, though perhaps while we have the special case that a user is not onboarded and redirected to the onboarding, we end up with a login redirect to a redirect to the onboarding redirect, to a reroute back to their starting point. As I understand it, the starting point is invalidated by the initial redirect to onboarding.

PR #2235, opened 2025-05-08
Adjust default env values for mit-learn

mbertrand:
Related to https://github.com/mitodl/hq/issues/6756

Shuffles default env values around so that keycloak/apisix integration should work without adjusting any local env values.

### How can this be tested?
- Run `docker compose up` and go to http://open.odl.local:8062
- Log in, you should be redirected to keycloak, then back to the home page (or user personalization page if signing in for the 1st time)
- Log out, the home page should eventually reload and you should be logged out.


env/shared.env
```diff
@@ -13,3 +12,6 @@ NGINX_PORT=8063
 APISIX_PORT=8065
 KEYCLOAK_PORT=8066
 KEYCLOAK_SSL_PORT=8067
+
+# Run all containers by default
+COMPOSE_PROFILES=backend,frontend,keycloak,apisix
```
ChristopherChudzicki, line 17: I don't actually think this will work. For `docker compose` to recognize `COMPOSE_PROFILES`, they need to either be in your shell env or in a `.env` file adjacent to the compose file.

So that's a little awkward. Otherwise everything looks good.

ChristopherChudzicki (approved):
We should probably change the defaults to `learn.ol.mit.edu` at some point, but that's a separate issue. 

Worked for me, and cleaned up my backend.local.env! Thanks!

env/shared.env
mbertrand: removed, and updated the README's

PR #2236, opened 2025-05-08
Less noisy openapi-diff action.

ChristopherChudzicki:
None

Makes the `openapi-diff` workflow a bit less noisy on PRs:
- It will always run.
- It only post an initial comment if it detects openapi changes.
    - After which point, it will *always* update this comment (even if no changes detected) and will not post a new one.

### How can this be tested?
I have tested this thoroughly on this branch which is why there are so many commits.

My recommendation to test it is:
1. Make a copy of this branch and open your own PR.
    ```sh
    git fetch origin
    git checkout cc/less-noisy-openapi-diff
    git checkout -b cc/less-noisy-openapi-diff-copy
    git commit -m 'empty commit' --allow-empty
    git push origin cc/less-noisy-openapi-diff-copy
    ```
2.  Open the PR with `cc/less-noisy-openapi-diff-copy`. 
    - **Expectation:** No comment is posted (no breaking change, no previous comment).
3. Now make a breaking change. (E.g., remove `title` from `LearningResourceMetadataDisplaySerializer`, re-run `./scripts generate_openapi`, commit your changes, and push.
    - **Expectation:** A comment is posted with changes AND the OpenAPI Diff action fails. (breaking changes)
4. Undo your change from (3) and push. (Or revert the commit, I guess).
    - **Expectation:** The comment is updated to indicate no breaking changes.
5. Close your PR.

### DIFF
--- a/.github/workflows/openapi-diff.yml
+++ b/.github/workflows/openapi-diff.yml
@@ -32,12 +32,21 @@ jobs:
         run: |
           echo summary=$(echo "${{ steps.oasdif_changelog.outputs.changelog }}" | head -1 ) \
           >> $GITHUB_OUTPUT
+      - name: Find existing comment
+        id: find_comment
+        uses: peter-evans/find-comment@v3
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+          repository: ${{ github.repository }}
+          issue-number: ${{ github.event.pull_request.number }}
+          body-includes: "## OpenAPI Changes"
       - name: Post changes as comment
         uses: peter-evans/create-or-update-comment@v3
         with:
           token: ${{ secrets.GITHUB_TOKEN }}
           repository: ${{ github.repository }}
           issue-number: ${{ github.event.pull_request.number }}
+          comment-id: ${{ steps.find_comment.outputs.comment-id }}
           body: |
             ## OpenAPI Changes
 
@@ -48,7 +57,6 @@ jobs:
             ${{ steps.oasdif_changelog.outputs.changelog }}
             ```
             </details>
-          body-includes: "## OpenAPI Changes"
       - name: Check for breaking changes
         id: oasdif_breaking
         run: |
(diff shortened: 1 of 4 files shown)


gumaerc (approved):
👍 Works great

2 things I noticed that aren't a problem but maybe worth noting:

- When breaking changes are detected, the Python tests also fail. This is expected of course, but I thought it was worth noting
- When the comment is updated in the last step after removing the breaking change, it says;

```
2 changes: 0 error, 0 warning, 2 info
info	[endpoint-added] at head/openapi/specs/v0.yaml	
	in API POST /api/v0/chat_agent/
		endpoint added

info	[endpoint-added] at head/openapi/specs/v0.yaml	
	in API POST /api/v0/syllabus_agent/
		endpoint added
```

ChristopherChudzicki:
> When breaking changes are detected, the Python tests also fail. This is expected of course, but I thought it was worth noting

I think this was because you deleted a bit of the serializer. This (A) is a breaking change, but (B) is also logic that was tested in the python tests. If you were really going to do this on purpose, you would have updated the python tests to not use title. [The error was](https://github.com/mitodl/mit-learn/actions/runs/14932879800/job/41953333671#step:11:528):
```
FAILED learning_resources/views_test.py::test_learning_resources_display_info_list_view - KeyError: 'title'
```

> When the comment is updated in the last step after removing the breaking change, it says;
>
>```
>2 changes: 0 error, 0 warning, 2 info
>info	[endpoint-added] at head/openapi/specs/v0.yaml	
>	in API POST /api/v0/chat_agent/
>		endpoint added
>
>info	[endpoint-added] at head/openapi/specs/v0.yaml	
>	in API POST /api/v0/syllabus_agent/
>		endpoint added
>```

That's interesting. I believe this happened because:
- Matt recently removed those endpoints, https://github.com/mitodl/mit-learn/pull/2238
- The branch in https://github.com/mitodl/mit-learn/pull/2241 still had those APIs in it, so when oas-diff ran between the branches, I think it detected it as an "add". (I think it just looks at diff, not how git would resolve a merge). 

I'm going to add a note to the comment that if you notice unexpected changes, make sure your branch is up-to-date with main.

PR #2237, opened 2025-05-08
Tracking user account creation event in posthog 

rachellougee:

https://github.com/mitodl/hq/issues/7269

Adds `user_account_created` event to PostHog, when accounts are created via apisix middleware. 
Q: Is there any other way to register?

### How can this be tested?

- Make sure you have `POSTHOG_ENABLED`, `POSTHOG_PERSONAL_API_KEY`, `POSTHOG_PROJECT_API_KEY`, `POSTHOG_PROJECT_ID` and `POSTHOG_API_HOST` in shared.local.env
- Visit http://open.odl.local:8062/, click on "log In", then click on "Register"
-  Complete the Register on keycloak
-  Check your Posthog activity and verify that you see `user_account_created` event
image

### DIFF
--- a/main/middleware/apisix_user.py
+++ b/main/middleware/apisix_user.py
@@ -107,7 +107,7 @@ def get_user_from_apisix_headers(request, decoded_headers, original_header):
         )
         posthog.capture(
             user.id,
-            event=PostHogEvents.account_created.value,
+            event=PostHogEvents.ACCOUNT_CREATED,
             properties={
                 "$current_url": request.build_absolute_uri(),
                 "global_id": global_id,
--- a/main/middleware/apisix_user.py
+++ b/main/middleware/apisix_user.py
@@ -107,7 +107,7 @@ def get_user_from_apisix_headers(request, decoded_headers, original_header):
         )
         posthog.capture(
             user.id,
-            event=PostHogEvents.ACCOUNT_CREATED,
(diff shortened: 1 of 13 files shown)


main/middleware/apisix_user.py
```diff
@@ -99,6 +100,22 @@ def get_user_from_apisix_headers(request, decoded_headers, original_header):
             "get_user_from_apisix_headers: User %s not found, created new",
             global_id,
         )
+        # Send user creation event to PostHog
+        posthog = Posthog(
+            settings.POSTHOG_PROJECT_API_KEY, host=settings.POSTHOG_API_HOST
+        )
+        posthog.capture(
+            user.id,
+            event="user_account_created",
```
gumaerc, line 109: Maybe we want to move this into a constants file? I don't think we do any other custom event capturing in Python yet, but if we're getting started on that it might be a good idea to get off on the right foot with organizing the keys for the different events in one place.

gumaerc (approved):
👍 Looks good to me! I was able to see the event come through in Posthog:

image

I just left one small suggestion, take it or leave it.

main/middleware/apisix_user.py
rachellougee: Yes, I think its a good idea. I added PostHogEvents to define event constants. Can you take another look? Thanks

PR #2238, opened 2025-05-08
Remove ai_chat app

mbertrand:
N/A

Removes the ai_chat app since it is no longer used, all that functionality is in the learn-ai repo now.

### How can this be tested?
The app should start up locally and work as before.


abeglova (approved):
THis works as long as the OpenAPI error does not interfere with the deploy and does not reappear in subsequent prs

PR #2246, opened 2025-05-12
Fix flaky test

shanbady:

Closes https://github.com/mitodl/hq/issues/7312

This is a tiny PR to sort lists in profiles/serializers_test.py::test_serialize_profile_preference_search_filters to resolve flakiness

### How can this be tested?
Tests should always pass


poetry.lock
```diff
@@ -1,12 +1,11 @@
-# This file is automatically @generated by Poetry 2.1.2 and should not be changed by hand.
+# This file is automatically @generated by Poetry 1.8.5 and should not be changed by hand.
```
abeglova, line 1: I think the poetry.lock file was committed by mistake

abeglova (changes requested):
Change to the test looks good

poetry.lock
shanbady: good call. must have snuck in while doing renovate items and switching branches. removed

abeglova (approved):
👍

PR #2250, opened 2025-05-14
config: Remove syslog handler

blarghmatey:
N/A

Remove syslog handler for logging

### How can this be tested?

Verify that logs are still displayed in docker-compose output

### Additional Context

Now that this app is not running on Heroku there's no need to use the syslog handler


rhysyngsun (approved):
LGTM

PR #2256, opened 2025-05-20
integrate mitxonline api

gumaerc:

- [ ] Desktop screenshots


frontends/main/src/app-pages/DashboardPage/OrganizationContent.tsx
```diff
@@ -95,6 +66,7 @@ const OrganizationHeader: React.FC<{ org: Organization }> = ({ org }) => {
  * For an array of programs, fetch the associated courses.
  * [program1, program2] => [[...courses1], [...courses2]]
  */
+
```
jkachel, line 69: Since it won't let me leave this on the correct line, it's here instead. A few lines down (line 76) this loads the course list by ID. This fails for me and after some red herring chasing, it turned out it's because the org ID isn't also specified. So, the `courseQueries.coursesList` call below also needs to include the `org_id`. 

It fails for me because my local setup has courses that _only_ have B2B course runs - this would work as-is if you have courses with regular course runs (which you do if you use the test data). (But it wouldn't be completely correct if you just had regular runs anyway, because the course runs wouldn't align with the user's enrollments.)

jkachel (changes requested):
Minor change needed. Otherwise seems to work OK.

gumaerc:
@jkachel Thanks for the review and fixing the api client. I updated it here and the org page should now be filtering courses based on `org_id`.

PR #2284, opened 2025-06-03
Feature flag syllabus chat for programs

shanbady:

Related to https://github.com/mitodl/hq/issues/7370

This PR makes it so that the Syllabus chat for programs have a separate feature flag than the flag used for courses.

### How can this be tested?
1. Checkout this branch.
2. Note that the syllabus bot is disabled by default on program resource drawers
3. If you have a local posthob setup - enable/add a "pr-drawer-chatbot" - otherwise you can short-circuit it by manually setting programChatbotEnabled in LearningResourceExpanded.tsx to true
4. note that the chatbot is visible.


mbertrand (approved):
👍

PR #2289, opened 2025-06-05
dialing back get youtube data task

shanbady:

Closes https://github.com/mitodl/hq/issues/7487

This PR dials back the get youtube data task from 30 minutes to 4 hours

### How can this be tested?
Verify the schedule in celery_settings.py and make sure it is set to 4 hours.

### Additional Context
This was a change that was previously in place but appears to have been reverted at some point.


shanbady:
it does appear to still be 4 hours (set via env vars)

PR #2292, opened 2025-06-06
Allow files from multiple course runs to be indexed

abeglova:
closes https://github.com/mitodl/hq/issues/7505

This pr updates the etl to import content files from multiple runs

### How can this be tested?
1. run `docker-compose run web ./manage.py backpopulate_mitxonline_data`
2. find the resource id for "Paradox and Infinity"
3. run `docker-compose run web ./manage.py backpopulate_mitxonline_files --resource-ids=the_id_from_step_2 --overwrite`
4. Check that you have content files from multiple runs from that resource with
`ContentFile.objects.filter(run__learning_resource_id=the_id_from_step_2).values('run_learning_resource_id').annotate(Count('run_id', distinct=True))`
5. Verify that search and the learning_resource apis work correctly

### DIFF
--- a/learning_resources/etl/edx_shared.py
+++ b/learning_resources/etl/edx_shared.py
@@ -16,7 +16,6 @@ from learning_resources.etl.utils import (
     transform_content_files,
 )
 from learning_resources.models import LearningResourceRun
-from learning_resources.utils import content_files_loaded_actions
 
 log = logging.getLogger(__name__)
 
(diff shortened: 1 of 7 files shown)


shanbady (changes requested):
I'm wondering if it is a good idea to initially constrain enabling multiple runs to just courses that are in test_mode? 
My concern is with how much new content we will now regularly have to embed and process - our celery queues are already having issues keeping up with the backlog of tasks. Do we have some idea of how much new content enabling this across the board introduces?

shanbady (approved):
My concern was overblown - seems we dont have that many courses with multiple runs at the moment

PR #2305, opened 2025-06-13
Feature Flag for MITxOnline API Call

ChristopherChudzicki:

Closes https://github.com/mitodl/hq/issues/7573

This PR fixes some issues with the `mitlearn-organization-dashboard` feature flag. Now:
- No current user mitxonline API call is made if flag is disabled
- Non-organization dashboard tabs load immediately

### How can this be tested?

**Reminder:** Tab = label on left, TabPanel = thing that holds content.

Visit http://learn.odl.local:8062/dashboard
1. On `main` you will initially see 0 tabs. On this branch, you should see all non-org tabs immediately.
 2. On `main`, mitxonline endpoint `/api/v0/users/current_user/` is always called. On this branch, it should only be called if `mitlearn-organization-dashboard` flag is enabled in your posthog instance.
3. The carousels should load faster on this branch.
    - Previously, API call order was `api/v0/users/me` THEN `api/v0/profiles/me` THEN carousels load. The profile API call no longer happens on home tab.

### DIFF
--- a/frontends/api/src/mitxonline/hooks/user/index.ts
+++ b/frontends/api/src/mitxonline/hooks/user/index.ts
@@ -2,7 +2,7 @@ import { useQuery } from "@tanstack/react-query"
 import { usersApi } from "../../clients"
 import type { User } from "@mitodl/mitxonline-api-axios/v1"
 
-const useMitxOnlineCurrentUser = () =>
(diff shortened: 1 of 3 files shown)


gumaerc (approved):
👍 LGTM

PR #2309, opened 2025-06-17
unpin qdrant-client

shanbady:

Closes  https://github.com/mitodl/hq/issues/7614

This PR unpins qdrant-client from a specific version which we initially did to fix https://github.com/mitodl/hq/issues/7519

### How can this be tested?
Tests should pass without errors. previously running `python manage.py learning_resources/etl/loaders_test.py` would fail


mbertrand (approved):
LGTM

PR #2322, opened 2025-06-27
Add favorites list plugin to application defaults

shanbady:
Closes https://github.com/mitodl/hq/issues/7544

This PR adds the FavoritesListPlugin plugin to application defaults. FavoritesListPlugin is responsible for automatically creating a placeholder favorites list for users upon signup.

#### NOTE: this already works locally. this fix is intended for deployed environments

1. checkout this branch
2. create a new user or login as existing
3. post-auth check to see that they have at least one favorites list via:
```python
User = get_user_model()
user = User.objects.get(id=<your user id>)
assert user.user_lists.count() >= 1
```


mbertrand (approved):
👍

PR #2339, opened 2025-07-03
Shanbady/summary flashcard sync

shanbady:

Closes https://github.com/mitodl/hq/issues/7714

This PR ensures that content that already has an existing summary/flashcard stays in sync when content is updated

### How can this be tested?
1. Checkout this branch
2. make sure QDRANT_ENABLE_INDEXING_PLUGIN_HOOKS is set to True in settings
3. ensure you also have an open ai api key configured.
4. configure a summarizer for an etl source locally (lets pick mitxonline) - you can reference the summarizer currently configured on production here https://api.learn.mit.edu/admin/learning_resources/contentsummarizerconfiguration/1/change/
5. pull down contentfiles ```python manage.py backpopulate_mitxonline_files```
6. note that we dont have summaries for any contentfiles yet:
```python
ContentFile.objects.exclude(summary="").all()
```
7. manually set the "summary" field for a contentfile to "test" - also change the checksum for the associated run and contentfile so it thinks we have an update.
```python
cf = ContentFile.objects.filter(file_extension=".srt").first()
run = cf.run
lr = run.learning_resource

cf.summary = "test"
cf.checksum = "test"
run.checksum = "test"
cf.save()
run.save()
print(f"resource id - {lr.id}")
```
8. pull down contentfiles for that particular learning resource again ```python manage.py backpopulate_mitxonline_files ---resource-ids {resourceid}```
10. once complete you should see a generated summary populated for just that one contentfile

**note**: Dont forget to unset your openai api key once done testing.

### Additional Context
On live systems, the ETL plugin hooks will call generate_embeddings with the overwrite flag which is what we are testing here.

### DIFF
--- a/vector_search/utils.py
+++ b/vector_search/utils.py
@@ -106,7 +106,9 @@ def create_qdrant_collections(force_recreate):
                     ),
                 ),
             },
-            sparse_vectors_config=client.get_fastembed_sparse_vector_params(),
+            sparse_vectors_config={
+                "bm25": models.SparseVectorParams(modifier=models.Modifier.IDF)
+            },
             optimizers_config=models.OptimizersConfigDiff(default_segment_number=2),
             quantization_config=models.ScalarQuantization(
                 scalar=models.ScalarQuantizationConfig(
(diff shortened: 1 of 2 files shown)


mbertrand (approved):
👍

PR #2340, opened 2025-07-07
Return to page after sign up

jonkafton:

Closes https://github.com/mitodl/hq/issues/7734

Returns new users to their initial page after completing the sign up and onboarding flow.

### How can this be tested?

Start on a page other than the homepage and not signed in. Search params on the return URL should also be preserved, so we can start with a drawer open, e.g http://open.odl.local:8062/search?resource=186.

- Hit a bookmark button.

- Note that the "Sign Up" button link includes a next parameter to the current URL (encoded), e.g. http://api.open.odl.local:8065/login?next=http://open.odl.local:8062/search%3Fresource%3D186

- Hit the link button and register a new account.

- Note that the onboarding page URL includes the next parameter, e.g. http://open.odl.local:8062/onboarding?next=http%3A%2F%2Fopen.odl.local%3A8062%2Fsearch%3Fresource%3D186. Will need attention on RC where email verification is enabled.

- Step through the onboarding screens.

- After hitting the "Finish", confirm that you land on the initial page (client only navigation).

### Additional Context

The acceptance criteria on https://github.com/mitodl/hq/issues/7734 includes the same behavior for previously registered users however that already works great - no changes included.

### DIFF
--- a/frontends/main/src/app-pages/OnboardingPage/OnboardingPage.test.tsx
+++ b/frontends/main/src/app-pages/OnboardingPage/OnboardingPage.test.tsx
@@ -59,7 +59,7 @@ const STEP_TITLES = [
 
 const PROFILES_FOR_STEPS = times(STEPS_DATA.length, profileForStep)
 
-const setup = async (profile: Profile) => {
+const setup = async (profile: Profile, url?: string) => {
   allowConsoleErrors()
   setMockResponse.get(urls.userMe.get(), factories.user.user())
   setMockResponse.get(urls.profileMe.get(), profile)
@@ -68,12 +68,12 @@ const setup = async (profile: Profile) => {
     ...req,
   }))
 
-  renderWithProviders(<OnboardingPage />)
+  renderWithProviders(<OnboardingPage />, url ? { url } : undefined)
 }
 
 // this function sets up the test and progresses the UI to the designated step
-const setupAndProgressToStep = async (step: number) => {
-  await setup(PROFILES_FOR_STEPS[step])
+const setupAndProgressToStep = async (step: number, url?: string) => {
+  await setup(PROFILES_FOR_STEPS[step], url)
 
   for (let stepIdx = 0; stepIdx < step; stepIdx++) {
     await user.click(await findNextButton())
@@ -152,4 +152,16 @@ describe("OnboardingPage", () => {
       )
     },
   )
+
+  test("Redirects to next url after completion if present", async () => {
+    const nextUrl = `${process.env.NEXT_PUBLIC_ORIGIN}/search?resource=184`
+    await setupAndProgressToStep(STEPS_DATA.length - 1, nextUrl)
+    const finishButton = await findFinishButton()
+    await user.click(finishButton)
+    await waitFor(() =>
+      expect(window.location.pathname + window.location.search).toBe(
+        "/search?resource=184",
+      ),
+    )
+  })
 })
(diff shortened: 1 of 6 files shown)


frontends/main/src/app-pages/OnboardingPage/OnboardingPage.test.tsx
```diff
@@ -152,4 +152,16 @@ describe("OnboardingPage", () => {
       )
     },
   )
+
+  test("Redirects to next url after completion if present", async () => {
+    const nextUrl = `${process.env.NEXT_PUBLIC_ORIGIN}/search?resource=184`
+    await setupAndProgressToStep(STEPS_DATA.length - 1, nextUrl)
+    const finishButton = await findFinishButton()
+    await user.click(finishButton)
```
ChristopherChudzicki, line 160: I took a quick look at this test. I don't think it's testing anything right now—it passes if you comment out these two lines.

I think the test should be more akin to:
```ts
const pushSpy = jest.spyOn(mockRouter, "push")
const nextUrl = `${process.env.NEXT_PUBLIC_ORIGIN}/search?resource=184`
const url = `${process.env.NEXT_PUBLIC_ORIGIN}/onboarding?next=${encodeURIComponent(nextUrl)}`
await setupAndProgressToStep(STEPS_DATA.length - 1, url)
const finishButton = await findFinishButton()
await user.click(finishButton)
expect(pushSpy).toHaveBeenCalledWith(nextUrl)
```
- I did try ☝️ and got errors about missing prefetch. Not sure why? I think mock router has a no-op prefetch... https://github.com/scottrippey/next-router-mock/blob/aa891b21af1168e32bb6eb5a4279e86ccf1dcca9/src/MemoryRouter.tsx#L73
- Semi related... I'd like to get rid of `frontends/ol-test-utilities/src/mocks/nextNavigation.ts` all together. Our mock router is based on `next-router-mock` v0, which only supported NextJS' Pages router (i.e., `next/router`, not `next/navigation`). We recently updated to the (recently released) v1 #2267 which **does** support the App router. Ideally we can get rid of all our customization of the mock router.

frontends/main/src/app-pages/OnboardingPage/OnboardingPage.tsx
```diff
@@ -169,13 +172,23 @@ const OnboardingPage: React.FC = () => {
...
     },
     validateOnChange: false,
     validateOnBlur: false,
   })
 
+  useEffect(() => {
+    if (nextUrl) {
+      router.prefetch(nextUrl)
+    }
+  }, [nextUrl])
```
ChristopherChudzicki, line 190: Good idea.

I'm with github... I'd add `router` to the prefetch array. It's probably a singleton anyway?

ChristopherChudzicki (changes requested):
Works great! Cool.

I did notice one thing with the test—not sure it's testing what it was intended to.

frontends/main/src/app-pages/OnboardingPage/OnboardingPage.test.tsx
jonkafton: Hmm, yes - I was loading the page with the return URL and rejoicing that it was indeed still on the same URL 🤔

I was under the impression that our our `pushState` mock gets called and updates window.location, but not the case.

I've fixed the test to load on the onboarding URL and actually test the navigation. Now using next-router-mock for the OnboardingPage.test.tsx spec - works well and no prefetch complaints (using its /navigation submodule, https://github.com/scottrippey/next-router-mock?tab=readme-ov-file#usage-with-nextnavigation-beta for mocking "next/navigation").

PR #2346, opened 2025-07-10
Set first run language on drawer and card titles and description

jonkafton:

Closes https://github.com/mitodl/hq/issues/6831

Sets a lang attribute on the learning drawer title and description and resource card titles for where available to improve machine readability.

This will provide occasional benefit, though we currently have limited data on the API:

- We have a `resource.language` attribute, though I haven't come across any resources where this is not null. This course has `resource.language` null and a title and description in Turkish: https://api.learn.mit.edu/api/v1/learning_resources/3048/.

- I'm using the value on `resource.runs[0].languages.[0]`, the first language on the fist run. This is also what we're using for the languages in the drawer info section (though listing all languages for all runs). If a course has runs in several languages, this may not be the language of the parent title and description and in any case there's not guarantee the course title and description are in the same language as the first run.

- Some of our data is missing or incorrect. This course offered in Spanish does not have languages set on its run: https://api.learn.mit.edu/api/v1/learning_resources/16541/. This course offered in Mandarin has "en-us" set on its run languages: https://api.learn.mit.edu/api/v1/learning_resources/2911/

The fix needed, if available on the source data, is to ensure that the language of the course title and description is always set on the on the resource so we're not selecting from the runs in hope of a match. An example of a _good_ resource where the title matches the run language: https://api.learn.mit.edu/api/v1/learning_resources/2908/. Side note: It would be really useful to provide a language filter on the search for foreign language learners.

### How can this be tested?

- Search for some foreign language courses. Some searches:
  - "Kullanıcıdan" (Turkish / tr)
  - "스타트업 기업가정신" (Korean / ko)

- Confirm that the list card titles have a `lang` attribute set to the first run language.

- Open the resource drawer and confirm that both the title and description have a `lang` attribute set.

### DIFF
--- a/frontends/main/src/page-components/LearningResourceExpanded/LearningResourceExpanded.test.tsx
+++ b/frontends/main/src/page-components/LearningResourceExpanded/LearningResourceExpanded.test.tsx
@@ -114,6 +114,25 @@ describe("Learning Resource Expanded", () => {
     screen.getByText(resource.title)
   })
 
+  test("Sets lang attribute on title and description", () => {
+    const resource = factories.learningResources.resource({
+      resource_type: ResourceTypeEnum.Course,
+      runs: [
+        factories.learningResources.run({
+          languages: ["en-us"],
+        }),
+      ],
+    })
+
+    setup({ resource })
+
+    const title = screen.getByText(resource.title)
+    expect(title).toHaveAttribute("lang", "en-us")
+
+    const description = screen.getByText(resource.description!)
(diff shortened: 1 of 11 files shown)


abeglova (approved):
This works for now although i agree that we need to fix the etl to propagate run language data to the learning resource and also more consistently  populate language data on runs. In particular i noticed MIT Professional Education has lots of foreign language courses and none of them have the language correctly set in the api on either the resource or the runs

PR #2347, opened 2025-07-10
canvas - separate courses per session

shanbady:

Closes https://github.com/mitodl/hq/issues/7772

This PR makes it so that the course folder is included in the readable id when creating canvas learning resource objects. 

### How can this be tested?
1. Checkout this branch
2. restart celery
3. ingest canvas courses via `python manage.py backpopulate_canvas_courses`
4. once finished, inspect the readable ids of the learning resource objects where etl_source="canvas"
```
list(LearningResource.objects.filter(etl_source="canvas").values_list("readable_id"))
```
or check via admin: http://open.odl.local:8063/admin/learning_resources/learningresource/?etl_source=canvas
5. note that the readable ids should look like "{folder_name}-{course_id}"

### Additional Context
In the initial implementation we expected the course id in the course_settings.xml to be unique which turned out to not be the case between sessions. This pr resolves that issue.

### DIFF
--- a/learning_resources/etl/canvas_test.py
+++ b/learning_resources/etl/canvas_test.py
@@ -83,6 +83,7 @@ def test_run_for_canvas_archive_creates_resource_and_run(tmp_path, mocker):
     Test that run_for_canvas_archive creates a LearningResource and Run
     when given a valid canvas archive.
     """
+    course_folder = "test"
     mocker.patch(
         "learning_resources.etl.canvas.parse_canvas_settings",
         return_value={"title": "Test Course", "course_code": "TEST101"},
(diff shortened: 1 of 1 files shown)


learning_resources/etl/canvas.py
```diff
@@ -57,7 +62,9 @@ def run_for_canvas_archive(course_archive_path, overwrite):
     )
     if resource.runs.count() == 0:
         LearningResourceRun.objects.create(
-            run_id=f"{readable_id}+canvas", learning_resource=resource, published=True
+            run_id=f"{readable_id}+canvas",
+            learning_resource=resource,
+            published=True,
```
mbertrand, line 67: New session-specific courses were created, but the 2 older courses from an earlier command run are still present with `test_mode=True`.  I think any courses not found in the archives during a command run should probably be deleted and/or have test_mode set to `False` (in addition to `published=False`), with all their contentfiles deleted or unpublished from regular and vector search.

mbertrand (changes requested):

learning_resources/etl/canvas.py
shanbady: I added the deletion of stale canvas courses. Retrieving the current course ids required going into the archive so i funnel up the relevant readable ids and at the end of the command do a deletion+deindex. This also only applies when running the management command since the webhook just tells us to add a new entry.

mbertrand (approved):
LGTM! I wonder if we should eventually add an unpublish webhook like OCW has, but that's an issue for another day/PR.

PR #2357, opened 2025-07-14
add a healthcheck route

ChristopherChudzicki:
None (?)

Adds a healthcheck endpoint for use in our kubernetes deployments

### How can this be tested?
With docker up, `curl -v http://learn.odl.local:8062/healthcheck`. You should get a 200 response.

### DIFF
--- a/frontends/main/src/app/healthcheck/route.ts
+++ b/frontends/main/src/app/healthcheck/route.ts
@@ -1,0 +1,3 @@
+export async function GET() {
+  return Response.json({ status: "ok" }, { status: 200 })
+}


frontends/main/src/app/healthcheck/route.ts
```diff
@@ -0,0 +1,3 @@
+export async function GET() {
+  return Response.json({ status: "ok" }, { status: 200 })
+}
```
ChristopherChudzicki, line 3: BTW: This is a NextJS route handler file for making APIs. We don't use these much, since our APIs are all on Django.

(The only other one we have is for sitemap index)

https://nextjs.org/docs/app/getting-started/route-handlers-and-middleware

mbertrand (approved):
👍

PR #2312, opened 2025-06-18
dashboard email settings dialog functionality

gumaerc:
### What are the relevant tickets?
Closes https://github.com/mitodl/hq/issues/7637

### Description (What does it do?)
This PR adds the hooks necessary to update email communication settings in MITx Online (and subsequently OpenEdx) and implements them on the email settings dialog in the Dashboard.

### How can this be tested?
In order to test this, you need a basic installation of `mitxonline` up and running with example data in it. You may be able to skip one or more steps if you have already done them:
- Ensure you have local `hosts` redirects for the following domains, replacing the example IP with your __local__ IP address (Google how to get this if unsure, mine is 192.168.1.50)
```
192.168.1.50 open.odl.local
192.168.1.50 api.open.odl.local
192.168.1.50 kc.ol.local
192.168.1.50 mitxonline.odl.local
```
- Clone `mitxonline`: https://github.com/mitodl/mitxonline
- Create a `.env` file with the following values:
```
CELERY_TASK_ALWAYS_EAGER=True
DJANGO_LOG_LEVEL=INFO
LOG_LEVEL=INFO
SENTRY_LOG_LEVEL=ERROR
MAILGUN_KEY=fake
MAILGUN_URL=
MAILGUN_RECIPIENT_OVERRIDE=
MAILGUN_SENDER_DOMAIN=.odl.local
SECRET_KEY=
STATUS_TOKEN=
UWSGI_THREADS=5
SENTRY_DSN=
MITX_ONLINE_BASE_URL=http://open.odl.local:8065/mitxonline
MITX_ONLINE_ADMIN_CLIENT_ID=refine-local-client-id
MITX_ONLINE_ADMIN_BASE_URL=http://mitxonline.odl.local:8016
POSTHOG_PROJECT_API_KEY=
POSTHOG_API_HOST=https://app.posthog.com/
HUBSPOT_HOME_PAGE_FORM_GUID=
HUBSPOT_PORTAL_ID=
APISIX_PORT=9080

# APISIX/Keycloak settings
APISIX_LOGOUT_URL=http://api.open.odl.local:8065/logout/
APISIX_SESSION_SECRET_KEY=supertopsecret1234
KC_SPI_THEME_WELCOME_THEME=scim
KC_SPI_REALM_RESTAPI_EXTENSION_SCIM_LICENSE_KEY=
KEYCLOAK_BASE_URL=http://kc.ol.local:8066
KEYCLOAK_CLIENT_ID=apisix
[...]

PR #2360, opened 2025-07-14
Add path prefixed MITx Online API config to local dev APISIX

gumaerc:
Related to https://github.com/mitodl/hq/issues/7796

This PR adds a separate route in the APISIX configuration used for local development. The route is a path prefixed route to the MITx Online API, so you can access it at http://open.odl.local:8065/mitxonline/api.

### How can this be tested?
- In order to test this PR, you should have a running instance of MITx Online that is using the same Keycloak / APISIX as MIT Learn
- Assuming you are using the default local domains (`open.odl.local` and `mitxonline.odl.local`) access Learn at http://open.odl.local:8062/ and Log In using Keycloak
- Open a new tab and browse to http://open.odl.local:8065/mitxonline/api/v0/users/me
- Verify that you see the response from the MITx Online API


ChristopherChudzicki (approved):
👍 

But... Could add some docs similar to directions in https://github.com/mitodl/mit-learn/pull/2312 ?

The key points to me seem to be:
- etc/host using ip for inter-network communication
- keycloak vars must match between mitxonline and learn
- use learn's apigateway
- mitxonline's cors_allowed_origins

shanbady (approved):
👍

PR #2361, opened 2025-07-15
remove unused MITOL_NOINDEX from django settings [Only used on nextjs]

ChristopherChudzicki:
None

Removes an unused django setting.

### How can this be tested?
Tests should pass. Search `MITOL_NOINDEX` in codebase; only used in NextJS app.

### Additional Context
Noticed this while working on https://github.com/mitodl/ol-infrastructure/pull/3380; the backend env var that was setting this value is reoved in that PR.

### DIFF
--- a/main/settings.py
+++ b/main/settings.py
@@ -747,9 +747,6 @@ POSTHOG_PROJECT_ID = get_int(
     default=None,
 )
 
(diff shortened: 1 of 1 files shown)


abeglova (approved):
LGTM

PR #2364, opened 2025-07-15
favorites list fix

shanbady:

Closes https://github.com/mitodl/hq/issues/7544

I connected with @rhysyngsun who pointed me towards a solution. Locally we have an authentication middleware that calls the appropriate plugin methods. In our deployed environments the authentication bypasses it since it is done via scim. This PR adds the plugin calls to the scim user adapter.

### How can this be tested?
I'm unsure exactly how to go about testing this locally but I've added a test to ensure when the scim adapter saves, the default favorites list and profile get created.


mbertrand (approved):
👍

PR #2370, opened 2025-07-18
prevent hero layout shift

ChristopherChudzicki:
Closes https://github.com/mitodl/hq/issues/7818

Fixes min height of AskTim hero area container to prevent layout shift as feature flag loads.

### How can this be tested?
1. Load the homepage with AskTim recommendation bot turned on. There should be no layout shift as hero section loads.
2. Open/close the drawer. It should work same.
3. AskTim button should have same dimensions as RC / Prod

### DIFF
--- a/frontends/main/src/page-components/AiChat/AskTimDrawerButton.tsx
+++ b/frontends/main/src/page-components/AiChat/AskTimDrawerButton.tsx
@@ -1,11 +1,10 @@
 import React from "react"
-import { Typography, styled } from "ol-components"
-import { ButtonLink } from "@mitodl/smoot-design"
+import { Typography, styled, LinkAdapter } from "ol-components"
 import { RiSparkling2Line } from "@remixicon/react"
 import AiRecommendationBotDrawer from "./AiRecommendationBotDrawer"
 import { RECOMMENDER_QUERY_PARAM } from "@/common/urls"
 
-const StyledButton = styled(ButtonLink)(({ theme }) => ({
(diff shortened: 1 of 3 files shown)


frontends/main/src/page-components/AiChat/AskTimDrawerButton.tsx
```diff
@@ -1,44 +1,29 @@
 import React from "react"
-import { Typography, styled } from "ol-components"
-import { ButtonLink } from "@mitodl/smoot-design"
+import { Typography, styled, LinkAdapter } from "ol-components"
 import { RiSparkling2Line } from "@remixicon/react"
 import AiRecommendationBotDrawer from "./AiRecommendationBotDrawer"
 import { RECOMMENDER_QUERY_PARAM } from "@/common/urls"
 
-const StyledButton = styled(ButtonLink)(({ theme }) => ({
+const StyledButton = styled(LinkAdapter)(({ theme }) => ({
```
ChristopherChudzicki, line 7: We shouldn't use ButtonLink here. We don't want any of its styling.

As far as I can tell, we only used it because we wanted the `shallow` prop, which comes from LinkAdapter.

ChristopherChudzicki:
That's interesting. I see the slight placeholder shift very clearly in your video, but it doesn't seem to happen in Chrome or Firefox for me, having tested at various zoom levels.

Not sure what to do about that, but, as you say, not very noticeable.

PR #2372, opened 2025-07-18
update nextjs

ChristopherChudzicki:
None

Updates NextJS

### How can this be tested?
Tests should pass, site should look/behave the same.

### DIFF
--- a/frontends/main/src/app/page.tsx
+++ b/frontends/main/src/app/page.tsx
@@ -29,6 +29,18 @@ export async function generateMetadata({
   })
 }
 
+/**
+ * Fallback to dynamic rendering to load as much of the page as possible.
+ *
+ * We could alternatively wrap the carousels (which use useSearchParams) in a
+ * suspense boundary. This ostensibly leads to a faster response, since the
+ * server can render the rest of the homepage at build time.
+ *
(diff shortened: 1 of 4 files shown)


frontends/main/src/app/page.tsx
```diff
@@ -29,6 +29,18 @@ export async function generateMetadata({
...
+ * Fallback to dynamic rendering to load as much of the page as possible.
+ *
+ * We could alternatively wrap the carousels (which use useSearchParams) in a
+ * suspense boundary. This ostensibly leads to a faster response, since the
+ * server can render the rest of the homepage at build time.
+ *
+ * But... We ache the result on CDN anyway, so it's essentially pre-build for
+ * most requests, anyway.
+ */
+export const dynamic = "force-dynamic"
```
ChristopherChudzicki, line 42: This fixes a build error on NextJS 15.4.1; IMO it's a NextJS bug that we weren't getting this error before.

jonkafton: Ah the carousels use `useSearchParams()` as per your comment and the previous NextJS version was ignoring silently.

jonkafton (approved):

PR #2374, opened 2025-07-19
unpublish contentfiles webhook endpoint

shanbady:

Closes https://github.com/mitodl/hq/issues/7810

This PR adds the following:
- Ability to remove/unpublish contentfiles (canvas only at the moment) via webhook
- A celery task that runs 1 time per week to keep all canvas courses in sync

### How can this be tested?
1. checkout this branch
2. restart celery
3. Run the following snippet to ingest a canvas course via webhook
```python
import hmac
import hashlib
import json
from time import time
import requests

data = {
	'content_path': 'canvas/course_content/28751/fc0418f679c7c7f5b7e9ec73c40665e1118d36d2e48f96df651e26a70e0e658c.imscc',
	'source':'canvas',
	'timestamp':time()
}
payload = bytes(json.dumps(data), 'utf-8')
secret = settings.WEBHOOK_SECRET
computed_signature = hmac.new(secret.encode(), payload, hashlib.sha256).hexdigest()

requests.post('http://open.odl.local:8063/api/v1/webhooks/content_files/', json=data, headers={'X-MITLearn-Signature': computed_signature})
```
4. That call should result in a 200 response. Give it a moment for celery to finish downloading and processing then check to confirm there is a canvas learning resource via shell or django admin: `LearningResource.objects.filter(etl_source="canvas")`
5. Now send a request to delete the canvas course via webhook via:
```python
import hmac
import hashlib
import json
import requests
from time import time
data = {
	'course_id': '28751',
	'source':'canvas',
	'timestamp':time()
}
payload = bytes(json.dumps(data), 'utf-8')
secret = "please-change-this"
computed_signature = hmac.new(secret.encode(), payload, hashlib.sha256).hexdigest()

requests.post('http://open.odl.local:8063/api/v1/webhooks/content_files/delete/', json=data, headers={'X-MITLearn-Signature': computed_signature})
```
6.  `LearningResource.objects.filter(etl_source="canvas")` should return empty (or be missing that particular course if you have other canvas courses)


webhooks/views.py
```diff
@@ -45,35 +50,83 @@ class ContentFileWebhookView(BaseWebhookView):
...
+    """
+    Process a content file DELETE webhhook request based on the ETL source
+    """
+    etl_source = data.get("source")
+    if etl_source == ETLSource.canvas.name:
+        course_id = data.get("course_id")
+        if course_id:
+            try:
+                resource = LearningResource.objects.get(
+                    readable_id__istartswith=course_id, etl_source=ETLSource.canvas.name
```
mbertrand, line 124: This could potentially cause problems if the course_id is `287` and there is also a canvas course_id of `2871`.  Since the `readable_id` is `<canvas_course_id>_<course_code>`, might be best to change this slightly to include the `_`:  

```python
readable_id__istartswith=f"{course_id}_"
```

webhooks/views.py
```diff
@@ -45,35 +50,83 @@ class ContentFileWebhookView(BaseWebhookView):
...
+                )
+                resource.published = False
+                resource.test_mode = False
+                resource.save()
+                resource_unpublished_actions(resource)
+                resource.delete()
```
mbertrand, line 130: If it's going to be deleted, you don't need to change the values of `published` and `test_mode`.  Or you can not delete it and just change the above values, I think either approach is fine, though usually we unpublish old resources instead of deleting them.

If you prefer to delete, there is a  `learning_resources.utils.resource_delete_actions` function which should call the `resource_before_delete` hook, triggering plugins of the same name in `channels.plugins` (remove any topics exclusive to that resource, then delete the resource) and `learning_resources_search.plugins` (remove from ES and qdrant).  So you could replace the 2 lines immediately above with `resource_delete_actions(resource)` though I didn't test this to verify.

mbertrand (changes requested):

PR #2384, opened 2025-07-22
separate login and signup redirects

ChristopherChudzicki:
Closes https://github.com/mitodl/hq/issues/7734

This PR adds separate redirect parameters for login and signup. The goal is to allow different login/signup links to behave differently from one another:
- some will redirect to `/onboarding?next={original-url}` upon **signup**, and to `original-url` upon **login**.
    - achieved via `signup_next=/onboarding?next={original-url}` and `next=original-url`
- others will redirect to `/onboarding?next=/dashboard` upon **signup**, and to `original-url` upon **login**.
    - achieved via `signup_next=/onboarding?next=dashboard` and `next=original-url`

### How can this be tested?
Locally, for non-org users, check that:

1. Login redirects you to the current page, always.
2. Signup from the "Signup popover" redirects you to the current page AFTER onboarding.
    - Signup popover can be triggered by clicking the bookmark buttons on cards, or by clicking "Subscribe" on channel pages.
3. Signup from all other signup links (Header on any page; "Sign up" in blue section on homepage) redirects you to dashboard AFTER onboarding.

For org users, separate logic not touched in this PR redirects to `/dashboard/org-name` after signup.

### DIFF
--- a/frontends/main/src/app-pages/B2BAttachPage/B2BAttachPage.tsx
+++ b/frontends/main/src/app-pages/B2BAttachPage/B2BAttachPage.tsx
@@ -42,9 +42,11 @@ const B2BAttachPage: React.FC<B2BAttachPageProps> = ({ code }) => {
       return
     }
     if (!user?.is_authenticated) {
-      const loginUrlString = urls.login({
-        pathname: urls.b2bAttachView(code),
-        searchParams: new URLSearchParams(),
+      const loginUrlString = urls.auth({
+        loginNext: {
+          pathname: urls.b2bAttachView(code),
+          searchParams: null,
+        },
       })
       const loginUrl = new URL(loginUrlString)
       loginUrl.searchParams.set("skip_onboarding", "1")


authentication/views.py
```diff
@@ -73,15 +73,18 @@ def get(
...
-                params = urlencode({"next": redirect_url})
-                redirect_url = f"{settings.MITOL_NEW_USER_LOGIN_URL}?{params}"
+            if request.GET.get("skip_onboarding", "0") == "0":
+                params = urlencode({"next": signup_redirect_url})
+                login_redirect_url = f"{settings.MITOL_NEW_USER_LOGIN_URL}?{params}"
                 profile.completed_onboarding = True
                 profile.save()
-        return redirect(redirect_url)
+            else:
+                return redirect(signup_redirect_url)
```
rhysyngsun, line 89: It's possible that `signup_next` isn't specified and I think it's reasonable to have this implemented so that it's not required. In that case, it should fall back to `next`, otherwise the default (`/app`).

rhysyngsun (changes requested):

rhysyngsun (commented):
Code looks good, will finish functional testing tomorrow

rhysyngsun (approved):
LGTM

PR #2388, opened 2025-07-24
Video Shorts / Media Section

jonkafton:

https://github.com/mitodl/hq/issues/7824

Implements the initial designs for the new Video Shorts section on the homepage with vertical scrolling video player. This is feature flagged with Posthog though releasable with backend support to ingest videos from YouTube.

It currently shows a selection of videos from https://www.youtube.com/@MITOpenLearning/shorts that are converted to mp4s and loaded to a temporary bucket.

We are introducing a new carousel, powered by [Embla](https://www.embla-carousel.com/), as our existing Slick carousels have limited support for gesture/swipe navigations. The new carousels are also applied to all resource carousels.

- Wheel/swipe gestures:

https://github.com/user-attachments/assets/180e8c8b-8812-48ca-aec8-962c5098c317

- Player:

https://github.com/user-attachments/assets/da8caff4-4cf4-44f5-bfff-8bc84541a0f2

(no audio)

- Fixes the drawer carousel right margin:

| Before    | After |
| -------- | ------- |
| image  |  image |

### How can this be tested?

- In your PostHog instance, create a feature flag "video-shorts" and set it to 100% rollout. Alternatively, assign `videoShortsEnabled` in MediaSection.tsx to true.

- Attention has been given to touch gestures - point to your locally running instance in your mobile browser to test there in addition to desktop.

- All resource carousels and the new Video Shorts carousel on the homepage should respond smoothly to scroll and swipe gestures on touchpads and mobile. The carousels have natural feeling inertia and do not snap into alignment to avoiding the jarring sensation of resisting the scroll movement. Carousels are also draggable, maintaining the start point under the pointer or finger and support inertial flinging.

- Ensure that prev/next arrows are disabled at scroll extents.

- Ensure that prev/next arrows scroll the number of elements visible on the page and snap with left alignment to their container.

- Click/tap a video thumbnail in the Video Section to open the vertical carousel player.

  - Videos should initially autoplay muted (autoplay policies prevent videos from playing with sound unless they're triggered by a user interaction).

  - The mute/unmute button should mute/unmute the video.
 
  - The next video should autoplay mutes/unmuted as per previous.
  
  - Videos can be navigated with up and down keypress.

  - Space keypress plays/pauses the video.
 
  - Escape closes the player.
  
  - Swipe to navigate prev/next video. We're using swipe gesture detection here as opposed to momentum scroll on the horizontal carousels. This gives us control to ensure only a single video is navigated per gesture to avoid scrolling past the next video and causing unnatural snaps under more aggressive gestures.
  
  - Only the current video and two adjacent videos are loaded to the page.
  
  - The close button closes the video player.

- Turn the "video-shorts" flag off in PostHog and confirm that the homepage Media section has not changed / matches Prod, with the exception that we are now rendering in the client only as a limitation due to loading the Posthog feature flag in the front end and otherwise causing an SSR hydration mismatch.

### Additional Context

Next steps:

- Design review @steven-hatch @mbilalmughal. I've added a mute button bottom-right, that we don't have designs for, to accommodate autoplay restrictions that require user interaction. Perhaps it's best in all cases that we don't autoplay with sound. I've also added a close button. These could use some design treatment - one issue is that they will not show over a white video background.

- UX / designs around the autoplay issue. Browsers will not autoplay a video that is not muted, enforced at varying degrees that are not too predictable. Chrome scores a Media Engagement Index on the degree to which the user has already engaged with the site, among other factors[^1]. Generally once a video has been unmuted, a subsequent video with autoplay with sound. On iOS though, each new video has to be interacted with. The current behavior is that the newly loaded video does not autoplay while audio is on and needs to be clicked to play. It's not necessarily obvious that the video is clickable to play/pause, so it would be good to have a design for a play/pause button or further investigation into a workaround for this limitation (it should be possible - YouTube shorts loaded in the browser play with sound).

- Deprecate Slick carousels once design approved and user tested and remove from codebase. The new carousel, living alongside the existing in ol-components/CarouselV2 implement the same API and are swappable.

- Move CarouselV2 to Smoot Design (can rename to just Carousel).

- Backend support:
  - We're using native HTML `<video>` elements pointing to `.mp4` files. Alternatively, videos could be embedded via [YouTube's iframe API](https://developers.google.com/youtube/iframe_api_reference), but iframe limitations on pointer events reduce support for gestures and responsiveness.
  - Videos must be converted to `.mp4` and hosted in a cloud bucket behind a media CDN[^2].
  - We may query the YouTube Data API directly, but integrating into the ETL pipelines and exposing via the Django API would ensure consistency and type safety.

[^1]: https://developer.chrome.com/blog/autoplay
[^2]: https://d1.awsstatic.com/whitepapers/amazon-cloudfront-for-media.pdf


frontends/ol-components/package.json
```diff
@@ -8,7 +8,9 @@
   },
   "exports": {
     ".": "./src/index.ts",
-    "./deprecated": "./src/components/deprecated/index.ts"
+    "./deprecated": "./src/components/deprecated/index.ts",
+    "./CarouselV2": "./src/components/CarouselV2/CarouselV2.tsx",
+    "./CarouselV2Vertical": "./src/components/CarouselV2/CarouselV2Vertical.tsx"
```
jonkafton, line 13: Using direct exports here as was seeing a circular dependency type issue with our barrel file around simultaneous imports and smoot design re-exports (tests fail with e.g. ReferenceError: Cannot access '_LearningResourceListCard' before initialization)

It looks something like:
- ol-components exports components that import from @mitodl/smoot-design.
- ol-components also exports deprecated aliases (Migrated exports) that are essentially re-exports of @mitodl/smoot-design components.
- When Jest tries to mock ol-components, it triggers the initialization of both the components AND the aliases.
- The aliases try to access @mitodl/smoot-design while the components are also trying to import from it.
- This creates a circular reference.

frontends/main/src/app-pages/HomePage/VideoShortsModal.tsx
```diff
@@ -0,0 +1,246 @@
...
+                ref={(el) => {
+                  if (videosRef.current && el) {
+                    videosRef.current[index] = el
+                    el.addEventListener("error", (e: Event) => {
+                      console.error("Video error:", e)
+                    })
+                  }
+                }}
+                onClick={handleVideoClick}
+                // TODO: Using a temporary bucket on GCP owned by jk
```
shanbady, line 223: Is this something that will be resolved in a follow up or can we move this to s3?

shanbady (changes requested):
Looks good. functions well. I tried a variety of browsers and devices. left a question about the temporary GCP bucket

shanbady (approved):
LGTM

PR #2406, opened 2025-07-31
change problem set list permissions

abeglova:
None

Currently both the api for listing problem sets for a course run and the api for showing problem set content requires a login and permissions. It will simplify the ui for the canvas tutorbot if the problem set list api was public. The problem set and solution content should still require a login and special permissions.

This would be consistent with how we treat content files too. Content file titles are available to anonymous users/ users without special permissions but content file content requires a login and permissions.

### How can this be tested?
Run docker compose exec web ./manage.py backpopulate_canvas_courses --canvas-ids=14566 --overwrite

As an anonymous user go to 
http://api.open.odl.local:8063/api/v0/tutor/problems/14566-kaleba:20211202+canvas/
you should see results
go to
http://api.open.odl.local:8063/api/v0/tutor/problems/14566-kaleba:20211202+canvas/Problem%20Set%201

you should get NotAuthenticated error,

Log in as an admin

You should see results at both the above urls

### DIFF
--- a/learning_resources/views_test.py
+++ b/learning_resources/views_test.py
@@ -1418,7 +1418,17 @@ def test_learning_resources_summary_listing_endpoint_large_pagesize():
     ],
 )
 def test_course_run_problems_endpoint(client, user_role, django_user_model):
-    """Test course run problems endpoint"""
+    """
+    Test course run problems endpoint.
+
+    All types of users should be able to get a list of problem set titles
+    for a course run from api/v0/tutorproblems/<run_readable_id>
+
+    Only admin and group_tutor_problem_viewer users should be able to get
+    problem and solution content from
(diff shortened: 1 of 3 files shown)


learning_resources/views_test.py
```diff
@@ -1463,20 +1463,7 @@ def test_course_run_problems_endpoint(client, user_role, django_user_model):
         reverse("lr:v0:tutorproblem_api-list-problems", args=[course_run.run_id])
     )
 
-    if user_role in ["admin", "group_tutor_problem_viewer"]:
```
shanbady, line 1466: minor nit - the docstring for this test could be a bit more descriptive (include details about what is actually being tested like permissions and content of response etc)

shanbady (changes requested):
works as intended. just some minor nits

shanbady (approved):
LGTM 👍

PR #2408, opened 2025-07-31
raise page size of userlists and learning paths

gumaerc:
https://github.com/mitodl/hq/issues/7972

This PR raises the page size in requests to get user list and learning path items. Previously, due to a report about not being able to see all learning path items, we set this limit to 100. A limit was not set on user lists, so that was defaulting to 10 for default pagination. The problem is, we implemented pagination in the API and the frontend queries, but never in the UI itself. This PR sets both limits to 1000, to allow for the size of most reasonable lists to work for now, until we can actually implement pagination on the frontend.

image
image

### How can this be tested?
- Log in as a superuser and create a UserList
- Add more than 10 items to the UserList
- Visit the My Lists section of the Dashboard and verify that all items are shown
- Test the reorder functionality, dragging items around
- Repeat the process for a Learning Path


jonkafton (approved):
👍

PR #2410, opened 2025-08-01
add mitxonline certificate api

gumaerc:
TBD

This PR sets up the `mitxonline` API integration to start using the v2 endpoints, and adds the course and program certificate endpoints

### How can this be tested?
TBD


jonkafton (approved):
Looks good and thanks @gumaerc for looking into the issue with signature images on the mitxonline django (do we need to add some tolerance there?). Just the one change request.

PR #2411, opened 2025-08-01
Add new delivery method "other"

shanbady:

Closes https://github.com/mitodl/hq/issues/7962

This pr adds a new "Other" delivery method we get from xpro. Other appears to indicate a combination of the other delivery methods.

### How can this be tested?
1. checkout main
2. running the mitx etl should give you a key error:
```
from learning_resources.tasks import get_xpro_data
get_xpro_data.run()
```
3. checkout this branch, run migrations, and re-run the command above and it should succeed.


shanbady:
closing since we are switching to changing it to "hybrid" in xpro as per discussion https://github.com/mitodl/hq/issues/7962#issuecomment-3141281745

cachob:
Course has been updated to format Hybrid

PR #2418, opened 2025-08-05
config: Add OTEL setup for NextJS

blarghmatey:
N/A

Adds OTEL integration for the NextJS app so that we can capture tracing data for it in Grafana.

### How can this be tested?

Run the app and verify that it starts generating trace spans

### Additional Context

We have OTel tracing enabled in the Django layer but not yet in the NextJS. Adding it to the frontend will help provide more end-to-end context of requests throughout our application layers.

### DIFF
--- a/frontends/main/src/instrumentation.ts
+++ b/frontends/main/src/instrumentation.ts
@@ -11,7 +11,7 @@ export async function register() {
     await import("../sentry.edge.config")
   }
 
-  registerOTel({ serviceName: "mitlearn-frontend" })
+  registerOTel({ serviceName: process.env.OTEL_SERVICE_NAME || "mitlearn-frontend" })
 }
 
 export const onRequestError = Sentry.captureRequestError


ChristopherChudzicki:
@blarghmatey I tried to get this to work locally today. There seems to be some interaction between Sentry and `@vercel/otel` going on. After failing to get data into into [otel-lgtm](https://github.com/grafana/docker-otel-lgtm) locally, I tried simply using `ConsoleSpanExporter`:

```ts
export async function register() {
  if (process.env.NEXT_RUNTIME === "nodejs") {
    await import("../sentry.server.config")
  }

  const serviceName = process.env.OTEL_SERVICE_NAME || "mitlearn-frontend"
  registerOTel({ serviceName, traceExporter: new ConsoleSpanExporter() })
}
```

This logs nothing. But when the sentry code is removed, it starts behaving as expected. I haven't looked into the conflict too much, but did notice these two issues on Sentry's github:
- https://github.com/getsentry/sentry-javascript/issues/11046
- https://github.com/getsentry/sentry-javascript/issues/17273

PR #2426, opened 2025-08-08
ContentFile vector search grouping

shanbady:

Closes  https://github.com/mitodl/hq/issues/6491

This PR adds a "group_by" parameter to the vector contentfiles endpoint which uses qdrant's [groupby api](https://api.qdrant.tech/api-reference/search/query-points-groups) to group chunks by a particular payload field. The intended end use of this is for "search in documents" functionality that executes a vector search and returns results where each row corresponds to a single document.

### How can this be tested?
1. checkout this branch
2. ensure you have contentfiles locally and have also generated embeddings. if not, run ```python manage.py generate_embeddings --all```
3. visit the contentfile [vector endpoint](http://open.odl.local:8063/api/v0/vector_content_files_search/?q=testing) with some query (q= parameter). 
4. now group the results by document by adding group_by=key - also add group_size=10 (to make it easier to see the grouping in action). the group_size determines the max number of matches per group (1 is default).  http://open.odl.local:8063/api/v0/vector_content_files_search/?q=testing&group_by=key&group_size=5
5. note that the results now have "chunk_content" which is a concatenation of all the matching chunks and a "chunks" field which is an array of individual matches:
```
{
...
"chunk_content": 'chunk 1 chunk 2 chunk 3 chunk 4 chunk 5',
"chunks":['chunk 1', 'chunk 2', chunk 3', chunk 4', 'chunk 5']
...
}
```

### Additional Context
1. Grouping only works when a query (q= param) is specified since it is not compatable with the qdrant scroll api
2. when grouping, we lose pagination (offset param). We can however specify the number of documents to be returned using the limit= param.

### DIFF
--- a/frontends/api/src/generated/v0/api.ts
+++ b/frontends/api/src/generated/v0/api.ts
@@ -11020,6 +11020,8 @@ export const VectorContentFilesSearchApiAxiosParamCreator = function (
      * @param {Array<string>} [course_number] Course number of the content file
      * @param {Array<string>} [edx_module_id] The edx_module_id of the content file
      * @param {Array<string>} [file_extension] The extension of the content file.
+     * @param {string} [group_by] The attribute to group results by
+     * @param {number} [group_size] The number of chunks in each group. Only relevant when group_by is used
      * @param {Array<string>} [key] The filename of the content file
      * @param {number} [limit] Number of results to return per page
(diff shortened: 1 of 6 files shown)


abeglova (approved):
LGTM - as discussed chunk_content is confusing since it concatenates unrelated chunks that are out of order and should just be null when there is a group by

PR #2434, opened 2025-08-13
Update the social media image

jonkafton:

- Closes https://github.com/mitodl/hq/issues/8086

Replaces the social media image

### How can this be tested?

Load http://open.odl.local:8062/images/learn-og-image.jpg and confirm you see the new image (MIT logo top right as opposed to https://learn.mit.edu/images/learn-og-image.jpg).

We're also now setting the known width and height. You should see the following in the HTML source:

```
<meta property="og:image:height" content="511">
<meta property="og:image:width" content="967">
```

### Additional Context

I considered providing a new image filename/URL; from gpt-4.1:

"Is it better to replace the image at the same URL?"

Yes, in most cases, it is better to replace the image at the same URL if you want new shares to immediately use the new image and to avoid confusion or broken links.

Here’s why:

- Social platforms cache the image at the OG image URL. If you keep the URL the same and just update the image file, new shares will eventually show the new image once the cache expires or is manually refreshed.
- If you change the URL, platforms will treat it as a new image and fetch it immediately for new shares, but old shares will still show the old cached image (or the old URL, which may break if deleted).


frontends/main/src/common/metadata.ts
```diff
@@ -98,8 +98,8 @@ export const standardizeMetadata = ({
           images: [
             {
               url: image,
-              width: image === DEFAULT_OG_IMAGE ? "" : 967,
-              height: image === DEFAULT_OG_IMAGE ? "" : 511,
+              width: image === DEFAULT_OG_IMAGE ? 967 : "",
+              height: image === DEFAULT_OG_IMAGE ? 511 : "",
```
jonkafton, line 102: Switched as the intention is to provide the dimensions for our known image.

PR #2457, opened 2025-08-25
Add (citation) urls for canvas contentfiles

shanbady:

Closes https://github.com/mitodl/hq/issues/8124

This PR adds urls for canvas contentfiles so that citations can be displayed in ai responses.

### How can this be tested?
1. Checkout this branch
2. restart celery
3. ingest a canvas course via ```python manage.py backpopulate_canvas_courses --canvas-id 28839 --overwrite```
4. once complete, note that the course has contentfiles with urls populated for resources:
```python
list(ContentFile.objects.filter(run__learning_resource__readable_id__istartswith='28839').exclude(url=None).values_list("url",flat=True))
```

### Additional Context
- It seems that currently (for canvas courses so far) only files that appear under web_resources have urls in the url config


mbertrand (approved):
👍

PR #2458, opened 2025-08-25
Minor tweaks to initial setup instructions, add memory env var for web container, note minimum required domain config for things to "work out of box"

dsubak:
Includes a few small changes to initial setup docs, as well as one or two QoL changes.
- Attempts to clarify that running keycloak and apisix is the "default" supported config for authentication.
- Adds some information about docker resource limits and how to check for OOMKills
- Enumerate the `/etc/hosts` entries required if you're relying on the example env values.
- Adjusts a few default and example values for memory and env vars, adds a new env var for setting web container mem_limit.

### How can this be tested?
Majority of this is documentation, but the memory changes can be observed by running `docker compose up` and observing that the upper limits are reflective of the increased thresholds in the Docker Desktop container stats tab:
Screenshot 2025-08-25 at 3 30 32 PM

### Additional Context
Additional context, clarifying questions, ongoing discussion about issues encountered can be found [here](https://mitodl.slack.com/archives/C03K5HYGPT9/p1755872165330549)

### DIFF
--- a/env/backend.local.example.env
+++ b/env/backend.local.example.env
@@ -21,7 +21,7 @@ MAILGUN_SENDER_DOMAIN=open.odl.local
 MAILGUN_KEY=fake
 
 # Set this to False if apisix/keycloak are running locally
-DISABLE_APISIX_USER_MIDDLEWARE=True
+DISABLE_APISIX_USER_MIDDLEWARE=False
 
 # APISIX/Keycloak settings
 APISIX_LOGOUT_URL=http://api.open.odl.local:8065/logout/
--- a/env/backend.local.example.env
+++ b/env/backend.local.example.env
@@ -20,9 +20,6 @@
 MAILGUN_SENDER_DOMAIN=open.odl.local
 MAILGUN_KEY=fake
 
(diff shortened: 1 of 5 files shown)


env/backend.local.example.env
```diff
@@ -21,7 +21,7 @@ MAILGUN_SENDER_DOMAIN=open.odl.local
 MAILGUN_KEY=fake
 
 # Set this to False if apisix/keycloak are running locally
-DISABLE_APISIX_USER_MIDDLEWARE=True
+DISABLE_APISIX_USER_MIDDLEWARE=False
```
abeglova, line 24: you can probably just remove it from example.env since DISABLE_APISIX_USER_MIDDLEWARE=False is the default and also the normal way to run the app now. Also (definitely not in this pr) we can consider a followup to remove the variable entirely, since this var was added for local development and now apisix/keycloak is sufficiently integrated into the app that local development with users that are authenticated without keycloak doesn't really work anymore

abeglova (approved):
lgtm

PR #2460, opened 2025-08-26
UAI Landing Page

gumaerc:
Closes https://github.com/mitodl/hq/issues/8248

This PR adds a landing page for the UAI launch as described in the above issue. A Hubspot form is being injected into the page via a script they provided. The components are being styled in a hacky way, rendering invisible placeholder components for each UI control the form needs to use, then hijacking their styles and applying them wholesale to the controls inside the Hubspot container. Hubspot forms in general should not be inserted this way into a production website, but due to the expediency surrounding getting this page up, this method was used. The way this should be done is by fetching the form data using Hubspot's forms API:

https://developers.hubspot.com/docs/reference/api/marketing/forms/v3

This API unfortunately requires OAuth2 authentication, and can only be done by a backend service. In order to do this correctly, it would probably be best to implement usage of the forms API into https://github.com/mitodl/ol-django/tree/main/src/hubspot_api. This would allow us to add it to any of our Django applications and implement an API endpoint to fetch Hubspot forms and return the data to the frontend. From there we could build a UI control, something like `HubspotForm` that would build out the UI using our customized MUI components.

image
image
image
image
image
image
image
image
image

### How can this be tested?
 - Review the page at `/universal-learning/ai` and ensure that the design has been faithfully implemented
 - Ensure that the "Learn More" CTA's successfully put the Hubspot form into view
 - Ensure that you can submit the Hubspot form


frontends/main/src/app/universal-learning/ai/page.tsx
```diff
@@ -0,0 +1,14 @@
+import React from "react"
+import { Metadata } from "next"
+import { UAILandingPage } from "@/app-pages/UAILandingPage/UAILandingPage"
+import { standardizeMetadata } from "@/common/metadata"
+
+export const metadata: Metadata = standardizeMetadata({
+  title: "UAI Landing",
```
shanbady, line 7: minor nit and not even sure if it matters but maybe changing this to something more descriptive since it shows up in browser tabs etc. maybe "Universal Learning"etc

shanbady (approved):
it does what it needs to. left a minor nit as a comment. also noticed some stuff in console that might be worth cleaning up but doesnt seem to affect functionality

Image

PR #2471, opened 2025-08-27
Ingest non-module canvas files

shanbady:

Closes https://github.com/mitodl/hq/issues/8275

This PR makes sure that when we process canvas archives, we also pull in files that are published and not explicitly part of a module.

### How can this be tested?
1. checkout main
2. restart celery and grab a specific canvas archive ``` python manage.py backpopulate_canvas_courses --canvas-ids 28755```
3. once complete, check the count of contentfiles ```LearningResource.objects.get(readable_id__istartswith="28755").runs.first().content_files.count()```
4. get the source_paths of the contentfiles ```LearningResource.objects.get(readable_id__istartswith="28755").runs.first().content_files.values_list("source_path", flat=True)```
5. use this helper script to manually download the archive:
```python
import boto3
s3 = boto3.resource(
            "s3",
            aws_access_key_id=settings.AWS_ACCESS_KEY_ID,
            aws_secret_access_key=settings.AWS_SECRET_ACCESS_KEY,
)
bucket.download_file('canvas/course_content/28755/55ae6687662c6eb38e5ced60437d74f4cb63daa1630a2f755a1955f2736a1097.imscc', Path("canvas_archive.zip"))
```
6. unzp the archive and note that files in course_settings/files_meta.xml dont show up as contentfiles
7. switch to this branch and repeat. note that the files from files_meta now exist.


learning_resources/etl/canvas.py
```diff
@@ -58,11 +59,14 @@ def sync_canvas_archive(bucket, key: str, overwrite):
         )
         checksum = calc_checksum(course_archive_path)
         if run:
-            load_content_files(
-                run,
+            canvas_content_files = list(
```
mbertrand, line 62: This seems to defeat the purpose of `transform_canvas_content_files` yielding results one at a time instead of returning everything at once, and could potentially cause memory issues for archives containing lots of contentfiles.  My first reaction was to revert it, but I get different counts with this code (139) vs as it was (82).  Not sure why, any ideas?

shanbady: i stumbled on the same quirk. I initially thought it had to do with the file handlers closing prematurely but that doesnt seem to be the case. Ill see if i can suss it out.

shanbady: fixed

mbertrand (approved):
👍

PR #2477, opened 2025-09-02
Fix flaky test

mbertrand:
Closes https://github.com/mitodl/hq/issues/8291

I believe the test was occasionally failing because 1 or both user ids generated by the `UserListFactory` happened to match the hardcoded "fake" ids used for the test. 

Changed the fake ids to negative integers so they will always fail.  Also incremented the fake resource id by 1 mroe than the real resource id so that should always fail.

### How can this be tested?
Run the test a few times, it should always fail.


shanbady (approved):
👍

PR #2478, opened 2025-09-02
remove unpublished resources with duplicate readable_ids

shanbady:

Closes  https://github.com/mitodl/hq/issues/8303

This PR adds a task that runs once a day to remove unpublished courses that occasionally show up with duplicate ids. 

### How can this be tested?
1. checkout this branch
2. restart celery
3. find some course and grab the readable id
4. create some unpublished learning resource objects with that same readable id:
```python
for i in range(3):
    LearningResource.objects.create(readable_id=<the id>, published=False)
```
5. run the task that removes the duplicate resources:
```python
from learning_resources.tasks import remove_duplicate_resources
remove_duplicate_resources.apply()
```
6. verify the duplicate resources have been removed. The published instance of that course should still be there

### Additional Context
This is just a simple and temporary stop-gap until we have the more permanent (and involved) solution in place which would involve factoring in the platform as part of the ids we generate in qdrant

### DIFF
--- a/learning_resources/tasks_test.py
+++ b/learning_resources/tasks_test.py
@@ -679,7 +679,7 @@ def test_sync_canvas_courses(settings, mocker, django_assert_num_queries, canvas
         assert mock_ingest_course.call_count == 2
 
 
-def test_remove_duplicate_resources(mocker):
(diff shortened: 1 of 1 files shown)


mbertrand (approved):
👍

PR #2481, opened 2025-09-03
Qdrant collection and parameter updates

shanbady:

Closes https://github.com/mitodl/hq/issues/8164

Over time we have tweaked many parameters on qdrant to make it performant in production. Those changes havent made their way back to the code which is what this PR does.

The updates:
- enable binary quantization for both collections (as suggested by qdrant if we are using openai text-embedding large)
- enable strict mode (so we are alerted when we have unindexed fields or other issues causing degraded performance) 
- remove the unused multi-vector configuration
- setup a replication factor of >1 (and shards to some factor of 2) so we can have zero downtime upgrades
- Add missing payload fields to schema
- make it so that when new payload fields are added, they are automatically added to the index

#### Verify Collection Params
1. checkout this branch
2. regenerate embeddings ```python manage.py generate_embeddings --all --recreate-collections```
3. visit the info page for the [contentfile](http://localhost:6333/dashboard#/collections/resource_embeddings.content_files#info) and [resource](http://localhost:6333/dashboard#/collections/resource_embeddings.resources#info) collection
4. verify that the fields that show up in the "payload_schema" match the payloads defined in vector_search/constants.py (QDRANT_LEARNING_RESOURCE_INDEXES and QDRANT_CONTENT_FILE_INDEXES)
5. verify that in the "config" section, the replication factor is 2 and and shard number is 6
6. verify that quantization is "binary" and that "on_disk_payload" on the contentfiles collection is true.  

#### Verify Updates to Payload Schema
1. delete some existing indexed payload fields from a collection via:
```python
from vector_search.utils import qdrant_client
from vector_search.constants import RESOURCES_COLLECTION_NAME, CONTENT_FILES_COLLECTION_NAME
client = qdrant_client()
client.delete_payload_index(collection_name=CONTENT_FILES_COLLECTION_NAME, field_name="course_number")
client.delete_payload_index(collection_name=CONTENT_FILES_COLLECTION_NAME, field_name="course_feature")
```
2. go into the qdrant dashboard and verify the deleted fields do not show up in the payload index
3. re-run generate_embeddings (without recreating the collection or overwriting) ```python manage.py generate_embeddings --all```
4. go back into the dashboard and see that the payload fields have re-appeared.

### Checklist:
- [ ] collection/parameter changes are valid
- [ ] everything works normally


abeglova (approved):
lgtm

PR #2485, opened 2025-09-04
Fix Sloan news/events ETL pipeline

mbertrand:
Blocked, waiting on Sloan to provide a proper feed source so that HTML scraping won't be necessary

Closes https://github.com/mitodl/hq/issues/8370

- Replaces the separate news and events ETL pipelines for Sloan with a single pipeline that retrieves both.
- The pipeline is based on the new HTML structure which has significantly changed.  The old URL that returned JSON is no longer accessible.

### How can this be tested?
Run `./manage.py backpopulate_news_events` - all the relevant ETL pipelines should run and complete successfully, including for Sloan.

Depending on when you run this, it is likely that at least 1 Sloan news article will appear at the bottom of the home page.

You can also go to http://open.odl.local:8065/admin/news_events/feeditem/ and check that there are items for Sloan news and events

I initially considered keeping separate pipelines for news and events, and parsing https://executive.mit.edu/webinars.html for events because it contains a carousel of future webinars.  But no date is specified on that page, only on the linked in page for the event.  And that in turn only displays when logged in to LinkedIn.  So only webinars that occurred in the past and are found at https://executive.mit.edu/insights will be ingested.

### DIFF
--- a/news_events/etl/ol_events.py
+++ b/news_events/etl/ol_events.py
@@ -185,6 +185,7 @@ def transform_event(event_data: dict) -> dict or None:
                 extract_relationship(event_data, "field_event_category")
             ),
             "event_datetime": dt_utc,
+            "event_end_datetime": dt_utc,
         },
     }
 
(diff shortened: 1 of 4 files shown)


news_events/etl/ol_events.py
```diff
@@ -185,6 +185,7 @@ def transform_event(event_data: dict) -> dict or None:
                 extract_relationship(event_data, "field_event_category")
             ),
             "event_datetime": dt_utc,
+            "event_end_datetime": dt_utc,
```
mbertrand, line 188: Added to prevent OLL events from showing up after the datetime has passed.

PR #2488, opened 2025-09-04
track first login on profile

gumaerc:
Closes https://github.com/mitodl/hq/issues/8386

This PR adds a boolean field to the `Profile` model called `has_logged_in`. This is `False` by default, and set `True` after the first login.

### How can this be tested?
- Spin up `mit-learn` on this branch, making sure to run migrations if you are not restarting your containers
- Log in as a superuser
- Navigate to Django admin, to the Profiles section
- Find the profile for your logged in user and verify that `has_logged_in` is true
- Create a new superuser account using `createsuperuser`
- In your browser, refresh Django admin and find the account you just created
- Verify that `has_logged_in` is false for that user
- Log in with the new user you created, navigate back to Django admin and verify that `has_logged_in` is now true


ChristopherChudzicki (approved):
👍

PR #2501, opened 2025-09-09
Fix userlist stale data

ChristopherChudzicki:
Fixes https://github.com/mitodl/hq/issues/8465

Resolves two separate issues with userlist data being outdated on the frontend.

### How can this be tested?
Test two things:
1. Visit one of your lists, at `/dashboard/my-lists/{list-id}`. It should have some items.
2. **First thing:** Reorder the items in the list and click "Done". After clicking "Done", the order should be correctly updated.
3. **Second Thing:** On the `/dashboard/my-lists/{list-id}` page, try removing an item from a list. The UI should update correctly. Before it would not.

### DIFF
--- a/frontends/main/src/app/dashboard/my-lists/[id]/page.tsx
+++ b/frontends/main/src/app/dashboard/my-lists/[id]/page.tsx
@@ -8,7 +8,7 @@ const Page: React.FC<PageParams<never, { id: string }>> = async ({
 }) => {
   const resolved = await params!
   const id = Number(resolved.id)
-  invariant(!id, "id is required")
+  invariant(id, "id is required")
   return <UserListDetailsContent userListId={id} />
 }
 
--- a/frontends/main/src/app/dashboard/my-lists/[id]/page.tsx
+++ b/frontends/main/src/app/dashboard/my-lists/[id]/page.tsx
@@ -1,14 +1,15 @@
 import React from "react"
 import { UserListDetailsContent } from "@/app-pages/DashboardPage/UserListDetailsContent"
-import { PageParams } from "@/app/types"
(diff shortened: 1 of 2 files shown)


frontends/main/src/app/dashboard/my-lists/[id]/page.tsx
```diff
@@ -1,14 +1,15 @@
...
+const Page: React.FC<PageParams<never, { id: string }>> = async ({
   params,
 }) => {
-  const resolved = await params
-  invariant(resolved?.id, "id is required")
-  return <UserListDetailsContent userListId={resolved.id} />
+  const resolved = await params!
+  const id = Number(resolved.id)
+  invariant(!id, "id is required")
+  return <UserListDetailsContent userListId={id} />
```
Comment (ChristopherChudzicki) [inline frontends/main/src/app/dashboard/my-lists/[id]/page.tsx:12]: What was happening was... the list data was being fetched based on the URL parameter, which was a string. But when we invalidated the query cache, we were using a number. So:

```ts
// query stored under key
["userlist", "detail", "16", "items", listingParams]
// invalidated query used a number for the id:
["userlist", "detail", 16]
```
So we weren't invalidating the correct query.

---------

Manual typings are error prone... Putting `{ id: number }` as the page parameter types was reasonable, but it really shouldve been `{ id: string }`, in which case Typescript wouldve caught the issue.

NextJS 15.5 largely removes the need for manual typing of page params. See #2500

jonkafton (approved):
Looks good :+1:

While waiting for a slow API response locally (I have the mitxonline project also running in Docker and fans spinning) - it would be worth us looking into optimistic updates, at least on some mutations. There's plenty of code overhead to snapshot existing values, but we have well organized query keys - there should be some reuse patterns.

PR #2502, opened 2025-09-09
ingest canvas html content

shanbady:

Closes https://github.com/mitodl/hq/issues/8436

This PR makes it so:
- we include published html/wiki content and assignment pages as part of ingesting a canvas course.
- explicitly filter out contents of the tutorbot folder even if it is somehow included in the files_meta.xml (out of abundance of caution)

### How can this be tested?
1. checkout main
2. restart celery
3. ingest a canvas course: `python manage.py backpopulate_canvas_courses --canvas-ids 34062 --overwrite`
4. view the list of files that were ingested: ` LearningResource.objects.filter(readable_id__istartswith="34062").first().runs.first().content_files.all().values_list("source_path", flat=True)`
5. note that it is empty for this particular course.
6. do the same for another with published files:
 `python manage.py backpopulate_canvas_courses --canvas-ids 34062 --overwrite`
```python
LearningResource.objects.filter(readable_id__istartswith="14566").first().runs.first().content_files.all().values_list("source_path", flat=True)
```
7. note the files and filetypes
8. checkout this branch
9. restart celery
10. re-run the backpopulate commmands for each
11. note that published html content exists for both courses

### Additional Context
- This PR also includes an extra check to filter out web_resources/ai/tutor just in case it were to inadvertently  end up in files manifest.
- The files that have assignment_settings.xml also have an extra check to see if it is only visible to instructors

### DIFF
--- a/learning_resources/etl/canvas.py
+++ b/learning_resources/etl/canvas.py
@@ -314,7 +314,6 @@ def is_file_published(file_meta: dict) -> bool:
         bool: True if file is published/visible, False otherwise.
     """
     now = now_in_utc()
-
     hidden = str(file_meta.get("hidden", "false")).lower() == "true"
     locked = str(file_meta.get("locked", "false")).lower() == "true"
     unlock_at = file_meta.get("unlock_at")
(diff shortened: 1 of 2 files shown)


mbertrand (approved):
LGTM 👍

PR #2518, opened 2025-09-17
Fix posthog import

abeglova:
https://github.com/mitodl/hq/issues/8319

While testing https://github.com/mitodl/mit-learn/pull/2489 i  noticed
1) An error in this line https://github.com/mitodl/mit-learn/blob/main/learning_resources/etl/posthog.py#L96C13-L96C66 because `platform` is None for learning paths
2) Lots of noisy warnings because the export files contain posthog traces for llm calls

This pr fixes those issues 

### How can this be tested?
set 
POSTHOG_EVENT_S3_BUCKET=ol-data-lake-landing-zone-qa
POSTHOG_EVENT_S3_PREFIX=thirdparty/posthog/learn/events/
and AWS keys from rc

from the shell run
```
from learning_resources.tasks import get_learning_resource_views
get_learning_resource_views()
```
you should get some messages such as
`WARNING: skipping event for resource ID 17235 - resource not found`
since the resource ids in your local database don't match rc.

You should not see any messages saying
`WARNING: skipping event for resource ID None - resource not found`
and you should not see errors

### DIFF
--- a/learning_resources/etl/posthog.py
+++ b/learning_resources/etl/posthog.py
@@ -26,10 +26,7 @@ class PostHogLearningResourceViewEvent:
     the lrd_view specific properties.
     """
 
-    resourceType: str  # noqa: N815
-    platformCode: str  # noqa: N815
-    resourceId: int  # noqa: N815
-    readableId: str  # noqa: N815
+    resource_id: int
     event_date: datetime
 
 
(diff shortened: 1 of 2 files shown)


learning_resources/etl/posthog.py
```diff
@@ -26,10 +26,7 @@ class PostHogLearningResourceViewEvent:
     the lrd_view specific properties.
     """
 
-    resourceType: str  # noqa: N815
-    platformCode: str  # noqa: N815
-    resourceId: int  # noqa: N815
-    readableId: str  # noqa: N815
```
abeglova, line 32: Most of these fields are not used

PR #2519, opened 2025-09-17
Fix flaky test

mbertrand:
Closes https://github.com/mitodl/hq/issues/8583

Fixes the intermittent test failures

### How can this be tested?
- On docker container, `pip install pytest-repeat`
- Run `pytest learning_resources/tasks_test.py::test_get_content_tasks --count 100 -x`, test should pass every time.

### DIFF
--- a/learning_resources/tasks.py
+++ b/learning_resources/tasks.py
@@ -207,9 +207,13 @@ def get_content_tasks(  # noqa: PLR0913
     )
 
     if learning_resource_ids:
-        learning_resources = LearningResource.objects.filter(
-            id__in=learning_resource_ids, etl_source=etl_source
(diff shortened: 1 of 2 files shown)


learning_resources/tasks.py
```diff
@@ -209,7 +209,7 @@ def get_content_tasks(  # noqa: PLR0913
     if learning_resource_ids:
         learning_resources = LearningResource.objects.filter(
             id__in=learning_resource_ids, etl_source=etl_source
-        ).values_list("id", flat=True)
+        ).order_by("-id").values_list("id", flat=True)
```
mbertrand: Since `else` block below is ordered by `-id`, I decided to do it here too for consistency.

shanbady (approved):
LGTM 👍

PR #2521, opened 2025-09-18
Shanbady/canvas citation html pages

shanbady:

Closes https://github.com/mitodl/hq/issues/8490

This PR is responsible for adding citation urls for html content ingested from canvas archives (modules, pages, assignments etc)

### How can this be tested?
1. checkout main.
2. ingest a canvas course
3. view urls associated with html files from that course
```

list(LearningResource.objects.filter(readable_id__istartswith="34716").first().runs.first().content_files.filter(file_extension=".html").values_list("url","source_path"))
```
4. checkout this branch
5. restart celery
6. repeat steps above and see that we have urls for html content

### Additional Context
Due to limitations with what the canvas api gives us, we are only able to match up content in the archive to items in the url metadata file based on the content's title - canvas does allow you to name things the same which can cause issues. We are assuming that whoever puts the course together will have uniquely named modules, pages etc.


abeglova (approved):
lgtm

PR #2524, opened 2025-09-19
Certificate social media / URL sharing

jonkafton:

Closes https://github.com/mitodl/hq/issues/8616

Adds a share button and popover from the designs in [Figma](https://www.figma.com/proto/Eux3guSenAFVvNHGi1Y9Wm/MIT-Design-System?page-id=107%3A2479&node-id=21743-65112).

Adds Open Graph metadata for social media previews:
- Title: "Joe Learner's Module Certificate | MIT Learn"
- Description: "Joe Learner has successfully completed the Universal Artificial Intelligence Module Certificate: The Course Name"

image

image

### How can this be tested?

Hopefully you have mitxonline running locally with a certificate set up (otherwise please message me for a run through).

- Load the certificate page at http://open.odl.local:8062/certificate/{"course"|"program"}/{uuid}
- Hit the "Share" button.
- Check that the popover matches the designs in [Figma](https://www.figma.com/design/Eux3guSenAFVvNHGi1Y9Wm/MIT-Design-System?node-id=22738-68471&m=dev).
- Check that the mobile layout looks reasonable (we don't have a mobile design cc. @steven-hatch).
- Check that the popover closes when clicked outside.
- Check that the textbox selects url content when clicked and is copyable.
- Check that the "Copy Link" button updates to "Copied!"
  - `navigator.clipboard` is only available in secure contexts, not HTTP, so locally this fails silently.
- Check that the social media buttons link to the respective share URLs.
- Check the Open Graph metadata is correctly set in the header (og:title and og:description).
  - The social media previews won't work locally (unless you have some tunnelling for your localhost to be reachable).
  - I used this extension to emulate the preview: https://socialsharepreview.com/browser-extensions

image


frontends/main/src/app/certificate/[certificateType]/[uuid]/page.tsx
```diff
@@ -1,16 +1,69 @@
...
+      certificateType,
+      uuid,
+      error,
+    })
+    return standardizeMetadata({})
+  }
+
+  return standardizeMetadata({
+    title: `${userName}'s ${displayType}`,
+    description: `${userName} has successfully completed the Universal Artificial Intelligence ${displayType}: ${title}`,
```
Comment (pdpinch) [inline frontends/main/src/app/certificate/[certificateType]/[uuid]/page.tsx:63]: Can you give me an example of how this renders? 

We have to be careful about hard-coding "Universal Artificial Intelligence" because this won't stay forever. In a case like this, which isn't very visible, I would just drop "Universal Artificial Intelligence" altogether for better future-compatibility. 

```suggestion
    description: `${userName} has successfully completed the  ${displayType}: ${title}`,
```

Comment (jonkafton) [reply frontends/main/src/app/certificate/[certificateType]/[uuid]/page.tsx:63]: That currently looks like: "Joe Learner has successfully completed the Universal Artificial Intelligence Module Certificate: The Course Name".

With your suggested change: "Joe Learner has successfully completed the Module Certificate: The Course Name".

`displayType` is either "Module Certificate", "Series Certificate" or "MicroMasters® Certificate".

We do have "Universal Artificial Intelligence" hardcoded on the certificate text itself ("has successfully completed all requirements of the Universal Artificial Intelligence module").

ChristopherChudzicki (changes requested):
Still need aria-labels on the links, plus some comments

ChristopherChudzicki (approved):

PR #2554, opened 2025-09-30
Update MITxOnline API Client

ChristopherChudzicki:
Toward https://github.com/mitodl/hq/issues/8715

This PR updates the MITxOnline API client and fixes an issue with URL query parameter sorting in our network request mocking utilities. 

### How can this be tested?
1. Tests should pass.
2. If you want, check that http://learn.odl.local:8062/courses/course-v1:MITxT+7.28.1x-fake is still there. (Replace with appropriate readable id of some mitxonline course you have locally; it will need to have `include_in_learn_catalog=true` in the wagtail CMS.)
    - You'll need to be logged into mitxonline or on branch https://github.com/mitodl/mitxonline/pull/2975
    - product pages are currently feature flagged under `product-page-course`

### DIFF
--- a/frontends/api/src/test-utils/mockAxios.ts
+++ b/frontends/api/src/test-utils/mockAxios.ts
@@ -45,7 +45,7 @@ const standardizeUrl = (url: string) => {
  * NOTE: URLs called by this function are first
  * ```
  */
-const makeRequest: RequestMaker = jest.fn(alwaysError)
+const makeRequest = jest.fn(alwaysError)
 const makeSortedRequest: RequestMaker = (method, url, body) =>
   makeRequest(method, standardizeUrl(url), body)
 


frontends/main/src/app-pages/ProductPages/CoursePage.tsx
```diff
@@ -510,62 +510,71 @@ const CoursePage: React.FC<CoursePageProps> = ({ readableId }) => {
...
-                Prerequisites
-              </Typography>
-              <RawHTML html={page.prerequisites} />
-            </PrerequisitesSection>
-            <InstructorsSection aria-labelledby={HeadingIds.Instructors}>
-              <Typography
-                variant="h4"
-                component="h2"
-                id={HeadingIds.Instructors}
+            {page.about ? (
```
ChristopherChudzicki, line 513: `page.about`, etc, are actually nullable. The old MITxOnline client incorrectly marked them as non-null.

This change, and the others in this file, only render the relevant section if the property exists.

gumaerc (approved):
:+1:

PR #2559, opened 2025-10-01
Fix featured list channel sorting

ChristopherChudzicki:
Closes https://github.com/mitodl/hq/issues/6278

Fixes some sorting we did on the frontend to handle numbers properly.

### How can this be tested?
1. Run `./manage.py populate_featured_lists` management command (I changed the default to 12, which is better for our carousels in general). **Note:** This will override your featured lists. If you want to keep them, manually add at least 12 items to one of them.
    - must use at least 12 items (or 11, actually) to test the lexicographical (string vs number) sorting issue. 
2. Visit http://learn.odl.local:8062/c/unit/ocw
    - The featured list carousel items should be in the same order as shown in API response from http://api.learn.odl.local:8065/api/v1/featured/?offered_by=ocw&limit=12 (` curl http://api.learn.odl.local:8065/api/v1/featured/\?offered_by\=ocw\&limit\=12 | jq '.results.[].title'`)
    - ⚠️ NOTE: The featured list API responses are cached by django for anonymous users, which includes the NextJS backend and your terminal if using curl above. You may need to restart the docker containers or clear the cache some other way.

### DIFF
--- a/frontends/api/src/hooks/learningResources/queries.ts
+++ b/frontends/api/src/hooks/learningResources/queries.ts
@@ -20,6 +20,7 @@ import type {
   LearningResourcesSearchResponse,
 } from "../../generated/v1"
 import { queryOptions } from "@tanstack/react-query"
+import { hasPosition, randomizeGroups } from "./util"
 
 /* List memberships were previously determined in the learningResourcesApi
  * from user_list_parents and learning_path_parents on each resource.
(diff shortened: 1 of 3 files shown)


frontends/api/src/hooks/learningResources/util.ts
```diff
@@ -0,0 +1,58 @@
...
+  results.forEach((result) => {
+    const pos = result?.position
+    if (!resultsByPosition[pos]) {
+      resultsByPosition[pos] = []
+    }
+    resultsByPosition[pos]?.push(result)
+  })
+  Object.keys(resultsByPosition)
+    .sort(
+      (a, b) => Number(a) - Number(b), // Sort positions numerically
```
ChristopherChudzicki, line 49: This is the only functional change vs the old implementation. But I decided to move it to a separate file and have copilot add some tests.

frontends/api/src/hooks/learningResources/util.test.ts
```diff
@@ -0,0 +1,97 @@
...
+    const items: Array<{ id: string; position: number }> = []
+    const result = randomizeGroups(items)
+    expect(result).toEqual([])
+  })
+})
+
+describe("hasPosition", () => {
+  it("should return true for objects with non-null position", () => {
+    const obj = { id: "test", position: 5 }
+    expect(hasPosition(obj)).toBe(true)
```
ChristopherChudzicki, line 72: copilot really wanted to add this test......

frontends/api/src/hooks/learningResources/queries.ts
```diff
@@ -37,35 +38,6 @@ export const clearListMemberships = (
...
-  } = {}
-  const randomizedResults: LearningResource[] = []
-  results.forEach((result) => {
-    if (!resultsByPosition[result?.position]) {
-      resultsByPosition[result?.position] = []
-    }
-    resultsByPosition[result?.position ?? ""]?.push(result)
-  })
-  Object.keys(resultsByPosition)
-    .sort()
```
ChristopherChudzicki, line 61: **Explanation:** Object keys are strings, and these were stringified numbers, so they were being sorted like `0, 1, 10, 11, 2, 3, 4, ...`.

This wasn't a visible problem on the homepage because that carusel uses items from every offeror... to get up to 12 items, that carousel only needs positions 0, 1, 2 from all five offerors.

The featured list query is also used on channel pages (We don't query the learning path directly because, at least right now, those are private; only visible to non-staff via `/featured` endpoint). And on the channel pages, 12 items are shown, so the sorting was getting messed up.

PR #2574, opened 2025-10-07
Cc/update jest

ChristopherChudzicki:
None

This updates jest to v30. My immediate motivation was that pathname filtering ("p" in `yarn test-watch`) was not working very well in v29. Dunno why/when that started, but this seems to fix it.

Also updated some other dev dependencies related to testing.


frontends/api/jest.config.ts
```diff
@@ -1,3 +1,4 @@
+/** @jest-config-loader esbuild-register */
```
ChristopherChudzicki, line 1: I was getting errors on node ^v22.20 that seemed related to
- https://github.com/jestjs/jest/issues/15682
- https://github.com/nodejs/typescript/issues/37

By default when using TS config files, jest uses `ts-node`, and that has some issue I don't fully understand starting in Node 22.18.

It can be worked around with `NODE_OPTIONS='--no-experimental-strip-types' yarn test`, but that didn't seem great. Supposedly the issue was fixed in Node v24.4, but trying that locally it still was still failing to parse the TS config files.

Switching from ts-node to es-build fixed the issue, and ts-node is kinda dead anyway.

Note: We actually already had esbuild-register installed via storybook, though I added it explicitly as a dev dep now.

ChristopherChudzicki:
Timeout problem showing up on CI, too, and in general Jest 30 seems slower.

Others seem to have this issue.
- https://github.com/jestjs/jest/issues/15718

Closing this for now.

PR #2589, opened 2025-10-10
MIT CLimate Portal ETL

shanbady:

Closes https://github.com/mitodl/hq/issues/8777

note to reviewer: I think* the failing openai-diff is something we can ignore -  I didn't find a way to suppress it but it should go away once this is merged in. cc @ChristopherChudzicki 

This PR adds MIT Climate Portal articles to Learn. It adds the following:
- A new "Article" resource type 
- Climate offerer and platform
- a "backpopulate_climate_articles" command to manually pull in articles
- a new celery task "get_mit_climate_data" that runs once per day (1:30am est)

### How can this be tested?
1. checkout this branch
2. restart celery
3. run the climate article backpopulate command: `python manage.py backpopulate_mit_climate_articles`
4. you should see new articles populate in search results under "learning materials"
5. you should be able to click the article and see that the resource loads in the learning resource tab and that it matches the [figma](https://www.figma.com/proto/Eux3guSenAFVvNHGi1Y9Wm/MIT-Design-System?page-id=107%3A2479&node-id=21172-64363&viewport=-986%2C1026%2C0.4&t=YxOmmK2MljGiQI24-1&scaling=min-zoom&content-scaling=fixed&starting-point-node-id=3852%3A51162&show-proto-sidebar=1)
6. The topics for the articles should be mapped according to the [spreadsheet](https://docs.google.com/spreadsheets/d/1mjF8LbfYtHg6uw1EqRyqMjP_2KYZmRNm7-zR-F6CVVM/edit?gid=0#gid=0)

### Additional Notes
There is an open-api error which after looking into - it wants us to move the addition of the Article resource type into a new version of the api which is not what i think we want to do here so I think* it can be ignored

### DIFF
--- a/learning_resources/etl/loaders.py
+++ b/learning_resources/etl/loaders.py
@@ -1145,6 +1145,14 @@ def load_articles(articles_data: iter) -> list[LearningResource]:
             )
         else:
             article_resources.append(article_resource)
+    unpublished_articles = Article.objects.exclude(
+        learning_resource__id__in=[resource.id for resource in article_resources]
+    )
+    # remove articles that no longer exist
+    bulk_resources_unpublished_actions(
+        unpublished_articles.values_list("learning_resource__id", flat=True),
+        LearningResourceType.article.name,
+    )
     return article_resources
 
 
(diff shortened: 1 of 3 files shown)


shanbady:
@mbertrand I resolved the article embedding generation (also similar resources issue) in addition to unpublishing articles that don't exist. 



It seems the filtering via facet and in the url via query string are parts of course-search-utils which should get updated with the new specs once this gets releases. I'm thinking once we merge this we should:
- deploy it and hold off on running the management command to populate articles - This will ensure the article option is hidden in the interface.
- then make update course-search-utils to new version
- redeploy and run the backpopulate

PR #2594, opened 2025-10-15
add mitxonline networking directions

ChristopherChudzicki:
All the PRs we make where we copy-and-paste the setup directions for mitxonline <> Learn.

This PR adds directions for running MITxOnline and MIT Learn locally in a way that they can communicate.

### How can this be tested?
1. Follow the directions in the README
2. Sanity checks:
    - http://api.open.odl.local:8065/mitxonline/api/v2/courses/ should return same as  http://mitxonline.open.odl.local:8013/mitxonline/api/v2/courses/
    - **After authenticating on Learn via frontend login** http://api.open.odl.local:8065/mitxonline/ap/v0/users/me should show your user info (note: no trailing slash!).
3. Create a program with some required and elective courses. Useful things to know:
    - Create courses and programs with `./manage.py create_courseware`, `./manage.py create_courseware_page`; see `--help` for details. Note: program readable ids should start with `program-v1:`.
    - Add instructors/relevant data for your program in Wagtail at http://mitxonline.odl.local:8013/cms
4. Run MITxOnline ETL against your local MITxOnline; `./manage.py backpopulate_mitxonline_data --delete`, and `./manage.py backpopulate_mitxonline_data`
5. Ensure `product-page-course` feature flag is on in your personal posthog
6. Try backpopulating mitxonline data against local, `./manage.py backpopulate_mitxonline_data`
7. Try Viewing a program page or course page, e.g., http://open.odl.local:8062/courses/COURSE_ID_HERE or http://open.odl.local:8062/programs/PROGRAM_ID_HERE

### DIFF
--- a/README.md
+++ b/README.md
@@ -288,11 +288,13 @@ KEYCLOAK_DISCOVERY_URL=http://kc.ol.local:8066/realms/ol-local/.well-known/openi
 KEYCLOAK_REALM_NAME=ol-local
 ```
 
-Then, in Learn, set:
+Then, in Learn, set these values:
 
 ```env
+# MIT Learn, shared.local.env (ensure not overwritten in backend/frontend env)
+MITX_ONLINE_BASE_URL=http://api.open.odl.local:8065/mitxonline
+
 # MIT Learn, backend.local.env
-MITX_ONLINE_BASE_URL=http://mitxonline.odl.local:8013/
 MITX_ONLINE_COURSES_API_URL=http://mitxonline.odl.local:8013/api/v2/courses/
 MITX_ONLINE_PROGRAMS_API_URL=http://mitxonline.odl.local:8013/api/v2/programs/
 ```


docker-compose.services.yml
```diff
@@ -1,6 +1,13 @@
 include:
   - docker-compose.opensearch.${OPENSEARCH_CLUSTER_TYPE:-single-node}.yml
 
+# Define the extra_hosts as a reusable anchor
+x-host-access: &host-access
+  extra_hosts:
+    # This allows containers to access services running on a host linux machine
+    # Unnecessary but harmless for mac/windows users
+    - "host.docker.internal:host-gateway"
```
ChristopherChudzicki: Lacking a linux machine, i have not actually tested that this enables it, but Copilot and SO https://stackoverflow.com/questions/70725881/what-is-the-equivalent-of-add-host-host-docker-internalhost-gateway-in-a-comp suggest it should

README.md
```diff
@@ -256,6 +258,51 @@ Personal API keys only need read permission to Query. When creating a personal A
 
 Once these are set (and you've restarted the app), you should see events flowing into the PostHog dashboard.
 
+### Connecting with MITxOnline
+
+Set up the [mitxonline](https://github.com/mitodl/mitxonline) as indicated there. _Note: A working OpenEdx installation is not necessary for integration with Learn._
```
mbertrand: The MITx Online README should probably indicate that keycloak and apisix are required for the login button to work correctly, otherwise the `/admin` page can be used to log in.  Maybe mention that MITxOnline shouldn't be started up with apisix/keycloak here (instead of or in addition to doing so in the tip below) and  that logging in via /admin should be done if anything needs to be set up like CMS pages.

docker-compose.services.yml
```diff
@@ -129,8 +137,8 @@ services:
       - APISIX_SESSION_SECRET_KEY=${APISIX_SESSION_SECRET_KEY:-something_at_least_16_characters}
       - APISIX_LOGOUT_URL=${APISIX_LOGOUT_URL:-http://open.odl.local:8065/}
       - MITOL_API_DOMAIN=${MITOL_API_DOMAIN:-open.odl.local}
-      - MITX_ONLINE_UPSTREAM=${MITX_ONLINE_UPSTREAM:-mitxonline.odl.local:8013}
-      - MITX_ONLINE_DOMAIN=${MITX_ONLINE_DOMAIN:-mitxonline.odl.local}
+      - MITX_ONLINE_UPSTREAM=${MITX_ONLINE_UPSTREAM:-host.docker.internal:8013}
+      - MITX_ONLINE_DOMAIN=${MITX_ONLINE_DOMAIN:-host.docker.internal}
```
gumaerc: I had to set `MITX_ONLINE_DOMAIN` to `mitxonline.odl.local`. This allows you to access http://mitxonline.odl.local:8065/ in a browser and then, using your already logged in session in APISIX, you are automatically authenticated to use Django admin in MITx Online.

README.md
mbertrand: FWIW, even after setting up the env values below, I get into an endless loop when clicking the MITx Online login button.  Ditto for using MITx Online's own keycloak/apisix instance.

README.md
```diff
@@ -256,6 +258,51 @@ Personal API keys only need read permission to Query. When creating a personal A
...
+MITOL_APIGATEWAY_DISABLE_MIDDLEWARE=False
+MITOL_APIGATEWAY_USERINFO_CREATE=True
+MITOL_APIGATEWAY_USERINFO_UPDATE=True
+
+# Keycloak settings, these should match mit-learn
+KEYCLOAK_BASE_URL=http://kc.ol.local:8066
+KEYCLOAK_CLIENT_ID=apisix
+KEYCLOAK_CLIENT_SECRET=HckCZXToXfaetbBx0Fo3xbjnC468oMi4 # pragma: allowlist-secret
+KEYCLOAK_DISCOVERY_URL=http://kc.ol.local:8066/realms/ol-local/.well-known/openid-configuration
+KEYCLOAK_REALM_NAME=ol-local
```
mbertrand: I could not log in with these settings.   But the backpopulate command works regardless.

gumaerc (changes requested):
I've been testing this on Ubuntu Linux 24.04. I was able to eventually get the dashboard connected and see my programs / enrollments on the frontend, after changing some of the settings I noted below. However, I was unable to get the program page in Learn working. When I try and load the page for my test program, I get "Error fetching course page metadata for programv1%3Aprogam-v1-test-program" (Yes, that is the readable ID of my test program). The message says "connect ECONNREFUSED 127.0.0.1:8065." This indicates that the API in the `watch` server (since the metadata seems to be queried by the server and not the client) has mitxonline.odl.local as the MITx Online base url, which is going to evaluate to 127.0.0.1. I think in the docker configuration, if we somehow get the `watch` container itself to use `host.docker.internal:8013` for that, then it should work?

README.md
ChristopherChudzicki: > Not sure if this PR should also enable mitxonline users to log in via mit-learn's apisix/keycloak instance, but if so I could not validate that because something seems off with my local setup and I was unable to log in that way.

Do you mean click login via MITxOnline login button?  I had forgotten that does a surprising infinite loop. I'm not sure what causes that, but logging in via MIT Learn works.

@jkachel Do you have any idea? IMO not critical though.

README.md
```diff
@@ -256,6 +258,51 @@ Personal API keys only need read permission to Query. When creating a personal A
...
+and in MITxOnline, set:
+
+> [!TIP]
+> Be sure that COMPOSE_PROFILES is NOT set in your .env file; that way, MITxOnline's keycloak and apisix containers will not be used.
+
+```env
+# MITxOnline, .env
+CSRF_COOKIE_DOMAIN=.odl.local
+CORS_ALLOWED_ORIGINS=http://mitxonline.odl.local:8065, http://learn.odl.local:8062
+CSRF_TRUSTED_ORIGINS=http://mitxonline.odl.local:8065, http://learn.odl.local:8062
```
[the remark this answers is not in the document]
ChristopherChudzicki: For now I've reverted my local setup to use `open.odl.local` and any changes in this PR will use that host.

docker-compose.services.yml
ChristopherChudzicki: I'm not entirely sure what the issue was here, but I've changed so that `docker.host.internal` is not longer used and we just use `mitxonline.odl.local` everywhere.

ChristopherChudzicki:
@mbertrand @gumaerc This is ready for another look.

@mbertrand I tried to clarify the readme a bit, both in order and expected behavior of mitxonline.

I also dropped `host.docker.internal` for a scheme more similar https://github.com/mitodl/mitxonline/blob/29e6356817c1b88d34a08b442df293787873712b/docker-compose.yml#L21

docker-compose.services.yml
```diff
@@ -1,6 +1,11 @@
 include:
   - docker-compose.opensearch.${OPENSEARCH_CLUSTER_TYPE:-single-node}.yml
 
+# Define the extra_hosts as a reusable anchor
```
ChristopherChudzicki:
```suggestion
```

docker-compose.yml
```diff
@@ -26,6 +31,10 @@ services:
     extends:
       file: docker-compose.apps.yml
       service: watch
+    extra_hosts:
+      # Both the browser and NextJS server (containerized) make requests to
+      # api.open.odl.local, forward those to host
+      - "api.open.odl.local:host-gateway"
```
ChristopherChudzicki: @gumaerc re
> When I try and load the page for my test program, I get "Error fetching course page metadata for programv1%3Aprogam-v1-test-program" (Yes, that is the readable ID of my test program). 

I suspect that was happening for all api-dependent metadata generation (though some of them may have fallbacks to static metadata w/o an api call), not just the mitxonline-specific ones.

This should fix it.

dsubak:
FWIW, I took another look at this and ran through the tests. Still looks good (I like the new extra_hosts approach)!

I'll note that I did hit 404s when attempting to hit `http://learn.odl.local:8062/courses/COURSE_ID_HERE` in step 10, but everything got populated at the database level afaict. Given my lack of familiarity, I wanted to note it, but unless anyone else hits the same, I'm operating on the assumption that I've messed something up and it's not an issue w/ these changes.

PR #2595, opened 2025-10-15
Disable all scheduled tasks with a setting

mbertrand:
N/A

Adds a new optional setting (`CELERY_BEAT_DISABLED`), default of `False`, that if `True` will set `CELERY_BEAT_SCHEDULE={}`

I decided to add this because some of the daily scheduled tasks can take awhile to complete and interfere with development, testing, and code reviews.

### How can this be tested?
- Run `docker compose up` as usual. 
- In a shell:
  ```python
    settings.CELERY_BEAT_SCHEDULE
  ```
  You should get back a dict with multiple scheduled tasks
- Add   `CELERY_BEAT_DISABLED=True` to your `backend.local.env` file
- Restart containers, then rerun the shell code.  You should get back an empty dict,


ChristopherChudzicki (approved):

mbertrand:
Sorry @ChristopherChudzicki , forgot about this.  I talked with Shankar about it and decided that the new setting should apply to all the scheduled tasks, regardless of the `DEV_ENV` setting (there were a couple handled by that one previously).  That change is in the latest commit.

PR #2596, opened 2025-10-15
Server side query retry and caching with request scoped QueryClient

jonkafton:

- Closes https://github.com/mitodl/hq/issues/8704
- Fixes `Error: Functions cannot be passed directly to Client Components unless you explicitly expose it by marking it with "use server". Or maybe you meant to call this function rather than return it.
  [function]
   ^^^^^^^^` https://mit-office-of-digital-learning.sentry.io/issues/6770939422/?environment=production&project=4506260054540288&query=is%3Aunresolved%20function&referrer=issue-stream

- Reuses a single query client instance for API calls made during a request lifecycle
  - Duplicate requests in prefetching and metadata generation are only made once.
  - Retry logic is applied consistently and in one place.
- Uses the shared query client in metadata generation for retry.
  - Direct calls into the API client are replaced with `queryClient.fetchQuery()`, use is enforced with ESLint rules.
- Provides a `safeGenerateMetadata` utility that catches errors. Uncaught errors here were responsible for the error page being shown as a result of the non-serializable error being passed to client components.
  - Enforce use with an ESLint rule.

The approach is to reuse the [QueryClient](https://tanstack.com/query/latest/docs/reference/QueryClient) for all API requests made from the server (page server components, `generateMetadata`, client components during the SSR render phase). The QueryClient contains our retry logic, so we able to specify it in one place. React's server component [cache](https://react.dev/reference/react/cache) 

Reusing a QueryClient instance and therefore its [QueryCache](https://tanstack.com/query/latest/docs/reference/QueryCache) enables us, to share a single instance across a request lifecycle using React [cache](https://react.dev/reference/react/cache). This provides a cache layer that prevents duplicate API requests made during SSR prefetching and metadata generation. The instance is memoized and invalidated for each server request.

### How can this be tested?

Request caching
- We need to run Next.js in production mode as behaviors are different in development mode.
- Run services without the watch container `COMPOSE_PROFILES=backend,apisix,keycloak docker compose up` (or use local proxy described below).
- Build the Next.js container, `cd frontends/main && yarn build`
- Run with `PORT=8062 yarn start`
- Monitor the apisix logs, `docker compose logs -f apigateway`
- Load a channel page, e.g. http://open.odl.local:8062/c/unit/ocw
- The server makes a prefetch request for the channel detail and another to generate metadata. 
  - Check that only one request has been made to the channel detail endpoint from the server, e.g. GET /api/v0/channels/type/unit/ocw/
- Check that a single request is made for each refresh.

Retry mechanism and graceful fallback to client side query
- To test retry I've been using a proxy script that emulates errors responses: https://gist.github.com/jonkafton/278f80fcf5219f4af53a0ad0257bdef2. This is set to error for calls from the server to GET /api/v0/channels/type/unit/ocw/ for the first two attempts, but edit accordingly.
- The script proxies to RC, so we can stop the docker services.
- Drop the 2 files in a directory and run `yarn && yarn proxy:watch`.
- Load the channel page, http://open.odl.local:8062/c/unit/ocw
  - Check the proxy logs to make sure that the server makes two requests and succeeds on the third.
  - Check that the page loads correctly.
  - The page meta tags should fall back to default values, "MIT Learn" (og:title) and Learn with MIT (og:description).
- Edit proxy.ts to continue to error after 3 attempts / indefinitely.
  - Check that after the server requests fail, the page loads correctly, falling back to client side API call.

### Additional Context

I also looked at using native fetch on the Axios instance - it can be set as the adapter on the config. The goal is that we make use of Next.js' [data cache](https://nextjs.org/docs/app/guides/caching#data-cache) and route segment config for caching rules, however the fetch adapter is not compatible with React Query - the retry mechanism is called, but no actual requests are made; nor is it compatible with Next.js - requests are persisted (in ./main/.next/cache/fetch-cache), however config is ignored and all requests have a max-age of 1 year, which is not what we want.


frontends/api/src/ssr/serverQueryClient.ts
```diff
@@ -0,0 +1,84 @@
...
+ *
+ * Uses React's cache() to ensure the same QueryClient instance is reused
+ * throughout a single HTTP request, enabling:
+ *
+ * - Server API calls share the same QueryClient:
+ *   - Prefetch runs in page server components
+ *   - generateMetadata()
+ * - No duplicate API calls within the same request
+ * - Automatic cleanup when the request completes
+ * - Isolation between different HTTP requests
```
ChristopherChudzicki: > Isolation between different HTTP requests

That comes from React's `cache`, right? 

> React will invalidate the cache for all memoized functions for each server request.
> –[`cache` docs ](https://react.dev/reference/react/cache#caveats:~:text=React%20will%20invalidate%20the%20cache%20for%20all%20memoized%20functions%20for%20each%20server%20request.)

frontends/main/src/common/metadata.ts
```diff
@@ -18,6 +20,38 @@ type MetadataAsyncProps = {
...
+ * Axios error cannot be serialized as they contain function values and circular references.
+ * These result in "Functions cannot be passed directly to Client Components" errors.
+ *
+ * Instead, we catch the error and return the fallback of default metadata.
+ *
+ * If the error is a 404, we show the not found page.
+ */
+export async function safeGenerateMetadata(
+  fn: () => Promise<Metadata>,
+  fallback?: Metadata | (() => Promise<Metadata>),
```
ChristopherChudzicki: Are we using the fallback anywhere? Should we be?

NextJS merges metadata iirc, so we will always fall back to the root metadata, i think?

ChristopherChudzicki (changes requested):
_mean to click "Request changes"_

frontends/api/src/ssr/serverQueryClient.ts
```diff
@@ -0,0 +1,84 @@
+import { cache } from "react"
+import { QueryClient } from "@tanstack/react-query"
+import { AxiosError } from "axios"
+
+const MAX_RETRIES = 3
+const NO_RETRY_CODES = [400, 401, 403, 404, 405, 409, 422]
+
+export interface TaggedQueryClient extends QueryClient {
+  __requestId?: string
+  __createdAt?: number
```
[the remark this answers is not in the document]
jonkafton: I kept them as I kept needing to at back in to verify which instance was being used. Useful for future debugging.

frontends/api/src/ssr/serverQueryClient.ts
jonkafton: Yes

frontends/main/src/common/metadata.ts
jonkafton: Not used. We could always add back in if the need presents.

frontends/api/src/ssr/serverQueryClient.ts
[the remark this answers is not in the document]
ChristopherChudzicki: But we don't need them or anything? I asked ChatGPT able `__requestId` and react's `cache`, and it said it was internal. I am 90% sure that was a hallucination.

Can we rename them, like `__debug_id` or something?

Or at least add a comment that they are for debugging?

frontends/main/src/app/certificate/[certificateType]/[uuid]/page.tsx
```diff
@@ -22,24 +21,28 @@ export async function generateMetadata({
 }: PageProps<"/certificate/[certificateType]/[uuid]">): Promise<Metadata> {
   const { certificateType, uuid } = await params
 
-  let title, displayType, userName
+  return safeGenerateMetadata(async () => {
+    let title, displayType, userName
+
+    const queryClient = getServerQueryClient()
```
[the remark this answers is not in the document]
Comment (ChristopherChudzicki) [reply frontends/main/src/app/certificate/[certificateType]/[uuid]/page.tsx]: IMO, let's do this now so as not to to introduce two separate query clients.

ChristopherChudzicki (changes requested):
This looks good 👍  Just TS error from test file and
- https://github.com/mitodl/mit-learn/pull/2596/files#r2457215605

PR #2621, opened 2025-10-22
Remove usage of 'next' url cookie

rhysyngsun:

Fixes https://github.com/mitodl/hq/issues/8977

This removes the usage of the `next` parameter being stored and then being used for login and logout. This was causing issues because it was redirecting to auth-required urls on logout. The `next` param works without it anyway.

### How can this be tested?

- You should be able to login and the next urls passed should still work
- You should be able to logout from a page like `/dashboard` and not be prompted to reauthenticate.

### DIFF
--- a/authentication/views.py
+++ b/authentication/views.py
@@ -26,7 +26,7 @@ def get_redirect_url(request, param_names):
         str: Redirect URL
     """
     for param_name in param_names:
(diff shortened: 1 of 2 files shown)


ChristopherChudzicki (approved):
👍 Working well.

PR #2629, opened 2025-10-23
Fix python unit test

mbertrand:
N/A

Fixes a flaky test temporarily (long term fix: get rid of n+1 queries for this view)

### How can this be tested?
Tests should all pass

### DIFF
--- a/pytest.ini
+++ b/pytest.ini
@@ -3,6 +3,7 @@ addopts = --cov . --cov-report term --cov-report html --cov-report xml --ds=main
 norecursedirs = node_modules .git .tox static templates .* CVS _darcs {arch} *.egg
 markers =
   betamax: test requires betamax
+  skip_nplusone_check: test requires skipping n+1 query checks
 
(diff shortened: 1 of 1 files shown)


ChristopherChudzicki (approved):
This improves the situation, so 👍 

I did notice two things:

First, lots of
```
  /src/learning_resources/views_test.py:1245: PytestUnknownMarkWarning: Unknown pytest.mark.skip_nplusone_check - is this a typo?  You can register custom marks to avoid this warning - for details, see https://docs.pytest.org/en/stable/how-to/mark.html
    @pytest.mark.skip_nplusone_check
```

Second: `test_similar_resources_endpoint_does_not_return_self` is similarly flakey. Not sure if it's the same issue, though.

mbertrand:
> It looks like in mitxonline the `skip_nplusone_check` is registered in `pytest.ini` but not in Learn. Maybe we should add it to reduce the warnings?

Added

PR #2631, opened 2025-10-24
Video shorts webhook and api

mbertrand:
Part of https://github.com/mitodl/hq/issues/8880

- Adds an API endpoint for video shorts
- Adds a webhook endpoint to process new video short metadata coming from ol-data-platform
- Adds a management command to backpopulate video shorts from S3

### How can this be tested?
With your mit-learn containers up, run this in a python shell in your host OS:

```python
import hmac
import hashlib
import json
import requests

data =  {
    "video_id": "k_AA4_fQIHc",
    "youtube_metadata": <use sample_youtube_metadata dict in the parent conftest.py file>,
    "source": "youtube"
}

secret = "please-change-this"
payload_string = json.dumps(data, separators=(",", ":"))
signature = hmac.new(secret.encode(), payload_string.encode(), hashlib.sha256).hexdigest()
headers = {
    "X-MITLearn-Signature": signature,
    "Content-Type": "application/json",
}

response = requests.post("http://open.odl.local:8063/api/v1/webhooks/video_shorts/", data=payload_string, headers=headers)
print(response)
```

- The status code should be 200
- Go to http://open.odl.local:8063/api/v1/video_shorts/ - you should get 1 result
- Rerun the above again, changing the "title" field in the youtube metadata
- Go to http://open.odl.local:8063/api/v1/video_shorts/, you should see the new title

You can also try out the new backpopulate management command: `./manage.py backpopulate_video_shorts`

### TODO
- devops will need to set up fastly so that `/shorts/...` urls will route to the correct bucket location

### DIFF
--- a/frontends/api/src/generated/v1/api.ts
+++ b/frontends/api/src/generated/v1/api.ts
@@ -8619,10 +8619,10 @@ export interface VideoShortWebhookRequestRequest {
   video_id: string
   /**
    *
-   * @type {YouTubeMetadataRequest}
+   * @type {{ [key: string]: any; }}
    * @memberof VideoShortWebhookRequestRequest
    */
-  youtube_metadata: YouTubeMetadataRequest
+  youtube_metadata: { [key: string]: any }
   /**
    *
    * @type {string}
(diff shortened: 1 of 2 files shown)


shanbady (approved):
Looks good. only question i had is around why this is a completely separate model (not a learning resource) - just making sure its intentional we dont want embeddings or for this to show up in the search index

PR #2636, opened 2025-10-27
Fixes top margin at <md breakpoint

jonkafton:

N/A

Fixes issue with extra top margin at screen width >600px and <900px.

Before:
image

After:
image

### How can this be tested?

Size your browser window width to somewhere between 600px and 900px. Check that there is no gap between the header and the main content background. The issue on all pages, though less noticeable on the dashboard pages as they have no background <900px.

### Additional Context

Introduced in https://github.com/mitodl/mit-learn/pull/2620, which increased the header layout responsive breakpoint from 600px to 900px, but did not apply the same to margin calculation on the page wrapper top margin.

### DIFF
--- a/frontends/ol-components/src/components/NavDrawer/NavDrawer.tsx
+++ b/frontends/ol-components/src/components/NavDrawer/NavDrawer.tsx
@@ -6,7 +6,10 @@ import Link from "next/link"
 import React, { ReactElement } from "react"
 import { RiCloseLargeLine } from "@remixicon/react"
 import { ActionButton } from "@mitodl/smoot-design"
-import { HEADER_HEIGHT, HEADER_HEIGHT_MD } from "ol-components"
+import {
+  HEADER_HEIGHT,
+  HEADER_HEIGHT_MD,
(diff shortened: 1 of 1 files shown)


frontends/main/src/app/styled.tsx
```diff
@@ -13,7 +13,7 @@ export const PageWrapper = styled.div(({ theme }) => ({
   flexDirection: "column",
   height: "calc(100vh - 72px)",
   marginTop: "72px",
-  [theme.breakpoints.down("sm")]: {
+  [theme.breakpoints.down("md")]: {
     marginTop: "60px",
     height: "calc(100vh - 60px)",
```
ChristopherChudzicki: Could you switch this and the 72px above to `theme.custom.dimensions.headerHeightSm`, `theme.custom.dimensions.headerHeight`. (Maybe smoot should have a `theme.custom.dimensions.headerHeightMd`.)

I'd also do a find-all for a few other places that need to be changed. One is the left-side nav bar.

PR #2644, opened 2025-10-28
do not optimize in background images if optimize is off

ChristopherChudzicki:
None

This fixes an issue with our app when image optimization is turned off.
- When image opt is turned off, do not try to optimize the background images for css
- rename env var to NEXT_PUBLIC_OPTIMIZE_IMAGES

### How can this be tested?
1. On `main`, set `OPTIMIZE_IMAGES=false` in your frontend.local.env, and `docker compose up`.
2. Visit the homepage. Right click an image (like hero images) and click "copy image address". The image address should be akin to `http://open.odl.local:8062/images/hero/hero-3.png` (no query params). But if you scroll down, **the blue wavy background is missing** from the "Personalize" section. 
3. Try again on this branch **but with `NEXT_PUBLIC_OPTIMIZE_IMAGES=false`**.
    - hero image address should still be akin to `http://open.odl.local:8062/images/hero/hero-3.png`
    - blue wavy background should appear

### Checklist:
- [ ]  https://github.com/mitodl/ol-infrastructure/pull/3763

### DIFF
--- a/frontends/ol-utilities/src/images/backgroundImages.ts
+++ b/frontends/ol-utilities/src/images/backgroundImages.ts
@@ -19,8 +19,6 @@ export const backgroundSrcSetCSS = (image: ImageProps["src"]) => {
     src: image,
   })
 
-  console.log(props)
-
   if (!NEXT_PUBLIC_OPTIMIZE_IMAGES || !props.srcSet) {
     return `url("${props.src}")`
   }


frontends/ol-utilities/src/images/backgroundImages.ts
ChristopherChudzicki: The `backgroundSrcSetCSS` function runs client-side sometimes, so I had to rename the env var to expose it to the client.

**Aside:** We don't have any automated tests for this function, and I haven't tried to add any here. It depends a bit too heavily on NextJS's bundling to work well with Jest/JSDom, I think.

PR #2660, opened 2025-10-29
Eliminate build-time API calls, default to dynamic rendering

ChristopherChudzicki:
https://github.com/mitodl/hq/issues/9058

This PR:
- makes all pages use dynamic rendering, with a comment explaining why
- changes the docker container to run in production mode (build + preview) if NODE_ENV=production.

### How can this be tested?
1. I suggest testing on the penultimate commit, https://github.com/mitodl/mit-learn/pull/2660/commits/3b896dfed710c70e193b0ab4bebab48a2fbaa8fc, or at least copying the associated logging lines to `axios.ts` and `layout.ts`
2. Run in production mode (if using docker, set `NODE_ENV=production in watch container)
   - you should NOT see any api calls during the build
3. Load the site. it should work like before.
    - Navigate around. You should see "Rendering root layout" logged only once per full-page reload.
    -  you should see some server-side API calls in docker logs.

### DIFF
--- a/frontends/main/src/app/sitemaps/channels/sitemap.ts
+++ b/frontends/main/src/app/sitemaps/channels/sitemap.ts
@@ -10,14 +10,11 @@ invariant(BASE_URL, "NEXT_PUBLIC_ORIGIN must be defined")
 const PAGE_SIZE = 100
 
 /**
- * As of NextJS 15.5.3, sitemaps are ALWAYS generated via incremental
- * static regeneration (ISR), even with the force-dyanamic below.
- *
- * Setting revalidate to 0 forces regeneration on each request.
- * We handle caching via CDN.
+ * As of NextJS 15.5.3, sitemaps are ALWAYS generated at build time, even with
+ * the force-dynamic below (this may be a NextJS bug?). However, the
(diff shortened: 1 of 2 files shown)


frontends/main/src/app/sitemaps/channels/sitemap.ts
```diff
@@ -10,19 +10,19 @@ invariant(BASE_URL, "NEXT_PUBLIC_ORIGIN must be defined")
...
  */
 export const dynamic = "force-dynamic"
+export const revalidate = 0
 
 export async function generateSitemaps(): Promise<GenerateSitemapResult[]> {
   /**
-   * NextJS runs this at build time (despite force-dynamic below).
-   * Early exist here to avoid the useless build-time API calls.
+   * NextJS runs this at build time (despite force-dynamic above).
+   * Early exit here to avoid the useless build-time API calls.
```
gumaerc: Where is this early exit happening? Is that what `revalidate = 0` does? That tells it not to revalidate the cache?

ChristopherChudzicki: It's the `if (dangerouslyDetectProductionBuildPhase()) return`  statement below.

PR #2662, opened 2025-10-30
Avoid n+1 queries on video.playlists serializer field

mbertrand:
N/A

- Removes an n+1 query on `video.playlists` that was causing an intermittent test failure
- Adjusts that test so the created resource is always a video

### How can this be tested?
Tests should pass


shanbady (approved):
looks good. i re-ran the initially flaky test with pytest repeat=100 and it passes

PR #2663, opened 2025-10-30
Ab/hybrid search

abeglova:
part of https://github.com/mitodl/hq/issues/9135

This is a v0 implementation for vector search. This uses the huggingface/sentence-transformers/msmarco-distilbert-base-tas-b model. For now only title and description are vectorized the search performance in terms of returning the most relevant results definitely has room for improvement. The new index is only updated  and the new search mode only works if the model into opensearch and for now the plan is to only run it on rc. Also the new combined hybrid index is only updated via the recreate_index command for now and not upsert actions when learning resources are created or updated. Partially this is because this pr is already large and partially i wanted to make sure that vector embedding do not  affect the existing search until the hybrid search is closer to being production ready.  Additionally ,for now the new hybrid index is not linked to contentfiles and content file content is not used in the search

### How can this be tested?
Verify that the search page works normally. Also searching  "intro to ai class" returns no results

Login as an admin and select "hybrid" as the search mode from the admin options or go to http://open.odl.local:8062/search?search_mode=hybrid. The search will be empty for now regardless of whether you have a term but the site won't crash

Run ` docker-compose run web ./manage.py recreate_index --combined_hybrid` The task should finish right away and you should see `Skipping indexing hybrid index reindexing because no vector model is configured.` in the logs

from the shell run
```
from learning_resources_search.indexing_api import update_local_index_settings_for_hybrid_search, register_model, create_ingest_pipeline

update_local_index_settings_for_hybrid_search()
register_model()
create_ingest_pipeline()
```

Run

`docker-compose run web ./manage.py recreate_index --combined_hybrid`

Go to 
http://open.odl.local:8062/search

Search should still work normally

Go to 
http://open.odl.local:8062/search?search_mode=hybrid

Search should work and facets should work
Searching "intro to ai class" should return results.

### DIFF
--- a/learning_resources_search/tasks.py
+++ b/learning_resources_search/tasks.py
@@ -649,7 +649,6 @@ def start_recreate_index(self, indexes, remove_existing_reindexing_tags):  # noq
                 )
             ]
 
-
         for resource_type in set(LEARNING_RESOURCE_TYPES) - {COURSE_TYPE}:
             if resource_type in indexes:
                 index_tasks = index_tasks + [
--- a/learning_resources_search/tasks_test.py
+++ b/learning_resources_search/tasks_test.py
@@ -26,8 +26,8 @@ from learning_resources_search.constants import (
     COMBINED_INDEX,
     CONTENT_FILE_TYPE,
     COURSE_TYPE,
-    PROGRAM_TYPE,
     LEARNING_RESOURCE_TYPES,
+    PROGRAM_TYPE,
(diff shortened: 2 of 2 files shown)


abeglova:
I'm not sure  if the  `micromasters_` indexes are still used. In any case, they will be ignored by learn since learn ignores any indexes that don't have prefix settings.OPENSEARCH_INDEX

PR #2669, opened 2025-11-03
Fix OpenAPI Client Generation for multivalue query params

ChristopherChudzicki:
Closes https://github.com/mitodl/hq/issues/9124

This PR: changes our OpenAPI spec and generated client, but has no functional changes to any APIs.

This PR changes our OpenAPI spec to correctly use `explode: true` for query params generated via `CharInFilter` and `NumberInFilter`. `explode: true` means that multivalued query params should be passed as `name=value1&name=value2,...`, whereas `explode:false` indicates a comma-separated list `name=value1,value2`
- These filters inherit from Django Filter's `BaseCSVFilter`; as the name implies, that is expected to support comma separated values
- But we modified our filters and filter backend to explicitly not support commas as a separator because it caused problems in scenarios like filtering on the topic `"Data Science, Analytics & Computer Technology"` or `"Atomic, Molecular & Optical Physics"`, where the filter value contains commas.
- So currently on main, our multivalued filters only work with `explode: true` syntax[^1], but OpenAPI spec indicates the opposite. This breaks our generated client. 

[^1]: Actually, we use three multivalued filters: `MultipleChoiceFilter` works in comma-separated or exploded mode at the API level, but OpenAPI has always used `explode:true` for this one.

### How can this be tested?

1. In `frontends/api/src/clients.ts`, add the line `globalThis.learningResourcesApi = learningResourcesApi`
2. On the frontend, try these queries in the JS dev console, on both `main` and this branch, `cc/openapi-explode`
    ```ts
     learningResourcesApi.learningResourcesList({
         readable_id: ["course-v1:xPRO+SysEngx1", "course-v1:xPRO+SysEngx3"]
     })
     learningResourcesApi.learningResourcesList({
         topic: ["Data Science, Analytics & Computer Technology", "Engineering"]
     })
     ```
    - expectation: On `main`, these queries produce comma-separated network calls that do not work correctly.
    - expectation: On `cc/openapi-explode`, the queries should produce network calls like
        - http://api.open.odl.local:8065/api/v1/learning_resources/?readable_id=course-v1%3AxPRO%2BSysEngx1&readable_id=course-v1%3AxPRO%2BSysEngx3 (note: from xpro, so backpopulate that)
        - http://api.open.odl.local:8065/api/v1/learning_resources/?topic=Data+Science%2C+Analytics+%26+Computer+Technology&topic=Engineering
3. Try the URLs (`api.open.odl....`) directly on `main` and on this `cc/openapi-explode`. **You should get exactly the same results on both branches** (No change in actual API behavior)

**Note:** In words, ☝️ means that our generated client doesn't work on `main` when multiple values are passed to `CharInFilter`-powered and `NumberInFilter`-powered params. This currently never happens on `main` as far as I know, but we need to pass multiple readable IDs for the upcoming program pages.

### DIFF
--- a/openapi/specs/v1.yaml
+++ b/openapi/specs/v1.yaml
@@ -452,7 +452,7 @@ paths:
         schema:
           type: array
           items:
-            type: number
+            type: integer
         description: The id of the learning resource the content file belongs to
         explode: true
         style: form
@@ -1091,7 +1091,7 @@ paths:
         schema:
           type: array
           items:
-            type: number
+            type: integer
         description: The id of the learning resource the content file belongs to
         explode: true
         style: form
@@ -3355,7 +3355,7 @@ paths:
         schema:
           type: array
           items:
-            type: number
+            type: integer
         description: The id of the learning resource the content file belongs to
         explode: true
         style: form


abeglova (approved):
👍

PR #2671, opened 2025-11-04
feat: add article create page and configure ckeditor with no custom plugin

ahtesham-quraish:
In reference to this ticket https://github.com/mitodl/hq/issues/9108

- [ ] A minimal WYSIWYG text editor, CKEditor with no custom plugins
- [ ] Page `/articles/new` for creating a new article via the WYSIWYG editor, https://github.com/mitodl/hq/issues/9108

### How can this be tested?
- [ ] run yarn to install the dependencies
- [ ] after installing the dependencies you need to up the frontend next app
- [ ] then hit the following url `/articles/new` for accessing the page  

image

image

image


frontends/ol-ckeditor/src/components/CkEditorClient.tsx
```diff
@@ -0,0 +1,378 @@
...
+          _CKEditorModules: CKEditorModules,
+        })
+      } catch (error) {
+        console.error("CKEditor token fetch failed:", error)
+        setIsLoader(false)
+      }
+    })()
+  }, [])
+
+  const editorConfig: EditorConfig = useMemo(() => {
```
ChristopherChudzicki: Can't we define everything that is token-independent at the module top level?

**Actually:** Since there's a chance we will replace CKEditor, I would suggest.... Define a static config at the top level. We don't need the token for anything other than image uploading, right? Let's just leave that out for now.

frontends/main/package.json
```diff
@@ -32,6 +32,7 @@
     "moment": "^2.30.1",
     "next": "^15.5.2",
     "next-nprogress-bar": "^2.4.2",
+    "ol-ckeditor": "0.0.0",
```
ChristopherChudzicki: very minor... `workspace:*` is preferable. feel free to change the ones below, too.

`workspace:*` will **always** use the local workspace, even if a package named `ol-ckeditor` exists on the npm registry.

ChristopherChudzicki (changes requested):
Left some requests and suggestions. The most important of these is around permissioning and feature flags.

It would also be good to add a simple test page for NewArticlePage.tsx; things I would test are:
- only shown if you have the correct permission
- makes correct POST call given title + body

You would likely need to mock the WYSIWYG component with a textarea—we don't have any browser-based testing environments right now.

frontends/main/src/app/articles/new/page.tsx
```diff
@@ -0,0 +1,14 @@
+import React from "react"
+import { Metadata } from "next"
+import { standardizeMetadata } from "@/common/metadata"
+import { NewArticlePage } from "@/app-pages/Articles/NewArticlePage"
+
+export const metadata: Metadata = standardizeMetadata({
+  title: "Articles",
```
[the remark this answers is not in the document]
ahtesham-quraish: ok

frontends/main/package.json
ahtesham-quraish: Done

frontends/ol-ckeditor/src/components/CkeditorArticle.tsx
```diff
@@ -0,0 +1,95 @@
...
+          key={alertText}
+          severity={severity}
+          className="info-alert"
+          closable
+        >
+          <Typography variant="body2" color="textPrimary">
+            {alertText}
+          </Typography>
+        </Alert>
+      )}
```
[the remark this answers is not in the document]
ahtesham-quraish: This stuff will go away when we redirect the user to detail page/list page of articles don't worry about them for now

frontends/ol-ckeditor/src/components/CkEditorClient.tsx
ahtesham-quraish: I have removed the cloud service for now and replaced it with `Base64UploadAdapter` which will convert the image into base64 and add it into html. We can later on enable it when we would be sure that which editor we are going with.

PR #2676, opened 2025-11-05
embeddings healthcheck

shanbady:

Closes https://github.com/mitodl/hq/issues/6978

This PR adds a comprehensive periodic (1/week) healthcheck for missing embeddings and summaries/flashcards. Any missing records posted via sentry which we can setup alerts for.

### How can this be tested?
1. checkout this branch
2. copy down settings.SENTRY_DSN from rc (if you are setting it directly in the settings.py file instead of as an env var make sure it is defined before the call to `init_sentry()`)
3. restart celery
4. make sure you have some resources populated locally - also make sure you are locally missing embeddings for some (or all) resources. If they are all embedded locally, manually delete records from the [local qdrant dashboard](http://localhost:6333/dashboard) (easier to do with the resources collection since contentfiles are chunked)
5. manually run the healthcheck task (may take a long time depending on how much data you have):
```python
from vector_search.tasks import embeddings_healthcheck
embeddings_healthcheck()
```
6. visit the [sentry dashboard](https://mit-office-of-digital-learning.sentry.io/explore/discover/homepage/?dataset=errors&environment=dev&field=title&field=project&field=user.display&field=timestamp&name=All%20Errors&project=4506260054540288&query=healthcheck%3Aembeddings&queryDataset=error-events&sort=-timestamp&statsPeriod=24h&yAxis=count%28%29) for the "dev" environment and you should see items for missing embeddings:

### Additional Context
1. The current implementation of the checks for embeddings iterates through all resources and all contentfiles (that should be embedded)
    - for resources we directly check each record in qdrant
    - for contentfiles we just check that the very first chunk for a contentfile exists in qdrant
2. This check is set to run on friday when there are fewer etl pipelines running and everything should be embedded

### DIFF
--- a/RELEASE.rst
+++ b/RELEASE.rst
@@ -1,6 +1,16 @@
 Release Notes
 =============
 
+Version 0.47.9 (Released November 05, 2025)
+--------------
+
+- don't use next_start_date for "anytime" courses & programs (#2670)
+- Shanbady/fix subscription groups (#2664)
+- Server side query retry and caching with request scoped QueryClient (#2596)
+- alter org page header based on new requirements (#2659)
+- Get rid of debug code (#2661)
+- Eliminate build-time API calls, default to dynamic rendering (#2660)
+
 Version 0.47.8 (Released October 29, 2025)
 --------------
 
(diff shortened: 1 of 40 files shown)


abeglova (approved):
I couldn't get it to post to sentry from my local environment but i'm pretty sure it's a issue with my environment. The rest of it works correctly

PR #2685, opened 2025-11-06
add error message for enrollment code issues

gumaerc:
Closes https://github.com/mitodl/hq/issues/9176

This PR adds functionality that will display an error message if there was a problem redeeming an enrollment code. This is done by comparing the initial load of the MITx Online user's `b2b_organizations` property versus after redeeming the enrollment code. If the orgs haven't changed, that's an indication that something went wrong. It is done this way because the API endpoint always returns a 200. This is done so that nefarious actors can't probe the endpoint in an attempt to find a code that works. If the orgs haven't changed, you are redirected to the dashboard home page with a query string arg that triggers the dismissable error at the top of the screen.

image
image

### How can this be tested?
- Ensure that you have an instance of MITx Online up and running and configured to use the same Keycloak / APISIX instances as your instance of Learn
- In your instance of MITx Online, make sure you have a B2B organization set up with a contract with some courses added to it
- Generate enrollment codes for said contract, and keep one of them handy
- Spin up MIT Learn, and make sure it is configured to integrate with your instance of MITx Online (more on that in the README)
- Log in as a user that is not part of any B2B orgs
- Visit the enrollment code redemption page with your code from earlier, at `/enrollmentcode/<code here>`
- You should see a brief message that the code is being redeemed, and then be redirected to the dashboard home and see a tab for the org you just enrolled in
- Verify that no error message is displayed
- Try and redeem the enrollment code again with the same user
- Verify that you are redirected to the dashboard again, but this time you should see the error message indicated in the screenshots

### Additional Context
In the Figma design, you may see a different look for the alert banner than what you see during testing. This is because the design for the `Alert` component in `smoot-design` has pending style updates that have not been made yet.

### DIFF
--- a/frontends/main/src/app-pages/EnrollmentCodePage/EnrollmentCodePage.tsx
+++ b/frontends/main/src/app-pages/EnrollmentCodePage/EnrollmentCodePage.tsx
@@ -17,7 +17,6 @@ const InterstitialMessage = styled(Typography)(({ theme }) => ({
 }))
 
 const EnrollmentCodePage: React.FC<EnrollmentCodePage> = ({ code }) => {
-  const [hasEnrolled, setHasEnrolled] = React.useState(false)
   const router = useRouter()
 
   const enrollment = useB2BAttachMutation({
@@ -30,11 +29,14 @@ const EnrollmentCodePage: React.FC<EnrollmentCodePage> = ({ code }) => {
   })
 
   React.useEffect(() => {
-    if (user?.is_authenticated && !hasEnrolled && !enrollment.isPending) {
-      setHasEnrolled(true)
+    if (
+      user?.is_authenticated &&
+      !enrollment.isPending &&
+      !enrollment.isSuccess
+    ) {
       enrollment.mutate()
     }
-  }, [user?.is_authenticated, hasEnrolled, enrollment])
+  }, [user?.is_authenticated, enrollment])
 
   // Handle redirect based on response status code
   // 201: Successfully attached to new contract(s) -> redirect to dashboard


frontends/main/src/app-pages/EnrollmentCodePage/EnrollmentCodePage.tsx
```diff
@@ -17,22 +18,53 @@ const InterstitialMessage = styled(Typography)(({ theme }) => ({
...
+      enrollment.mutate()
+    }
+  }, [user?.is_authenticated, hasEnrolled, enrollment])
+
+  // Handle redirect after mutation succeeds and query is refetched
+  React.useEffect(() => {
+    if (enrollment.isSuccess && !enrollment.isPending) {
+      const currentOrgCount =
+        mitxOnlineUser.data?.b2b_organizations?.length ?? 0
+      if (initialOrgsRef.current === currentOrgCount) {
```
ChristopherChudzicki: **Request:** I noticed the mitxonline slack discussion about why you're checking this (API basically never errors, but returns 200 even if the code is invalid). That's not obvious and probably worth mentioning here. If I saw this code, I'd want to change it to check that response status.

From the /attach PR:
> The endpoint returns the user's current contract associations, even if the code is not valid, so the API doesn't expose extra information about the discount.

**But really:** I think we should change the MITxOnline behavior.
- I'm skeptical that this adds any security. As demonstrated here, you can can just check if the original list of contracts has changed. If we need to prevent spamming the API, seems like rate-limiting would be better. 
- We currently cannot handle the following situation well: user clicks link to `open.odl.local:8062/enrollmentcode/code1` and joins `contract1`; then user clicks the same link again 
    - This PR shows an error to the user
    - The API is a no-op... they're already in the contract.
    - Of course, clicking the link again is unnecessary, but ... people click links.

ChristopherChudzicki: cc @jkachel

ChristopherChudzicki (changes requested):
Left some comments about race condition and simplifying things.

I do think it's worth checking with @jkachel about changing the mitxonline api.

gumaerc:
@ChristopherChudzicki This is ready for another look. I merged https://github.com/mitodl/mitxonline/pull/3084 in MITx Online, which added response codes to the attach API endpoint. The enrollment code page now bases its actions on these response codes as described, and I added some functionality that clears the query param after it is seen like you suggested.

ChristopherChudzicki (changes requested):
The error display isn't quite working because my suggestion for clearing URL param wasn't quite right. Left another comment, too, about enrollmentcode page

much simpler w/ proper error codes, though

ChristopherChudzicki (approved):
👍 

I still think https://github.com/mitodl/mit-learn/pull/2685/files#r2511242400 is worth changing, though.

PR #2688, opened 2025-11-07
Fix bug for staleness and completeness opensearch

abeglova:
closes https://github.com/mitodl/hq/issues/9243

Currently there is a bug on production causing errors if yearly_decay_percent or max_incompleteness_penalty is greater than zero. For example

https://learn.mit.edu/search?q=human+brain&yearly_decay_percent=5&resource_category=course

I was able to recreate the issue locally (but weirdly rc works correctly)

Changing the the format of the query to match the opensearch documentation fixed it for me locally https://docs.opensearch.org/latest/query-dsl/compound/function-score/

Hopefully it works on prod too and is a good idea regardless

There is also a  separate bug that is setting `next_start_date` to the latest etl run for ocw courses https://github.com/mitodl/hq/issues/9206 that will be fixed in a different pr

### How can this be tested?

Download single variable calculus from fall 2006
`
docker-compose run web ./manage.py backpopulate_ocw_data --course-name 18-01-single-variable-calculus-fall-2006 --skip-contentfiles
`

Clear out next_start_date from ocw courses 
`
 LearningResource.objects.filter(offered_by='ocw').update(next_start_date=None)
`

Update the course index to update `next_start_date` in search
`
docker-compose run web ./manage.py update_index --courses
`
go to  http://open.odl.local:8062/search  and login

Search for "single variable calculus" 

You should be able to set "Resource Score Staleness Penalty" from the admin section to any value. 18-01-single-variable-calculus-fall-2006 should move down in the search results as "Resource Score Staleness Penalty"  increases

### DIFF
--- a/learning_resources_search/api.py
+++ b/learning_resources_search/api.py
@@ -592,38 +592,37 @@ def add_text_query_to_search(search, text, search_params, query_type_query):
 
     if yearly_decay_percent or max_incompleteness_penalty:
         script_query = {
-            "script_score": {
+            "function_score": {
                 "query": {"bool": {"must": [text_query], "filter": query_type_query}}
             }
         }
 
-        completeness_term = (
-            "(doc['completeness'].value * params.max_incompleteness_penalty + "
-            "(1-params.max_incompleteness_penalty))"
-        )
-
-        staleness_term = (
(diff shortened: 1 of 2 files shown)


shanbady (approved):
👍

PR #2691, opened 2025-11-10
Initial Tiptap Editor

jonkafton:

N/A

Adds a basic Tiptap editor component. See https://github.com/mitodl/mit-learn/pull/2665 for POC/background.

 - New article edit page at /article/new. This PR https://github.com/mitodl/mit-learn/pull/2686 adds /article**s**/new. This lives adjacent for the moment.
 - Editor is self contained in ol-components/TiptapEditor.
 - Tiptap's [Simple Editor](https://tiptap.dev/docs/ui-components/templates/simple-editor) is installed. This eject's code to the project expecting import aliases for `@/components`, `@/hooks` and `@/lib`, which conflict with our main workspace existing. Some attempts were made to configure our TS and build config to avoid modifying vendor code, though fixes were not trivial - files are updated with relative imports and fixes applies for ESLint, Prettier and type checking.
 - Editor is not hooked up to the API to save.
 - Sibling article detail page not implemented.

image

### How can this be tested?

Navigate to /article/new

### DIFF
--- a/RELEASE.rst
+++ b/RELEASE.rst
@@ -1,6 +1,17 @@
 Release Notes
 =============
 
+Version 0.47.11 (Released November 10, 2025)
+---------------
+
+- set h5 on the welcome message (#2682)
+- Avoid errors caused by None tags (#2680)
+- org contract welcome message display (#2677)
+- Add periodic topic collection sync task (#2673)
+- Resource card headings for screen reader navigation (#2658)
+- fix CTA text logic (#2675)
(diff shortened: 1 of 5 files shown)


jonkafton:
> While I like the idea of minimizing the changes to the template, the alias imports may be more effort than they are worth...

Agree - in the end it does not seem possible to support conflicting aliases in a nested tsconfig without changing project config and adding not-so-nice workarounds. I've updated the ejected tiptap code to use relative imports.

PR #2695, opened 2025-11-12
Ensure re-indexing and re-embedding operations include all non-course learning resource types

shanbady:

Closes https://github.com/mitodl/hq/issues/9233

This PR resolves (and prevents something similar from happening moving forward) an issue where the newly added Article learning resource type was not being indexed or updated during recreate_index and update_index operations in both vector search and learning_resources search apps. There are places where we hardcode a list of non-course resource type which should actually just be automatically derived from LEARNING_RESOURCE_TYPES constant.

### How can this be tested?
1. checkout main
2. backpopulate mit climate articles via `python manage.py backpopulate_mit_climate_articles`
3. recreate the opensearch index `python manage.py recreate_index --all`
4. go to your [local opensearch indexes](http://localhost:9100/_cat/indices?v) and note that the article index is empty/has 0 records
5. checkout this branch
6. restart celery 
7. repeat the steps and note that it is populated

You can also see this behavior with the update_index command as well as the generate_embeddings management command

### Additional notes
The reason this is not an issue on our live instances is that the the missing content types still get added to the index upon insertion so this only affected recreation and updates to the entire index.

### DIFF
--- a/learning_resources_search/tasks_test.py
+++ b/learning_resources_search/tasks_test.py
@@ -134,7 +134,7 @@ def test_system_exit_retry(mocker):
 
 @pytest.mark.parametrize(
     "indexes",
-    [["course"], ["program"]],
+    [["course"], ["program"], list(LEARNING_RESOURCE_TYPES)],
 )
 def test_start_recreate_index(mocker, mocked_celery, user, indexes):
     """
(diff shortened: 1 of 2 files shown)


learning_resources_search/tasks_test.py
```diff
@@ -134,7 +134,7 @@ def test_system_exit_retry(mocker):
 
 @pytest.mark.parametrize(
     "indexes",
-    [["course"], ["program"]],
+    [["course"], ["program"], list(LEARNING_RESOURCE_TYPES)],
```
mbertrand: Should this be ` [["course"], ["program"], *list(LEARNING_RESOURCE_TYPES)],` ?

PR #2699, opened 2025-11-13
Article editor refactor for reuse and layout updates

jonkafton:

N/A

- Refactors common create/update logic to the new ArticleEditor component (replaces EditorContainer).
- Updates layout for full width header fixed toolbar.
- Title input moved into shared component.
- Save button moved to toolbar.
- Error messaging from query hooks (replaces onError handling).
- Removes /article/new route from initial editor PR, https://github.com/mitodl/mit-learn/pull/2691.

image


ahtesham-quraish:
After editing the article it should lead me to article detail page which is not happening RN

Issue #2698, opened 2025-11-12
rename search cache


abeglova:
In the learn codebase SEARCH_PAGE_CACHE_DURATION affects all api pages not just the search page. Also there isn't any way to clear the cache for non-search learning resource api views since clear_search_cache() is limited to search views.

PR #2700, opened 2025-11-13
Ab/rename SEARCH_PAGE_CACHE_DURATION

abeglova:
https://github.com/mitodl/mit-learn/issues/2698

The current cache is confusingly named. The variable is SEARCH_PAGE_CACHE_DURATION, the caches are saved in redis as "search" and the function to clear it is `clear_search_cache` but the cache affects all api views. This renames it. It also sets the local default SEARCH_PAGE_CACHE_DURATION in shared.env to 0 since having a cache locally while developing is really confusing

### How can this be tested?
set SEARCH_PAGE_CACHE_DURATION=160 in shared.local.env
Open some api views. Also go to the front page and the search page

Run
```
cache = caches["redis"]
 cache.keys("views.*")
```
you should see some caches and they should have different names depending on what api call is being cached

Run
```
from main.utils import clear_views_cache
clear_views_cache()

cache = caches["redis"]
cache.keys("views.*")
```
all the caches should be cleared


shanbady (approved):
looks good. left a very minor suggestion.

PR #2711, opened 2025-11-17
feat: Add Custom Media Embed Plugin for Tiptap Editor

ahtesham-quraish:
In reference to https://github.com/mitodl/hq/issues/9276

We want to provide content creators with the ability to insert and customize embedded media inside the Tiptap editor.
The custom plugin should support adding video URLs, resizing the embedded player for different screens like medium.com has for media.

image

Screenshot 2025-11-19 at 12 27 03 PM

Screenshot 2025-11-19 at 12 26 38 PM

### How can this be tested?
- Hit the following url '/articles/new' 
- You would see the video embed icon in toolbar, click on that 
- A modal will open where you paste the url
- then give title of the article and if you want to change layout of video from control right above the video 
- Give caption of the video 
- click save, after that saving the content you will be redirecting to detail page of the article


jonkafton (approved):
Looks good :+1:

We were hoping to avoid changes to vendor code to ease upgrades, though I'm not sure how practical that is given TipTap's workflow that adds base code to the project. We should extend where possible or consider refactoring extension code into the project. I do want to better delineate the vendor code by putting it in its own folder.

A merge from main should fix the failing OpenAPI diff job.

PR #2716, opened 2025-11-19
[WIP] Fix feature flag loading issue on product pages

ChristopherChudzicki:
Closes https://github.com/mitodl/hq/issues/9317

This PR fixes an issue where the feature-flagged product pages would occasionally show 404, particularly on slow networks.

### How can this be tested?
If you really want to see (A) the issue on `main` and (B) that this PR fixes it:

1. **Prerequisites:**
    - Learn <> MITxOnline integration set up
    - Posthog setup with `product-course-page` flag enabled
    - `NODE_ENV=production` in `frontend.local.env` (I find that page loads too slowly in dev mode, meaning that posthog always loads latest; the race condition is much more reproducible in production mode.)
2. On `main`:
    1. visit `http://open.odl.local:8062/programs/<readable-id>`
    2. enable network throttling (4G should be good enough) and reload the page. If you don't get 404, keep reloading, but I've found it very reproducible on 4G throttling in production mode.
3. Repeat Step 2 on this branch.

### DIFF
--- a/frontends/main/package.json
+++ b/frontends/main/package.json
@@ -34,7 +34,7 @@
     "next-nprogress-bar": "^2.4.2",
     "ol-components": "0.0.0",
     "ol-utilities": "0.0.0",
-    "posthog-js": "^1.157.2",
+    "posthog-js": "^1.296.1",
     "react": "^19.0.0",
     "react-dom": "^19.0.0",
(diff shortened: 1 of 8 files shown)


pdpinch:
Will this close https://github.com/mitodl/hq/issues/9335 ?

PR #2718, opened 2025-11-20
use qdrant vectors in hybrid search

abeglova:
closes https://github.com/mitodl/hq/issues/9148

This pr updates the hybrid search index in opensearch to reuse vectors from qdrant. The hybrid search is currently behind a flag (search_mode=hybrid)

### How can this be tested?
set 
OPENAI_API_KEY to the value from rc
and
QDRANT_ENCODER=vector_search.encoders.litellm.LiteLLMEncoder

run `docker-compose run web ./manage.py generate_embeddings --skip-contentfiles` to ensure you have resource embedding in qdrant

run `docker-compose run web ./manage.py recreate_index --combined_hybrid`

Got to
http://open.odl.local:8062/search?q=intro+to+calculus&search_mode=hybrid

you should see results


learning_resources_search/indexing_api.py
```diff
@@ -192,23 +190,15 @@ def clear_and_create_index(*, index_name=None, skip_mapping=False, object_type=N
     }
 
     if object_type == COMBINED_INDEX:
```
shanbady: imo the naming of these indexes are getting a bit confusing (constants.COMBINED_INDEX vs constants.ALL_INDEXES vs constants.BOTH_INDEXES) might be better to call this HYBRID_INDEX or HYBRID_COMBINED_INDEX etc

PR #2720, opened 2025-11-21
update posthog, add useFeatureFlagsLoaded

ChristopherChudzicki:
Closes https://github.com/mitodl/hq/issues/9317

This PR fixes an issue where the feature-flagged product pages would occasionally show 404, particularly on slow networks.

### How can this be tested?
1. **Prerequisites:**
    - Learn <> MITxOnline integration set up
    - Posthog setup with `product-course-page` flag enabled
    - `NODE_ENV=production` in `frontend.local.env` (I find that page loads too slowly in dev mode, meaning that posthog always loads latest; the race condition is much more reproducible in production mode.)
2. On `main`:
    1. visit `http://open.odl.local:8062/programs/<readable-id>`
    2. enable network throttling (4G should be good enough) and reload the page. If you don't get 404, keep reloading, but I've found it very reproducible on 4G throttling in production mode.
3. Repeat Step 2 on this branch
    - I suggest checking both dev mode (NODE_ENV=development) and production build (`NODE_ENV=production`).

### DIFF
--- a/frontends/main/src/common/useFeatureFlagsLoaded.ts
+++ b/frontends/main/src/common/useFeatureFlagsLoaded.ts
@@ -18,6 +18,12 @@ import { useEffect, useState } from "react"
  *     loaded.
  * 2. Unlike posthog's `onFeatureFlags` callback, this hook enables you to
  *   distinguish between bootstrapped flags and flags loaded from PostHog server.
+ *
+ * IMPLEMENTATION:
+ * Posthog does not make detecting "flags have loaded from server" easy.
+ * This approach relies on the fact that bootstrapped flags are completely
+ * discarded after flags are loaded from server, so `useFlagEnabled` will
+ * return `undefined` for any flag that was only bootstrapped locally.
  */
 const useFeatureFlagsLoaded = () => {
   const posthog = usePostHog()


frontends/main/src/app-pages/ProductPages/ProgramPage.tsx
```diff
@@ -89,10 +90,11 @@ const ProgramPage: React.FC<ProgramPageProps> = ({ readableId }) => {
   const program = programs.data?.results?.[0]
   const programResource = programResources.data?.results?.[0]
   const enabled = useFeatureFlagEnabled(FeatureFlags.ProductPageCourse)
-  if (enabled === false) {
-    return notFound()
+  const flagsLoaded = useFeatureFlagsLoaded()
```
ChristopherChudzicki: If you change these lines to
```ts
  const enabled = useFeatureFlagEnabled(FeatureFlags.ProductPageCourse)
  const bootstrapped = useFeatureFlagEnabled(INTERNAL_BOOTSTRAPPING_FLAG)
  const flagsLoaded = useFeatureFlagsLoaded()

  console.log({ enabled, bootstrapped, flagsLoaded })
```
you would see values something like
```ts
// posthog not initialized
{enabled: undefined, bootstrapped: undefined, flagsLoaded: false}
// posthog initialized, flags loaded from localstorage / bootstrapping
{enabled: true, bootstrapped: true, flagsLoaded: false}
// posthog initialized, flags loaded from server
{enabled: true, bootstrapped: undefined, flagsLoaded: true}
```

frontends/main/src/page-components/ConfiguredPostHogProvider/ConfiguredPostHogProvider.tsx
```diff
@@ -2,20 +2,12 @@ import React, { useEffect } from "react"
 import posthog from "posthog-js"
 import { PostHogProvider, usePostHog } from "posthog-js/react"
 import { useUserMe } from "api/hooks/user"
+import { INTERNAL_BOOTSTRAPPING_FLAG } from "@/common/feature_flags"
 
 const POSTHOG_API_KEY = process.env.NEXT_PUBLIC_POSTHOG_API_KEY
 const POSTHOG_API_HOST = process.env.NEXT_PUBLIC_POSTHOG_API_HOST
 const FEATURE_FLAGS = process.env.FEATURE_FLAGS
 
-if (POSTHOG_API_KEY) {
```
ChristopherChudzicki: For nextjs, Posthog docs recommend initializing either
- `instrumentation-client.ts`
    - I tried this; it caused a bunch of hydration errors. I think that in order for the `instrumentation-client.ts` approach to work w/o hydration errors, you need also to initialize in the server and load flags server-side. Since we don't have user info on the server, we can't do that.
- provider component inside a `useEffect`

The current top-level initialization caused hydration errors, too. Initializing inside the provider effect does not.

jonkafton (approved):
Works a charm! 👍

frontends/main/src/page-components/ConfiguredPostHogProvider/ConfiguredPostHogProvider.tsx
ChristopherChudzicki: It does. Though that also wasn't happening with the current placement of posthog init code. On RC currently:
- code is module level
- `NEXT_PUBLIC_FEATURE_home_page_recommendation_bot` is set to true. 

I do think we could get server-side flagging of the bootstrapped values to work, though.

PR #2777, opened 2025-12-01
Canvas - Transcribe handwritten notes and pdf contentfiles 

shanbady:

Closes https://github.com/mitodl/hq/issues/9288

This PR lets use use an LLM to OCR content from PDFs that is not parseable via traditional means such as handwriting, diagrams and formulas embedded as images.

**How it works:**
A pdf document is typically a composite of many different data types. It can contain text, embedded metadata, embedded images, pages that are entirely an image, tables, diagrams and more. We leverage [docling](https://docling-project.github.io/docling/) which is a framework for parsing many different file formats to break apart the pdf and send off the un-parseable content for an LLM to transcribe.

For any given pdf (in a canvas course):
1. create the docling document tree
2. iterate page by page (in batches)
3. For each page, grab all the text that we can
4. when we encounter a photo, code block, or formula block -> send it for the LLM to ocr and return docling text elements
5. If there happens to be a page with 0 text elements - we OCR the entire page as an image
6. Stitch the docling document back together and render as markdown

This approach gives us many advantages over the previous approach of OCR'ing every single page:
- its faster
- its cheaper since we only ocr pieces that need it
- Eventually expanding this to process all document types (powerpoints, word docs, etc etc) is plug and play since docling handles the format-specific complexities and markdown conversion logic.

### How can this be tested?

**Testing via canvas course backpopulation**
1. Checkout this branch
2. go to the [openai dashboard](https://platform.openai.com/api-keys) and create a personal developer api key under the "semantic poc" project
3. set you OPENAI_API_KEY to your key and set OCR_MODEL to "gpt-5-nano-2025-08-07"
4. rebuild your web and celery containers `docker compose build celery web`
5. in [learning_resources/etl/utils.py](https://github.com/mitodl/mit-learn/pull/2777/files#diff-b8026e606dd8a6c55dcfd6f59feb6411dad157e9c0437ae17d54b3d476159a1bR1082) set debug_mode=True to see the actual content being converted as well as the resulting markdown - `converter = DoclingLLMConverter(pdf_path,debug_mode=True)`
6. ingest some canvas courses ```python manage.py backpopulate_canvas_courses --canvas-ids 28841,28842```
7. as the ingestion runs, you should see debug files/images get populated in the "ocr_debug" folder. each pdf has its own folder with sections that were OCR'd as well as the final markdown file that was generated.

**manually testing individual pdfs**
1. perform steps 1-4 from above
2. in a django shell run the following script to run this on specific pdfs:
```python
from pathlib import Path
import time
from learning_resources.converters.docling_llm_converter import DoclingLLMConverter
start = time.perf_counter()
converter = DoclingLLMConverter(Path("notes2.pdf"), debug_mode=True)
converter.convert_to_markdown()
end = time.perf_counter()
print(f"Took {end - start:.4f} seconds")
```

### Additional Context
1 - settings.OCR_PDF_MAX_PAGE_THRESHOLD currently controls the max number of pages a pdf can have for OCR. It is currently set to 15 but can be increased as we get a better sense of how long the process takes and costs (its mainly there to prevent us from processing very large pdfs). This setting is ignored for tutor problem files where we process all of them regardless of the number of pages


shanbady (commented):
left a comment about the change to CI config

PR #2778, opened 2025-12-01
display not found on dashboard program page if not enrolled

gumaerc:
Closes https://github.com/mitodl/hq/issues/9439

This PR adds a check of the user's program enrollments to `EnrollmentDisplay` before displaying program requirements. If the user is not enrolled in a program and they are trying to view it, `NotFoundPage` is returned instead.

### How can this be tested?
- Connect your instance of MIT Learn to an instance of MITx Online
- Make sure you have a program with some courses in it in your instance of MITx Online, and get the integer ID of said program
- Make another progam, but add a program enrollment for that program to your user
- Back on Learn, visit `/dashboard/programs/program-id` replacing `program-id` with the integer ID's of your programs
- Verify that the program you are enrolled in displays its requirements, and the one you are not enrolled in doesn't


jonkafton (changes requested):
Looks good. Just the change to return null or display a load skeleton while items are loading.

PR #2781, opened 2025-12-03
Article design layout and refactor

jonkafton:

Closes https://github.com/mitodl/hq/issues/9353

- Applies the article layout designs in Figma ([link](https://www.figma.com/design/Eux3guSenAFVvNHGi1Y9Wm/MIT-Design-System?node-id=27000-77728&m=dev))
  - Introducing a new article layout template with banner section.
  - Tiptap/ProseMirror schema driven layout provides for edit-in-place.
- Refactors all custom extension code to improve node organization and consistency with project guidelines.

Editor:
image

---
View:
image
image
image
image
image
image
image

---
Mobile:
image

image

### How can this be tested?

Run `COMPOSE_PROFILES=backend,frontend,keycloak,apisix docker compose up`

Open the editor view at http://open.odl.local:8062/articles/new.

Compose an article using all toolbar items with a focus our design elements https://www.figma.com/design/Eux3guSenAFVvNHGi1Y9Wm/MIT-Design-System?node-id=27000-77728&m=dev and:
- Images at default, wide and full views
- Headings
- Links
- Lists
- Blockquotes (code icon)

Save the article and view at http://open.odl.local:8062/articles/<id>
- Ensure elements are rendered according to designs
- Ensure the authorship/byline bar renders as expected 
I've added a sample article document here https://gist.github.com/jonkafton/e386b7a5d7ff205f9a7a7b648edb54f2 that reflects the design in Figma, which you can swap into ArticleDetailPage.tsx.

The change also refactors our custom extensions to date. These were built closer to the Tiptap library conventions while we have consolidating out approaches for code separation. These are now reworked towards our naming conventions, structure and styling and theming practices. Regression checks needed for:
- Editor controls for image embedding
- Video/media embedding and rendering
- Divider node (. . .)
- Resource list card embedding

### Additional Context

Not implemented / needed for production readiness:
- ~~The container scroll issue remains (footer can scroll up from window bottom). I have not found a solution for this that is self-contained in the editor components. The solution I have changes the base layout. As that is a global change I would like to do it in a separate PR.~~ Fixed
- "Save as draft" is nearing completion - https://github.com/mitodl/mit-learn/pull/2780
- Article slugs. Not strictly a blocker for production, but if articles are pubished before this is in place we will need extra code to redirect from the legacy URLs - https://github.com/mitodl/hq/issues/9460

### DIFF
--- a/frontends/ol-components/src/components/TiptapEditor/ArticleEditor.tsx
+++ b/frontends/ol-components/src/components/TiptapEditor/ArticleEditor.tsx
@@ -355,7 +355,7 @@ const ArticleEditor = ({ onSave, readOnly, article }: ArticleEditorProps) => {
               {(!article || !article?.is_published) && (
                 <Button
                   variant="secondary"
-                  disabled={isPending || !touched}
+                  disabled={isPending || !touched || !title}
                   onClick={() => {
                     handleSave(false)
                     setIsPublishing(false)


frontends/ol-components/src/components/TiptapEditor/extensions/node/ArticleByLineInfoBar/ArticleByLineInfoBar.tsx
```diff
@@ -0,0 +1,116 @@
...
+      <StyledWrapper>
+        <InnerContainer noAuthor={!author}>
+          {author && (
+            <InfoContainer>
+              <Avatar
+                alt={`${user?.first_name} ${user?.last_name}`}
+                src={avatarUrl}
+              >
+                JK
+              </Avatar>
```
ChristopherChudzicki: IMO the alt text is annoying here; the full name follows anyway.

Is JK a fallback? If I have avatarUrl set, I would see that instead?

frontends/ol-components/src/components/TiptapEditor/extensions/node/ArticleByLineInfoBar/ArticleByLineInfoBar.tsx
```diff
@@ -0,0 +1,116 @@
...
+}))
+
+const Spacer = styled.div({
+  marginBottom: "56px",
+})
+
+const ArticleByLineInfoBar = ({ node }: ReactNodeViewProps) => {
+  const { authorName, avatarUrl, readTime, publishedDate, editable } =
+    node.attrs
+  const { isFetching: isLoadingUser, data: user } = useUserMe()
```
ChristopherChudzicki: **Request:** `isLoading` would be better to use here.

- `isLoading` = true only on first load; if data is refreshing, it will be false
- `isFetching` = true during first load or any subsequent refresh.

And one thing that triggers user refetching is window focus. So the byline disappears if you switch tabs and switch back.

-----------------------

Also.... does this mean that the byline is tied to the current user viewing the post? Not the user who created it?

That's definitely not what we want, but might be an issue to handle separately.

frontends/ol-components/src/components/TiptapEditor/extensions/node/Image/ImageWithCaption.tsx
```diff
@@ -0,0 +1,216 @@
...
+}
+
+const Image = styled.img<{ layout: Layout }>(({ layout }) => ({
+  "&&": {
+    borderRadius: layout === Layout.full ? 0 : "8px",
+  },
+}))
+
+const Caption = styled.p(({ theme }) => ({
+  "&&&&&": {
```
ChristopherChudzicki: Seems like we are fighting with the vendored stylesheets.

Do you think they are generally useful, or should we remove them? Or should we keep the vendor discussion closed...

frontends/ol-components/src/components/TiptapEditor/ArticleEditor.tsx
```diff
@@ -127,26 +129,35 @@ const ArticleEditor = ({ onSave, readOnly, article }: ArticleEditorProps) => {
...
+          type: "byline",
         },
+        { type: "paragraph", content: [] },
       ],
     },
   )
   const [touched, setTouched] = useState(false)
 
+  useEffect(() => {
+    const title = content?.content?.[0].content?.[0]?.content?.[0]?.text || ""
```
ChristopherChudzicki: This seems fragile to me / unclear why it is the way it is.

How about
```ts
const searchContent = (
  content: JSONContent,
  matcher: (node: JSONContent) => boolean,
): JSONContent | undefined => {
  if (matcher(content)) {
    return content
  }
  if (Array.isArray(content.content)) {
    for (const child of content.content) {
      const result = searchContent(child, matcher)
      if (result) {
        return result
      }
    }
  }
  return undefined
}

  useEffect(() => {
    const heading = searchContent(
      content,
      (node) => node.type === "heading" && node.attrs?.level === 1,
    )
    const title = heading?.content?.[0]?.text
    setTitle(title)
  }, [content])
```

ChristopherChudzicki (commented):
Noticed one height issue.

------



Also, maybe we should throw `object-fit: cover` on images so if the aspect ratio isn't quite right, they don't get distorted?

I know there was a whole discussion about images and aspect ratios, but right now it's very easy to upload something that gets distorted badly.

Image

frontends/ol-components/src/components/TiptapEditor/extensions/node/ArticleByLineInfoBar/ArticleByLineInfoBar.tsx
jonkafton: No, I missed that on the refactor. It should display the current user only if in edit mode ie. current user is the editor (we don't actually display the bar in edit mode). Pushed the fix.

PR #2793, opened 2025-12-09
feature: add slug for article feature

ahtesham-quraish:
N/A

This feature is about adding slug, Have added the backend slug field and frontend will convert the title into slug on its first published. We have introduced new api which will return the article detail based on slug or id of the article. 

https://github.com/user-attachments/assets/0dc84916-3051-44c9-a420-45e3f7a9d1dd

### How can this be tested?
- Need to go to new article url
- As long as you have not published the article the slug will not be created once article is published then slug will be created and saved in backed
- You need to test the copy paste into editor as well because it was bug in code which I resolved

### DIFF
--- a/articles/views.py
+++ b/articles/views.py
@@ -1,20 +1,18 @@
 from django.conf import settings
+from django.shortcuts import get_object_or_404
 from django.utils.decorators import method_decorator
 from drf_spectacular.utils import (
+    OpenApiParameter,
     OpenApiResponse,
     extend_schema,
     extend_schema_view,
-    OpenApiParameter,
 )
 from rest_framework import status, viewsets
+from rest_framework.decorators import action
 from rest_framework.pagination import LimitOffsetPagination
 from rest_framework.permissions import IsAuthenticated
 from rest_framework.response import Response
 from rest_framework.views import APIView
-from django.shortcuts import get_object_or_404
-from rest_framework.decorators import action
-from drf_spectacular.utils import extend_schema, OpenApiParameter, OpenApiResponse
-
 
 from articles.models import Article
 from articles.serializers import RichTextArticleSerializer
--- a/articles/models.py
+++ b/articles/models.py
@@ -2,10 +2,10 @@
 
 from django.conf import settings
 from django.db import models
+from django.utils.text import slugify
 
 from main.models import TimestampedModel
 from profiles.utils import article_image_upload_uri
-from django.utils.text import slugify
 
 
 class Article(TimestampedModel):
--- a/articles/urls.py
+++ b/articles/urls.py
@@ -2,6 +2,7 @@ from django.urls import include, path, re_path
 from rest_framework.routers import SimpleRouter
 
 from articles import views
+
 from .views import MediaUploadView
 
 v1_router = SimpleRouter()


articles/views.py
```diff
@@ -78,6 +81,47 @@ def destroy(self, request, *args, **kwargs):
...
+            200: RichTextArticleSerializer,
+            404: OpenApiResponse(description="Not found"),
+        },
+    )
+    def get(self, request, identifier):
+        qs = Article.objects.all()
+
+        # Admins/staff/groups see everything
+        if not (is_admin_user(request) or is_article_group_user(request)):
+            qs = qs.filter(is_published=True)
```
ChristopherChudzicki: should this be in `get_queryset` ?

frontends/ol-components/src/components/TiptapEditor/extensions/lib/utils.ts
```diff
@@ -0,0 +1,109 @@
...
+          publishedDate: publishedDate || new Date().toLocaleDateString(),
+          readTime: "5 min read",
+          avatarUrl: "",
+        },
+      })
+      .run()
+  }
+}
+
+export const slugify = (title: string) => {
```
ChristopherChudzicki: I would lean toward slugifying on the backend... We usually use django's `slugify` util.

I noticed that the current slug field is also nullable. We could make it required and set it to the title always pre-publish:

```
from django.utils.text import slugify

class Article(TimestampedModel):

    def save(self, *args, **kwargs):
        previous = Article.objects.get(pk=self.pk) if self.pk else None
        was_published = getattr(previous, "is_published", None)
        if not was_published:
            self.slug = slugify(self.title)
        super().save(*args, **kwargs)
```

May also want to deal with duplicate slugs (that seems easier on the backend).

ahtesham-quraish: I have incorporated the above code and removed slugify from frontend. we have created the slug field with unique constraint and slug will be created only once so why we need to deal with duplicate?
```js
slug = models.SlugField(max_length=255, unique=True, blank=True, null=True)
```

articles/views.py
```diff
@@ -78,6 +80,47 @@ def destroy(self, request, *args, **kwargs):
         return super().destroy(request, *args, **kwargs)
 
 
+class ArticleDetailByIdOrSlugAPIView(APIView):
```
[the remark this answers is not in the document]
ChristopherChudzicki: After discussion, this was moved into an `@action` in order to reuse queryset and permissions.

ChristopherChudzicki (changes requested):
The slug changes look good 👍 

Could you fix the linting issue and also either 
- move image changes to separate PR (preferred)
- or add testing instructions for them (is there an issue?)

frontends/ol-components/src/components/TiptapEditor/extensions/node/Image/ImageWithCaption.tsx
```diff
@@ -196,7 +246,13 @@ export function ImageWithCaption({
           </div>
         )}
 
-        <Image src={src} alt={alt || caption} layout={layout} />
+        <Image
+          src={src}
+          alt={alt || caption}
+          layout={layout}
+          ref={imgRef}
+          className={`${!canExpand ? "img-contained" : ""}`}
```
ChristopherChudzicki: We should not do `className={`${!canExpand ? "img-contained" : ""}`}`.

The display of an image should only depend on the node `attrs`, not on the non-node state.

This leads to some odd behavior right now where:
1. Insert a small image
2. It looks ok, good.
3. Delete it
4. Press undo
5. Now the image looks huge. (I guess the value of `canExpand` is wrong in this scenario).

The above would be avoided if the display is a function of only the node attributes. 

**Suggestion:** Should this instead be based on `node.layout === "default"` ?

frontends/ol-components/src/components/TiptapEditor/extensions/node/Image/ImageWithCaption.tsx
```diff
@@ -196,7 +246,13 @@ export function ImageWithCaption({
           </div>
         )}
 
-        <Image src={src} alt={alt || caption} layout={layout} />
+        <Image
+          src={src}
+          alt={alt || caption}
```
ChristopherChudzicki: A screenreader will read the alt text then the caption, so using the caption as a fallback for the alt text doesn't seem beneficial to me. It would just result in the caption being read twice.

I'd leave this as `alt={alt}`; if there's no alt text...well, at least it will read the caption.

(That said, at some point, we should figure out how to make alt text required.)

ChristopherChudzicki (changes requested):
Noticed a couple issues with the image uploading.

**Suggestion:** Move the image-related changes to a separate branch, we can merge the slug stuff as-is.

In general, it would be good to do distinct features on distinct branches.

frontends/ol-components/src/components/TiptapEditor/extensions/node/Image/ImageWithCaption.tsx
ahtesham-quraish: I have changed the logic you can check now

PR #2803, opened 2025-12-12
feat: add callout node as per design along with its button

ahtesham-quraish:
 N/A

This is about creating callout feature in editor which has black background and it will be available in insert dropdown as a new control user can click and and area look like code block will be displayed and user can start typing in it. 
https://www.figma.com/design/Eux3guSenAFVvNHGi1Y9Wm/MIT-Design-System?node-id=27000-77728&t=KHRXrwUPBrwGTRup-0

Screenshot 2025-12-18 at 4 34 29 PM

Screenshot 2025-12-18 at 4 34 38 PM

### How can this be tested?
- Go to new article page where you will see the insert dropdown click on that button 
- dropdown will be displayed where you will see button along with text `Callout`
- click on that button and callout area will be rendered

### DIFF
--- a/frontends/ol-components/src/components/TiptapEditor/extensions/node/Callout/Callout.ts
+++ b/frontends/ol-components/src/components/TiptapEditor/extensions/node/Callout/Callout.ts
@@ -1,16 +1,20 @@
-import Blockquote from "@tiptap/extension-blockquote"
+import { Node, mergeAttributes } from "@tiptap/core"
 
-export const Callout = Blockquote.extend({
+export const Callout = Node.create({
   name: "callout",
+
+  group: "block",
+  content: "block+",
+  defining: true,
 
   parseHTML() {
     return [
-      { tag: "callout" }, // Custom tag
+      { tag: "callout" },
+      { tag: "div[data-type='callout']" }, // optional fallback
     ]
   },
 
-  // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  renderHTML({ HTMLAttributes }: { HTMLAttributes: { [key: string]: any } }) {
-    return ["callout", HTMLAttributes, 0]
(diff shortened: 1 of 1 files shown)


frontends/ol-components/src/components/TiptapEditor/extensions/node/Quote/Quote.ts
```diff
@@ -0,0 +1,17 @@
+import Blockquote from "@tiptap/extension-blockquote"
+
+export const Quote = Blockquote.extend({
+  name: "quote",
+
+  parseHTML() {
+    return [
+      { tag: "quote" }, // Custom tag
```
ChristopherChudzicki: General comment: Let's stick to regular / valid HTML tags unless there's a very good reason.

ahtesham-quraish:
> It would be good if we could try and recover the behavior in https://tiptap.dev/docs/examples/basics/default-text-editor... I took a quick look and if I disable our custom schema, it behaves the "good" way. Unsure why, though.

A: I have made the change in which I am ensuring that the quote node must have paragraph node so that user can move away from quote node easily like in https://tiptap.dev/docs/examples/basics/default-text-editor 

B: The other approach that can be that we can insert the paragraph node after the quote node and user can delete it if he/she wants to delete the paragraph this approach is slightly different than the above (A)

ahtesham-quraish:
> I noticed that there is an item for block quotes in https://github.com/mitodl/hq/issues/9558, but it seems to have been mostly implemented on RC already (it is missing the red quotation mark):

Yes agree the `blockquote` already there but it is vendor code and our one is custom extension with customization like inserting paragraph after quote node and inverted quotes as per design, so point is we have separated out the vendor code and custom extension code. What we can do is we can remove the default one and keep the customized one. what do you think?

PR #2804, opened 2025-12-12
More efficient S3 key retrieval for edX contentfile archives

mbertrand:
Short-term fix for https://github.com/mitodl/hq/issues/9587

Refactors the code so that just the relevant attributes of each S3 object (key, last modified) are added to the list used to find the most recent archives.

### How can this be tested?
- In a shell, run the following:
  ```
  from learning_resources.etl.edx_shared import *
  get_most_recent_course_archives("mitxonline")
  ```

While that is running, use `docker stats` in separate terminal to monitor memory usage.  It might go up by ~150MB or so.

It should complete successfully and return a list of strings with a prefix close to today's date (ex: `20251212/courses/course-v1:MITx+Template.PC.Firefox+04_05_2023.tar.gz')

If you try the same thing on the main branch, the memory usage will be significantly higher.

### DIFF
--- a/learning_resources/etl/edx_shared.py
+++ b/learning_resources/etl/edx_shared.py
@@ -91,7 +91,7 @@ def get_most_recent_course_archives(
             reversed(  # noqa: C413
                 sorted(
                     [
-                        obj
+                        (obj.key, obj.last_modified)
                         for obj in bucket.objects.filter(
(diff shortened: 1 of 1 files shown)


abeglova (approved):
lgtm

PR #2810, opened 2025-12-15
Adjust org dashboard loading skeletons

gumaerc:
Follow-up to https://github.com/mitodl/mit-learn/pull/2792

After we merged the above PR and got it into RC for testing, an issue was noticed where because of the significant volume of data in RC the initial loading skeleton was displayed for longer than when testing locally. The enrollments and program enrollments endpoints were to blame, and because the above PR shows an initial loading skeleton until all of that information is available, it seems to be extra slow. This PR removes enrollments from that check, allowing loading of the downstream components to continue while waiting for enrollments data to come in.

### How can this be tested?
- Follow the instructions in the above PR to set up some test data
- Load the org dashboard with throttling turned on in your browser and watch the network tab
- You should see an initial loading skeleton while programs and program collections are being fetched, then the program titles should be displayed with loading skeletons under them while it loads the course data for each course in the program / collection

### DIFF
--- a/frontends/main/src/app-pages/DashboardPage/OrganizationContent.tsx
+++ b/frontends/main/src/app-pages/DashboardPage/OrganizationContent.tsx
@@ -462,13 +462,8 @@ const OrganizationContentInternal: React.FC<
     </Stack>
   )
 
-  // Wait for all top-level queries to complete before rendering content
-  if (
-    programs.isLoading ||
-    programCollections.isLoading ||
-    courseRunEnrollments.isLoading ||
-    programEnrollments.isLoading
-  ) {
+  // Wait for all program and collection data to load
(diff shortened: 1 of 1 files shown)


ChristopherChudzicki:
I think this is good enough to unblock the current RC—revert to the old behavior.

I think we have two issues.
1. UI loads slowly because using slow APIs
2. UI looks janky while it loads.

Obviously (1) is the biggest issue, but (2) could be addressed a bit better, too: The slowest ui-blocking api is `/v2/coures/?ids=lots_of_ids`. At that point, we know how many cards should be displaying. We could show the appropriate number of dashoard cards in a loading state. (And prior to that, I'd probably show some fixed number in a loading state, like 5 or 6.)

PR #2827, opened 2025-12-19
Prevent query devtools from erroring on image upload

ChristopherChudzicki:
None

This addresses a small issue with our image upload functionality in conjunction with Tanstack Query devtools.

### How can this be tested?
This error is only visible if you have the Tanstack Query devtools enabled. These can be installed from LINK
1. Try uploading an image on  https://rc.learn.mit.edu/articles/new You'll see an error like:
    > 4821-2c216e271775abe8.js:12 Error sending message to content script: DataCloneError: Failed to execute 'postMessage' on 'Window': e=>{var l;let r=Math.round(100*e.loaded/(null!=(l=e.total)?l:1));null==t||t(r)} could not be cloned.
2. Try again locally on this branch. No error.

### DIFF
--- a/frontends/api/src/hooks/articles/index.ts
+++ b/frontends/api/src/hooks/articles/index.ts
@@ -77,6 +77,15 @@ export const useMediaUpload = () => {
 
   return {
     ...mutation,
+    /**
+     * Set a callback to be called on the next upload progress event.
+     *
+     * NOTES:
+     * - This callback will be cleared after the mutation settles (either success or error).
+     * - This is a separate method, not part of the mutate/mutateAsync options,
+     *   to avoid errors with function serialization. (E.g., Tanstack Query
+     *   devtools attempt to serialize mutation options.)
+     */
     setNextProgressCallback: (
       callback: ((percent: number) => void) | undefined,
     ) => {


frontends/ol-components/src/components/TiptapEditor/ArticleEditor.tsx
```diff
@@ -190,13 +190,10 @@ const ArticleEditor = ({ onSave, readOnly, article }: ArticleEditorProps) => {
...
       async (file: File, progressCb?: (percent: number) => void) => {
         try {
-          const response = await uploadImage.mutateAsync({
-            file,
-            onUploadProgress: (e) => {
-              const percent = Math.round((e.loaded * 100) / (e.total ?? 1))
-              progressCb?.(percent)
-            },
-          })
+          uploadImage.setNextProgressCallback(progressCb)
```
ChristopherChudzicki: So the reason the error happens is:
1. Tanstack query devtools are a chrome extension, so they run in a webworker
2. The devtools track mutation state, which includes the args we pass to `mutate` (or `mutateAsync`).
3. In order for that data to get into the webworker, the data must be serializable. Previously we were passing a function, which was not serializable. 

Now `onUploadProgress` is no longer an arg of mutateasync, which avoids the issue.

I don't love changing the application code to avoid a devtool error. However, it is useful to not have these errors showing up, and this alternative implementation seems fine.

PR #2832, opened 2025-12-23
fix: improvements in learning resource

ahtesham-quraish:
After discussion with Chris we finalized the following thing https://github.com/mitodl/hq/issues/9684#issuecomment-3682869381

There are improvements we need to do like given below
- Consolidate multiple calls into single one on reader side
- Remove the dialog that is used for taking the input from user instead we should copy paste the resource url directly inside the editor and hit enter it will add the resource card node.

https://github.com/user-attachments/assets/823b8e88-f4fd-4429-b2df-8600be38112b

### How can this be tested?
- Need to insert the url of resource for local you need to insert something like this `http://open.odl.local:8062/search?resource=297` and click enter it will add the resource card node in editor. On reader side you can refresh the page and see the network `bulk` api call which will have all the resource id.
- On hover the resource card you would see the cross button at the right top corner of the resource card which will be used for delete.


zamanafzal (approved):
Minor tweaks needed. LGTM 👍🏻

PR #2834, opened 2025-12-31
feat: add course detail in resource card

ahtesham-quraish:
https://github.com/mitodl/hq/issues/9737

The Learning Resource Card has been introduced in the TipTap editor to improve the authoring experience and enable richer, reusable learning content. However, the current implementation is missing key interactive features that already exist elsewhere in the platform—specifically in the MIT Learn Search and Dashboard experiences.

https://github.com/user-attachments/assets/ae0085a6-b0dc-4e33-95f1-431a06f51caa

### How can this be tested?
- You need to embed the following url `http://open.odl.local:8062/search?resource=<resource_id>` and it will insert the course card inside the editor
- After the above step, you need to click on the course card it will open the drawer from the right 
- You need to click on bookmark icon on the course card it will open the user list.


ahtesham-quraish:
> > We have lost the insert menu item at some point. The ability to paste a URL to insert a resource card is certainly a nice-to-have, but editors have no way of knowing about this functionality without documentation.
> 
> That is a good point. We should find a way so we can keep the item in the menu as optional. When users click the item ..
> 
> Screenshot 2026-01-07 at 12 17 56 PM

Ok I will create new PR to bring that option back

PR #2836, opened 2026-01-05
fix: add delete button for image and media node

ahtesham-quraish:
https://github.com/mitodl/hq/issues/9747

Currently, image and media nodes in the TipTap editor can only be removed by selecting the node and using keyboard actions. To improve usability and maintain consistency with the existing Learning Resource Card node, we want to add a visible delete (cross) icon to image and media nodes.

The cross icon should appear when the editor hovers over the node and allow the editor to remove the node with a single click. This provides a more intuitive and discoverable interaction, especially for non-technical editors.

https://github.com/user-attachments/assets/09f536d7-3629-4f93-874d-c7a0de2effc2

### How can this be tested?
For this you just need to insert image or media node and hover on them you would see the cross icon which will be used to delete the icon


zamanafzal:
I’ve tested the changes, and the cross icon appears for both videos and images. 
One issue to note: when an image is uploading and we try to add a video, the image gets automatically replaced by the video instead of allowing both to be added together.

zamanafzal (approved):
LGTM 👍🏻

PR #2850, opened 2026-01-12
Article editor tests

jonkafton:

Close https://github.com/mitodl/hq/issues/9863

Adds coverage for the Article Editor basic functionality and rendering.

Uses [Happy DOM](https://github.com/capricorn86/happy-dom/wiki/) as the test environment for testing edit mode as TipTap accesses DOM APIs not made available by JSDOM.

### How can this be tested?

Tests should run successfully on CI. Locally run `yarn test ArticleViewer` and `yarn test ArticleEditor`

### DIFF
--- a/frontends/main/src/page-components/TiptapEditor/ArticleEditor.happydom.test.tsx
+++ b/frontends/main/src/page-components/TiptapEditor/ArticleEditor.happydom.test.tsx
@@ -19,20 +19,6 @@ jest.mock("posthog-js/react", () => ({
 
 const mockOnSave = jest.fn()
 
-const getMakeRequestCalls = () => {
-  return (makeRequest as unknown as ReturnType<typeof jest.fn>).mock.calls
-}
-
-const findApiCall = (
-  method: string,
-  url: string,
-): [string, string, unknown] | undefined => {
-  const calls = getMakeRequestCalls()
-  return calls.find((call) => call[0] === method && call[1] === url) as
-    | [string, string, unknown]
-    | undefined
-}
-
 describe("ArticleEditor - Content Editing and Saving", () => {
   beforeEach(() => {
     mockOnSave.mockClear()
@@ -109,12 +95,14 @@ describe("ArticleEditor - Content Editing and Saving", () => {
               content: [
                 {
                   type: "heading",
-                  attrs: { level: 1 },
+                  attrs: { level: 1, textAlign: null },
                   content: [{ type: "text", text: "Updated Title" }],
                 },
                 {
                   type: "paragraph",
-                  content: [],
+                  attrs: {
+                    textAlign: null,
+                  },
                 },
               ],
             },
@@ -123,7 +111,9 @@ describe("ArticleEditor - Content Editing and Saving", () => {
             },
             {
               type: "paragraph",
-              content: [],
+              attrs: {
+                textAlign: null,
+              },
             },
           ],
         },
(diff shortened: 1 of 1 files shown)


jonkafton:
Setting to draft for now as this needs to go in first https://github.com/mitodl/mit-learn/pull/2847 (this PR branches off that one).

frontends/main/src/page-components/TiptapEditor/ArticleEditor.happydom.test.tsx
```diff
@@ -0,0 +1,1510 @@
...
+ * elements not supported by JSDOM, the default environment in Jest.
+ */
+import React from "react"
+import { screen, waitFor, fireEvent } from "@testing-library/react"
+import userEvent from "@testing-library/user-event"
+import { QueryClient, QueryClientProvider } from "@tanstack/react-query"
+import { Provider as NiceModalProvider } from "@ebay/nice-modal-react"
+import { setMockResponse, factories, urls, makeRequest } from "api/test-utils"
+import { userQueries } from "api/hooks/user"
+import { renderWithTheme } from "../../../../ol-components/src/test-utils"
```
ChristopherChudzicki: **Question:** Can we use regular `renderWithProviders`? This file uses a customized one without the theme, then also adds the theme via renderWithTheme.

**Request:** If we can use it, let's use the regular renderWithProviders. If we can't, we should get rid of this import and just add ThemePRovider manually or import from `ol-components` without the relative stuff.

frontends/main/src/page-components/TiptapEditor/ArticleEditor.happydom.test.tsx
```diff
@@ -0,0 +1,1510 @@
+/**
+ * @jest-environment @happy-dom/jest-environment
+ *
+ * Using the Happy DOM environment as the editor accesses DOM APIs and uses contenteditable
+ * elements not supported by JSDOM, the default environment in Jest.
```
ChristopherChudzicki: Random comment: This is so odd to me. The internet always insists JSDOM is more complete.

I dunno if that's an outdated assessment, or if happy-dom just happens to better mimic browsers for this use case.

frontends/main/src/page-components/TiptapEditor/ArticleEditor.happydom.test.tsx
```diff
@@ -0,0 +1,1510 @@
...
+    <QueryClientProvider client={queryClient}>
+      <NiceModalProvider>{children}</NiceModalProvider>
+    </QueryClientProvider>
+  )
+
+  return renderWithTheme(<Wrapper>{component}</Wrapper>)
+}
+
+const getMakeRequestCalls = () => {
+  return (makeRequest as unknown as ReturnType<typeof jest.fn>).mock.calls
```
ChristopherChudzicki: Shouldn't need all the typecasting here.

Image

frontends/main/src/page-components/TiptapEditor/ArticleEditor.happydom.test.tsx
```diff
@@ -0,0 +1,1510 @@
...
+            name: /Update|Publish/,
+          })
+          expect(updateButton).not.toBeDisabled()
+        },
+        { timeout: 3000 },
+      )
+
+      const updateButton = screen.getByRole("button", {
+        name: /Update|Publish/,
+      })
```
ChristopherChudzicki: **Request:** Should be able to replace all of this by
```ts
      const updateButton = await screen.findByRole(
        "button",
        { name: /Update|Publish/ },
        { timeout: 3000 },
      )
```

I'd also suggest asking Copilot to look for other such replacements.

**Aside:** Sometimes Copilot itself writes code like the original for me. I do think it's worth sticking for `findBy` to make things less verbose / easier to read. When I have copilot write tests, usually mys trategy is to closely look at one or two of its tests then ask it to apply any changes I suggest everywhere possible.

frontends/jest-shared-setup.ts
```diff
@@ -61,13 +61,11 @@ class FakeResizeObserver {
 }
 const polyfillResizeObserver = () => {
   if (window.ResizeObserver !== undefined) {
-    /**
-     * If this throws... I guess our test env supports it natively now.
```
ChristopherChudzicki: ha, i forgot about that.

If you want to keep the error, I think you could use
```ts
  if (
    window.ResizeObserver !== undefined &&
    // @ts-expect-error happyDOM specific property
    window.happyDOM === undefined
  ) {
```

ChristopherChudzicki (changes requested):
This is great. I left a few suggestions/requests.

ChristopherChudzicki:
@jonkafton you may want a rebase, too... looks like openapi diff is picking up some changes.

frontends/main/src/page-components/TiptapEditor/ArticleEditor.happydom.test.tsx
jonkafton: That's removed completely, no longer used.

frontends/main/src/page-components/TiptapEditor/ArticleEditor.happydom.test.tsx
jonkafton: Hopefully will improve with your new readme. I'm finding the same with Cursor + Composer 1. This did get me thinking that as copilot and friends get better at cranking out tests, the manual work to carefully craft them feels increasingly redundant. I wonder if sloppy AI driven tests (that work and quickly provide good coverage, if themselves decreasingly succinct or manually maintainable) perhaps even test code at all, are a transitional artifacts as the test effort delegates towards AI; which ultimately promises to test systems on our behalf based off feature spec. Your approach of adding documentation to instruct agents feels exactly right - setting good constraints and patterns for clarity and signal, which are likely to become more important than the actual detail in the test code. If tests become less durable a document of system behavior for human consumption, I wouldn't be surprised to see a push back towards detailed human-readable plain language feature specs (remember BDD!) that express sufficient intent for fully automated no-code test systems to gateway acceptance with confidence; which would be a good thing.

PR #2860, opened 2026-01-13
fix program collection ordering

gumaerc:
Closes https://github.com/mitodl/hq/issues/9866

This PR sets up the program collections on B2B org dashboards to be ordered by the explicit `order` property returned along with each program in the collection. In https://github.com/mitodl/mitxonline/pull/3007, this `order` property was added to the API, reflecting the order set when adding programs to the collection through Wagtail. This didn't take effect immediately because of the way the dashboard has to query the courses based on those programs. Program collections are displayed in the dashboard by fetching the first course from each program in the collection and displaying a card for that course. Both the program and course listing endpoints in the MITx Online API accept a list of ID's in the query string to filter the request on a specific subset of data. The API does not return these results in a deterministic order, which is why the order property was added to program collections. This PR sets up the frontend here in Learn to take that into account and properly order the courses shown in the dashboard as they are related to the program collection.

### How can this be tested?
- Connect your instance of Learn to an instance of MITx Online as described in the Readme
- Create a program with a few courses as requirements
- Create two other programs that require:
  - The first program we created
  - A unique course not included in the first program
- In your instance of MITx Online, create a B2B organization and add your user to it, then use the B2B management commands described in the MITx Online docs to create a B2B contract and add all of the programs we created above to the contract
- In the MITx Online CMS, create a program collection and add the "other" programs we created besides the first one and publish it
- View the Learn dashboard, and click on the organization you created
- You should see:
  - The first program we created
  - The program collection with a card representing each program in the collection, showing the course that you added that's not part of the first program in the order that you added the programs to the collection
- Back in the MITx Online CMS, re-order the programs in the collection
- Refresh the Learn org dashboard and verify that the order changes to reflect what you set

### DIFF
--- a/frontends/main/src/app-pages/DashboardPage/ContractContent.tsx
+++ b/frontends/main/src/app-pages/DashboardPage/ContractContent.tsx
@@ -206,19 +206,14 @@ const useProgramCollectionCourses = (
   })
   const isLoading = programsQuery.isLoading
 
-  const programsWithCourses = programsQuery.data?.results
-    .map((program) => {
-      if (!program) {
-        return null
-      }
-      const transformedProgram = transform.mitxonlineProgram(program)
-      return {
-        programId: program.id,
-        program: transformedProgram,
-        hasCourses: program.courses && program.courses.length > 0,
-      }
-    })
-    .filter(Boolean)
+  const programsWithCourses = programsQuery.data?.results.map((program) => {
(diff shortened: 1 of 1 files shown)


ChristopherChudzicki (changes requested):
small request (remove unnecessary sort) and a suggestion

PR #2868, opened 2026-01-16
remove intermediate dashboard classes and use mitxonline api classes directly

gumaerc:
Closes https://github.com/mitodl/hq/issues/9865

This PR removes the intermediate classes the dashboard uses when dealing with data coming from the MITx Online API. These classes were defined in `types.ts` and a series of functions in `transform.ts` would take the raw API object from the MITx Online Axios API and generate the intermediate objects. This was originally done because it was assumed that data would be coming from multiple upstream API's, but in reality all the data for our courseware will be coming from MITx Online. Instead, the objects returned by the Axios package are used directly.

### How can this be tested?
TBD

### DIFF
--- a/frontends/main/src/app-pages/DashboardPage/CoursewareDisplay/DashboardCard.tsx
+++ b/frontends/main/src/app-pages/DashboardPage/CoursewareDisplay/DashboardCard.tsx
@@ -585,11 +585,7 @@ const DashboardCard: React.FC<DashboardCardProps> = ({
       : undefined
 
   const hasValidCertificate =
-    resource.type === DashboardType.CourseRunEnrollment
-      ? !!resource.data.certificate?.uuid
-      : resource.type === DashboardType.ProgramEnrollment
-        ? !!resource.data.certificate?.uuid
-        : false
+    resource.type !== DashboardType.Course && !!resource.data.certificate?.uuid
 
   const enrollmentStatus: EnrollmentStatus = isAnyCourse
     ? hasValidCertificate


frontends/main/src/app-pages/DashboardPage/CoursewareDisplay/DashboardCard.tsx
```diff
@@ -489,95 +481,169 @@ const CourseStartCountdown: React.FC<{
...
+  const hasValidCertificate = !!enrollmentData?.certificate?.uuid
+  const enrollmentStatus: EnrollmentStatus = isAnyCourse
+    ? hasValidCertificate
+      ? EnrollmentStatus.Completed
+      : enrollmentData && "run" in enrollmentData
+        ? getEnrollmentStatus(enrollmentData)
+        : EnrollmentStatus.NotEnrolled
+    : enrollmentData && "status" in enrollmentData
+      ? (enrollmentData.status as EnrollmentStatus)
+      : EnrollmentStatus.NotEnrolled
```
ChristopherChudzicki: minor suggestion: Maybe would be simpler as `getEnrollmentStatus(resource: DashboardCourse): EnrollmentStatus`. Could remove the existing `getEnrollmentStatus` (which only handles courserunenrollments) or expand it.

frontends/main/src/app-pages/DashboardPage/CoursewareDisplay/DashboardCard.tsx
```diff
@@ -603,65 +669,91 @@ const DashboardCard: React.FC<DashboardCardProps> = ({
...
 
   const menuItems = contextMenuItems.concat(
-    isCourse && enrollment?.id
-      ? getDefaultContextMenuItems(title, enrollment)
+    isAnyCourse &&
+      enrollmentData &&
+      "run" in enrollmentData &&
+      enrollmentData.id
+      ? getDefaultContextMenuItems(title, enrollmentData)
       : [],
```
ChristopherChudzicki: IMO, this component would be simpler if you got rid of `enrollmentData` and just used `resource`.

For example, `isAnyCourse && enrollmentData && "run" in enrollmentData &&  enrollmentData.id` is the same as `resource.type === "DashboardType.CourseRunEnrollment"`, so the above could just be
```ts
  const menuItems = contextMenuItems.concat(
    resource.type === DashboardType.CourseRunEnrollment
      ? getDefaultContextMenuItems(title, resource.data)
      : [],
  )
```
The certificate link calculation could be
```ts
const getCertificateLink = (resource: DashboardResource): string | null => {
  if (resource.type === DashboardType.CourseRunEnrollment) {
    const uuid = resource.data.certificate?.uuid
    // OR use a url from common?
    return uuid ? `/certificate/course/${uuid}` : null
  }
  if (resource.type === DashboardType.ProgramEnrollment) {
    const uuid = resource.data.certificate?.uuid
    return uuid ? `/certificate/program/${uuid}` : null
  }
  return null
}
```
etc

PR #2870, opened 2026-01-20
feat: add dialog and button for learning resource input

ahtesham-quraish:
N/A

We need to implement the learning resource dialog that is used to take the resource url input from editor and also the button inside the insert dropdown control. The purpose of this is to provide support for manually handing the input in case editor is not aware of pasting the url.

### How can this be tested?
For testing you need go to new article page where in controls menu you would see the insert dropdown that is where you will see Add course button that is used to insert the url of learning resource card

### DIFF
--- a/frontends/main/src/page-components/TiptapEditor/extensions/ui/LearningResource/useLearningResourceEmbed.ts
+++ b/frontends/main/src/page-components/TiptapEditor/extensions/ui/LearningResource/useLearningResourceEmbed.ts
@@ -3,7 +3,7 @@ import type { Editor } from "@tiptap/react"
 import { useTiptapEditor } from "../../../vendor/hooks/use-tiptap-editor"
 import NiceModal from "@ebay/nice-modal-react"
 import CourseUrlInputDialog from "./ResourceUrlInputDialog"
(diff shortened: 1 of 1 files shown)


frontends/main/src/page-components/TiptapEditor/extensions/ui/LearningResource/useLearningResourceEmbed.ts
```diff
@@ -0,0 +1,43 @@
+import { useCallback } from "react"
+import type { Editor } from "@tiptap/react"
+import { useTiptapEditor } from "../../../vendor/hooks/use-tiptap-editor"
+import NiceModal from "@ebay/nice-modal-react"
+import CourseUrlInputDialog from "./ResourceUrlInputDialog"
+import { Icon } from "./Icon" // you create this SVG
```
zamanafzal: I think this comment was added mistakenly.

ahtesham-quraish:
> I’ve tested the changes, and they look good to me. One thing we should look into is the flow for adding multiple courses. Currently, the user has to open the Insert dropdown, add a course card, return to the editor, and then manually add space. Otherwise, adding another course card replaces the existing one.

We can improve this later on

PR #2884, opened 2026-01-22
Default QDRANT_DENSE_MODEL to None

shanbady:

Closes https://github.com/mitodl/hq/issues/9953

This PR reverts the default dense model back to None (instead of the openai specific text-embedding-3-small) so developer environments have a working embedding setup without any extra configuration or dependencies.

### How can this be tested?
1. git clone this repo into a fresh directory
2. do a docker compose up
3. in the django shell run the following:
```python
from vector_search.utils import dense_encoder
encoder = dense_encoder()
```
4. note that it fails with the following error:
```python
ValueError: Model text-embedding-3-small is not supported in TextEmbedding. Please check the supported models using `TextEmbedding.list_supported_models()`
```
5. checkout this branch and note that it instantiates with out errors

### DIFF
--- a/learning_resources_search/api_test.py
+++ b/learning_resources_search/api_test.py
@@ -2414,6 +2414,7 @@ def test_execute_learn_search_with_hybrid_search(mocker, settings, opensearch):
     }
 
     settings.DEFAULT_SEARCH_MODE = "best_fields"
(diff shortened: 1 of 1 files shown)


rhysyngsun (approved):
LGTM

PR #2920, opened 2026-02-04
Remove some dependencies, clean up others

ChristopherChudzicki:
None

This PR cleans up some dependencies:
- removes a few unused dependencies (e.g., `tinycolor2`, `react-select`)
- moves some deps from `ol-components` to `main` that are for tiptap vendor code. (Normally our linting ensures deps are listed in correct package.json, but we exempted tiptap vendor code from those checks.)
- Fixes `syncpack`, which was supposed to ensure consistent deps across our workspaces.
    - This ensures different workspaces have same version, avoiding issues like recent test failures in #2372 which was due to conflicting versions of nextjs.
    - I think syncpack broke when we started having 3 levels of workspaces. Ideally we'd have only 2 levels, but that's a much igger change. For now, we run it from `/frontends` dir.

### How can this be tested?
Tests should pass, app should run/build.

### DIFF
--- a/frontends/ol-components/package.json
+++ b/frontends/ol-components/package.json
@@ -57,6 +57,7 @@
     "@testing-library/user-event": "^14.5.2",
     "@types/lodash.throttle": "^4.1.9",
     "@types/react-slick": "^0.23.13",
+    "@types/validator": "^13.7.6",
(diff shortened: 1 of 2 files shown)


jonkafton (approved):
:+1: no issues.

PR #2922, opened 2026-02-04
Remove user_lists, learning_path_parents entirely from serializers

mbertrand:
Closes https://github.com/mitodl/hq/issues/10089

Removes the `learning_path_parents` and `user_list_parents` fields entirely from learning resource serializers.
This will cause lots of failures on the openapi test but they can be ignored.

### How can this be tested?
- Log in as an admin, go to the search page, add some courses to your Favorites and learning paths, refresh the page.  You should see courses in your paths/lists marked as such before and after the refresh.
- Go to http://open.odl.local:8062/learningpaths.  You should still be able to see your learning paths, the list of resources in each, etc.
- Go to http://open.odl.local:8062/dashboard/my-lists.  You should still be able to see your userlists, the list of resources in each, etc.

### DIFF
--- a/frontends/api/src/hooks/learningResources/queries.ts
+++ b/frontends/api/src/hooks/learningResources/queries.ts
@@ -16,7 +16,6 @@ import type {
   PlatformsApiPlatformsListRequest,
   FeaturedApiFeaturedListRequest as FeaturedListParams,
   LearningResourcesApiLearningResourcesItemsListRequest as ItemsListRequest,
-  LearningResourcesSearchResponse,
   LearningResourcesApiLearningResourcesSummaryListRequest as LearningResourcesSummaryListRequest,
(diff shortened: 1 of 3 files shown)


frontends/api/src/hooks/learningResources/queries.ts
```diff
@@ -191,7 +172,7 @@ const learningResourceQueries = {
           (res) =>
```
ChristopherChudzicki: same, this could jsut return `res.data`, and no need for the typecast to LearningResourcesSearchResponse anymore

same comment a few other places

ChristopherChudzicki (approved):
👍 I think it's worth just returning `res.data` in the couple places you had to change, but this is ok too.

Thanks for getting rid of these!

PR #3135, opened 2026-03-31
feat: video playlist page design implementation

ahtesham-quraish:
https://github.com/mitodl/hq/issues/10734

MIT Learn already has a video playlist object, but it does not exist as a meaningful user-facing thing.

Right now playlists live as backend metadata. Users mostly land on individual videos. There is no real page for the playlist itself, and no clear way to see or navigate the grouping. Even when the content is clearly meant to belong together, it shows up as isolated videos.

Desktop:
image (2)

Mobile:
image

### How can this be tested?

if you dont have playlist data locally then in frontend env file you should add the following variable `NEXT_PUBLIC_MITOL_API_BASE_URL="https://api.rc.learn.mit.edu"` it will connect with RC learn backend
then we need to enable the flag which is `video-playlist-page` and then visit the following url `http://open.odl.local:8062/playlist/6831`


frontends/main/src/app/playlist/[id]/page.tsx
```diff
@@ -0,0 +1,41 @@
...
+
+  await Promise.all([
+    queryClient.prefetchQuery(videoPlaylistQueries.detail(playlistId)),
+    queryClient.prefetchQuery(
+      learningResourceQueries.items(playlistId, {
+        learning_resource_id: playlistId,
+      }),
+    ),
+  ])
+
```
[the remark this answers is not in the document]
Comment (ahtesham-quraish) [reply frontends/main/src/app/playlist/[id]/page.tsx]: Not sure about it need to think more on it

arslanashraf7:
The design in this PR has been approved by @Ferdi and @mbilalmughal as per @ahtesham-quraish . The conversation was done over Slack.

arslanashraf7:
@zamanafzal could you review this PR once ready?

frontends/main/src/app-pages/VideoPlaylistCollectionPage/VideoPageHeader.tsx
```diff
@@ -0,0 +1,116 @@
...
+
+  return (
+    <>
+      <BreadcrumbBar>
+        <VideoContainer>
+          <StyledBreadcrumbs
+            variant="light"
+            ancestors={[
+              { href: "/", label: "Home" },
+              { href: "/videos", label: "Videos" },
```
zamanafzal: This breadcrumb points to /videos, but there is no matching route in the app router. Clicking "Videos" leads to a dead page. Can we update this to a valid destination (or remove it) to avoid broken navigation?

frontends/main/src/app-pages/VideoPage/VideoDetailPage.tsx
```diff
@@ -0,0 +1,543 @@
...
+  const isLoading = videoLoading || (!!playlistId && playlistLoading)
+
+  return (
+    <PageWrapper>
+      {/* Breadcrumb */}
+      <BreadcrumbBar>
+        <Container>
+          <StyledBreadcrumbs
+            variant="light"
+            ancestors={[{ href: `/playlist/${playlistId}`, label: "Playlist" }]}
```
zamanafzal: If playlistId is missing from search params, this renders /playlist/null in the breadcrumb link. Repro: open /playlist/detail/<videoId> without ?playlist=. Suggest guarding this link when playlistId is null or falling back to a safe route.

Image

frontends/main/src/app-pages/VideoPage/VideoDetailPage.tsx
```diff
@@ -0,0 +1,543 @@
...
+                width="55%"
+                height={48}
+                style={{ marginBottom: 12 }}
+              />
+            </>
+          ) : (
+            <VideoTitle>{video?.title}</VideoTitle>
+          )}
+
+          {/* Description */}
```
zamanafzal: nit: I think we should remove these comments.

frontends/main/src/app-pages/VideoPage/FeaturedVideo.tsx
```diff
@@ -0,0 +1,160 @@
...
+  zIndex: 1,
+}))
+
+const TextSide = styled.div(({ theme }) => ({
+  [theme.breakpoints.down("md")]: {
+    padding: "0 0 0 36px",
+  },
+  [theme.breakpoints.down("sm")]: {
+    padding: "16px 0 0",
+    letteSpacing: "inherit",
```
zamanafzal: typo: `letteSpacing` should be `letterSpacing`.

frontends/main/src/app-pages/VideoPage/FeaturedVideo.tsx
```diff
@@ -0,0 +1,160 @@
...
+  const truncatedTitle =
+    video.title.length > FEATURED_TITLE_MAX_CHARS
+      ? `${video.title.slice(0, FEATURED_TITLE_MAX_CHARS).trimEnd()}...`
+      : video.title
+
+  return (
+    <Section>
+      <VideoContainer>
+        <FeaturedGrid>
+          <ImageWrapper onClick={() => onPlay(video)}>
```
zamanafzal: Could we switch to semantic button/link behaviour (or add role/tabIndex + Enter/Space handlers) for accessibility parity?

frontends/main/src/app-pages/VideoPage/FeaturedVideo.tsx
```diff
@@ -0,0 +1,160 @@
...
+              alt={video.title}
+              fill
+              sizes="(max-width: 600px) 100vw, 50vw"
+              style={{ objectFit: "cover" }}
+            />
+            {duration && <DurationBadge>{duration}</DurationBadge>}
+          </ImageWrapper>
+
+          <TextSide>
+            <FeaturedTitle onClick={() => onPlay(video)}>
```
zamanafzal: Same as above.

frontends/main/src/app-pages/VideoPage/VideoJsPlayer.tsx
```diff
@@ -0,0 +1,99 @@
+"use client"
+
+import React, { useEffect, useRef } from "react"
+import videojs from "video.js"
+import Player from "video.js/dist/types/player"
+import "video.js/dist/video-js.css"
+// Register YouTube tech so video.js can play youtube:// sources
```
zamanafzal: I think this comment is a bit misleading. Code passes normal YouTube URLs (e.g. https://youtube.com/...), not necessarily youtube://

frontends/main/src/app-pages/VideoPage/VideoDetailPage.tsx
```diff
@@ -0,0 +1,543 @@
...
+  "& > span": {
+    paddingBottom: 0,
+  },
+}))
+
+// ---------------------------------------------------------------------------
+// Helper: extract instructor info
+// ---------------------------------------------------------------------------
+
+function getInstructorInfo(video: VideoResource): {
```
zamanafzal: How about renaming it to `getPrimaryInstructorInfo` for better clarity?

frontends/main/src/app-pages/VideoPage/FeaturedVideo.tsx
```diff
@@ -0,0 +1,160 @@
...
+  },
+}))
+
+type FeaturedVideoProps = {
+  videos: VideoResource[]
+  onPlay: (video: VideoResource) => void
+}
+
+const FeaturedVideo: React.FC<FeaturedVideoProps> = ({ videos, onPlay }) => {
+  const [currentIndex] = useState(0)
```
zamanafzal: nit: `currentIndex` is always 0, so useState here adds cognitive overhead. Consider replacing with `const video = videos[0]` for clearer intent.

zamanafzal (changes requested):
I've left a few comments and the UI doesn't look good on two screens. 

Here are the screenshots of the screens. 

Image

Image

frontends/main/src/app-pages/VideoPage/VideoPage.tsx
```diff
@@ -0,0 +1,118 @@
...
+
+  const showVideoPlaylistPage = useFeatureFlagEnabled(
+    FeatureFlags.VideoPlaylistPage,
+  )
+  const flagsLoaded = useFeatureFlagsLoaded()
+
+  const { data: playlistData, isLoading: playlistLoading } = useQuery(
+    videoPlaylistQueries.detail(playlistId),
+  )
+  const playlist = playlistData as VideoPlaylistResource | undefined
```
ChristopherChudzicki: **Request:** Let's remove this. The type is already `VideoPlaylistResource | undefined`, the cast does nothing.

frontends/main/src/app-pages/VideoPlaylistCollectionPage/VideoPlaylistCollectionPage.tsx
```diff
@@ -0,0 +1,118 @@
...
+
+  const { data: items, isLoading: itemsLoading } = useQuery(
+    learningResourceQueries.items(playlistId, {
+      learning_resource_id: playlistId,
+    }),
+  )
+
+  const { data: similarData, isLoading: similarLoading } = useQuery(
+    learningResourceQueries.vectorSimilar(playlistId),
+  )
```
ChristopherChudzicki: I noticed that:
- below you did `otherCollections.slice(0,6)`
- and there is filtering on the results of vectorSimilar

This has the downside that you're getting an unpredictable number of results. (E.g., if vectorSimilar returns mostly courses or podcasts or whatever, they are filtered out by your filter) 

It would be better to let the APIs handle the trimming and filtering. You can do that by changing the vectorSimilar query...for some reason right now it only supports a single `id` param, but the API supports much more.

Suggest changing to:
```diff
--- a/frontends/api/src/hooks/learningResources/queries.ts
+++ b/frontends/api/src/hooks/learningResources/queries.ts
@@ -20,6 +20,7 @@ import type {
   LearningResourcesApiLearningResourcesItemsListRequest as ItemsListRequest,
   LearningResourcesApiLearningResourcesSummaryListRequest as LearningResourcesSummaryListRequest,
   VideoPlaylistResource,
+  LearningResourcesApiLearningResourcesVectorSimilarListRequest,
 } from "../../generated/v1"
 import type { VectorLearningResourcesSearchApiVectorLearningResourcesSearchRetrieveRequest as VectorLearningResourcesSearchRetrieveRequest } from "../../generated/v0"
 import { queryOptions } from "@tanstack/react-query"
@@ -42,10 +43,9 @@ const learningResourceKeys = {
   detailsRoot: () => [...learningResourceKeys.root, "detail"],
   detail: (id: number) => [...learningResourceKeys.detailsRoot(), id],
   similar: (id: number) => [...learningResourceKeys.detail(id), "similar"],
-  vectorSimilar: (id: number) => [
-    ...learningResourceKeys.detail(id),
-    "vector_similar",
-  ],
+  vectorSimilar: (
+    params: LearningResourcesApiLearningResourcesVectorSimilarListRequest,
+  ) => [...learningResourceKeys.detail(params.id), "vector_similar", params],
   itemsRoot: (id: number) => [...learningResourceKeys.detail(id), "items"],
   items: (id: number, params: ItemsListRequest) => [
     ...learningResourceKeys.itemsRoot(id),
@@ -128,12 +128,14 @@ const learningResourceQueries = {
           .learningResourcesSimilarList({ id })
           .then((res) => res.data),
     }),
-  vectorSimilar: (id: number) =>
+  vectorSimilar: (
+    params: LearningResourcesApiLearningResourcesVectorSimilarListRequest,
+  ) =>
     queryOptions({
-      queryKey: learningResourceKeys.vectorSimilar(id),
+      queryKey: learningResourceKeys.vectorSimilar(params),
       queryFn: () =>
         learningResourcesApi
-          .learningResourcesVectorSimilarList({ id })
+          .learningResourcesVectorSimilarList(params)
           .then((res) => res.data),
     }),
   list: (params: LearningResourcesListRequest) =>
```

then using
```ts
  const { data: similarData, isLoading: similarLoading } = useQuery(
    learningResourceQueries.vectorSimilar({
      id: playlistId,
      resource_type: [ResourceTypeEnum.VideoPlaylist],
      limit: 6,
    }),
  )
```
to get exactly the data you want.

frontends/main/src/app-pages/VideoPage/FeaturedVideo.tsx
ChristopherChudzicki: **Request:** similar to Zaman's comment, change the `videos` prop to `video`. We are only using one. No need to pass the whole array. Confusing for "FeaturedVideo" to take an array, IMO.

frontends/main/src/app-pages/VideoPage/VideoPageHeader.tsx
```diff
@@ -0,0 +1,59 @@
...
+  return (
+    <PageHeader>
+      <Container>
+        <PageHeaderTop>
+          <PageHeaderContent>
+            <PageTitle>{playlist?.title ?? <Skeleton width={320} />}</PageTitle>
+            <PageDescription>
+              {playlist?.description ??
+                "Conversations with MIT faculty on the future of science, technology, and society."}
+            </PageDescription>
```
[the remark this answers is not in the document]
ChristopherChudzicki: is this resolved? I still see the hard-coded description. (Maybe forgot a push?)

frontends/main/src/app-pages/VideoPage/VideoPage.tsx
```diff
@@ -0,0 +1,118 @@
...
+  const isLoading = playlistLoading || itemsLoading
+  const videos = (items ?? []).filter(
+    (item): item is VideoResource =>
+      item.resource_type === VideoResourceResourceTypeEnum.Video,
+  )
+  const collectionVideos = videos.slice(1)
+
+  const otherCollections = getResults(similarData).filter(
+    (resource): resource is VideoPlaylistResource =>
+      resource.resource_type === ResourceTypeEnum.VideoPlaylist &&
```
ChristopherChudzicki: See https://github.com/mitodl/mit-learn/pull/3135/changes#r3053866744 ... would be better to filter at API level.

frontends/main/src/app-pages/VideoPage/VideoPage.tsx
```diff
@@ -0,0 +1,118 @@
...
+    return data.results
+  }
+  return data
+}
+
+type VideoPageProps = {
+  playlistId: number
+}
+
+const VideoPage: React.FC<VideoPageProps> = ({ playlistId }) => {
```
ChristopherChudzicki: VideoPage doesn't seem like a great fit for the name of this file/component. (It's not showing a video, it's showing a playlist).

The issue
- https://github.com/mitodl/hq/issues/10734

describes two types of playlists... "Series" and "Collections". I would expect them to have different templates sharing lots of components.

This should be VideoPlaylistCollectionPage right?

frontends/main/src/app-pages/VideoPage/VideoCard.tsx
```diff
@@ -0,0 +1,196 @@
...
+  const [imgError, setImgError] = useState(false)
+  const imageUrl =
+    !imgError && resource.image?.url ? resource.image.url : PLACEHOLDER_IMG
+  const description = resource.description ?? ""
+  const duration = resource.video?.duration
+    ? formatDurationClockTime(resource.video.duration)
+    : null
+
+  return (
+    <VideoCardItem
```
ChristopherChudzicki: Currently:
- these are `role="button"` divs and we manually have to handle keyboard controls (tab, plus adding handler for click and enter/spacebar). 
- And then the handler we pass is just a `router.push(...)` call

👍 for getting the keyboard controls working. That said, it would be better to use native button or anchor elements. Since these are functioning as links, we should use anchor. Native anchor elements provide standard capability like
- right click to copy url
- command/ctrl + click to open in a new tab

Neither of these things works with the button approach. In general, if it's behaving like a link, we should try to use a link.

frontends/main/src/app-pages/VideoPage/VideoPage.tsx
```diff
@@ -0,0 +1,118 @@
...
+import OtherCollections from "./OtherCollections"
+import { FeatureFlags } from "@/common/feature_flags"
+import { useFeatureFlagsLoaded } from "@/common/useFeatureFlagsLoaded"
+
+const Page = styled.div(({ theme }) => ({
+  backgroundColor: theme.custom.colors.lightGray1,
+  minHeight: "100vh",
+}))
+
+const getResults = (
```
ChristopherChudzicki: The only current usage of `getResults` is `getResults(similarData)`, and `similarData` is already a plain array, `LearningResource[]`. I'd suggest we remove this for now.

frontends/main/src/app-pages/VideoPlaylistCollectionPage/FeaturedVideo.tsx
```diff
@@ -0,0 +1,160 @@
...
+  const imageUrl = video.image?.url ?? null
+  const duration = video.video?.duration
+    ? formatDurationClockTime(video.video.duration)
+    : null
+  const description = video.description ?? ""
+  const FEATURED_TITLE_MAX_CHARS = 30
+  const truncatedTitle =
+    video.title.length > FEATURED_TITLE_MAX_CHARS
+      ? `${video.title.slice(0, FEATURED_TITLE_MAX_CHARS).trimEnd()}...`
+      : video.title
```
ChristopherChudzicki: It would be better to use CSS line clamping (you do that elsewhere in this PR). We actually have a helper defined for this, https://github.com/mitodl/mit-learn/blob/a4e56332d09f203810dc50915f54681dc1c986b6/frontends/ol-components/src/components/TruncateText/TruncateText.tsx#L11

frontends/main/src/app-pages/VideoPage/FeaturedVideo.tsx
ChristopherChudzicki: Semantic elements would be great!

frontends/main/src/app-pages/VideoPage/OtherCollections.tsx
```diff
@@ -0,0 +1,186 @@
...
+  },
+})
+
+const CollectionMeta = styled(Typography)({
+  ...theme.typography.body3,
+  color: theme.custom.colors.silverGrayDark,
+  lineHeight: "16px" /* 133.333% */,
+})
+
+const parseDurationToHoursAndMinutes = (duration?: string): string | null => {
```
ChristopherChudzicki: We have a utility function [`formatDurationClockTime`](https://github.com/mitodl/mit-learn/blob/a4e56332d09f203810dc50915f54681dc1c986b6/frontends/ol-utilities/src/date/utils.ts#L32)  that is used to do this elsewhere. 

Could we use that instead?

frontends/main/src/app-pages/VideoPage/OtherCollections.tsx
```diff
@@ -0,0 +1,186 @@
...
+
+type OtherCollectionsProps = {
+  collections: VideoPlaylistResource[]
+  isLoading: boolean
+}
+
+const OtherCollections: React.FC<OtherCollectionsProps> = ({
+  collections,
+  isLoading,
+}) => {
```
ChristopherChudzicki: Looking at the Figma, I'm guessing the component naming was chosen to match Figma.  In theory I think that is a good idea, but it kind of just moves the burden of picking good names from us to Bilal. And naming is hard.

Looking at the two playlist pages (Series and Collection) I see they both use this component, and this component displays both types of playlists:

Image

IMO, "RelatedPlaylist" would be a better name for this component. cc @mbilalmughal in case you want to update yours. (Doesn't need to be that name in particular, but it seems this isn't particular to collections).

ChristopherChudzicki (changes requested):
This will be a great feature. I left some comments on naming, existing utility funcs that might help, queries, and questions about files that seem to be unused.

frontends/main/src/app/video-collection/[id]/page.tsx
Comment (ChristopherChudzicki) [inline frontends/main/src/app/video-collection/[id]/page.tsx:1]: @ahtesham-quraish This is a video playlist collection page, right? What do you think about a url like `/video-collections/:id` ?  Ideally we'd use a slug, but I don't think we have the data in place for that.

frontends/main/src/app/video-collection/[id]/page.tsx
```diff
@@ -0,0 +1,41 @@
...
+  learningResourceQueries,
+  videoPlaylistQueries,
+} from "api/hooks/learningResources"
+import { getQueryClient } from "@/app/getQueryClient"
+import VideoPage from "@/app-pages/VideoPage/VideoPage"
+import { notFound } from "next/navigation"
+
+export const metadata: Metadata = standardizeMetadata({
+  title: "Video Playlist",
+})
```
Comment (ChristopherChudzicki) [inline frontends/main/src/app/video-collection/[id]/page.tsx:16]: I'd add noindex for now until we are ready for these to be live.

frontends/main/src/app-pages/VideoPlaylistCollectionPage/VideoPageHeader.tsx
ahtesham-quraish: @zamanafzal Ferdi has provided me the design in which video is part of breadcrumb. I already asked him about it but he did not reply me let me ask him again where it should point to.

frontends/main/src/app-pages/VideoPage/VideoDetailPage.tsx
ahtesham-quraish: Currently I have removed the detail page from this PR and moved it here https://github.com/mitodl/mit-learn/pull/3173 so when we click on the video then for now we are showing 404 page which obviously will available in new PR

frontends/main/src/app-pages/VideoPage/VideoDetailPage.tsx
ahtesham-quraish: I have removed the videoDetailPage which will be available in new PR https://github.com/mitodl/mit-learn/pull/3173

frontends/main/src/app-pages/VideoPage/FeaturedVideo.tsx
ahtesham-quraish: Done

frontends/main/src/app-pages/VideoPage/FeaturedVideo.tsx
ahtesham-quraish: Done

frontends/main/src/app-pages/VideoPage/FeaturedVideo.tsx
ahtesham-quraish: Done

frontends/main/src/app-pages/VideoPage/VideoJsPlayer.tsx
ahtesham-quraish: I have removed this comment

frontends/main/src/app-pages/VideoPage/VideoDetailPage.tsx
ahtesham-quraish: I have removed the videoDetailPage from this PR

frontends/main/src/app-pages/VideoPage/FeaturedVideo.tsx
ahtesham-quraish: Done

frontends/main/src/app-pages/VideoPage/VideoPage.tsx
ahtesham-quraish: Done

frontends/main/src/app-pages/VideoPlaylistCollectionPage/VideoPlaylistCollectionPage.tsx
ahtesham-quraish: done

frontends/main/src/app-pages/VideoPage/VideoPage.tsx
ahtesham-quraish: done

frontends/main/src/app-pages/VideoPage/VideoCard.tsx
ahtesham-quraish: Done

frontends/main/src/app-pages/VideoPlaylistCollectionPage/FeaturedVideo.tsx
ahtesham-quraish: Ferdi wants to have this title truncate on the basis of characters, the approach you are talking is based on lines that is why I have created this function to show only 30 characters.

frontends/main/src/app-pages/VideoPage/VideoPage.tsx
ahtesham-quraish: I have updated the code can you please check

frontends/main/src/app-pages/VideoPage/OtherCollections.tsx
ahtesham-quraish: I have created new util which will display the date time in human form like in the design 
image

frontends/main/src/app-pages/VideoPage/OtherCollections.tsx
ahtesham-quraish: Done

frontends/main/src/app-pages/VideoPlaylistCollectionPage/VideoPlaylistCollectionPage.tsx
ahtesham-quraish: Can you please check I have changed the implementation and let me know in case i have done wrong

frontends/main/src/app-pages/VideoPlaylistCollectionPage/VideoJsPlayer.tsx
```diff
@@ -0,0 +1,98 @@
...
+  poster,
+  autoplay = true,
+  controls = true,
+  fluid = true,
+  onReady,
+}) => {
+  const videoRef = useRef<HTMLDivElement>(null)
+  const playerRef = useRef<Player | null>(null)
+
+  useEffect(() => {
```
zamanafzal: The useEffect has unstable dependencies that could cause the video player to be recreated unnecessarily, leading to memory leaks and performance issues.

**Suggestion:** 

```
useEffect(() => {
  // Only initialise once
  if (playerRef.current) return

  const videoEl = document.createElement("video-js")
  videoEl.classList.add("vjs-big-play-centered")
  videoEl.style.width = "100%"
  videoEl.style.height = "100%"
  videoRef.current!.appendChild(videoEl)

  const player = videojs(
    videoEl,
    {
      autoplay,
      controls,
      fluid,
      fill: !fluid,
      responsive: true,
      poster: poster ?? undefined,
      sources,
      techOrder: ["youtube", "html5"],
    },
    function (this: Player) {
      onReady?.(this)
    },
  )

  playerRef.current = player
}, []) // Remove all dependencies to prevent re-initialization

// Handle updates in separate effects
useEffect(() => {
  const player = playerRef.current
  if (!player) return
  player.src(sources)
}, [sources])

useEffect(() => {
  const player = playerRef.current
  if (!player || !poster) return
  player.poster(poster)
}, [poster])
```

frontends/main/src/app-pages/VideoPlaylistCollectionPage/FeaturedVideo.tsx
```diff
@@ -0,0 +1,163 @@
...
+const FeaturedVideo: React.FC<FeaturedVideoProps> = ({ video, href }) => {
+  if (!video) return null
+
+  const imageUrl = video.image?.url ?? null
+  const duration = video.video?.duration
+    ? formatDurationClockTime(video.video.duration)
+    : null
+  const description = video.description ?? ""
+  const FEATURED_TITLE_MAX_CHARS = 30
+  const truncatedTitle =
```
zamanafzal: The title truncation is calculated on every render. How about memoized this for better performance?

```
const FEATURED_TITLE_MAX_CHARS = 30
const truncatedTitle = useMemo(() => 
  video.title.length > FEATURED_TITLE_MAX_CHARS
    ? `${video.title.slice(0, FEATURED_TITLE_MAX_CHARS).trimEnd()}...`
    : video.title
, [video.title])
```

frontends/main/src/app-pages/VideoPlaylistCollectionPage/VideoPlaylistCollectionPage.tsx
```diff
@@ -0,0 +1,97 @@
...
+> = ({ playlistId }) => {
+  const getVideoHref = (resource: VideoResource) =>
+    `/playlist/detail/${resource.id}?playlist=${playlistId}`
+
+  const showVideoPlaylistPage = useFeatureFlagEnabled(
+    FeatureFlags.VideoPlaylistPage,
+  )
+  const flagsLoaded = useFeatureFlagsLoaded()
+
+  const { data: playlist, isLoading: playlistLoading } = useQuery(
```
zamanafzal: I think we need to add error handling for API failures.

**Suggestion:** 
```
const { data: playlist, isLoading: playlistLoading, error: playlistError } = useQuery(
  videoPlaylistQueries.detail(playlistId),
)

const { data: items, isLoading: itemsLoading, error: itemsError } = useQuery(
  learningResourceQueries.items(playlistId, {
    learning_resource_id: playlistId,
  }),
)

if (playlistError || itemsError) {
  return (
    <Page>
      <VideoContainer>
        <Typography variant="h4">Failed to load playlist</Typography>
        <Typography>Please try refreshing the page.</Typography>
      </VideoContainer>
    </Page>
  )
}
```

zamanafzal (changes requested):
I've left some minor suggestions.

frontends/main/src/app-pages/VideoPlaylistCollectionPage/VideoJsPlayer.tsx
ahtesham-quraish: Done

frontends/main/src/app-pages/VideoPlaylistCollectionPage/VideoPlaylistCollectionPage.tsx
ahtesham-quraish: @ChristopherChudzicki what do you think about it?

zamanafzal:
@ahtesham-quraish On a 1024 screen size, the content(Title, BreadCrumbs) is not properly responsive. 

Screenshot 2026-04-10 at 5 07 13 PM

frontends/main/src/app-pages/VideoPlaylistCollectionPage/VideoPlaylistCollectionPage.tsx
ChristopherChudzicki: - **Good:** Well, these changes look great. Exactly what I was hoping for. 

- **Disappointing**: Thank you for making this change. Unfortunately, it seems that the OpenAPI spec is wrong, and the resource_type filter doesn't actually work. I've opened https://github.com/mitodl/hq/issues/10896 about this.


**My recommendation for now:** would be to leave the change as-is for the moment BUT restore a resource_type="video_playlist" filter to prevent crashing. (⚠️ ⚠️ it's currently crashing for some playlists for me, like playlist 6382 from prod):

```ts
  const { data: similarData, isLoading: similarLoading } = useQuery({
    ...learningResourceQueries.vectorSimilar({
      id: playlistId,
      resource_type: [ResourceTypeEnum.VideoPlaylist],
      limit: 6,
    }),
    select: (resources) =>
      resources.filter(
        (r) => r.resource_type === ResourceTypeEnum.VideoPlaylist,
      ),
  })
```
A selector like this wouldn't be bad to leave in even once the backend filtering is in place. It should be a no-op, but would also give you the correct TS type without typecasting.


Then we can coordinate on Monday about supporting resource_type filters on the similarity endpoint.

frontends/main/src/app-pages/VideoPlaylistCollectionPage/VideoPlaylistCollectionPage.tsx
```diff
@@ -0,0 +1,97 @@
...
+  const isLoading = playlistLoading || itemsLoading
+  const videos = (items ?? []).filter(
+    (item): item is VideoResource =>
+      item.resource_type === VideoResourceResourceTypeEnum.Video,
+  )
+  const collectionVideos = videos.slice(1)
+
+  const otherCollections = (
+    (similarData ?? []) as VideoPlaylistResource[]
+  ).filter((resource) => resource.id !== playlistId)
```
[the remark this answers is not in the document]
ChristopherChudzicki: This is a reasonable assumption on Sentry's part, but is incorrect. the vectorsimilar api returns a raw arra and is not paginated.

frontends/ol-utilities/src/date/utils.ts
```diff
@@ -52,6 +52,31 @@ const formatDurationClockTime = (value: string | moment.Duration) => {
...
+ * Suitable for displaying total collection/playlist durations.
+ * Seconds are omitted; only the two largest non-zero units are shown.
+ *
+ * Examples:
+ *  PT8H24M  → "8h 24m"
+ *  PT1H     → "1h"
+ *  PT45M    → "45m"
+ *  PT0S     → ""
+ */
+const formatDurationHuman = (value: string | moment.Duration): string => {
```
ChristopherChudzicki, line 66: Could you add some simple tests for this?

frontends/main/src/app-pages/VideoPlaylistCollectionPage/FeaturedVideo.tsx
```diff
@@ -0,0 +1,167 @@
...
+
+const FeaturedVideo: React.FC<FeaturedVideoProps> = ({ video, href }) => {
+  const truncatedTitle = useMemo(
+    () =>
+      video?.title.length > FEATURED_TITLE_MAX_CHARS
+        ? `${video.title.slice(0, FEATURED_TITLE_MAX_CHARS).trimEnd()}...`
+        : video?.title,
+    [video?.title],
+  )
+  if (!video) return null
```
ChristopherChudzicki: This prop is required, so no need for this early exit.

frontends/main/src/app-pages/VideoPlaylistCollectionPage/FeaturedVideo.tsx
ChristopherChudzicki: See 

OK. We really shouldn't do it this way:
- CSS is more robust visually, see screenshot below ... awkward partially filled line.
- True truncation (not CSS based) has accessibility implications, e.g., screenreaders now get incomplete headings.

Screenshot 2026-04-10 at 5 25 26 PM

I posted something in the slack thread about it.

> [!NOTE]
> It also seems to me that slack thread is mostly about the **description** not the title?

ChristopherChudzicki (changes requested):
Changes look good.. Added a few comemnts. Two things I'd highlight especially:


1. Some files seem unused... `VideoPlayerModal`, `videoSources`, `VideoJsPlayer`. Let's remove these until they are actually used. (You could keep them on a different branch if you want to save them.)
2.  See https://github.com/mitodl/mit-learn/pull/3135/changes#r3066507911 ... important comment there, unfortunately. Page is crashing right now sometimes 😦

frontends/main/src/app/video-collection/[id]/page.tsx
Comment (ahtesham-quraish) [reply frontends/main/src/app/video-collection/[id]/page.tsx:16]: Done

frontends/ol-utilities/src/date/utils.ts
ahtesham-quraish: Done

frontends/main/src/app-pages/VideoPlaylistCollectionPage/FeaturedVideo.tsx
ahtesham-quraish: Done

frontends/main/src/app/video-collection/[id]/page.tsx
Comment (ahtesham-quraish) [reply frontends/main/src/app/video-collection/[id]/page.tsx:1]: @ChristopherChudzicki i dont mind to change it but i think it should be more general because we will have video series as well

frontends/main/src/app/video-collection/[id]/page.tsx
```diff
@@ -0,0 +1,41 @@
...
+
+  await Promise.all([
+    queryClient.prefetchQuery(videoPlaylistQueries.detail(playlistId)),
+    queryClient.prefetchQuery(
+      learningResourceQueries.items(playlistId, {
+        learning_resource_id: playlistId,
+      }),
+    ),
+  ])
+
```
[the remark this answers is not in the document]
Comment (ChristopherChudzicki) [reply frontends/main/src/app/video-collection/[id]/page.tsx:36]: this is a good idea

frontends/main/src/app/video-collection/[id]/page.tsx
Comment (ChristopherChudzicki) [reply frontends/main/src/app/video-collection/[id]/page.tsx:1]: @ahtesham-quraish But those will be a different URL, right?

From the issue:

> This distinction carries through to the video pages.
> 
> On a series video page... On a collection video page...
>
> ...Do not collapse this into one template just because it is easier to implement. Components can be shared. Behavior should not.

frontends/main/src/app-pages/VideoPlaylistCollectionPage/VideoPlaylistCollectionPage.tsx
ChristopherChudzicki: Probably best to handle this on the nextjs server for proper 404s, etc.

https://github.com/mitodl/mit-learn/pull/3135/changes#r3015405136

frontends/main/src/app-pages/VideoPlaylistCollectionPage/RelatedPlaylist.tsx
```diff
@@ -0,0 +1,159 @@
...
+    ? "Video Series"
+    : "Collection"
+}
+
+type OtherCollectionsProps = {
+  collections: VideoPlaylistResource[]
+  isLoading: boolean
+}
+
+const RelatedPlaylist: React.FC<OtherCollectionsProps> = ({
```
ChristopherChudzicki: Should make `OtherCollectionsProps` match the component name.

frontends/main/src/app-pages/VideoPlaylistCollectionPage/VideoPlaylistCollectionPage.tsx
```diff
@@ -0,0 +1,93 @@
...
+  const isLoading = playlistLoading || itemsLoading
+  const videos = (items ?? []).filter(
+    (item): item is VideoResource =>
+      item.resource_type === VideoResourceResourceTypeEnum.Video,
+  )
+  const collectionVideos = videos.slice(1)
+
+  const otherCollections = ((similarData ?? []) as VideoPlaylistResource[])
+    .filter((resource) => resource.id !== playlistId)
+    .slice(0, 6)
```
ChristopherChudzicki, line 77: Currently this list can include things other than video playlists, and that causes the page to crash.

For example, using production base URL, it crashes for pretty consistently for http://open.odl.local:8062/playlist/6382

I'd suggest doing the resource_type filtering at the query call, as mentioned in https://github.com/mitodl/mit-learn/pull/3135/changes#r3066507911

That will make sure everywhere is getting playlists only.

ChristopherChudzicki (changes requested):
Changes generally look good but I'm still seeing it crash. See https://github.com/mitodl/mit-learn/pull/3135/changes#r3073950679

In addition to that, I think a collection-specific URL would be good, as would the 404 handling.

frontends/main/src/app-pages/VideoPlaylistCollectionPage/VideoPlaylistCollectionPage.tsx
ChristopherChudzicki: Suggest doing the filtering when you make the query:

```ts
  const { data: similarData, isLoading: similarLoading } = useQuery({
    ...learningResourceQueries.vectorSimilar(playlistId),
    select: (data) =>
      data.filter(
        (resource) => resource.resource_type === ResourceTypeEnum.VideoPlaylist,
      ),
  })
```
This:
1. Filters on the frontend—we should filter on backend once supported
2. Refines the type.

(1) will be irrelevant once we improve the backend, (2) is still useful, though

ahtesham-quraish: done

frontends/main/src/app-pages/VideoPlaylistCollectionPage/RelatedPlaylist.tsx
ahtesham-quraish: done

frontends/main/src/app-pages/VideoPlaylistCollectionPage/VideoPlaylistCollectionPage.tsx
```diff
@@ -38,23 +38,33 @@ const VideoPlaylistCollectionPage: React.FC<
...
+      ),
+  })
 
   if (!showVideoPlaylistPage) {
     return flagsLoaded ? notFound() : null
   }
+
+  if (isError) {
+    return notFound()
+  }
```
ChristopherChudzicki, line 67: This can be addressed in a followup PR, but this is best to do on the server side (See copilot's suggestion, https://github.com/mitodl/mit-learn/pull/3135#discussion_r3015405136).

ChristopherChudzicki (approved):
👍  one suggestion for followup PR https://github.com/mitodl/mit-learn/pull/3135/changes/62d2fdf19f44c06e549a3cabfa48a5c724a06432#r3074389926

zamanafzal (approved):
LGTM

PR #3152, opened 2026-04-02
10735 add hubspot tracking code to the learn front end

cp-at-mit:
https://github.com/mitodl/hq/issues/10735

Adds the hubspot tracking script alogn with this envirionment variable to specifying the Hubspot application: PUBLIC_HUBSPOT_PORTAL_ID=23128026

### How can this be tested?
Load any page in the application and verify that the request to the Hubspot script is made successfully.  Interacting with buttons on the page should result in more requests being made to the hubspot domains.


env/frontend.env
```diff
@@ -29,6 +29,9 @@ NEXT_PUBLIC_LEARN_AI_SYLLABUS_ENDPOINT=https://api.rc.learn.mit.edu/ai/http/syll
 NEXT_PUBLIC_MITX_ONLINE_BASE_URL=${MITX_ONLINE_BASE_URL}
 NEXT_PUBLIC_VERSION="local-dev"
 
+# Hubspot tracking - xprodev account for dev/RC environments
+NEXT_PUBLIC_HUBSPOT_PORTAL_ID=23128026
```
mbertrand: Should this be added to `frontend.local.env` manually only if/when needed?  That way we would not all be sending local data to the hubspot dev/rc account all the time.

mbertrand (changes requested):

mbertrand (approved):
👍

PR #3171, opened 2026-04-07
Add user verification step to all data mgmt commands if --delete option is specified

mbertrand:
Related to https://github.com/mitodl/hq/issues/10847

- Adds a verification step to all `backpopulate_<source>_data` commands, requesting the user to confirm that the deletion should proceed in the specified environment (dev, qa, production).
- Makes the ovs and youtube commands more consistent with others (`--delete`, `--transcripts`, no option leads to former `fetch` behavior)

### How can this be tested?
- `python manage.py backpopulate_mitpe_data` - should proceed normally
- `python manage.py backpopulate_mitpe_data --delete` - you should be asked to confirm.  Enter anything except `y` (upper/lower) and the command should be aborted
- Repeat the above step but enter `y` or `Y`, the deletion should proceed

### Additional Context
This is just in case someone uses the `--delete` option on RC/production when they think they're in their local docker container

### DIFF
--- a/learning_resources/management/commands/backpopulate_micromasters_data.py
+++ b/learning_resources/management/commands/backpopulate_micromasters_data.py
@@ -3,14 +3,17 @@
 from django.core.management import BaseCommand
 
 from learning_resources.etl.constants import ETLSource
-from learning_resources.management.commands.mixins import TestResourceConfigurationMixin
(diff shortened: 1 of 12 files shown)


abeglova (approved):
👍

PR #3182, opened 2026-04-09
feat: enable course access to staff user before the start date

Anas12091101:
https://github.com/mitodl/hq/issues/10519

This PR allows staff users to click into a course from the Learn enrollment dashboard before the course has started. Previously, the courseware button was disabled for all users when the course start date was in the future. With this change, users with is_staff: true on their MITx Online account bypass that restriction and see an enabled, navigable link to the courseware.

Staff
Screenshot 2026-04-10 at 2 45 11 PM

Non staff
Screenshot 2026-04-10 at 3 41 14 PM

### How can this be tested?

Manual testing (requires local MITx Online + MIT Learn setup):

- Log in as a user who has is_staff: true on their MITx Online account.
- Enroll that user in a course whose start date is in the future (e.g., 2030).
- Navigate to the Learn enrollment dashboard (/dashboard).
- Confirm the courseware button ("Continue") on the future course card is enabled and clicking it navigates to the courseware URL.
- Log out and log in as a non-staff user enrolled in the same future course.
- Confirm the courseware button is disabled for the non-staff user, as before.


jkachel (approved):
LGTM 👍

PR #3184, opened 2026-04-10
feat: implementing the podcast detail page

ahtesham-quraish:
https://github.com/mitodl/hq/issues/10522#issuecomment-4221379279

Implement the podcast detail page

image

### How can this be tested?
if you dont have podcast data locally then in frontend env file you should add the following variable `NEXT_PUBLIC_MITOL_API_BASE_URL="https://api.rc.learn.mit.edu"` it will connect with RC learn backend
then we need to enable the flag which is `video-playlist-page` and then visit the following url `http://open.odl.local:8062/podcast/15260`


frontends/main/src/app-pages/PodcastPage/PodcastDetailPage.tsx
```diff
@@ -0,0 +1,599 @@
...
+  () =>
+    import("@/page-components/LearningResourceDrawer/LearningResourceDrawer"),
+)
+
+/* ── Header ── */
+
+const HeaderSection = styled.div(({ theme }) => ({
+  borderBottom: `1px solid ${theme.custom.colors.lightGray2}`,
+  marginBottom: "56px",
+  paddingTop: "64px 0",
```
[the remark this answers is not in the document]
ahtesham-quraish: done

frontends/main/src/app-pages/PodcastPage/PodcastDetailPage.tsx
```diff
@@ -0,0 +1,599 @@
...
+    enabled: !!resource,
+  })
+
+  useEffect(() => {
+    if (episodesPage) {
+      setAllEpisodes((prev) => [...prev, ...episodesPage])
+    }
+  }, [episodesPage])
+
+  const hasMore = (episodesPage?.length ?? 0) === EPISODES_PAGE_SIZE
```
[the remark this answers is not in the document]
ahtesham-quraish: done

frontends/main/src/app-pages/PodcastPage/PodcastDetailPage.tsx
```diff
@@ -0,0 +1,599 @@
...
+
+  const offeredBy = resource?.offered_by?.name
+  const lastModified = resource?.last_modified
+    ? formatDate(resource.last_modified, "MMM YYYY")
+    : null
+  const episodeCount = podcast?.episode_count
+
+  const metaParts = [
+    offeredBy,
+    episodeCount ? `MIT OpenCourseWare ·${episodeCount} episodes` : null,
```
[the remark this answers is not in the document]
ahtesham-quraish: Will discuss with the reviewer leave it as it is

frontends/main/src/app-pages/PodcastPage/PodcastDetailPage.tsx
```diff
@@ -0,0 +1,599 @@
...
+    }),
+    enabled: !!resource,
+  })
+
+  useEffect(() => {
+    if (episodesPage) {
+      setAllEpisodes((prev) => [...prev, ...episodesPage])
+    }
+  }, [episodesPage])
+
```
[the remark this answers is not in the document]
ahtesham-quraish: done

frontends/main/src/app-pages/PodcastPage/PodcastDetailPage.tsx
```diff
@@ -0,0 +1,599 @@
...
+    : null
+
+  const metaParts = [duration ? `${duration} min` : null, date].filter(Boolean)
+
+  return (
+    <EpisodeRow>
+      <EpisodeInfo>
+        <EpisodeTitleLink href={drawerHref}>{episode.title}</EpisodeTitleLink>
+        {episode.description && (
+          <EpisodeDescription variant="body3">
```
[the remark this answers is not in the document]
ahtesham-quraish: done

frontends/main/src/app-pages/PodcastPage/PodcastDetailPage.tsx
```diff
@@ -0,0 +1,599 @@
...
+      <PageSection>
+        <HeaderSection>
+          <BreadcrumbBar>
+            <PodcastContainer>
+              <Breadcrumbs
+                variant="light"
+                ancestors={[
+                  { href: HOME, label: "Home" },
+                  { href: "/podcasts", label: "Podcasts" },
+                ]}
```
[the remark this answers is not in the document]
ahtesham-quraish: I removed it

frontends/main/src/app/podcast/[id]/page.tsx
```diff
@@ -0,0 +1,37 @@
...
+import { learningResourceQueries } from "api/hooks/learningResources"
+
+export const generateMetadata = async (props: PageProps<"/podcast/[id]">) => {
+  const { id } = await props.params
+  const queryClient = getQueryClient()
+
+  return safeGenerateMetadata(async () => {
+    const resource = await queryClient.fetchQuery(
+      learningResourceQueries.detail(Number(id)),
+    )
```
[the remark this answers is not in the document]
Comment (ahtesham-quraish) [reply frontends/main/src/app/podcast/[id]/page.tsx:17]: done

frontends/main/src/app-pages/PodcastPage/PodcastDetailPage.tsx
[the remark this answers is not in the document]
ahtesham-quraish: done

frontends/main/src/app-pages/PodcastPage/PodcastDetailPage.tsx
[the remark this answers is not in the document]
ahtesham-quraish: done

frontends/main/src/app-pages/PodcastPage/PodcastDetailPage.tsx
```diff
@@ -0,0 +1,599 @@
...
+    ? formatDate(episode.last_modified, "MMM D")
+    : null
+
+  const metaParts = [duration ? `${duration} min` : null, date].filter(Boolean)
+
+  return (
+    <EpisodeRow>
+      <EpisodeInfo>
+        <EpisodeTitleLink href={drawerHref}>{episode.title}</EpisodeTitleLink>
+        {episode.description && (
```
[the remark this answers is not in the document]
ahtesham-quraish: fixed it

frontends/main/src/app-pages/PodcastPage/PodcastDetailPage.tsx
```diff
@@ -0,0 +1,599 @@
...
+    isFetching: episodesFetching,
+  } = useQuery({
+    ...learningResourceQueries.items(id, {
+      learning_resource_id: id,
+      limit: EPISODES_PAGE_SIZE,
+      offset,
+    }),
+    enabled: !!resource,
+  })
+
```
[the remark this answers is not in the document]
ahtesham-quraish: done

frontends/main/src/app/podcast/[id]/page.tsx
```diff
@@ -0,0 +1,37 @@
...
+      imageAlt: resource.image?.alt ?? undefined,
+    })
+  })
+}
+
+const Page: React.FC<PageProps<"/podcast/[id]">> = async (props) => {
+  const { id } = await props.params
+  const queryClient = getQueryClient()
+
+  await queryClient.prefetchQuery(learningResourceQueries.detail(Number(id)))
```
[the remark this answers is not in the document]
Comment (ahtesham-quraish) [reply frontends/main/src/app/podcast/[id]/page.tsx]: done

frontends/main/src/app-pages/PodcastPage/PodcastPlayer.tsx
```diff
@@ -0,0 +1,545 @@
...
+      setIsPlaying(false)
+      setIsBuffering(true)
+      const audio = audioRef.current
+      if (!audio) return
+      audio.load()
+      audio
+        .play()
+        .then(() => setIsPlaying(true))
+        .catch(() => {})
+    }, [track.audioUrl])
```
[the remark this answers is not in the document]
ahtesham-quraish: done

frontends/main/src/app-pages/PodcastPage/PodcastDetailPage.tsx
```diff
@@ -0,0 +1,584 @@
...
+                    <StyledButton
+                      href={subscribeUrl}
+                      target="_blank"
+                      rel="noopener noreferrer"
+                      variant="primary"
+                      startIcon={<RiMailLine />}
+                    >
+                      Play Latest Episode
+                    </StyledButton>
+                  )}
```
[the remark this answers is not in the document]
ahtesham-quraish: that is ok but i will confirm from author

frontends/main/src/app-pages/PodcastPage/PodcastDetailPage.tsx
```diff
@@ -0,0 +1,583 @@
...
+
+const PodcastTitle = styled(Typography)(({ theme }) => ({
+  marginBottom: "16px",
+  display: "inline-block",
+
+  [theme.breakpoints.down("sm")]: {
+    ...theme.typography.h3,
+    fontSize: "34px",
+    fontStyle: "normal",
+    lineHeight: "40px",
```
ChristopherChudzicki: **Request:** Could you look through all font sizes / styles? Generally we shouldn't need anything beyond the predefined typography styles. This one should be `h5`:

Image

Occasionally we override the line height for looser text in paragraphs, but h5 should be correct here.

frontends/main/src/app-pages/PodcastPage/PodcastDetailPage.tsx
```diff
@@ -0,0 +1,583 @@
...
+  color: theme.custom.colors.darkGray1,
+  whiteSpace: "nowrap",
+  textAlign: "right",
+}))
+
+interface PlayButtonProps {
+  $isPlaying?: boolean
+}
+
+const PlayButton = styled.a<PlayButtonProps>(({ theme, $isPlaying }) => ({
```
ChristopherChudzicki: **Request:** Could we make this an action button like figma has and please check the other new buttons against figma? 

**Explanation:**

This is a lot more custom CSS than I would expect for a button. Looking at Figma, it uses our standard ActionButton component, but it does have some custom overrides.

Image


I think something like this is more in-line with Figma:
```tsx
const PlayButton = styled(ActionButton)<{
  isPlaying: boolean
}>(({ theme, isPlaying }) => [
  {
    borderColor: "currentColor",
    "&:hover:not(:disabled)": {
      color: theme.custom.colors.red,
    },
  },
  isPlaying && {
    color: theme.custom.colors.red,
  },
])
```

then `<PlayButton variant="secondary" />`


In general, we should try to use the components we have from Figma. 

We definitely need to make this easier for our AI agents to get correct though... you might be interested in:
- https://github.com/mitodl/hq/issues/10453

frontends/main/src/app-pages/PodcastPage/PodcastDetailPage.tsx
```diff
@@ -0,0 +1,583 @@
...
+    width: "80px",
+    height: "48px",
+  },
+}))
+
+/* ── Episode row component ── */
+
+type EpisodeItemProps = {
+  episode: LearningResource
+  index: number
```
ChristopherChudzicki: Looks like `index` is unused. We should remove it.

frontends/main/src/app-pages/PodcastPage/PodcastDetailPage.tsx
```diff
@@ -0,0 +1,583 @@
...
+  } = useQuery({
+    ...learningResourceQueries.items(id, {
+      learning_resource_id: id,
+      limit: EPISODES_PAGE_SIZE,
+      offset,
+    }),
+    enabled: !!resource,
+  })
+
+  useEffect(() => {
```
ChristopherChudzicki: **Request:** React query actually has a standard pattern for "Load more" that we should use here. They call it [useInfiniteQuery](https://tanstack.com/query/v5/docs/framework/react/reference/useInfiniteQuery)

Could you ask Copilot to refactor this useQuery + useEffect to get rid to `useInfiniteQuery`? It should handle things like "is there a next page" and "fetch next page" automatically once you set up queryOptions. Here's what I would do for the queryOptions:

<details>
<summary>show diff</summary>

```diff
diff --git a/frontends/api/src/hooks/learningResources/queries.ts b/frontends/api/src/hooks/learningResources/queries.ts
index 79ee84705..2121a58d6 100644
--- a/frontends/api/src/hooks/learningResources/queries.ts
+++ b/frontends/api/src/hooks/learningResources/queries.ts
@@ -19,10 +19,12 @@ import type {
   FeaturedApiFeaturedListRequest as FeaturedListParams,
   LearningResourcesApiLearningResourcesItemsListRequest as ItemsListRequest,
   LearningResourcesApiLearningResourcesSummaryListRequest as LearningResourcesSummaryListRequest,
+  PaginatedLearningResourceRelationshipList,
   VideoPlaylistResource,
 } from "../../generated/v1"
 import type { VectorLearningResourcesSearchApiVectorLearningResourcesSearchRetrieveRequest as VectorLearningResourcesSearchRetrieveRequest } from "../../generated/v0"
-import { queryOptions } from "@tanstack/react-query"
+import { infiniteQueryOptions, queryOptions } from "@tanstack/react-query"
+import axiosInstance from "../../axios"
 import { hasPosition, randomizeGroups } from "./util"

 const timedPromise = async <T>(
@@ -60,6 +62,14 @@ const learningResourceKeys = {
     ...learningResourceKeys.itemsRoot(id),
     params,
   ],
+  infiniteItemsRoot: (id: number) => [
+    ...learningResourceKeys.detail(id),
+    "infiniteItems",
+  ],
+  infiniteItems: (id: number, params: ItemsListRequest) => [
+    ...learningResourceKeys.infiniteItemsRoot(id),
+    params,
+  ],
   // featured
   featuredRoot: () => [...learningResourceKeys.root, "featureds"],
   featured: (params: FeaturedListParams) => [
@@ -129,6 +139,22 @@ const learningResourceQueries = {
           .then((res) => res.data.results.map((rel) => rel.resource))
       },
     }),
+  infiniteItems: (id: number, params: ItemsListRequest) =>
+    infiniteQueryOptions({
+      queryKey: learningResourceKeys.infiniteItems(id, params),
+      queryFn: async ({ pageParam }) => {
+        const request = pageParam
+          ? axiosInstance.request<PaginatedLearningResourceRelationshipList>({
+              method: "get",
+              url: pageParam,
+            })
+          : learningResourcesApi.learningResourcesItemsList(params)
+        const { data } = await request
+        return data
+      },
+      initialPageParam: null as string | null,
+      getNextPageParam: (lastPage) => lastPage.next ?? undefined,
+    }),
   similar: (id: number) =>
     queryOptions({
       queryKey: learningResourceKeys.similar(id),
```

</details>

frontends/main/src/app-pages/PodcastPage/PodcastDetailPage.tsx
```diff
@@ -0,0 +1,583 @@
...
+            playingEpisode.podcast_episode?.episode_link) ||
+          "",
+        title: playingEpisode.title || "Untitled Episode",
+        podcastName: resource?.title || "Podcast",
+      }
+    : null
+
+  return (
+    <>
+      <LearningResourceDrawer />
```
ChristopherChudzicki: is this triggered on the page anywhere? sorry if i missed it.

frontends/main/src/app-pages/PodcastPage/PodcastPlayer.tsx
```diff
@@ -0,0 +1,548 @@
...
+          {/* Progress */}
+          <ProgressWrapper>
+            <TimeLabel variant="body3">{formatTime(currentTime)}</TimeLabel>
+            <ProgressTrack
+              onClick={handleProgressClick}
+              role="slider"
+              aria-label="Seek"
+              aria-valuemin={0}
+              aria-valuemax={duration}
+              aria-valuenow={currentTime}
```
ChristopherChudzicki: nice attributes here 👍 

ideally I think we should also add [aria-valuetext](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-valuetext) as a formatted timestamp or something like "10 minutes remaining"

ChristopherChudzicki (changes requested):
This looks really slick. The podcast player behaves well. I left some comments on the code—could you take a look?

**In addition to those comments:** I noticed that the title and episode titles are not headings.. could you take a look at that

**questions / comments**
1. I noticed no tests for the podcast player itself. Would it be feasible to test that here with jsdom/react-testing-library?
2. Should after you click the play button on an episode, should click the episode play button again do anything? (Like pause it)
3. Button label (screenshot below)... should this say subscribe?

Image

frontends/main/src/app-pages/PodcastPage/PodcastDetailPage.tsx
ahtesham-quraish: I have made the change what you have suggested

frontends/main/src/app-pages/PodcastPage/PodcastDetailPage.tsx
ahtesham-quraish: Done

frontends/main/src/app-pages/PodcastPage/PodcastDetailPage.tsx
ahtesham-quraish: done

frontends/main/src/app-pages/PodcastPage/PodcastDetailPage.tsx
ahtesham-quraish: Changed

frontends/main/src/app-pages/PodcastPage/PodcastDetailPage.tsx
ahtesham-quraish: Removed

frontends/main/src/app-pages/PodcastPage/PodcastPlayer.tsx
ahtesham-quraish: Done

frontends/main/src/app-pages/PodcastPage/PodcastDetailPage.tsx
```diff
@@ -0,0 +1,540 @@
...
+    },
+    "& .desktop-title": {
+      display: "none",
+    },
+  },
+}))
+
+const HeaderTextContent = styled.div(({ theme }) => ({
+  [theme.breakpoints.down("sm")]: {},
+}))
```
ChristopherChudzicki: I'd just use a div, but if you want the named styled component, i'd get rid of the empty styles.

frontends/main/src/app-pages/PodcastPage/PodcastDetailPage.tsx
```diff
@@ -0,0 +1,540 @@
...
+  display: "block",
+  marginBottom: "8px",
+  fontSize: "18px",
+  fontStyle: "normal",
+  fontWeight: theme.typography.fontWeightBold,
+  lineHeight: "26px",
+}))
+
+const StyledButton = styled(ButtonLink)(({ theme }) => ({
+  minWidth: "140px",
```
ChristopherChudzicki: is this minwidth coming from figma? Just curious... i didn't see it / wouldnt expect it.

frontends/main/src/app-pages/PodcastPage/PodcastDetailPage.tsx
```diff
@@ -0,0 +1,540 @@
...
+    : null
+
+  const date = episode.last_modified
+    ? formatDate(episode.last_modified, "MMM D")
+    : null
+
+  const metaParts = [duration ? `${duration} min` : null, date].filter(Boolean)
+
+  return (
+    <EpisodeRow>
```
ChristopherChudzicki: Looking at figma just now, I noticed that Figma has the whole row clickable / hover effect on the whole row, where click triggers the player (same as button)

Image

That probably is a significantly better experience, especially on mobile.

**Suggstion:** I can think of a few ways to handle that:
1. turn whole row into a proper button
    - if go this route, need to do something different with the Play button, since nested buttons won't work. 
    - probably could do `as="div"` on the action button to avoid nested buttons?
2. Keep the "Play" button as-is, turn add a click handler to the div
    - normally click handlers on a div aren't great since they aren't keyboard accessible, but keyboard users have the "Play" button icon already


Either one seems fine...

frontends/main/src/app-pages/PodcastPage/PodcastDetailPage.tsx
```diff
@@ -0,0 +1,540 @@
...
+  } = useInfiniteLearningResourceItems(
+    id,
+    { learning_resource_id: id, limit: EPISODES_PAGE_SIZE },
+    { enabled: !!resource },
+  )
+
+  const episodes = useMemo(
+    () =>
+      episodesData?.pages.flatMap((page) =>
+        page.results.map((rel) => rel.resource),
```
ChristopherChudzicki: **Suggestion:** If you do
```ts
  const episodes =
    episodesData?.pages.flatMap((page) =>
      page.results
        .map((rel) => rel.resource)
        .filter((r) => r.resource_type === ResourceTypeEnum.PodcastEpisode),
    ) ?? []
```

TS will refine the type and you won't have to do the resource_type checks below.

frontends/main/src/app-pages/PodcastPage/PodcastDetailPage.tsx
```diff
@@ -0,0 +1,540 @@
...
+
+  const { data: resource } = useLearningResourcesDetail(id)
+
+  const {
+    data: episodesData,
+    isLoading: episodesLoading,
+    isFetchingNextPage,
+    hasNextPage,
+    fetchNextPage,
+  } = useInfiniteLearningResourceItems(
```
ChristopherChudzicki, line 376: 👍

frontends/main/src/app-pages/PodcastPage/PodcastDetailPage.tsx
```diff
@@ -0,0 +1,540 @@
...
+                ancestors={[{ href: HOME, label: "Home" }]}
+                current={resource?.title}
+              />
+            </PodcastContainer>
+          </BreadcrumbBar>
+          <PodcastContainer>
+            <StyledHeaderSection>
+              <HeaderContent>
+                <HeaderTextContent>
+                  <PodcastTitle variant="h1" className="desktop-title">
```
ChristopherChudzicki: **Request:** This should be `component=h1.

frontends/main/src/app-pages/PodcastPage/PodcastDetailPage.tsx
```diff
@@ -0,0 +1,540 @@
...
+const HeaderContent = styled.div(({ theme }) => ({
+  display: "grid",
+  gridTemplateColumns: "12.95fr 0.6fr",
+  gap: "164px",
+  "& .mobile-title": {
+    display: "none",
+  },
+  "& .desktop-title": {
+    display: "inline-block",
+  },
```
ChristopherChudzicki: Right now, we have 2 copies of the title (one desktop, one mobile) and we show/hide one based on the breakpoint.

I asked Claude for a suggestion about how to do this with a single HTML structure (no repeated title element) and it suggested something like:
```tsx
  <HeaderContent>
    <PodcastTitle sx={{ gridArea: "title" }} />
    <PodcastImage sx={{ gridArea: "image" }} />
    <HeaderText sx={{ gridArea: "text" }}>
      {meta}{description}{latestLine}{button}
    </HeaderText>
  </HeaderContent>
```
>   // Desktop:               Mobile:
>  // "title image"          "title"
>  // "text  image"          "image"
>  //                        "text"
>
>  Desktop places image on the right spanning both rows; mobile stacks them in the order you want. One title node, one image node, no twin DOM.

This seems like a cleaner solution. 

**Question:** What do you think of ☝️ 

**Suggestion:** If not ☝️ , it might be better to repeat the Image element and show/hide that based on break point. Then the title can be the only h1 element on the page.

(Right now if you make title h1, there will be two h1s. I think that's OK since one of them is display none, but the CSS-only + single html structure version seems better.)

frontends/api/src/hooks/learningResources/queries.ts
```diff
@@ -129,6 +140,28 @@ const learningResourceQueries = {
...
+      ),
+      queryFn: async ({ pageParam }) => {
+        const request = pageParam
+          ? axiosInstance.request<PaginatedLearningResourceRelationshipList>({
+              method: "get",
+              url:
+                BASE_PATH +
+                new URL(pageParam, "https://x").pathname +
+                new URL(pageParam, "https://x").search,
+            })
```
ChristopherChudzicki, line 161: What's the goal here? The `next` urls are fully qualified absolute urls, so we should be able to do:

```suggestion
          ? axiosInstance.request<PaginatedLearningResourceRelationshipList>({
              method: "get",
              url: pageParam
```

frontends/main/src/app-pages/PodcastPage/PodcastDetailPage.tsx
```diff
@@ -0,0 +1,540 @@
...
+    () =>
+      episodesData?.pages.flatMap((page) =>
+        page.results.map((rel) => rel.resource),
+      ) ?? [],
+    [episodesData],
+  )
+
+  const isPodcast =
+    resource?.resource_type === ResourceTypeEnum.Podcast &&
+    resource.resource_type === "podcast"
```
ChristopherChudzicki: these two checks are the same.

frontends/main/src/app-pages/PodcastPage/PodcastPlayer.tsx
```diff
@@ -0,0 +1,550 @@
...
+})
+
+const ProgressTrack = styled.div(({ theme }) => ({
+  flex: 1,
+  height: "6px",
+  borderRadius: "3px",
+  backgroundColor: theme.custom.colors.lightGray2,
+  position: "relative",
+  cursor: "pointer",
+  "&:hover .thumb": { opacity: 1 },
```
ChristopherChudzicki: dead css? no `.thumb` used?

frontends/main/src/app-pages/PodcastPage/PodcastPlayer.tsx
```diff
@@ -0,0 +1,550 @@
...
+
+const formatTime = (seconds: number): string => {
+  const m = Math.floor(seconds / 60)
+  const s = Math.floor(seconds % 60)
+  return `${String(m).padStart(2, "0")}:${String(s).padStart(2, "0")}`
+}
+
+// ─── Component ────────────────────────────────────────────────────────────────
+
+const PodcastPlayer = forwardRef<PodcastPlayerHandle, PodcastPlayerProps>(
```
ChristopherChudzicki: `forwardRef` is deprecated, React 19 forwards refs automatically. (A linting rule should enforce this, ideally). So, you can remove this.

**Separately:** We're not using the ref anywhere are we? If so so I'd remove the ref entirely, including `useImperativeHandle`.

frontends/main/src/app-pages/PodcastPage/PodcastPlayer.tsx
```diff
@@ -0,0 +1,550 @@
...
+              <RiForward30Line size={24} />
+            </IconButton>
+          </Controls>
+
+          {/* Progress */}
+          <ProgressWrapper>
+            <TimeLabel variant="body3">{formatTime(currentTime)}</TimeLabel>
+            <ProgressTrack
+              onClick={handleProgressClick}
+              role="slider"
```
ChristopherChudzicki: Should we be using `<input type="range" />`? Or MUI's slider? (MUI's may have too much builtin css for our use here.)

Both of those should give us keyboard controls for free.

frontends/main/src/app-pages/PodcastPage/PodcastPlayer.tsx
```diff
@@ -0,0 +1,550 @@
...
+            setIsPlaying(false)
+          }}
+          onTimeUpdate={() =>
+            setCurrentTime(audioRef.current?.currentTime ?? 0)
+          }
+          onLoadedMetadata={() => setDuration(audioRef.current?.duration ?? 0)}
+          onEnded={() => setIsPlaying(false)}
+        />
+
+        {/* ── Desktop player bar (hidden on mobile) ── */}
```
ChristopherChudzicki: Similar to comment in PodcastDetailPAge, it would be really awesome if we can unify the HTML structure and not have two separate HTML layouts for mobile/desktop, switching between the two.

If we need to, we can, but it complicates testing (need to test both versions) and is extra code.

I suspect we could achieve the relevant differences similarly to idea in PodcastDetailPage via grid. Would you mind looking into it /  asking copilot/claude

ChristopherChudzicki (changes requested):
Changes look good, thanks for adding the PodcastPlayer test.

I left a few more comments. Generally a few buckets of comments:
1. small tweaks
2.  A note about Figma podcast episode row clicking https://github.com/mitodl/mit-learn/pull/3184/changes#r3099417605
3. Suggestions to clean up the HTML structure so we don't have separate HTML for desktop/mobile

Regarding (2) and (3)... These could be addressed in a followup if there's a desire to get this out to RC sooner.

**Bigger Question:** I noticed this isn't feature flagged. Has this been discussed yet? I would expect us to feature flag it. (Though possibly having it unflagged is OK, It's not linked to yet anywhere, right?)

frontends/main/src/app-pages/PodcastPage/PodcastDetailPage.tsx
ahtesham-quraish: removed

frontends/main/src/app-pages/PodcastPage/PodcastDetailPage.tsx
ahtesham-quraish: Removed empty style

frontends/main/src/app-pages/PodcastPage/PodcastDetailPage.tsx
ahtesham-quraish: I have matched with figma

frontends/main/src/app-pages/PodcastPage/PodcastDetailPage.tsx
ahtesham-quraish: check now

frontends/main/src/app-pages/PodcastPage/PodcastDetailPage.tsx
ahtesham-quraish: It does not take the component=h1

frontends/api/src/hooks/learningResources/queries.ts
ahtesham-quraish: locally it was not working I have tried the code always append the port that is why I had to do this

frontends/main/src/app-pages/PodcastPage/PodcastDetailPage.tsx
ahtesham-quraish: you are right

frontends/main/src/app-pages/PodcastPage/PodcastPlayer.tsx
ahtesham-quraish: removed

frontends/main/src/app-pages/PodcastPage/PodcastPlayer.tsx
ahtesham-quraish: done

frontends/main/src/app-pages/PodcastPage/PodcastPlayer.tsx
ahtesham-quraish: Done

frontends/main/src/app-pages/PodcastPage/PodcastDetailPage.tsx
ahtesham-quraish: Done

frontends/main/src/app-pages/PodcastPage/PodcastPlayer.tsx
ahtesham-quraish: Made the change

frontends/main/src/app/podcast/[id]/page.tsx
```diff
@@ -0,0 +1,42 @@
...
+
+const Page: React.FC<PageProps<"/podcast/[id]">> = async (props) => {
+  const { id } = await props.params
+  const queryClient = getQueryClient()
+
+  const podcastId = Number(id)
+  if (Number.isNaN(podcastId)) {
+    notFound()
+  }
+  await queryClient.fetchQueryOr404(learningResourceQueries.detail(podcastId))
```
Comment (ChristopherChudzicki) [inline frontends/main/src/app/podcast/[id]/page.tsx]: **Suggestion:** Might be worth doing
```ts
const resource =   await queryClient.fetchQueryOr404(learningResourceQueries.detail(podcastId))
if (resource.resource_type !== ResourceTypeEnum.Podcast){
  notFound() 
}
```

frontends/main/src/app-pages/PodcastPage/PodcastDetailPage.tsx
```diff
@@ -0,0 +1,566 @@
...
+
+  const metaParts = [
+    offeredBy,
+    episodeCount ? `${episodeCount} episodes` : null,
+    lastModified ? `Updated ${lastModified}` : null,
+  ].filter(Boolean)
+
+  const latestEpisode = episodes?.[0]
+  const latestEpisodeDuration =
+    latestEpisode?.resource_type === "podcast_episode" &&
```
ChristopherChudzicki: nitpick: This check is redundant now that you added the filtering to `episodes` selector. TS already knows it's a podcast episode. The optional chaining is still good, imo, lest `episodes` array is empty.

frontends/main/src/app-pages/PodcastPage/PodcastDetailPage.tsx
```diff
@@ -0,0 +1,566 @@
...
+          moment.duration(latestEpisode.podcast_episode.duration).asMinutes(),
+        )
+      : null
+  const latestEpisodeDate = latestEpisode?.last_modified
+    ? formatDate(latestEpisode.last_modified, "MMM D")
+    : null
+
+  const subscribeUrl =
+    podcast?.apple_podcasts_url ??
+    podcast?.google_podcasts_url ??
```
ChristopherChudzicki: **Question:** I'm still confused about this ... This is called `subscribeUrl`, and seems to go to a podcast platform (like apple podcasts), but the button text says "Play latest episode".

Can you check with @mbilalmughal or Ferdi about the desired behavior here?

**Either way:** We shouldn't use google podcast URLs. That service is no longer supported. I guess we do expose them on our API results, e.g., https://api.learn.mit.edu/api/v1/learning_resources/14144/ points to  https://podcasts.google.com/feed/aHR0cHM6Ly9mZWVkcy5zaW1wbGVjYXN0LmNvbS84ZlFkUzZEeA==
but that link doesn't go anywhere useful.

**Suggestion:** If we can't resolve the functionality question soon, just remove the google url and we'll revisit after people look at the page.

ChristopherChudzicki (changes requested):
The HTML unification/simplification via css grid is great. Very nice to have single HTML structure with mobile vs desktop handled only via css.

A few things I am still noticing:

1. The [pageParam](https://github.com/mitodl/mit-learn/pull/3184#discussion_r3099635950) suggestion is still worth doing. Can you elaborate what issue you're having locally? I'm happy to try and help debug that.
    - The API should be the source of truth here, and reading from next/prev is how our other infinite queries do this.
2. Regarding the [whole row clickability](https://github.com/mitodl/mit-learn/pull/3184#discussion_r3099417605) comment: The whole row now has hover, but in Figma, the whole row itself is clickable, not just the play button.
3. See [this comment] about Play button and google podcast urls.

Can you take a look at these? I can review first-thing in the morning while on my 2.5 hour  bus ride into the office.

ahtesham-quraish:
>The https://github.com/mitodl/mit-learn/pull/3184#discussion_r3099635950 suggestion is still worth doing. Can you elaborate what issue you're having locally? I'm happy to try and help debug that.
The API should be the source of truth here, and reading from next/prev is how our other infinite queries do this.

The issue is the pageParam comes null that is why proper next url is not being done

frontends/main/src/app-pages/PodcastPage/PodcastDetailPage.tsx
ahtesham-quraish: I have removed the google podcast url.

>Question: I'm still confused about this ... This is called subscribeUrl, and seems to go to a podcast platform (like apple podcasts), but the button text says "Play latest episode".


Yeah I have same concern

frontends/main/src/app-pages/PodcastPage/PodcastDetailPage.tsx
ahtesham-quraish: Done

frontends/main/src/app/podcast/[id]/page.tsx
Comment (ahtesham-quraish) [reply frontends/main/src/app/podcast/[id]/page.tsx]: Done

ahtesham-quraish:
>Regarding the https://github.com/mitodl/mit-learn/pull/3184#discussion_r3099417605 comment: The whole row now has hover, but in Figma, the whole row itself is clickable, not just the play button.

this is done

ChristopherChudzicki (approved):

PR #3186, opened 2026-04-10
Search response time indicator for admin users

shanbady:

Closes https://github.com/mitodl/hq/issues/10895

This PR adds a small "response time" indicator next to the search input when logged in as an admin user so we can see performance between searches, endpoints environments etc at a glance.

Screenshot 2026-04-10 at 2 55 24 PM

### How can this be tested?
1. checkout this branch
2. make sure you have data locally (also embeddings if possible) and log in as an admin/superuser
3. go to the search page and try searching for different things and changing filters. you should see the response times for the search (to the right of the search input)

### DIFF
--- a/frontends/main/src/page-components/SearchDisplay/SearchDisplay.tsx
+++ b/frontends/main/src/page-components/SearchDisplay/SearchDisplay.tsx
@@ -585,9 +585,6 @@ const SearchDisplay: React.FC<SearchDisplayProps> = ({
 }) => {
   const [searchParams] = useSearchParams()
   const [expandAdminOptions, setExpandAdminOptions] = useState(false)
-  const fetchStartRef = useRef<number | null>(
-    typeof performance !== "undefined" ? performance.now() : null,
-  )
 
   const { data: adminParams, isLoading: isAdminParamsLoading } =
     useAdminSearchParams(expandAdminOptions)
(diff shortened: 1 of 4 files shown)


frontends/main/src/page-components/SearchDisplay/SearchDisplay.tsx
```diff
@@ -663,6 +668,22 @@ const SearchDisplay: React.FC<SearchDisplayProps> = ({
     },
   })
 
+  const prevParamsRef = useRef(allParams)
+
+  useEffect(() => {
```
ChristopherChudzicki, line 672: Copilot's comment below is accurate. One way to see this is:
1. Visit search page
2. Search "math", then "physics", then "math" again
3. The response time is shown as 0ms because a cached query was used (but not a cachced response time).

@shanbady IMO, the most natural way to do all this timing stuff would be right in the queryFn in `frontends/api/src/hooks/learningResources/queries.ts`. Something like:

```ts
const timedPromise = async <T>(
  promise: Promise<T>,
): Promise<{ result: T; time: number }> => {
  const start = performance.now()
  const result = await promise
  const end = performance.now()
  return { result, time: end - start }
}

// then
      queryFn: () =>
        timedPromise(
          learningResourcesSearchApi
            .learningResourcesSearchRetrieve(params)
            .then((res) => res.data),
        )
```

No refs, no useEffect (well, you might still need it for the callback since we're setting the UI in a different component.) And now the time is directly on the query data, so it's stored in the cache, so in the scenario above it reset to the correct time on second math query.

Two comments about the approach I mention above:
1. It does actually change the query data, so if we were using search / vectorSearch queries in 100 places, it might be annoying. But we're only using it in one place.
    - And i guess if we were using it in 100 places, we could add a default `select` to the queryOptions, so that the new time response only shows up where we want it.
2. If you do a full reload on the search page, the initial time shown will be from the NextJS server to the backend, not your client to the backend.
    - I guess you could force the client to refetch for admin users via `refetchOnMount: isAdmin` or something.

ChristopherChudzicki (changes requested):
There is a small issue... I suggested a way to fix it. 

On the one hand, this is admin only. On the other hand, I think the fix is probably conceptually simpler (just time a promise instead of useEffect / useRef shenanigans). So IMO worth doing.

frontends/main/src/page-components/SearchDisplay/SearchDisplay.tsx
shanbady: That makes sense. I implemented wrapping things in the timedPromise as suggested

ChristopherChudzicki (approved):

PR #3174, opened 2026-04-08
Avoid inadvertent contentfile deletion

mbertrand:
### What are the relevant tickets?
Closes https://github.com/mitodl/hq/issues/10847

### Description (What does it do?)

- If processing a contentfile archive yields no results for whatever reason, don't unpublish/delete any existing contentfiles for that run
- If a run is changed from published=True to published=False, but test_mode=True, don't delete its contentfiles
- When any run is changed to published=False, null out its checksum, so if it ever changes to published=True again, contentfiles will be reingested.
- Wait for 14 days after contentfiles are unpublished before deleting them (via daily celery task)

#### Setup
- Set these in your backend.local.env file:
  ```bash
  MITX_ONLINE_BASE_URL=https://mitxonline.mit.edu/
  MITX_ONLINE_COURSES_API_URL=https://mitxonline.mit.edu/api/v2/courses/
  MITX_ONLINE_PROGRAMS_API_URL=https://mitxonline.mit.edu/api/v2/programs/
  ```
- In the web docker container, run `./manage.py backpopulate_mitxonline_data` if you haven't already done so.  `COURSE_ARCHIVE_BUCKET_NAME` should be unset to save time.
- Set AWS credentials and COURSE_ARCHIVE_BUCKET_NAME to the same values as on RC in your backend.local.env file and restart containers
- Pick a sample published mitxonline course and run `./manage.py backpopulate_mitxonline_files --resource-ids <TARGET_ID>`
- Run the following in a shell to establish a baseline:
```python
  from learning_resources.models import *
  TARGET_ID=<learning_resource__id>
  run = LearningResourceRun.objects.filter(
    learning_resource_id=TARGET_ID, published=True
  ).first()
  baseline_checksum = run.checksum
  baseline_cf_count = ContentFile.objects.filter(run=run, published=True).count()
  print(run.id, run.run_id, baseline_checksum, baseline_cf_count)
  
  > 8450 course-v1:MITxT+2.25.2x+3T2025 None 0
   ```

#### Change nothing if contentfile archives are empty
```python
from learning_resources.etl.loaders import load_content_files
result = load_content_files(run, iter([]), calc_completeness=False)
print("result:", result)
run.refresh_from_db()
print("checksum unchanged?", run.checksum == baseline_checksum)
print(
    "published cf preserved?",
    ContentFile.objects.filter(run=run, published=True).count() == baseline_cf_count,
)
  
[...]

PR #3192, opened 2026-04-13
Deindex unpublished test_mode resources, not contentfiles

mbertrand:
Related to https://github.com/mitodl/hq/issues/10847 and https://github.com/mitodl/mit-learn/pull/3174

If a `test_mode=True` learning resource is unpublished, the resource itself should be deindexed from open/vector search, but its contentfiles should not.

#### on the main branch
- Find a test_mode mitxonline course with contentfiles and set it to published=True:
  ```python
  from learning_resources.etl.loaders import update_index
   lr = LearningResource.objects.filter(
      etl_source="mitxonline", 
      published=False, 
      test_mode=True, 
      runs__content_files__isnull=False
  ).first()
  print(lr.id, lr.title)
  lr.published=True
  lr.save()
  update_index(lr, False)
  ```
- Wait a bit for indexing to occur, the do a search by title for that resource, it should show up.
- Verify that it has publised contentfiles:
  ```python
  ContentFile.objects.filter(run__learning_resource=lr, published=True).count()
  ```
- Unpublish the resource:
  ```python
  lr = LearningResource.objects.get(id=<resource_id>`
  lr.published=False
  lr.save()
  update_index(lr, False)
  ```
-  Wait a bit, then search by title again, it will still show up,
- It will still have published contentfiles:
  ```python
  ContentFile.objects.filter(run__learning_resource=lr, published=True).count()
  ```
  
####  This branch
- Switch to this branch and restart containers.
- Run this in a shell:
  ```python
  from learning_resources.etl.loaders import update_index
   lr = LearningResource.objects.get(id=<resource_id>)
   update_index(lr, False)    
-  Wait a bit, then search by title again, it should not show up,
- It should still have published contentfiles:
  ```python
  ContentFile.objects.filter(run__learning_resource=lr, published=True).count()
  ```

### DIFF
--- a/learning_resources/etl/loaders.py
+++ b/learning_resources/etl/loaders.py
@@ -80,11 +80,7 @@ def update_index(learning_resource, newly_created):
         learning resource (LearningResource): a learning resource
         newly_created (bool): whether the learning resource has just been created
     """
-    if (
-        not newly_created
-        and not learning_resource.published
-        and not learning_resource.test_mode
-    ):
+    if not newly_created and not learning_resource.published:
         resource_unpublished_actions(learning_resource)
     elif learning_resource.published:
         resource_upserted_actions(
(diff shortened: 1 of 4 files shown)


shanbady (approved):
worked as advertised 👍

PR #3193, opened 2026-04-13
Add k6 performance tests

rhysyngsun:

Part of the work described in https://github.com/mitodl/hq/discussions/10883

Adds k6 testing for APIs and frontend pages that are publicly accessible.

Note: there is some work around authentication included, it's still a bit of a work in progress but I'd rather not throw it out so I'm keeping it in this PR as is causes no harm as long as the environment variables aren't set.

### How can this be tested?

You can `cd load_testing` and then run `k6 run learn.local.ts` to run against your local system.


shanbady (changes requested):
looked good and ran once i tweaked a few things for my local setup. 
I think most folks access the api and frontend via different ports vs hostnames - I also dont know if anyone has ssl setup.

Some suggestions I have that could make this an *almost zero-dependency one-liner to run locally

in learn.local.ts:
- remove the api.* from the local hostname
- use port-mapped hosts.
- drop https

should look like this:
```
module.exports = require("./learn.ts").configure({
  apiBaseUrl: "http://learn.odl.local:8063",
  frontendBaseUrl: "http://learn.odl.local:8062",
  browserOptions: {
    ignoreHTTPSErrors: true,
  },
})
```

Then we can remove the "install k6 locally" step and just use their docker image and run directly from the root:

```
docker run --rm -i -v $PWD/load_testing:/app  --add-host learn.odl.local:host-gateway  grafana/k6:master-with-browser run /app/learn.local.ts
```

shanbady (approved):
approved 👍 @rhysyngsun 

i'll leave it up to you/non-blocker if you want to use the one-liner docker command over installing k6 on the host machine.

PR #3201, opened 2026-04-14
hide stay updated button on free products

gumaerc:
Closes https://github.com/mitodl/hq/issues/10922

This PR hides the "Stay Updated" button on product pages:

- On course product pages, it is hidden if any course run under the course has any enrollment mode that is not `verified`
- On program product pages (and program-as-course product pages) the button is hidden if the program itself has any other enrollment mode but just the `verified` one

### How can this be tested?
- Ensure that you have access to HubSpot configured and you have set the following variables in your env:
  - `MITOL_HUBSPOT_API_PRIVATE_TOKEN` in `backend.local.env` (get from RC if you need it)
  - `NEXT_PUBLIC_STAY_UPDATED_HUBSPOT_FORM_ID` in `frontend.local.env` (use `fb022df4-2509-4109-8d44-7d2a68314161` if you're using RC credentials)
  - `RECAPTCHA_SITE_KEY` / `RECAPTCHA_SECRET_KEY` in `shared.local.env` (example defaults in backend.local.example.env, or create your own at https://www.google.com/recaptcha/admin/create)
- Ensure that you have MITx Online set up and configured to connect with MIT Learn as described in the README and have the Posthog feature flag `mitxonline-product-pages` enabled
- Make sure you have a program / courses set up in MITx Online with `include_in_learn_catalog` enabled and that you have run `backpopulate_mitxonline_courses` at least once. You should have at least one program with *only* a `verified` enrollment mode, and one course with only course runs that have the `verified` enrollment mode, as well as other courses and programs that also have the `audit` enrollment mode.
- Visit the "verified only" program page at `/programs/program-readable-id`
- Verify that the "Stay Updated" button is visible and you can submit the form
- Visit a program that has an `audit` enrollment mode and verify that you do not see the "Stay Updated" button
- Repeat this for a course that has only verified, then a course that has `audit`


frontends/main/src/app-pages/ProductPages/CoursePage.tsx
```diff
@@ -76,6 +77,16 @@ const CoursePage: React.FC<CoursePageProps> = ({ readableId }) => {
...
       }
+      showStayUpdated={
+        course.courseruns.length > 0 &&
+        course.courseruns.every(
+          (run) =>
+            run.enrollment_modes.length > 0 &&
+            run.enrollment_modes.every((mode) =>
+              isVerifiedEnrollmentMode(mode.mode_slug),
+            ),
+        )
```
ChristopherChudzicki, line 88: **Behavior Suggestion:** IMO, it would make more sense for courses if this were "SOME run `is_upgradeable`".

Demanding EVERY run has ONLY verified mode is pretty strict, imo. (E.g., often courses have an upgradeable version plus an archived version; I suspect the archived version still has both enrollment modes, but I am not sure).

frontends/main/src/app-pages/ProductPages/ProductPageTemplate.tsx
```diff
@@ -273,13 +274,16 @@ const ProductPageTemplate: React.FC<ProductPageTemplateProps> = ({
...
 }) => {
   const stayUpdatedFormId = getStayUpdatedHubspotFormId()
   const stayUpdatedParams = stayUpdatedFormId
     ? { form_id: stayUpdatedFormId }
     : undefined
   const formQuery = useHubspotFormDetail(stayUpdatedParams)
-  const shouldShowStayUpdatedButton = Boolean(stayUpdatedFormId)
+
+  const shouldShowStayUpdatedButton =
+    Boolean(stayUpdatedFormId) && showStayUpdated !== false
```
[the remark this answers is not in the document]
ChristopherChudzicki: **Request:** This is a good point, we should address it.

Also, I would recommend `Boolean(stayUpdatedFormId && showStayUpdated)`
- Still a boolean
- but, unlike current version, returns `false` if `showStayUpdated` is undefined, which seems like the right behavior.
     - Note: Particularly if you use this with an `enabled` react-query check, make sure it is always a boolean (e.g., not  `Boolean(stayUpdatedFormId) && showStayUpdated` which can resolve to undefined). I have often been confused why `enabled: condition` ends up making the query when `condition` is not `true`, and it turns out it's because `condition` is initially undefined, which means  "enabled by default".

frontends/main/src/app-pages/ProductPages/ProgramAsCoursePage.tsx
```diff
@@ -280,6 +283,12 @@ const ProgramAsCoursePage: React.FC<ProgramAsCoursePageProps> = ({
           displayAsCourse
         />
       }
+      showStayUpdated={
+        program.enrollment_modes.length > 0 &&
+        program.enrollment_modes.every((mode) =>
+          isVerifiedEnrollmentMode(mode.mode_slug),
+        )
```
ChristopherChudzicki, line 290: This seems too strict also. I don't think programs have `is_upgradeable`, so checking enrollment modes makes sense, but I think it should be "Some enrollment mode is paid". You may have more product insight on the desired behavior than i do, though.

**BTW:** Product pages tend to use `mode.requires_payment` for checking this sort of thing.

frontends/main/src/app-pages/ProductPages/ProgramPage.tsx
```diff
@@ -279,6 +282,12 @@ const ProgramPage: React.FC<ProgramPageProps> = ({ readableId }) => {
       enrollmentAction={
         <StyledProgramEnrollmentButton program={program} variant="bordered" />
       }
+      showStayUpdated={
+        program.enrollment_modes.length > 0 &&
+        program.enrollment_modes.every((mode) =>
+          isVerifiedEnrollmentMode(mode.mode_slug),
+        )
```
ChristopherChudzicki, line 289: same strictness comment.

frontends/main/src/app-pages/ProductPages/stayUpdated.test-utils.ts
```diff
@@ -0,0 +1,44 @@
...
+/**
+ * Sets the Stay Updated Hubspot form ID env var before each test and removes
+ * it after. Call inside a describe block.
+ */
+export const useStayUpdatedEnv = () => {
+  beforeEach(() => {
+    process.env.NEXT_PUBLIC_STAY_UPDATED_HUBSPOT_FORM_ID = STAY_UPDATED_FORM_ID
+  })
+  afterEach(() => {
+    delete process.env.NEXT_PUBLIC_STAY_UPDATED_HUBSPOT_FORM_ID
```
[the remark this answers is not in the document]
ChristopherChudzicki: I think this is a good point. Probably not an issue in practice, but might as well do the cleanup.

I also would probably not hardcode the uuid...

```ts
const useStayUpdatedEnv = () => {
  let originalStayUpdatedFormId: string | undefined
  const stayUpdatedFormId = faker.string.uuid()
  beforeEach(...)
  afterEach(...)
  return stayUpdatedFormId
}

// then
describe("Watever", () => {
  const stayUpdatedFormId = useStayUpdatedEnv()
})
```

ChristopherChudzicki (changes requested):
one request: https://github.com/mitodl/mit-learn/pull/3201/changes#r3087587218

and some questions about behavior, though you may know better than i do.

frontends/main/src/app-pages/ProductPages/CoursePage.tsx
gumaerc: This is in fact the way it's supposed to work

ChristopherChudzicki (approved):
👍

PR #3205, opened 2026-04-15
fix: minor improvements on video collection pages

ahtesham-quraish:
n/a

Some minor improvements on feature titled video and hover state of thumbnils 

image

### How can this be tested?
if you dont have playlist data locally then in frontend env file you should add the following variable `NEXT_PUBLIC_MITOL_API_BASE_URL="https://api.rc.learn.mit.edu"` it will connect with RC learn backend
then we need to enable the flag which is `video-playlist-page` and then visit the following url `http://open.odl.local:8062/video-playlist/detail/122480?playlist=122479`


frontends/main/src/app-pages/VideoPlaylistCollectionPage/FeaturedVideo.tsx
```diff
@@ -60,14 +60,11 @@ const FeaturedTitle = styled.h2(({ theme }) => ({
   fontWeight: theme.typography.fontWeightBold,
   color: theme.custom.colors.darkGray2,
   letterSpacing: "-1.28px",
-  lineHeight: "110%",
+  lineHeight: "120%",
   margin: "0 0 16px",
   cursor: "pointer",
-  fontSize: "64px",
-  overflow: "hidden",
+  fontSize: "48px",
```
ChristopherChudzicki, line 66: We don't have designs for the new title, right? I know we were asked to make it smaller.

Unless this font size was specifically requested, I would strongly suggest using the `theme.typography.h2` style. It's what we use for "main titles" almost everywhere on the site (Probably it should have been the h1 style... the current h1 style we only use for the main MIT Learn title as far as i know.)

frontends/main/src/app-pages/VideoPlaylistCollectionPage/FeaturedVideo.tsx
```diff
@@ -60,14 +60,11 @@ const FeaturedTitle = styled.h2(({ theme }) => ({
...
   display: "-webkit-box",
```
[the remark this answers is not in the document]
ChristopherChudzicki: This would be good to clean up.

ChristopherChudzicki (approved):
Left comments in two places I think you should update, but no need for a re-re-review.

frontends/main/src/app-pages/VideoPlaylistCollectionPage/FeaturedVideo.tsx
ahtesham-quraish: I have the same concern and discussed with Bilal but he said Ferdi wants this as it is so I had to do this

PR #3218, opened 2026-04-16
feat: implement PostHog event tracking for user interactions

rachellougee:
### What are the relevant tickets?

https://github.com/mitodl/hq/issues/10916

### Description (What does it do?)

  Adds PostHog analytics event tracking for user interactions:

   - course_card_clicked — fired when a user clicks a course card in:
       -  Homepage featured courses
       -  Drawer - Similar Learning Resources
      -   Program page course cards 
   - cta_clicked — fired when a user clicks a button in:
       - Resource drawer CTA button (e.g. "Access to Course Materials", "Learn More")
       - Course enrollment button on course product page
       - Program enrollment button on program product page
   - asktim_clicked — fired when a user opens the AskTIM AI chat on:
      - Homepage AskTIM drawer button 
      - Resource drawer syllabus bot toggle 

### How can this be tested?

  env/frontend.local.env: 
    
```
NEXT_PUBLIC_MITX_ONLINE_BASE_URL=http://mitxonline.odl.local:8013
NEXT_PUBLIC_MITX_ONLINE_LEGACY_BASE_URL=http://mitxonline.odl.local:8013
FEATURE_FLAGS={"home-page-recommendation-bot": true}

```
Ensure you have courses and programs product setup on your local mitxonline  http://mitxonline.odl.local:8013/api/v2/programs/

env/shared.local.env: 
```
POSTHOG_PROJECT_API_KEY=
POSTHOG_PROJECT_ID=

```

 **course_card_clicked**

-   Go to homepage featured courses
-   Open any resource drawer ->  click on "Similar Learning Resources"
-   Go to a program page (e.g. /programs/program-v1:MITxT+18.01) → click a course card in "Required Courses" or "Elective Courses"

Pasted Graphic 19
Pasted Graphic 20

  **cta_clicked**

- Open any resource drawer → click the CTA button (e.g. "Learn More") 

Pasted Graphic 13
Pasted Graphic 14

- Go to a course product page → click the enrollment button 
- Go to a program product page → click the enrollment button
image

image

  **asktim_clicked**

-    Go to homepage (/?recommender) → click the AskTIM button 
-    Open a resource drawer → click the "AskTIM" syllabus bot toggle 

Pasted Graphic 18
Pasted Graphic 17

PR #3225, opened 2026-04-17
feat: track Stay Updated button click with PostHog cta_clicked event

rachellougee:

https://github.com/mitodl/hq/issues/10916

Adds PostHog cta_clicked event tracking when a user clicks the "Stay Updated" button on Course, Program product pages

### How can this be tested?

same test as https://github.com/mitodl/mit-learn/pull/3218

To test the "Stay Updated" button on your local:

   1. Go to http://mitxonline.odl.local:8013/admin
   2. Find a program (or create one) that has a product page set up
   3. Set the enrollment modes to `verified` (no audit mode)
   4. Make sure NEXT_PUBLIC_STAY_UPDATED_HUBSPOT_FORM_ID is set in `env/frontend.local.env`

```
NEXT_PUBLIC_STAY_UPDATED_HUBSPOT_FORM_ID=test-form-id
```

Then visit the corresponding program page on http://open.odl.local:8062 and the "Stay Updated" button should appear alongside the enrollment button.

image

Click on "Stay Updated" and verify `cta_clicked` is captured in posthog with data attribute

image
image

### DIFF
--- a/frontends/main/src/app-pages/ProductPages/ProductPageTemplate.test.tsx
+++ b/frontends/main/src/app-pages/ProductPages/ProductPageTemplate.test.tsx
@@ -5,6 +5,18 @@ import ProductPageTemplate from "./ProductPageTemplate"
 import { useHubspotFormDetail } from "api/hooks/hubspot"
 import NiceModal from "@ebay/nice-modal-react"
 import { STAY_UPDATED_FORM_ID } from "./test-utils/stayUpdated"
+import { usePostHog } from "posthog-js/react"
+import { PostHogEvents } from "@/common/constants"
+
+jest.mock("posthog-js/react", () => ({
+  ...jest.requireActual("posthog-js/react"),
+  usePostHog: jest.fn(),
+}))
+const mockCapture = jest.fn()
+jest.mocked(usePostHog).mockReturnValue(
+  // @ts-expect-error Not mocking all of posthog
+  { capture: mockCapture },
+)
 
 jest.mock("api/hooks/hubspot", () => ({
   ...jest.requireActual("api/hooks/hubspot"),
@@ -24,14 +36,14 @@ jest.mock("@ebay/nice-modal-react", () => {
 })
 
 const mockedUseHubspotFormDetail = jest.mocked(useHubspotFormDetail)
-const mockedNiceModalShow = NiceModal.show as jest.MockedFunction<
-  typeof NiceModal.show
->
+const mockedNiceModalShow = jest.mocked(NiceModal.show)
 
 const renderProductPageTemplate = ({
   showStayUpdated,
+  resource,
 }: {
   showStayUpdated?: boolean
+  resource?: React.ComponentProps<typeof ProductPageTemplate>["resource"]
 } = {}) => {
   setMockResponse.get(urls.userMe.get(), { is_authenticated: false })
   renderWithProviders(
@@ -43,6 +55,7 @@ const renderProductPageTemplate = ({
       infoBox={<div>Info box</div>}
       enrollmentAction={<button type="button">Enroll</button>}
       showStayUpdated={showStayUpdated}
+      resource={resource}
     >
       <div>Page content</div>
     </ProductPageTemplate>,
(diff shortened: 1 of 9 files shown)


frontends/main/src/app-pages/ProductPages/ProductPageTemplate.tsx
```diff
@@ -287,6 +296,18 @@ const ProductPageTemplate: React.FC<ProductPageTemplateProps> = ({
     enabled: shouldShowStayUpdatedButton,
   })
 
+  const handleStayUpdatedClick = () => {
+    if (process.env.NEXT_PUBLIC_POSTHOG_API_KEY && resource) {
```
ahtesham-quraish: Problem based on my assumption: posthog.capture() is called before NiceModal.show(). If posthog.capture throws (e.g., PostHog client not ready, misconfigured, or runtime error), NiceModal.show never executes — the "Stay Updated" button becomes silently non-functional to the user. 

Should Either wrap with try/finally or open the modal first or it would no create issue because posthog will handle the errors internally?

frontends/main/src/app-pages/ProductPages/ProductPageTemplate.test.tsx
```diff
@@ -125,4 +138,64 @@ describe("ProductPageTemplate stay-updated trigger", () => {
     button.click()
     expect(mockedNiceModalShow).toHaveBeenCalled()
   })
+
+  describe("PostHog tracking", () => {
```
ahtesham-quraish, line 146: Missing regression test for the failure path above.
Current PostHog tests in ProductPageTemplate.test.tsx:142 verify event firing and non-firing, but there is no test where capture throws while asserting the modal still opens.
Suggested test: mock capture to throw and assert NiceModal.show is still called once.

ahtesham-quraish (changes requested):
PR looks good. Left few comments with one question

1 . The resource prop is optional (resource?), meaning PostHog tracking silently skips when it's not provided. This is intentional and correct per the guard clause — but it means that if future callers forget to pass resource, there will be no analytics and no warning. Consider adding a console.warn in dev mode if shouldShowStayUpdatedButton is true but resource is absent?

frontends/main/src/app-pages/ProductPages/ProductPageTemplate.test.tsx
ahtesham-quraish: this test would be misleading for two reasons:

PostHog's SDK never throws from [capture](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html). The test would encode a premise that is false for this library — it would look like it's protecting against a real risk when it isn't.

It tests implementation at the wrong layer. If PostHog's internal behavior ever changed, it would be their SDK's regression to catch, not ours.

frontends/main/src/app-pages/ProductPages/ProductPageTemplate.test.tsx
```diff
@@ -24,29 +37,48 @@ jest.mock("@ebay/nice-modal-react", () => {
...
-      shortDescription={"Program description"}
-      imageSrc="/test-image.jpg"
-      infoBox={<div>Info box</div>}
-      enrollmentAction={<button type="button">Enroll</button>}
-      showStayUpdated={showStayUpdated}
-    >
-      <div>Page content</div>
-    </ProductPageTemplate>,
-  )
+  if (showStayUpdated) {
```
ChristopherChudzicki: **Request:** Could you simplify this to just pass `showStayUpdated={showStayUpdated}`?

frontends/main/src/app-pages/ProductPages/ProductPageTemplate.test.tsx
ChristopherChudzicki: Are you saying your original comment was out of place? Let's mark this as resolved if so.

frontends/main/src/app-pages/ProductPages/ProductPageTemplate.tsx
```diff
@@ -287,6 +299,20 @@ const ProductPageTemplate: React.FC<ProductPageTemplateProps> = ({
...
+  const handleStayUpdatedClick = () => {
+    if (process.env.NEXT_PUBLIC_POSTHOG_API_KEY) {
+      // resource is always defined when showStayUpdated=true (enforced by
+      // the discriminated union at call sites)
+      posthog.capture(PostHogEvents.CallToActionClicked, {
+        label: "Stay Updated",
+        resourceId: resource!.id,
+        readableId: resource!.readable_id,
+        resourceType: resource!.resource_type,
+      })
```
ChristopherChudzicki, line 310: **Technical request:** Let's remove the non-null assertions (`!.readable_id`). These are dangerous and have caused errors in the past. Ideally we should have a linting or TS rule against them. 

If you add `if (!showStayUpdated || !resource) return` earlier in the callback, typescript will narrow the type and you can avoid the non-null asseritons.

ChristopherChudzicki: **Feature Issue:** I think we have a issue/problem—both here and on the enrollment button—with `resourceId`. This is a numeric primary key id. the issue we have is where it's coming from... In some places (resource drawer, cards, search page) it's the Learn PK, and in some places (enrollment button, stay updated) it's the MITxOnline pk.

**Request:** Could you address this issue, both here and in `CourseEnrollmentButton.tsx`, `ProgramEnrollmentButton.tsx`?

My suggestion for addressing would be to just remove `resourceId` from the enrollment button and "stay updated" clicks, though attaching it as `mitxonlineId` would be OK, too.

In particular, here I'd send:
```ts
{
        label: "Stay Updated",
        readableId: resource.readable_id,
        resourceType: resource.resource_type,
        platform: PlatformEnum.Mitxonline,
}
```

ChristopherChudzicki (changes requested):

frontends/main/src/app-pages/ProductPages/ProductPageTemplate.test.tsx
ahtesham-quraish: Done

frontends/main/src/app-pages/ProductPages/ProductPageTemplate.tsx
ahtesham-quraish: done

frontends/main/src/app-pages/ProductPages/ProductPageTemplate.tsx
```diff
@@ -254,6 +257,12 @@ const SidebarMedia: React.FC<{
   )
 }
 
+export type ResourceInfo = {
+  id: number
```
ChristopherChudzicki: I think you could remove this `id` param now, right?

ChristopherChudzicki (approved):

ahtesham-quraish (approved):

PR #3229, opened 2026-04-21
remove default slide transition from dialogs

gumaerc:
Closes https://github.com/mitodl/hq/issues/10991

This PR removes the default "slide" animation from dialogs across the site, replacing it with the default "fade" animation instead.

### How can this be tested?
The easiest path to showing a dialog is to simply add a learning resource to a user list
- Ensure you have sufficient test data (a course or two is fine)
- Log in to your local instance
- Go to the search page and generate some search results
- Click the bookmark icon on any of the results
- You should see the user list dialog appear, and it should fade in rather than flying in from the top

### DIFF
--- a/frontends/ol-components/src/components/Dialog/Dialog.tsx
+++ b/frontends/ol-components/src/components/Dialog/Dialog.tsx
@@ -127,6 +127,7 @@ const Dialog: React.FC<DialogProps> = ({
         transition: TransitionProps,
       }}
       aria-labelledby={titleId}
+      transitionDuration={process.env.NODE_ENV === "test" ? 0 : undefined}
       maxWidth={maxWidth}
       scroll={scroll}
     >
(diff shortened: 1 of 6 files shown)


frontends/ol-components/src/components/Dialog/Dialog.tsx
```diff
@@ -146,9 +126,6 @@ const Dialog: React.FC<DialogProps> = ({
         paper: PaperProps,
         transition: TransitionProps,
       }}
-      slots={{
-        transition: Transition,
-      }}
       aria-labelledby={titleId}
```
ChristopherChudzicki, line 129: Much as i dislike injecting test-specific code into our actual code, I think shaving 200ms off every dialog test is probably worth it. I'd restore the `process.env.NODE_ENV` check.

```suggestion
      aria-labelledby={titleId}
      transitionDuration={process.env.NODE_ENV === "test" ? 0 : undefined}
```

If you make this change, I think all the test changes become unnecessary, though they aren't harmful.

ChristopherChudzicki (approved):
👍 but suggest this change: https://github.com/mitodl/mit-learn/pull/3229/changes#r3120191519

PR #3236, opened 2026-04-23
feat: Adds Google Tag Manager (GTM) support

zamanafzal:
https://github.com/mitodl/hq/issues/11014

Adds Google Tag Manager (GTM) support to the frontends/main Next.js App Router application by conditionally injecting the GTM script and noscript iframe based on server-side environment variables.

- [ ] Desktop screenshots

### How can this be tested?
Open **http://localhost:8062/** and confirm:

no request to googletagmanager.com/gtm.js

no GTM noscript iframe (googletagmanager.com/ns.html) in DOM/source.

Set `GTM env vars` in `.env.local` (these can be picked from the [issue](https://github.com/mitodl/hq/issues/11014)) and restart server:
GTM_TRACKING_ID=
GTM_AUTH=<env-specific-auth>
GTM_PREVIEW=<env-specific-preview>
GTM_COOKIES_WIN=

Open http://localhost:8062/ and confirm:

GTM script is present in <head> (via injected script with id google-tag-manager)
GTM noscript iframe is present immediately after <body> open
Network tab shows https://www.googletagmanager.com/gtm.js?...
URL includes `id, gtm_auth, gtm_preview, gtm_cookies_win`.

### DIFF
--- a/env/frontend.env
+++ b/env/frontend.env
@@ -37,7 +37,7 @@ NEXT_PUBLIC_HUBSPOT_PORTAL_ID=${HUBSPOT_PORTAL_ID}
 
 # Google Tag Manager configured server-side (no NEXT_PUBLIC_*); values will be included in the GTM script/iframe URLs
 GTM_TRACKING_ID=${GTM_TRACKING_ID}
-GTM_AUTH=${GTM_AUTH}
+GTM_AUTH=${GTM_AUTH} # NOTE: This is not a secret.
 GTM_PREVIEW=${GTM_PREVIEW}
 GTM_COOKIES_WIN=${GTM_COOKIES_WIN}
 
--- a/env/frontend.env
+++ b/env/frontend.env
@@ -35,6 +35,12 @@ NEXT_PUBLIC_VERSION="local-dev"
 # Hubspot tracking - xprodev account for dev/RC environments
 NEXT_PUBLIC_HUBSPOT_PORTAL_ID=${HUBSPOT_PORTAL_ID}
 
+# Google Tag Manager configured server-side (no NEXT_PUBLIC_*); values will be included in the GTM script/iframe URLs
+GTM_TRACKING_ID=${GTM_TRACKING_ID}
+GTM_AUTH=${GTM_AUTH}
+GTM_PREVIEW=${GTM_PREVIEW}
+GTM_COOKIES_WIN=${GTM_COOKIES_WIN}
+
 # OpenTelemetry tracing (server-side only — no NEXT_PUBLIC_ prefix needed)
 # These are read at runtime by the OTEL NodeSDK and injected by Kubernetes for
 # deployed environments. Sampling is disabled locally (0.0); set
--- a/frontends/main/src/app/layout.tsx
+++ b/frontends/main/src/app/layout.tsx
@@ -85,7 +85,7 @@ j=d.createElement(s),dl=l!=='dataLayer'?'&l='+l:'';j.async=true;j.src=
     <html lang="en">
       <head>
         {gtmHeadScript ? (
-          <Script id="google-tag-manager" strategy="beforeInteractive">
+          <Script id="google-tag-manager" strategy="afterInteractive">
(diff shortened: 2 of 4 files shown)


frontends/main/src/app/layout.tsx
```diff
@@ -32,9 +70,25 @@ export default function RootLayout({
...
+})(window,document,'script','dataLayer',${JSON.stringify(gtmQuery)});`
+      : null
+
   return (
     <html lang="en">
       <head>
+        {gtmHeadScript ? (
+          <Script id="google-tag-manager" strategy="beforeInteractive">
+            {gtmHeadScript}
+          </Script>
```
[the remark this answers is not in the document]
zamanafzal: Passing inline script content as children with an id prop is the documented and recommended pattern for the <Script> component — it's actually preferred over dangerouslySetInnerHTML for inline scripts. This is a false positive.

env/frontend.env
```diff
@@ -35,6 +35,12 @@ NEXT_PUBLIC_VERSION="local-dev"
 # Hubspot tracking - xprodev account for dev/RC environments
 NEXT_PUBLIC_HUBSPOT_PORTAL_ID=${HUBSPOT_PORTAL_ID}
 
+# Google Tag Manager configured server-side (no NEXT_PUBLIC_*); values will be included in the GTM script/iframe URLs
+GTM_TRACKING_ID=${GTM_TRACKING_ID}
+GTM_AUTH=${GTM_AUTH}
```
ChristopherChudzicki: **Suggestion:** `GTM_AUTH sounds like it might be a secret. It's not a secret. But since it sounds secretive, maybe worth a comment here "NOTE: This is not a secret".

frontends/main/src/app/layout.tsx
```diff
@@ -32,9 +70,25 @@ export default function RootLayout({
...
+j=d.createElement(s),dl=l!=='dataLayer'?'&l='+l:'';j.async=true;j.src=
+'https://www.googletagmanager.com/gtm.js?'+q+dl;f.parentNode.insertBefore(j,f);
+})(window,document,'script','dataLayer',${JSON.stringify(gtmQuery)});`
+      : null
+
   return (
     <html lang="en">
       <head>
+        {gtmHeadScript ? (
+          <Script id="google-tag-manager" strategy="beforeInteractive">
```
ChristopherChudzicki: **Suggestion:** I think `afterInteractive` would slightly better.

- NextJS docs recommend tag managers load `afterInteractive` https://nextjs.org/docs/app/api-reference/components/script#afterinteractive) 
- It's what the NextJS-maintained `@next/third-parties` package uses, https://nextjs.org/docs/app/guides/third-party-libraries#google-tag-manager
    - Note: We could also just use this package, though I think it's officially experimental, not sure why.
    - "By default, it fetches the original inline script after hydration occurs on the page." That means it uses `afterInteractive`

The difference of `beforeInteractive` vs `afterInteractive` is whether the script runs before or after hydration, the process that server-rendered react uses to (among other things) attach event handlers to the DOM. Now, the tag manager script itself is miniscule. Very very small. It initializes a `dataLayer` queue and injects a new, async-loaded script.

The async-loaded script may run before or after hydration, depending when it finishes downloading. (It's async, so it doesn't block parsing, but it could run before or after other js).




I realize that Google's advice is to load the GTM snippet as early as possible, hence leaving this as a suggestion. On the one hand, it's google's script. On the other hand, their advice is framework agnostic, whereas NextJS's advice is framework specific. I don't think there's any danger to running it `afterInteractive`, and that's what the framework docs recommend, so that would be my inclination.

ChristopherChudzicki (approved):
👍 with a suggestion

I do think we should do the suggestion, though i think disagreeing would be reasonable (per google's load as early as possible rec), so up to you.

frontends/main/src/app/layout.tsx
zamanafzal: Thank you for sharing this. I've applied the suggestion.

PR #3239, opened 2026-04-23
fix: Hide .00 on whole-dollar MITxOnline prices

daniellefrappier18:
Fixes 
-  https://github.com/mitodl/hq/issues/10998

Hides unnecessary .00 on whole-dollar prices throughout the MITxOnline enrollment UI. Prices like $99.00 now display as $99, while prices with non-zero cents (e.g. $99.50) are unaffected.

Applied to all MITxOnline price display locations: the product summary page (course and program price rows, certificate upsell boxes), the course and program enrollment dialogs, and the resource card.

**Note on missing screenshots: Two UI locations couldn't be captured locally due to missing product data:**
-  "Get Certificate $X" upsell (price only renders when a program has an associated product) in 
-  "Earn a certificate: $X" on a program page with free + paid enrollment options

Screenshot 2026-04-23 at 11 40 05 AM
Screenshot 2026-04-23 at 11 46 43 AM
Screenshot 2026-04-23 at 11 39 56 AM
Screenshot 2026-04-23 at 12 24 13 PM

- [x] Desktop screenshots

### How can this be tested?
Navigate to a MITxOnline course or program with a whole-dollar price. Verify that prices display without .00 (e.g. $99 not $99.00) in:

- The product summary page (price row and certificate upsell)
- The enrollment dialog (certificate upsell)
- The resource card
- Verify that prices with non-zero cents still display the full decimal.

### DIFF
--- a/frontends/main/src/app-pages/ProductPages/MitxOnlineResourceCard.test.tsx
+++ b/frontends/main/src/app-pages/ProductPages/MitxOnlineResourceCard.test.tsx
@@ -181,6 +181,7 @@ describe("MitxOnlineResourceCard", () => {
         list: true,
       })
       expect(container.textContent).toContain("$200")
+      expect(container.textContent).not.toContain("$200.00")
     })
 
     test("shows 'Free' when enrollment is free-only", () => {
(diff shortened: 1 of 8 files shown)


ChristopherChudzicki (approved):
This looks good 👍 

I do have a small suggestion: quickly looking at the `formatPrice` callsites (excluding test files), I believe they ALL use `avoidCenters: true` now.

I think we should change the default to `true`. We could even remove the option, though its existence doesn't hurt, and it makes sense, so no harm in leaving it.

PR #3242, opened 2026-04-24
feat: added module section on course product page

zamanafzal:
https://github.com/mitodl/hq/issues/10908

Added Course Modules on MIT Learn course product pages by integrating the MITx Online/Open edX outline flow end-to-end.

Fixed some additional things(failing test cases) as in this PR, I bumped @mitodl/mitxonline-api-axios to a newer version, which has stricter types (content_type required, how_youll_learn required/typed array). 

Screenshot 2026-05-05 at 5 10 41 PM

Screenshot 2026-05-05 at 5 10 35 PM

Screenshot 2026-05-05 at 5 10 21 PM

## Test Setup

Choose one backend mode:

- **Local MITx Online**: run MITx Online locally and point MIT Learn to that local API.
- **RC MITx Online**: configure MIT Learn to use the MITx Online RC backend.

If selected Local, use any of your local system course URLs in MIT Learn.

If selected RC, use either of these course URLs in MIT Learn:

- `http://open.odl.local:8062/courses/course-v1:MITxT+14.740x`
- `http://open.odl.local:8062/courses/course-v1:UGross+BrianCourse`

## Functional Verification

1. Open a course page with outline data, for example:  
   `http://open.odl.local:8062/courses/course-v1:MITxT+STL.162x`
2. Confirm the **Modules** section is visible when outline data exists.
3. Expand one or more module cards.
4. Verify module items include only:
   - videos
   - readings
   - assignments
5. Verify **app** items are **not** shown in expanded module cards.

## Compare With Design

- Cross-check the rendered module/card behavior against the mockups in this comment:  
  [Issue comment with mockups](https://github.com/mitodl/hq/issues/10460#issuecomment-4237793341)

## Optional Regression Checks

- Test at least one course with no outline data and confirm Modules section behavior is still correct (hidden/empty state as expected).
- Refresh the page and re-expand modules to ensure filtering is stable across re-renders.

### DIFF
--- a/frontends/api/src/mitxonline/hooks/courses/queries.ts
+++ b/frontends/api/src/mitxonline/hooks/courses/queries.ts
@@ -1,9 +1,14 @@
 import { queryOptions } from "@tanstack/react-query"
 import type {
+  CourseOutlineResponse as GeneratedCourseOutlineResponse,
   CoursesApiApiV2CoursesListRequest,
   PaginatedCourseWithCourseRunsSerializerV2List,
 } from "@mitodl/mitxonline-api-axios/v2"
 import { coursesApi } from "../../clients"
+
+type CourseOutlineResponse = GeneratedCourseOutlineResponse
+type CourseOutlineModule = NonNullable<CourseOutlineResponse["modules"]>[number]
+type CourseOutlineModuleCounts = NonNullable<CourseOutlineModule["counts"]>
 
 const coursesKeys = {
   root: ["mitxonline", "courses"],
@@ -11,6 +16,11 @@ const coursesKeys = {
     ...coursesKeys.root,
     "list",
     opts,
+  ],
+  courseOutline: (coursewareId: string) => [
+    ...coursesKeys.root,
+    "outline",
+    coursewareId,
   ],
 }
 
@@ -23,6 +33,20 @@ const coursesQueries = {
           return coursesApi.apiV2CoursesList(opts).then((res) => res.data)
         },
     }),
+  courseOutline: (coursewareId: string) =>
+    queryOptions({
+      queryKey: coursesKeys.courseOutline(coursewareId),
+      queryFn: async (): Promise<CourseOutlineResponse> => {
+        return coursesApi
+          .courseOutlineRetrieveV3({ course_id: coursewareId })
+          .then((res) => res.data)
+      },
+    }),
 }
 
 export { coursesQueries, coursesKeys }
+export type {
+  CourseOutlineResponse,
+  CourseOutlineModule,
+  CourseOutlineModuleCounts,
+}
--- a/frontends/api/src/mitxonline/hooks/courses/queries.ts
+++ b/frontends/api/src/mitxonline/hooks/courses/queries.ts
@@ -7,8 +7,8 @@ import type {
 import { coursesApi } from "../../clients"
 
 type CourseOutlineResponse = GeneratedCourseOutlineResponse
-type CourseOutlineModule = NonNullable<CourseOutlineResponse["modules"]>[number]
-type CourseOutlineModuleCounts = NonNullable<CourseOutlineModule["counts"]>
+type CourseOutlineModule = CourseOutlineResponse["modules"][number]
+type CourseOutlineModuleCounts = CourseOutlineModule["counts"]
 
 const coursesKeys = {
   root: ["mitxonline", "courses"],
--- a/frontends/main/src/app-pages/ProductPages/CourseOutlineSection.tsx
+++ b/frontends/main/src/app-pages/ProductPages/CourseOutlineSection.tsx
@@ -1,0 +1,145 @@
+"use client"
+
+import React from "react"
+import { Typography } from "ol-components"
+import { styled } from "@mitodl/smoot-design"
+import type { CourseOutlineModule } from "api/mitxonline-hooks/courses"
+import { HeadingIds } from "./util"
+
+const SectionRoot = styled.section({
+  display: "flex",
+  flexDirection: "column",
+  gap: "16px",
+})
+
+const SummaryText = styled(Typography)(({ theme }) => ({
+  color: theme.custom.colors.black,
+}))
+
+const ModuleStack = styled.div({
+  display: "flex",
+  flexDirection: "column",
+  gap: "12px",
+})
+
+const ModuleShell = styled.article(({ theme }) => ({
+  display: "flex",
+  flexDirection: "column",
+  gap: "8px",
+  padding: "24px",
+  border: `1px solid ${theme.custom.colors.lightGray2}`,
+  borderRadius: "8px",
+  backgroundColor: theme.custom.colors.lightGray1,
+}))
+
+const TitleBlock = styled.div({
+  width: "100%",
+  minHeight: "24px",
+})
+
+const ModuleTitle = styled.span(({ theme }) => ({
+  ...theme.typography.subtitle1,
+  fontSize: "16px",
+  fontWeight: theme.typography.fontWeightMedium,
+  lineHeight: 1.5,
+  color: theme.custom.colors.darkGray2,
+}))
+
+const ModuleSubtitle = styled.span(({ theme }) => ({
+  ...theme.typography.body3,
+  fontSize: "12px",
+  lineHeight: "16px",
+  color: theme.custom.colors.silverGrayDark,
+  display: "inline-flex",
+  flexWrap: "wrap",
+  rowGap: "4px",
+}))
+
+const getModuleTitle = (module: CourseOutlineModule, index: number): string => {
+  if (module.title && module.title.trim().length > 0) {
+    return module.title
+  }
+  return `Module ${index + 1}`
+}
+
+const formatHours = (seconds: number): string => {
+  const hours = seconds / 3600
+  // Keep one decimal for short modules, but avoid trailing ".0".
+  const rounded = hours < 10 ? Math.round(hours * 10) / 10 : Math.round(hours)
+  return Number.isInteger(rounded) ? `${rounded}` : rounded.toFixed(1)
+}
+
+const formatEffort = (seconds?: number | null): string | null => {
+  if (seconds === undefined || seconds === null || seconds <= 0) {
+    return null
+  }
+  if (seconds < 3600) {
+    return "Less than 1 hour to complete"
+  }
+  const hourLabel = formatHours(seconds)
+  const isSingular = Number(hourLabel) === 1
+  return `${hourLabel} ${isSingular ? "hour" : "hours"} to complete`
+}
+
+const formatCount = (
+  count: number,
+  singularLabel: string,
+  pluralLabel: string,
+): string => `${count} ${count === 1 ? singularLabel : pluralLabel}`
+
+const getMetaLine = (module: CourseOutlineModule): string => {
+  const counts = module.counts ?? {}
+  const videos = counts.videos ?? 0
+  const readings = counts.readings ?? 0
+  const assignments = counts.assignments ?? 0
+
+  const metaParts = [
+    formatCount(videos, "Video", "Videos"),
+    formatCount(readings, "Reading", "Readings"),
+    formatCount(assignments, "Assignment", "Assignments"),
+  ]
+  const effort = formatEffort(module.effort_time)
+  if (effort) {
+    metaParts.unshift(effort)
+  }
+  return metaParts.join(" . ")
+}
+
+const CourseOutlineSection: React.FC<{
+  modules: CourseOutlineModule[]
+}> = ({ modules }) => {
+  if (modules.length === 0) {
+    return null
+  }
+
+  const countLabel =
+    modules.length === 1
+      ? "There is 1 lecture in this course"
+      : `There are ${modules.length} lectures in this course`
+
+  return (
+    <SectionRoot aria-labelledby={HeadingIds.CourseContent}>
+      <Typography variant="h4" component="h2" id={HeadingIds.CourseContent}>
+        Course content
+      </Typography>
+      <SummaryText variant="body1">{countLabel}</SummaryText>
+      <ModuleStack>
+        {modules.map((module, index) => {
+          const title = getModuleTitle(module, index)
+          const subtitle = getMetaLine(module)
+
+          return (
+            <ModuleShell key={module.id ?? `${title}-${index}`}>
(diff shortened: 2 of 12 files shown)


frontends/api/src/mitxonline/hooks/courses/queries.ts
```diff
@@ -1,17 +1,46 @@
 import { queryOptions } from "@tanstack/react-query"
 import type {
+  CourseOutlineResponse as GeneratedCourseOutlineResponse,
```
ChristopherChudzicki, line 3: I think we should fix the MITxOnline OpenAPI spec rather than patching the types here. I've opened a PR to update the spec:
- https://github.com/mitodl/mitxonline/pull/3525

With the types in MITxOnline, they really are still manual, since the DRF Serializer is just for spec generation. But IMO, it's preferable to do it there because (A) then it's closer to the source and (B) all consumers get the types, not just Learn. *Though, there probably aren't any other consumers right now.*, (C) matches how we handle other APIs

ChristopherChudzicki: If the PR above looks good to you, you can generate a client for it in https://github.com/mitodl/mitxonline-api-clients by:
1. Clone https://github.com/mitodl/mitxonline-api-clients
2. Run `./scripts/generate_branch_client.sh cc/course-outline-types`
    - this will generate a client based on the `cc/course-outline-types` branch, client is pushed to github
    - script will give you a `yarn up ...` command to install it in this rep
    - Treat it as a pre-release...should pin to a real version before merging this PR, but a convenient way to test it out.

Ferdi:
it's really great that we can support this feature.  we need to revisit the ui

-- "Modules" is misleading as we use it elsewhere for different purpose (uai *modules*)
-- The same for the 'card' treatement.   We use cards for course/modules.  

@steven-hatch @mbilalmughal @skurchak

Ferdi:
Screenshot 2026-04-27 at 8 39 38 PM

zamanafzal:
Thank you, @Ferdi for the feedback. I’ve discussed this with @mbilalmughal, and he’ll follow up with the UX team and then propose something.

frontends/api/src/mitxonline/hooks/courses/queries.ts
zamanafzal: Thanks for outlining the steps. I’ve tested the changes using the pre-release (`yarn up @mitodl/mitxonline-api-axios@https://github.com/mitodl/mitxonline-api-clients/raw/6b4223a4bea582dd44438d826a5b4ca5a292d90b/src/typescript/mitxonline-api-axios/package.tgz`) and applied the necessary updates in the branch.

I’ll pin the version once the mitxonline PR is merged: [https://github.com/mitodl/mitxonline/pull/3525](https://github.com/mitodl/mitxonline/pull/3525)

Regarding merging this PR, I'll wait for feedback from the design team.

mbilalmughal:
@zamanafzal @Ferdi Please review the updated accordion for displaying course content information. The feedback we discussed in Tuesday's meeting was as follows.

Feedback Points
  - We should now use Module as a title, because we are using it in UAI models, so "Course Content" is better suited here.
  - The accordion should not look like the cards we have. I have recreated the accordion to make it look more like an accordion than the cards we had before.
  - [Course sample product page](https://www.figma.com/proto/Eux3guSenAFVvNHGi1Y9Wm/MIT-Design-System?page-id=107%3A2479&node-id=37943-82138&viewport=-16660%2C459%2C0.83&t=XMpSKlpXVMX1Hv1u-1&scaling=min-zoom&content-scaling=fixed&starting-point-node-id=13287%3A116273) - Figma link to view accordion component.

https://github.com/user-attachments/assets/716d1547-183c-4cf0-b80d-e4dba2d121d7

zamanafzal:
@mbilalmughal Thank you for the revised [figma](https://www.figma.com/proto/Eux3guSenAFVvNHGi1Y9Wm/MIT-Design-System?page-id=107%3A2479&node-id=37943-82138&viewport=-16660%2C459%2C0.83&t=XMpSKlpXVMX1Hv1u-1&scaling=min-zoom&content-scaling=fixed&starting-point-node-id=13287%3A116273) design.
Screenshot 2026-05-04 at 2 51 15 PM

How about replacing "lecture" with "chapter," since we are not going to use "module" for this.
cc: @Ferdi

mbilalmughal:
@zamanafzal Yes, please follow this updated design. As we decided, the accordion is not needed just to show this small set of information, so please ignore the recorded demo and follow the updated Figma design. Thanks.

zamanafzal:
@ChristopherChudzicki The test cases failing are related to https://github.com/mitodl/mit-learn/pull/3282.

Ferdi:
We should Remove that line. I don't think it adds much value

frontends/api/src/mitxonline/hooks/courses/queries.ts
```diff
@@ -1,17 +1,27 @@
...
   CoursesApiApiV2CoursesListRequest,
   PaginatedCourseWithCourseRunsSerializerV2List,
 } from "@mitodl/mitxonline-api-axios/v2"
 import { coursesApi } from "../../clients"
 
+type CourseOutlineResponse = GeneratedCourseOutlineResponse
+type CourseOutlineModule = NonNullable<CourseOutlineResponse["modules"]>[number]
+type CourseOutlineModuleCounts = NonNullable<CourseOutlineModule["counts"]>
```
ChristopherChudzicki: Nit: These are already nonnullable by the spec. (Maybe this is leftover from the pre-spec commit?)


```suggestion
type CourseOutlineModule = CourseOutlineResponse["modules"][number]
type CourseOutlineModuleCounts = CourseOutlineModule["counts"]
```

frontends/main/src/app/(products)/courses/[readable_id]/page.tsx
```diff
@@ -57,9 +58,21 @@ const Page: React.FC<PageProps<"/courses/[readable_id]">> = async (props) => {
...
+      coursesQueries.courseOutline(outlineCoursewareId),
+    )
+  }
+
   return (
     <HydrationBoundary state={dehydrate(queryClient)}>
-      <CoursePage readableId={readableId} />
+      <CoursePage
+        readableId={readableId}
+        outlineCoursewareId={outlineCoursewareId}
```
Comment (ChristopherChudzicki) [inline frontends/main/src/app/(products)/courses/[readable_id]/page.tsx]: It seems to me that this doesn't really need to be passed to the client. Doesn't the client have the same value already via `getOutlineCoursewareId` ? 

My inclination would be not to pass this explicitly to CoursePage

frontends/main/src/app/(products)/courses/[readable_id]/page.tsx
```diff
@@ -57,9 +58,21 @@ const Page: React.FC<PageProps<"/courses/[readable_id]">> = async (props) => {
     notFound()
   }
 
+  const [course] = courses.results
+  const outlineCoursewareId = getOutlineCoursewareId(course)
+
+  if (outlineCoursewareId) {
+    await queryClient.prefetchQuery(
+      coursesQueries.courseOutline(outlineCoursewareId),
+    )
```
Comment (ChristopherChudzicki) [inline frontends/main/src/app/(products)/courses/[readable_id]/page.tsx:67]: This is an extra (non-parallel) api endpoint fetch, but it should be very fast b/c of the caching on the outline API, right?

frontends/main/src/app-pages/ProductPages/CourseOutlineSection.tsx
```diff
@@ -0,0 +1,137 @@
...
+): string => `${count} ${count === 1 ? singularLabel : pluralLabel}`
+
+const getMetaLine = (module: CourseOutlineModule): string => {
+  const counts = module.counts ?? {}
+  const videos = counts.videos ?? 0
+  const readings = counts.readings ?? 0
+  const assignments = counts.assignments ?? 0
+
+  const metaParts = [
+    videos > 0 ? formatCount(videos, "Video", "Videos") : null,
```
ChristopherChudzicki, line 92: Nit:

```suggestion
  const metaParts = [
    formatEffort(module.effort_time),
    videos > 0 ? formatCount(videos, "Video", "Videos") : null,
```
and drop the effort / unshift below. Wouldn't that be the same? formatEffort already returns string/null, so the filter would catch it, right?

frontends/main/src/app-pages/ProductPages/CourseOutlineSection.tsx
```diff
@@ -0,0 +1,137 @@
...
+  display: "inline-flex",
+  flexWrap: "wrap",
+  rowGap: "4px",
+}))
+
+const getModuleTitle = (module: CourseOutlineModule, index: number): string => {
+  if (module.title && module.title.trim().length > 0) {
+    return module.title
+  }
+  return `Module ${index + 1}`
```
ChristopherChudzicki, line 58: Would you expect this to actually trigger? These are chapter titles, right? Does OpenEdx let you have an empty chapter title?

frontends/main/src/app-pages/ProductPages/CourseOutlineSection.tsx
```diff
@@ -0,0 +1,137 @@
...
+    readings > 0 ? formatCount(readings, "Reading", "Readings") : null,
+    assignments > 0
+      ? formatCount(assignments, "Assignment", "Assignments")
+      : null,
+  ].filter((part): part is string => Boolean(part))
+  const effort = formatEffort(module.effort_time)
+  if (effort) {
+    metaParts.unshift(effort)
+  }
+  return metaParts.join(" . ")
```
ChristopherChudzicki: A literal period looks odd here, IMO. I think we should check with @mbilalmughal if that's what we actually want. I realize it is what Figma shows:

Figma: https://www.figma.com/proto/Eux3guSenAFVvNHGi1Y9Wm/MIT-Design-System?page-id=107%3A2479&node-id=37943-82138&viewport=-16660%2C459%2C0.83&t=XMpSKlpXVMX1Hv1u-1&scaling=min-zoom&content-scaling=fixed&starting-point-node-id=13287%3A116273
Image

I would have expected a center dot.

frontends/main/src/app-pages/ProductPages/CourseOutlineSection.tsx
```diff
@@ -0,0 +1,137 @@
...
+}
+
+const formatEffort = (seconds?: number | null): string | null => {
+  if (seconds === undefined || seconds === null || seconds <= 0) {
+    return null
+  }
+  if (seconds < 3600) {
+    return "Less than 1 hour to complete"
+  }
+  const hourLabel = formatHours(seconds)
```
ChristopherChudzicki, line 73: For https://learn.mit.edu/courses/course-v1:UAI_SOURCE+UAI.ENT.1, this currently shows `2.7 hours` which is weirdly specific, IMO. 

Maybe we should round to nearest hour? @mbilalmughal 

Image

ChristopherChudzicki (approved):
I left some comments/questions but overall this looks good 👍 

Approving, but if you have time, I think it's worth getting @mbilalmughal feedback on these two points:
- https://github.com/mitodl/mit-learn/pull/3242/changes#r3198190633
- https://github.com/mitodl/mit-learn/pull/3242/changes#r3198167793

If you decide to make any of the suggested changes and want a re-review, happy to look again, and I won't take forever this time...

frontends/main/src/app-pages/ProductPages/CourseOutlineSection.tsx
mbilalmughal: @ChristopherChudzicki @zamanafzal 

Yes, I agree. These time estimates shouldn’t be overly precise; they’re meant to give a rough idea of how many hours a task may take, not an exact figure like 2.45 minutes.

frontends/main/src/app-pages/ProductPages/CourseOutlineSection.tsx
mbilalmughal: @zamanafzal @ChristopherChudzicki These are dot separators, I got them from the following UI. If they’re not actually dots but bullets, then that’s my mistake. I just wanted them to be smaller while still prominent enough to clearly act as dividers.
https://mit-learn-prototypes.vercel.app/course-content/

frontends/main/src/app-pages/ProductPages/CourseOutlineSection.tsx
zamanafzal: The outline plugin serializes `title` as `block.get("display_name") or ""`, so if `display_name` is `None` (its default in edx-platform), the API returns an empty string. The `Module N` fallback prevents blank headings in those cases.

frontends/main/src/app/(products)/courses/[readable_id]/page.tsx
Comment (zamanafzal) [reply frontends/main/src/app/(products)/courses/[readable_id]/page.tsx:67]: It should stay cheap because the outline endpoint is cached, and this avoids a client-side loading gap for course content.

pdpinch:
Is this behind a feature flag? It should be.

pdpinch:
I'm concerned about the UAI modules and how their outlines are written. We might need some time to address the coursware side of things. 

For example, UA.SE.1 looks like this: 

Screenshot 2026-05-07 at 1 59 09 PM

The interesting titles, that refect the content, are the next level down in this course structure.

zamanafzal:
@pdpinch It wasn’t behind a feature flag previously. I’ve now added a feature flag, and `CourseOutlineSection` is displayed only when the flag is enabled.

Are we good to merge this PR now, or should we wait for the UAI modules/courseware-related updates first?

zamanafzal:
@ChristopherChudzicki, I've addressed the feedback and made the design-related changes. Can you please take a second look?

frontends/main/src/app-pages/ProductPages/CourseOutlineSection.tsx
zamanafzal: done

frontends/main/src/app/(products)/courses/[readable_id]/page.tsx
Comment (zamanafzal) [reply frontends/main/src/app/(products)/courses/[readable_id]/page.tsx]: done

frontends/api/src/mitxonline/hooks/courses/queries.ts
zamanafzal: done

ChristopherChudzicki (approved):
👍 Looks good to me!

zamanafzal:
I'm merging this PR since this is behind a feature flag, and that is off by default.

PR #3246, opened 2026-04-24
Add authentication to load tests

rhysyngsun:

Part of https://github.com/mitodl/hq/issues/11039

This adds code that logs the tests in via Keycloak so they can poke around as an authenticated user.

### How can this be tested?

- Create a file for test user auth info in `load_testing/data/` (e.g. `load_testing/data/users.json`) per the readme instructions.
- Run the load tests with user auth:
  ```shell
  k6 run learn.smoke.ts \
    -e BACKEND_BASE_URL=https://api.learn.odl.local \
    -e FRONTEND_BASE_URL=https://learn.odl.local \
    -e USERS_JSON_FILE=data/users.json \
    -e SSO_BASE_URL=https://keycloak.odl.local \
    -e IGNORE_HTTPS_ERRORS=true \
    --console-output test.log
  ```
- Inspect the output in `test.log`, you should see log statements regarding successful logins

### DIFF
--- a/load_testing/auth.ts
+++ b/load_testing/auth.ts
@@ -14,7 +14,7 @@ export function hasAccessToken(): boolean {
 }
 
 function _validate_credentials(credentials) {
-  if (typeof credentials !== "array") {
+  if (!Array.isArray(credentials)) {
     throw Error("Expected an array of credentials")
   }
 
@@ -37,6 +37,8 @@ export const credentials: AuthCredential[] = new SharedArray(
     }
 
     const parsed = JSON.parse(open(__ENV.USERS_JSON_FILE))
+
+    console.log(parsed)
 
     _validate_credentials(parsed)
 
--- a/load_testing/auth.ts
+++ b/load_testing/auth.ts
@@ -38,8 +38,6 @@ export const credentials: AuthCredential[] = new SharedArray(
 
     const parsed = JSON.parse(open(__ENV.USERS_JSON_FILE))
 
-    console.log(parsed)
-
     _validate_credentials(parsed)
 
     return parsed
--- a/load_testing/auth.ts
+++ b/load_testing/auth.ts
@@ -1,12 +1,16 @@
 import { SharedArray } from "k6/data"
 
 export type AuthCredential = {
-  username: string
+  email: string
   password: string
 }
 
 export function getAccessToken(): string | null {
   return __ENV.AUTH_ACCESS_TOKEN
+}
+
+export function hasAccessToken(): boolean {
+  return !getAccessToken()
 }
 
 function _validate_credentials(credentials) {
--- a/load_testing/auth.ts
+++ b/load_testing/auth.ts
@@ -1,13 +1,41 @@
 import { SharedArray } from "k6/data"
 
-export function getAccessToken(): String {
+export type AuthCredential = {
+  username: string
+  password: string
+}
+
+export function getAccessToken(): string | null {
   return __ENV.AUTH_ACCESS_TOKEN
 }
 
-export const users = new SharedArray("users", function () {
-  if (!__ENV.USERS_JSON_FILE) {
-    return []
+function _validate_credentials(credentials) {
+  if (typeof credentials !== "array") {
+    throw Error("Expected an array of credentials")
   }
 
-  return JSON.parse(open(__ENV.USERS_JSON_FILE))
-})
+  for (let index = 0; index < credentials.length; index++) {
+    const credential = credentials[index]
+    if (!Object.hasOwn(credential, "email")) {
+      throw Error(`User entry is missing 'email' at index ${index}`)
+    }
+    if (!Object.hasOwn(credential, "password")) {
+      throw Error(`User entry is missing 'password' at index ${index}`)
+    }
+  }
+}
+
+export const credentials: AuthCredential[] = new SharedArray(
+  "credentials",
+  function () {
+    if (!__ENV.USERS_JSON_FILE) {
+      return []
+    }
+
+    const parsed = JSON.parse(open(__ENV.USERS_JSON_FILE))
+
+    _validate_credentials(parsed)
+
+    return parsed
+  },
+)
--- a/load_testing/auth.ts
+++ b/load_testing/auth.ts
@@ -10,7 +10,7 @@ export function getAccessToken(): string | null {
 }
 
 export function hasAccessToken(): boolean {
-  return !getAccessToken()
+  return !!getAccessToken()
 }
 
 function _validate_credentials(credentials) {
--- a/load_testing/backend/test.ts
+++ b/load_testing/backend/test.ts
@@ -16,23 +16,32 @@ export function testBackend() {
       exec.vu.metrics.tags.apiVersion = "v0"
 
       const client = createV0Client()
-      let res = client.videoShortsList({ limit: 50 })
+      group("video shorts", () => {
+        const res = client.videoShortsList({ limit: 50 })
 
-      check(res, {
-        "is status 200": (r) => r.response.status === 200,
-        "has results": (r) => r.response.json("results").length > 0,
+        group
+        check(res, {
+          "is status 200": (r) => r.response.status === 200,
+          "has results": (r) => r.response.json("results").length > 0,
+        })
       })
 
-      res = client.usersMeRetrieve()
-      check(res, {
-        "is status 200": (r) => r.response.status === 200,
+      group("users/me", () => {
+        const res = client.usersMeRetrieve()
+        check(res, {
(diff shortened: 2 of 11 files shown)


load_testing/auth.ts
```diff
@@ -1,13 +1,45 @@
...
+
+export function getAccessToken(): string | null {
   return __ENV.AUTH_ACCESS_TOKEN
 }
 
-export const users = new SharedArray("users", function () {
-  if (!__ENV.USERS_JSON_FILE) {
-    return []
+export function hasAccessToken(): boolean {
+  return !getAccessToken()
```
mbertrand: I got this:

```markdown
    ✗ expected auth state
      ↳  0% — ✓ 0 / ✗ 46
```

until I changed this function to:

```typescript
export function hasAccessToken(): boolean {
  return !!getAccessToken()
}
```

load_testing/backend/test.ts
```diff
@@ -8,30 +8,40 @@ import {
...
 
       const client = createV0Client()
-      let res = client.videoShortsList({ limit: 50 })
+      group("video shorts", () => {
+        const res = client.videoShortsList({ limit: 50 })
 
-      check(res, {
-        "is status 200": (r) => r.response.status === 200,
-        "has results": (r) => r.response.json("results").length > 0,
+        group
```
mbertrand: Stray `group` - delete?

load_testing/frontend/test.ts
```diff
@@ -18,35 +32,122 @@ async function home(page: Page) {
     .click()
 }
```
mbertrand, line 43: Race condition + multi-element matcher. The `data-testid` mounts immediately as an empty wrapper; article children only appear after React Query resolves, so `count()` reads `0` and `.nth(0).click()` then hangs ~30s on a missing element. Also, the home page renders several `ResourceCarousel` instances — `getByTestId` matches them all; scope to first.

```suggestion
async function home(page: Page, context: Context) {
  await page.goto(FRONTEND_BASE_URL)

  const carousel = page.getByTestId("resource-carousel").first()
  const articles = carousel.locator("article")

  try {
    await articles.first().waitFor({ timeout: 10_000 })
  } catch {
    // carousel never populated; let the assertion below report it
  }

  const articlesCount = await articles.count()
  await check(carousel, {
    "home page carousel has 12 items": () => articlesCount === 12,
  })

  if (articlesCount === 0) {
    return
  }

  await articles.nth(randomIntBetween(0, articlesCount - 1)).click()
}
```

load_testing/frontend/test.ts
```diff
@@ -18,35 +32,122 @@ async function home(page: Page) {
...
+
+  const loginButton = page.getByTestId("login-button-desktop")
+  await loginButton.first().click()
+
+  await loginKeycloak(page, credential, context)
+
+  await page.waitForURL(/.*\/dashboard.*/)
+
+  console.log("Login > reached dashboard")
+}
```
mbertrand, line 102: Ran into a few problems here:

1. **Missing `return`** after the no-credentials log — falls through to `credential.email` on line 65 and crashes. Out-of-the-box smoke runs (no `USERS_JSON_FILE`) fail every iteration.
2. **`waitForURL(/.*\/dashboard.*/)` is unreliable** — local APISIX drops the `next=` param through the OIDC handshake, so successful logins land on `/`, not `/dashboard`. The test marks all logins as failed.
3. **Login throws abort the iteration silently** — they bypass `check()`, so failures don't count toward the `checks` threshold. Test reports green even when 100% of logins fail.

```suggestion
async function login(page: Page, context: Context) {
  const credential: AuthCredential | undefined = randomItem(credentials)

  if (credential == null) {
    console.log("Login > skipping because no credentials provided")
    return
  }

  if (!SSO_BASE_URL) {
    throw Error("SSO_BASE_URL must be set to run the login flow")
  }

  if (page.url() === "about:blank") {
    await page.goto(FRONTEND_BASE_URL)
  }

  console.log(`Login > using '${credential.email}'`)

  const loginButton = page.getByTestId("login-button-desktop")
  await loginButton.first().click()

  let loggedIn = false
  try {
    await loginKeycloak(page, credential, context)
    await page.locator('[aria-label="User Menu"]').waitFor({ timeout: 10_000 })
    loggedIn = true
    console.log("Login > authenticated; user menu visible")
  } catch (err) {
    console.log(
      `Login > failed for '${credential.email}' at URL '${page.url()}': ${err}`,
    )
  }

  check(null, {
    "login succeeded": () => loggedIn,
  })
}
```

load_testing/frontend/test.ts
```diff
@@ -18,35 +32,122 @@ async function home(page: Page) {
...
+
+const ESCAPED_SSO_URL = escapeRegex(SSO_BASE_URL)
+
+const KEYCLOAK_USERNAME_URL_RE = new RegExp(
+  `${ESCAPED_SSO_URL}\/realms\/[a-z-]+\/protocol\/openid-connect\/auth.*`,
+)
+const KEYCLOAK_PASSWORD_URL_RE = new RegExp(
+  `${ESCAPED_SSO_URL}\/realms\/[a-z-]+\/login-actions\/authenticate.*`,
+)
```
mbertrand, line 111: `escapeRegex(SSO_BASE_URL)` runs at module load. When `SSO_BASE_URL` is unset (e.g., backend-only smoke runs), this throws `TypeError` and the entire frontend test fails to load. Compute lazily, and reference `patterns.username` / `patterns.password` from inside `loginKeycloak()`.

```suggestion
function getKeycloakUrlPatterns() {
  const escaped = escapeRegex(SSO_BASE_URL)
  return {
    username: new RegExp(
      `${escaped}/realms/[a-z-]+/protocol/openid-connect/auth.*`,
    ),
    password: new RegExp(
      `${escaped}/realms/[a-z-]+/login-actions/authenticate.*`,
    ),
  }
}
```

load_testing/frontend/test.ts
```diff
@@ -18,35 +32,122 @@ async function home(page: Page) {
...
+  const credentialnameInput = await page.locator("input[name=username]")
+  console.log("Login > Keycloak > found username input")
+  await credentialnameInput.focus()
+  await credentialnameInput.fill(credential.email)
+  console.log("Login > Keycloak > entered email address")
+
+  await page.locator("button[type=submit]").click()
+  console.log("Login > Keycloak > submitting email address")
+
+  await page.waitForNavigation(KEYCLOAK_PASSWORD_URL_RE)
```
mbertrand: `page.waitForNavigation` accepts `{waitUntil, timeout}` — not a URL pattern. The regex is silently ignored; this just waits for any navigation event. Use `waitForURL` for URL matching.

```suggestion
  await page.waitForURL(KEYCLOAK_PASSWORD_URL_RE)
```

(Also drop the redundant `await page.waitForNavigation()` on line 117 — the post-login wait in `login()` covers final arrival.)

load_testing/learn.smoke.ts
```diff
@@ -17,12 +23,13 @@ export const options = {
...
-      vus: 10,
-      duration: "30s",
+      vus: Math.floor(MAX_VUS * BACKEND_VU_SHARE),
+      duration: "1m",
     },
   },
   thresholds: {
     // the rate of successful checks should be higher than 90%
     checks: ["rate>0.9"],
   },
```
mbertrand, line 35: Only `checks` is bounded. HTTP failures and iteration crashes that bypass `check()` aren't reflected — the test can report green while requests are failing. Add failure-rate budgets. (Apply the same to `learn.average-load.ts` and `learn.stress.ts`.)

```suggestion
  thresholds: {
    // the rate of successful checks should be higher than 90%
    checks: ["rate>0.9"],
    http_req_failed: ["rate<0.01"],
    browser_http_req_failed: ["rate<0.05"],
  },
```

mbertrand (changes requested):
Got it working after some tweaks described here (running in WSL in case that matters).  I added two users to the JSON file, one real and one fake.  

The command I ran was slightly different since I don't have all the same aliases set up:
```bash
k6 run load_testing/learn.smoke.ts   \
-e BACKEND_BASE_URL=http://open.odl.local:8065   \
-e FRONTEND_BASE_URL=http://open.odl.local:8062   \
-e SSO_BASE_URL=http://kc.ol.local:8066   \
-e USERS_JSON_FILE=data/users.json \
-e IGNORE_HTTPS_ERRORS=true  \
--console-output test.log
```

load_testing/frontend/test.ts
rhysyngsun: I think to try to avoid potentially compiling the regexes multiple times I'm going to give a default for SSO_BASE_URL

mbertrand (approved):
👍

PR #3272, opened 2026-04-30
fix: remove h5 font weight override

daniellefrappier18:

Fixes https://github.com/mitodl/hq/issues/10977

Removes a local `fontWeight` override that was added to match Figma's h5 design. Now that the h5 typography definition in smoot-design has been updated to use `fontWeight: 700` (Neue Haas Grotesk 65 Bold), the override is redundant.

- [ ] Desktop screenshots
- [x] Mobile width screenshots

AFTER (override removed — inheriting h5 from current smoot-design): The title will render at 700 (bold) 
automatically — matching the "before" screenshot — without the local override.

Screenshot 2026-05-01 at 11 10 47 AM

### How can this be tested?

Using the browser's dev tools inspector, verify that the h5 element reflects a font-weight of 700.


ChristopherChudzicki (approved):
👍

PR #3273, opened 2026-04-30
qdrant: Automatically compute optimizer settings

shanbady:

Closes https://github.com/mitodl/hq/issues/11066

This PR automatically configures some optimizer settings on qdrant (namely the indexing_threshold and flush_interval_sec). These are settings that control when/how often the optimizer tries to "tidy up" and affects the number of segments created. Too many segments for small amounts of data is bad for performance (as well as the inverse)

### How can this be tested?
There is not too much testing that can be done locally without having large amounts of data in your local qdrant collection. I good local check might be just to verify things are embedded and that creating your qdrant collections from scratch cause no issues via ```python manage.py generate_embeddings --courses --recreate-collections --skip-contentfiles```

### Additional Context
This should not have any immediate noticeable impact when deployed - once the optimized settings are in place, it should lead to less fragmentation and under/over optimized segments over time. The impetus for this was that our collections had more segments than necessary leading to some performance hits (particularly noticeably on contentfiles)


vector_search/constants.py
```diff
@@ -142,3 +142,24 @@
...
+QDRANT_OPTIMIZER_SEGMENT_SMALL = 20_000
+QDRANT_OPTIMIZER_SEGMENT_MEDIUM = 60_000
+QDRANT_OPTIMIZER_SEGMENT_LARGE = 120_000
+QDRANT_OPTIMIZER_SEGMENT_XLARGE = 250_000
+
+# Flush intervals
+QDRANT_OPTIMIZER_FLUSH_INTERVAL_SMALL = 15
+QDRANT_OPTIMIZER_FLUSH_INTERVAL_MEDIUM = 20
+QDRANT_OPTIMIZER_FLUSH_INTERVAL_LARGE = 25
+QDRANT_OPTIMIZER_FLUSH_INTERVAL_XLARGE = 30
```
mbertrand: 12 module-level constants encoding what is really a 4-row lookup table — the threshold, segment size, and flush interval for each bucket have to be edited in three separate places and kept in sync, and the `if/elif` ladder in `compute_optimizer_settings` has to mirror that ordering. A single ordered list keeps each bucket's three values colocated and reduces the import block in `utils.py` from 12 names to 1:

```python
# vector_search/constants.py
# (max_points_per_shard, target_segment_size, flush_interval_sec)
# The last bucket's threshold is None — it's the tail.
QDRANT_OPTIMIZER_BUCKETS = [
    (50_000,    20_000,  15),
    (500_000,   60_000,  20),
    (2_000_000, 120_000, 25),
    (None,      250_000, 30),
]
QDRANT_OPTIMIZER_INDEXING_THRESHOLD_RATIO = 0.4
```

```python
# vector_search/utils.py
def compute_optimizer_settings(point_count: int, shard_number: int):
    points_per_shard = max(point_count // shard_number, 1)
    for threshold, target_segment, flush_interval in QDRANT_OPTIMIZER_BUCKETS:
        if threshold is None or points_per_shard < threshold:
            return {
                "indexing_threshold": int(
                    target_segment * QDRANT_OPTIMIZER_INDEXING_THRESHOLD_RATIO
                ),
                "flush_interval_sec": flush_interval,
            }
```

The existing parametrized test continues to work unchanged — it references the constants by value, not by name. Adding a bucket becomes a one-line edit instead of three constants + a new `elif` branch.

vector_search/utils.py
```diff
@@ -127,6 +139,44 @@ def points_generator(
...
+    info = client.get_collection(collection_name)
+    point_count = info.points_count
+    shard_number = info.config.params.shard_number
+
+    settings = compute_optimizer_settings(point_count, shard_number)
+
+    client.update_collection(
+        collection_name=collection_name,
+        optimizer_config=models.OptimizersConfigDiff(**settings),
+    )
```
mbertrand, line 181: `tune_collection` issues `update_collection` every time, with no comparison against the existing config. Since `embed_learning_resources` and `embed_topics` call `create_qdrant_collections(force_recreate=False)`, this could happen pretty frequently?

A small diff against `info.config.optimizer_config` could skip the write when settings already match:

```python
def tune_collection(client, collection_name):
    info = client.get_collection(collection_name)
    point_count = info.points_count or 0
    shard_number = info.config.params.shard_number
    desired = compute_optimizer_settings(point_count, shard_number)
    current = info.config.optimizer_config
    if (
        current.indexing_threshold == desired["indexing_threshold"]
        and current.flush_interval_sec == desired["flush_interval_sec"]
    ):
        return
    client.update_collection(
        collection_name=collection_name,
        optimizer_config=models.OptimizersConfigDiff(**desired),
    )
```

vector_search/constants.py
mbertrand, line 162: An alternate approach could be a single constant list of buckets like below, requires fewer constants but may be harder to decipher, so feel free to take or leave.

```python
# vector_search/constants.py
# (max_points_per_shard, target_segment_size, flush_interval_sec)
# The last bucket's threshold is None — it's the tail.
QDRANT_OPTIMIZER_BUCKETS = [
    (50_000,    20_000,  15),
    (500_000,   60_000,  20),
    (2_000_000, 120_000, 25),
    (None,      250_000, 30),
]
QDRANT_OPTIMIZER_INDEXING_THRESHOLD_RATIO = 0.4
```

```python
# vector_search/utils.py
def compute_optimizer_settings(point_count: int, shard_number: int):
    points_per_shard = max(point_count // shard_number, 1)
    for threshold, target_segment, flush_interval in QDRANT_OPTIMIZER_BUCKETS:
        if threshold is None or points_per_shard < threshold:
            return {
                "indexing_threshold": int(
                    target_segment * QDRANT_OPTIMIZER_INDEXING_THRESHOLD_RATIO
                ),
                "flush_interval_sec": flush_interval,
            }
```

mbertrand (approved):
LGTM, added a couple of non-critical suggestions

vector_search/utils.py
shanbady: good call out. fixed 👍

vector_search/constants.py
shanbady: I think worth doing if the list grows from here. will leave for later

PR #3278, opened 2026-05-01
Add to profile button linkedin share button for certificates and verifiable credentials

dsubak:

https://github.com/mitodl/hq/issues/10900

At a high level, this changes the links generated used in the SharePopover when used on Program and Course Certificate pages.
On certificate pages
- If the user does not have a verifiable credential or is not logged in, the LinkedIn url is changed to use the [Add to Profile](https://addtoprofile.linkedin.com/) query params derived from the verifiable credential. When clicked, this will redirect the user to LinkedIn and if they're signed in it'll prefill a form with the certificate name, ID and pageUrl. Other behaviors are unchanged.
- If the user does have a verifiable credential, the LinkedIn url is changed to use the [Add to Profile](https://addtoprofile.linkedin.com/) query params derived from the certificate. When clicked, this will redirect the user to LinkedIn and if they're signed in it'll prefill a form with the verifiable credential name and ID, issuance date and URL. Additionally, all other share links are replaced with a link to Verifierplus.org which allows for one-click validation of the verifiable credential.

When SharePopover is used on /news endpoints (which is the only other use of this component), existing behavior is unchanged.

When clicked, if logged in to linkedin, prepopulates a form so you can add a verifiable credential to your profile with one click.
Screenshot 2026-04-10 at 3 33 42 PM

### How can this be tested?
All automated tests should pass, including tests of link generation for the share popover.

Manual Testing:
 
Prereq: You'll need learn and mitxonline set up - the latter will need certificates and verifiable credentials provisioned to observe the full scope of new behavior.

Case 1 - News
- If you visit any /news pages, links should be unchanged vs main.

Case 2 - Certificate Page without any Verifiable Credentials provisioned. 
- Visit a certificate page (`http://learn.odl.local:9080/certificate/course/<uuid>`)
- Click the share button.
- Click LinkedIn, it should redirect you to linkedin with a partially prefilled form to add your certificate to your profile.
- Twitter, Facebook and the bare copy link button should all be unchanged vs main.

Case 3 - Certificate Page which does have a Verifiable Credential provisioned.
- Visit a certificate page (`http://learn.odl.local:9080/certificate/course/<uuid>`)
- Click the share button.
- Click LinkedIn, it should redirect you to linkedin with a partially prefilled form to add your certificate to your profile.
- Twitter, Facebook and the copy links should now be to a `verifierplus.org` link with a `vc` query param pointing to the mitxonline download API.
  - It's worth noting that while the experience of sharing that verifierplus link should work, actually visiting that verifierplus link _probably_ won't work. It's an external service which will go out to the link in the query param to verify the verifiable credential, so unless your local is externally accessible, you'll just get the normal experience of visiting that site where you can paste or upload a VC. This should be more fully verifiable on RC, but it's probably too much hassle to get right on locals for me to recommend more than manual inspection of the link.


frontends/main/src/common/certificateUtils.ts
```diff
@@ -6,3 +13,138 @@ export const getCertificateInfo = (): { displayType: string } => {
...
+
+export enum CertificateType {
+  Course = "course",
+  Program = "program",
+}
+
+/* Unsure if I should use NEXT_PUBLIC_MITX_ONLINE_LEGACY_BASE_URL or NEXT_PUBLIC_MITX_ONLINE_BASE_URL
+ * The below matches my naive expectations but we need to confirm that since we need unauthenticated access
+ */
+const API_BASE_URL = process.env.NEXT_PUBLIC_MITX_ONLINE_LEGACY_BASE_URL
```
daniellefrappier18, line 27: This looks kind of fragile. If the env var isn't set in any environment (staging, prod, CI preview), every certificate page silently generates broken links with no error thrown. `tiny-invariant` is already used in the codebase. A possible fix could be:
```
import invariant from "tiny-invariant"

const API_BASE_URL = process.env.NEXT_PUBLIC_MITX_ONLINE_LEGACY_BASE_URL
invariant(API_BASE_URL, "NEXT_PUBLIC_MITX_ONLINE_LEGACY_BASE_URL must be set")
```

frontends/main/src/common/certificateUtils.ts
```diff
@@ -6,3 +13,138 @@ export const getCertificateInfo = (): { displayType: string } => {
...
+}
+
+export const getVerifiableCredentialDownloadAPIURL = (
+  verifiableCredentialJson: VerifiableCredential,
+): string => {
+  const type =
+    verifiableCredentialJson["credentialSubject"]["achievement"][
+      "achievementType"
+    ].toLowerCase()
+  const certId = verifiableCredentialJson["id"].substring(9)
```
daniellefrappier18: Curious why 9 here? Both getVerifiableCredentialDownloadAPIURL and getVerifiableCredentialLinkedInURL  use .substring(9) with no explanation. Maybe add a comment explain why?

dsubak: Can do - fwiw, it's just because the [verifiable cred format](https://github.com/digitalcredentials/mit-learn-obv3-template/blob/main/certificates/courseCertificate.json#L7C9-L7C57) expects UUIDs to be prefixed with `urn:uuid:`. I'll add a comment clarifying as such.

We could also leave that in if we'd prefer, I just thought it was a bit cleaner. It's kind of meaningless out of the context of our systems, so very much up for debate!

daniellefrappier18 (commented):
I was able to verify all your test cases. I left a couple of comments. The API_BASE_URL  has no guard. If it's missing, the download URL, verifierplus.org link, and LinkedIn certUrl all become "undefined/api/v2/..." strings with no error.

dsubak:
@daniellefrappier18 Thanks for the feedback! Should all be addressed and ready for another look when you've got a moment!

daniellefrappier18 (approved):
Sorry I missed your comment that this was ready. LTGM

PR #3285, opened 2026-05-04
Min score for vector learning resources endpoint

shanbady:

Closes https://github.com/mitodl/hq/issues/11079

This PR adds a score_cutoff parameter to the vector learning resources endpoint that controls the minimum threshold score for results that are returned when there is a query string present.

There are some quirks in how this had to be implemented to get the facet counts to behave as expected (documented [here](https://github.com/mitodl/hq/issues/11079#issuecomment-4361549366))

#### How this works 
_when the "hybrid search" admin option is enabled_
1. If there is no query, we enable pagination on the api, facet counts are fetched directly from qdrant (existing default behavior)
2. if there **is a query** - the min-score is in effect. Instead of paginating we return all results above the min score for the query (+ any filters)
    - There is no pagination (we show all the results on a single page)
    - clicking on any of the facets applies the filter on the already retrieved results on the frontend

### How can this be tested?
1. checkout this branch
4. make sure you have courses/learning resources embedded. if not run ```python manage.py generate_embeddings --courses --skip-contentfiles```
5. login as an admin/superuser
6. visit the search page - enable the "hybrid search" admin option
7. try some queries. When using a query string you should get a small handful of results (results that score above the threshold). Try filtering by different facets and see that the counts line up.


vector_search/views.py
```diff
@@ -345,11 +457,74 @@ async def async_vector_search(  # noqa: PLR0913
...
+                order_by=order_by,
+                limit=total_count,
+                offset=0,
+                search_collection=search_collection,
+                score_cutoff=score_cutoff,
+                hybrid_search=hybrid_search,
+            )
+            counts = await self._async_vector_resource_counts(
+                hits, params, search_collection=search_collection
+            )
```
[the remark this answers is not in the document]
shanbady: fixed

vector_search/views.py
```diff
@@ -320,18 +334,116 @@ async def async_vector_search(  # noqa: PLR0913
...
+                            counter[key] += 1
+
+            aggregations[agg_key] = [
+                {"key": k, "doc_count": v} for k, v in counter.most_common()
+            ]
+
+        return {
+            "total": {"value": total_count},
+            "aggregations": aggregations,
+        }
```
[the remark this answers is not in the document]
shanbady: Added a test

frontends/main/src/page-components/SearchDisplay/SearchDisplay.tsx
```diff
@@ -566,6 +566,135 @@ const toVectorSearchParams = (
...
+  VECTOR_CLIENT_FILTER_FACETS.every((facet) => {
+    if (facet === excludedFacet) {
+      return true
+    }
+    const selectedValues = normalizeParamValues(params[facet])
+    if (selectedValues.length === 0) {
+      return true
+    }
+    const resourceValues = getResourceFacetValues(resource, facet)
+    return selectedValues.some((value) => resourceValues.includes(value))
```
abeglova, line 659: Why are you filtering in the UI and not the backend?  I think manually filtering for facets is a bad idea in general but at least it should be done in the backend so the results are cachable. Also filtering the results in python wouldn't be fast but at least it will be faster than filtering in the ui code

abeglova: Also, with the 0.1 cutoff i never had a query return more than 15 or so results even for very broad searches like "science" locally.  The facets worked as expected but many relevant results were not returned. When i lowered the cutoff so that i would get more results than would fit on a page for some queries, qdrant would still limit the number of results returned and some facets would say there were more results than were shown because some of the results were not displayed because they did not fit in the first page

frontends/main/src/page-components/SearchDisplay/SearchDisplay.tsx
```diff
@@ -1102,28 +1269,30 @@ const SearchDisplay: React.FC<SearchDisplayProps> = ({
...
-                    <PaginationItem
-                      slots={{
-                        previous: RiArrowLeftLine,
-                        next: RiArrowRightLine,
-                      }}
-                      {...item}
-                    />
-                  )}
-                />
+                {!isVectorQuerySearch && (
```
abeglova, line 1272: What is the plan for this long term? What should happen to searches that have more results than can fit on a page?

abeglova (changes requested):
I'm not sure how to review this. I don't think this is a good approach. Both the manual filtering in the ui and the manual facet aggregation seem not great and it doesn't seem like there is an easy fix that keeps the current approach.  The current score threshold is high enough that lots of relevant resources are just not returned if you make a query that should have lots of results like "AI". If you lower the the threshold and some queries return more than the number of results in a page the filtering does not work correctly. I think realistically we need to either do something else or rethink how facets work

frontends/main/src/page-components/SearchDisplay/SearchDisplay.tsx
abeglova: I tried again with score_cutoff 0.00001 and got the expected results. I previously tried setting score_cutoff=0 which caused to backend to still page but the UI to filter

shanbady: I went this route to get the interface and facet counts to line up (regardless of qdrant/opensearch endpoint) in addition to working around some fundamental quirks in how Qdrant applies filtering (as well as a limitation with their facet api (I left some notes on this [here](https://github.com/mitodl/hq/issues/11079#issuecomment-4361549366)).


###### Opensearch:
Search “Biology” -> I get a stable set of matching docs -> compute facets over that same set → clicking a facet ("Earth Science" etc) narrows that same set
###### Qdrant:
Search “Biology”-> I get the best candidates -> fuse/rank them ->  cut by score/limit ->  **apply filters over the surviving candidates**



- Qdrant's facet api does not support hybrid/complex searches involving a pre-filter or passing in a score.(presumably it would be just as expensive as performing the entire search and manually aggregating counts)
    - This was a non-issue without the score cutoff since all searches would yield all results  
- Qdrant performs neither a pre-filter or post-filter (do the search first - then apply the filters) like  opensearch.
    - It traverses a [filterable HNSW](https://yudhiesh.github.io/2025/05/09/the-achilles-heel-of-vector-search-filters/#filtering-in-hnsw-graph-based-indexes) as it performs the search. To further confound this Qdrant also dynamically selects how it decides to determine the filter counts depending on the cardinality of the field(s) being filtered

frontends/main/src/page-components/SearchDisplay/SearchDisplay.tsx
shanbady: As a first step to make it easier to iterate on vector specific logic/interface my plan was to properly subclass and separate out the vector specific logic on the frontend into its own component as part of [this ticket](https://github.com/mitodl/hq/issues/11167) 

Long term we could consider:
- what it currently does - limit the number of results by using a threshold score when dealing with the "brute force branch" (query is present + filters are applied + there is a cutoff_score > 0). its a non-issue if we are just dealing with the learning resources collection and the cutoff is appropriate (we are not dealing with thousands of results after applying a specific query+filter+cutoff_score). If we present the user with more than 10 results given the size of our learning resource catalog after query+filter+cutoff_score filters there is likely something else wrong IMO.
- changing the design of the search page/facet counts - consider hiding counts once a query is present  or even altogether

Long term I agree with what you said "I think realistically we need to either do something else or rethink how facets work" - this is a first dip (admin-only see how it works or doesnt)

shanbady:
@abeglova This is ready for another look. I resolved the min-score = 0 issue you spotted in addition to adding an absolute hard limit for the case where we need to fetch all results without pagination

vector_search/serializers.py
```diff
@@ -224,6 +225,12 @@ class LearningResourcesVectorSearchRequestSerializer(
         default=False,
         help_text="Whether to use a hybrid search",
     )
+    score_cutoff = serializers.FloatField(
+        required=False,
+        default=settings.VECTOR_SEARCH_MIN_SCORE,
+        min_value=settings.VECTOR_SEARCH_MIN_SCORE,
```
abeglova: Can you remove this from the serialized now that this is an environment variable? Or i think the better thing to do is to add it to the search params similar to the other admin variables

abeglova: Also why is min_value the same as default?

openapi/specs/v0.yaml
```diff
@@ -1423,6 +1423,14 @@ paths:
               * `learning_material` - Learning Material
         description: "The category of learning resource             \n\n* `course`\
           \ - Course\n* `program` - Program\n* `learning_material` - Learning Material"
+      - in: query
+        name: score_cutoff
+        schema:
+          type: number
+          format: double
+          minimum: 0.1
+          default: 0.1
```
abeglova, line 1432: You will be incorrect if VECTOR_SEARCH_MIN_SCORE changes.

abeglova (changes requested):

vector_search/serializers.py
shanbady: i was planning on factoring it out to be configurable in the next feature that lives on top of this. what do you think about for now setting min-value=0 and default=0.1 ?

openapi/specs/v0.yaml
shanbady: thats a good point. suggested hard coded values for now [here](https://github.com/mitodl/mit-learn/pull/3285#discussion_r3221963588)

vector_search/serializers.py
shanbady: in terms of removing this - I think we may* need to adjust this (have the frontend pass this in) depending on the interface (channel pages may need a slightely different score to be effective) just because of how this applies in qdrant. I'm going to move this to constants instead of a variable setting

abeglova (approved):
lgtm

PR #3269, opened 2026-04-30
dashboard translations UI

gumaerc:
### What are the relevant tickets?
Closes https://github.com/mitodl/hq/issues/11088

### Description (What does it do?)
This PR adds a language picker dropdown to both the B2B contract dashboard as well as the B2C program dashboard. A list of supported languages was added in `languageOptions.ts` that can be expanded upon in the future. It uses the new `language_options` data sent across with courses to display a dropdown on the B2B contract dashboard as well as the B2C program dashboard. Switching the language changes the underlying course run that is used for enrollment, pointing the user at language specific content.

### Screenshots (if appropriate):
image
image
image
image
image
image
image
image

### How can this be tested?
Ideally to test this end to end, you should have a course with multiple language runs set up in an instance of OpenEdx connected to MITx Online. Assuming you are using tutor, here is an example script for doing that:

```
cat > /tmp/setup_intltest_data.sh <<'BASH'
#!/usr/bin/env bash
set -euo pipefail

echo "[1/3] Creating courses in CMS (if missing)..."

cat > /tmp/create_intl_courses.py <<'PY'
from django.contrib.auth import get_user_model
from xmodule.modulestore.django import modulestore
from opaque_keys.edx.keys import CourseKey

User = get_user_model()
author = User.objects.filter(is_superuser=True).order_by("id").first()
if not author:
    raise RuntimeError("No superuser found")

ORG = "INTLTEST"
NUMBER = "TestCourse.001"

courses = [
    ("1T2026", "Test Course 001 (English)", "en"),
    ("1T2026_AR", "Test Course 001 (Arabic)", "ar"),
    ("1T2026_DE_DE", "Test Course 001 (German)", "de-de"),
    ("1T2026_EL", "Test Course 001 (Greek)", "el"),
    ("1T2026_ES", "Test Course 001 (Spanish)", "es"),
    ("1T2026_ES_419", "Test Course 001 (Latin American Spanish)", "es-419"),
    ("1T2026_FR", "Test Course 001 (French)", "fr"),
    ("1T2026_JA", "Test Course 001 (Japanese)", "ja"),
    ("1T2026_PT_BR", "Test Course 001 (Brazilian Portuguese)", "pt-br"),
    ("1T2026_ZH_HANS", "Test Course 001 (Simplified Chinese)", "zh-hans"),
]

store = modulestore()
print(f"Using author user: {author.username}")

for run, display_name, lang in courses:
    course_id = f"course-v1:{ORG}+{NUMBER}+{run}"
[...]

PR #3286, opened 2026-05-04
fix fallback when no language is selected

gumaerc:
Fixes a regression caused by https://github.com/mitodl/mit-learn/pull/3269

This PR fixes a regression caused by merging the above PR. A fallback was not properly set up on `getSelectedLanguageOption`, which caused course run association to fail when no language options are present. This PR restores the fallback so that if there are no language options, the proper course run is selected and associated with the dashboard card.

### How can this be tested?
- Ensure you have MIT Learn and MITx Online configured and connected
- Make sure you have at least one B2B contract that your user is enrolled in without language options, and it should contain at least one program with an enrollable course
- Go to the dashboard, click the org tab on the left and attempt to enroll in a course
- You should be redirected to the courseware
- If you go back to the dashboard, the card for that course should now reflect your enrollment status

### DIFF
--- a/frontends/main/src/app-pages/DashboardPage/CoursewareDisplay/languageOptions.ts
+++ b/frontends/main/src/app-pages/DashboardPage/CoursewareDisplay/languageOptions.ts
@@ -237,8 +237,6 @@ const getCourseRunForSelectedLanguage = (
   if (!languageOption) {
     return (
       getBestRun(course, { enrollableOnly: true }) ??
-      course.courseruns.find((run) => run.id === course.next_run_id) ??
-      course.courseruns.find((run) => run.is_enrollable) ??
       course.courseruns[0] ??
       null
     )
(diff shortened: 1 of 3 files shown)


frontends/main/src/app-pages/DashboardPage/CoursewareDisplay/languageOptions.ts
```diff
@@ -235,7 +235,13 @@ const getCourseRunForSelectedLanguage = (
 ): CourseRunV2 | null => {
   const languageOption = getSelectedLanguageOption(course, selectedLanguageKey)
   if (!languageOption) {
-    return null
+    return (
+      getBestRun(course, { enrollableOnly: true }) ??
+      course.courseruns.find((run) => run.id === course.next_run_id) ??
+      course.courseruns.find((run) => run.is_enrollable) ??
+      course.courseruns[0] ??
```
ChristopherChudzicki: I think you should remove these three fallbacks... getBestRun(course, { enrollableOnly: true }) returns first enrollable run, preferring the `next_run_id`.

The only fallback that might get triggered is `course.courseruns[0]`, and only if there were no enrollable runs (meaning  `course.courseruns[0]` would be unenrollable.)

ChristopherChudzicki (approved):

PR #3291, opened 2026-05-04
prefer existing B2B enrollment over next_run_id when picking same-lan…

ChristopherChudzicki:
Fix a bug on prod

This fixes a bug on prod where b2b contracts were not picking up enrollments correctly.

In particular, if a course has 2 runs and user is enrolled in the older once (i.e., not the `next_run_id`) then it was not picking up that enrollment.

**Summary of changes:**
- New `selectBestContractEnrollmentForLanguage` helper in `languageOptions.ts`: matches contract-scoped enrollments to the selected language via `language_options.id` / `courseware_id` (regardless of `is_enrollable`, so older/closed runs still match), with certificate > grade > first tiebreak.
- `ContractContent.tsx` calls the helper from both B2B render paths (`OrgProgramDisplay` and `OrgProgramCollectionDisplay`) before falling back to the existing `next_run_id`-pinned `getCourseRunForSelectedLanguage`. When a matching enrollment is found, `selectedRun` is seeded from the user's enrolled run; otherwise the unenrolled UX is unchanged.
- New tests in `languageOptions.test.ts` covering the regression case, no-enrollment-for-language, and the grade tiebreaker.

### How can this be tested? (On RC)

Assuming you have an admin user on RC:
1. Add some user to the MIT Org. The org is number [224 (admin page)](https://rc.mitxonline.mit.edu/admin/b2b/organizationpage/224/change/)), add your user by adding to row in https://rc.mitxonline.mit.edu/admin/b2b/userorganization/
2. Find your user (https://rc.mitxonline.mit.edu/admin/users/user/) add your user to contract to "550 - Cert Selection Test Contract 20260504230107"
3. View the org dashboard (https://rc.learn.mit.edu/dashboard/organization/mit-universal-ai/contract/cert-selection-test-20260504230107) You should not yet be enrolled in any of the org courses, so you should see all NEW RUN S (these are the runs specified by `course.next_run_id`).  
    ```
    Course 1 OLD Run ... 396
    Course 1 NEW Run ... 397
    Course 2 OLD Run ... 398
    Course 2 NEW Run ... 399
    Course 3 OLD Run ... 400
    Course 3 NEW Run ... 401
    ```
5. Create enrollments via Django admin. The runs for these courses are:
    - Create an enrollment in Course 1 Old Run
    - Create an enrollment in Course 2 New Run
6. Refresh dashboard. Your enrollments should be accurately reflected.
7. Try enrolling in Course 3... you should get the NEW Run. 

### How can this be tested? (local)
Set up a B2B contract in MITxOnline with at least three courses and manually enroll yourself in the course urns as follows:
```
Course 1 ... two runs , you are enrolled in OLD one ... NOT the next_run_id one
Course 2 ... two runs, you are enrolled in NEW one
Course 3 ... two runs, you are enrolled in neither
```
1. Check that the contract dashboard shows you enrolled for both Course 1 and Course 2
8. Check that enrolling in Course 3 goes to the NEW run.

### DIFF
--- a/frontends/main/src/app-pages/DashboardPage/CoursewareDisplay/languageOptions.ts
+++ b/frontends/main/src/app-pages/DashboardPage/CoursewareDisplay/languageOptions.ts
@@ -304,6 +304,65 @@ const getEnrollmentForSelectedLanguage = (
   )
 }
 
+/**
+ * Among contract-scoped enrollments, pick the user's best existing enrollment
+ * whose run is for the selected language on this course. Returns null when the
+ * user has no enrollment matching the language.
+ *
+ * Run language is determined via `course.language_options` (matching by id or
+ * courseware_id), independent of `is_enrollable` — we want to surface
+ * enrollments in older/closed runs too.
+ *
+ * Tiebreak across multiple matches: certificate > highest grade > first.
+ */
+const selectBestContractEnrollmentForLanguage = (
+  course: CourseWithCourseRunsSerializerV2,
+  enrollments: CourseRunEnrollmentV3[],
+  selectedLanguageKey: string,
+): CourseRunEnrollmentV3 | null => {
+  const resolvedKey =
+    selectedLanguageKey || getDefaultLanguageOptionKey(course) || ""
+  if (!resolvedKey) {
+    return null
+  }
+
+  const matchingOptions = (course.language_options ?? []).filter(
+    (option) => getLanguageOptionKey(option) === resolvedKey,
+  )
+  if (matchingOptions.length === 0) {
+    return null
+  }
+
+  const optionRunIds = new Set<number>()
+  const optionCoursewareIds = new Set<string>()
+  matchingOptions.forEach((option) => {
+    optionRunIds.add(option.id)
+    if (option.courseware_id) {
+      optionCoursewareIds.add(option.courseware_id)
+    }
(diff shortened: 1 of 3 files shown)


frontends/main/src/app-pages/DashboardPage/CoursewareDisplay/languageOptions.ts
```diff
@@ -304,6 +304,65 @@ const getEnrollmentForSelectedLanguage = (
...
+ * whose run is for the selected language on this course. Returns null when the
+ * user has no enrollment matching the language.
+ *
+ * Run language is determined via `course.language_options` (matching by id or
+ * courseware_id), independent of `is_enrollable` — we want to surface
+ * enrollments in older/closed runs too.
+ *
+ * Tiebreak across multiple matches: certificate > highest grade > first.
+ */
+const selectBestContractEnrollmentForLanguage = (
```
ChristopherChudzicki, line 318: I think that we should actually have this function handle non-b2b stuff, too for b2c people.

But keeping the change scoped as narrow as possible for the moment.

gumaerc (approved):
I tested this locally and was able to see proper enrollment selection given the provided test scenario

`main`
Image

`cc/lang-enrollment-pick-fix`
Image

PR #3293, opened 2026-05-05
fix b2b dashboard empty language code

gumaerc:
Another follow-up on https://github.com/mitodl/mit-learn/pull/3286

This PR fixes another issue with enrollment association on the B2B dashboard caused by introducing language selection. If language_code is blank, then null is returned from `selectBestContractEnrollmentForLanguage`. This fixes that by falling back to selectBestEnrollment

### How can this be tested?
- Ensure you have MIT Learn and MITx Online configured and connected
- Make sure you have at least one B2B contract that your user is enrolled in without language options, and it should contain at least one program with an enrollable course with the language_code set blank
- Go to the dashboard, click the org tab on the left and attempt to enroll in a course
- You should be redirected to the courseware
- If you go back to the dashboard, the card for that course should now reflect your enrollment status

### DIFF
--- a/frontends/main/src/app-pages/DashboardPage/CoursewareDisplay/languageOptions.ts
+++ b/frontends/main/src/app-pages/DashboardPage/CoursewareDisplay/languageOptions.ts
@@ -5,7 +5,7 @@ import type {
   CourseRunV2,
   CourseWithCourseRunsSerializerV2,
 } from "@mitodl/mitxonline-api-axios/v2"
-import { getBestRun } from "./helpers"
+import { getBestRun, selectBestEnrollment } from "./helpers"
 
 const LANGUAGE_CODE_TO_NATIVE_NAME: Record<string, string> = {
   ar: "العربية",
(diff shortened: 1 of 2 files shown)


frontends/main/src/app-pages/DashboardPage/CoursewareDisplay/languageOptions.ts
ChristopherChudzicki, line 348: If a contract has:
1. Course 1 ... two runs, both `""`
2. Course 2... two runs, both "en"`

Then this return makes you miss old enrollments in Course 1.
-  `resolvedKey` comes back as `'en'`
- `matchingOptions` comes back empty

ChristopherChudzicki (approved):
Approving this because it's an improvement over what's on main now, but https://github.com/mitodl/mit-learn/pull/3293/changes#r3185660558 might bite us if things are set up not quite right.

Ideally I think you are right that we should set courses to `"en"` in MITxOnline by default and make language a required field.

PR #3295, opened 2026-05-05
fix: 404 on product pages when live=false

daniellefrappier18:

Fixes https://github.com/mitodl/hq/issues/10597

**Problem**
Course and program product pages (/courses/[id], /courses/p/[id], /programs/[id]) were fetching from the MITxOnline API without the live=true filter. This caused a mismatch: the server-side route correctly filtered by live=true (returning 404 for non-live resources), but the client-side page components were fetching without the filter. When a resource had live=false, the client-side query would return results but the server would 404, leading to inconsistent behavior.

**Solution**
Add live: true to all MITxOnline API calls in the page components and their corresponding server-side routes, so client and server both consistently exclude non-live resources:

CoursePage.tsx — courses query
ProgramPage.tsx — programs query
ProgramAsCoursePage.tsx — programs query
/courses/[readable_id]/page.tsx — server-side courses fetch
/courses/p/[readable_id]/page.tsx — server-side programs fetch
/programs/[readable_id]/page.tsx — server-side programs fetch

- [x] Desktop screenshots

BEFORE Program
Program
UniversalAI

AFTER Program
UniversalAI

BEFORE Course 

Course

Case1

AFTER Course 
Case1404

BEFORE Program As Course 
ProgramAsCourse
Fundamentals ofDeepLearning

AFTER Program As Course 
Fundamentals ofDeepLearning404

### How can this be tested?

The product pages pull from two sources: the MITxOnline Programs/Courses API and the MITxOnline Wagtail CMS. Both must return data for a page to render.

Find a program with display_mode: "course" and live: true via http://mitxonline.odl.local:8013/api/v2/programs/?page_size=100 — note its readable_id
Confirm a Wagtail page exists for it: http://mitxonline.odl.local:8013/api/v2/pages/?fields=*&type=cms.programpage&readable_id=<readable_id> — if items is empty, create the page in the Wagtail admin at http://mitxonline.odl.local:8013/cms/ and publish it
Navigate to http://open.odl.local:8062/courses/p/<readable_id> — the page should render
Go to http://mitxonline.odl.local:8013/admin/ and find the program, uncheck the live field, and save
Reload http://open.odl.local:8062/courses/p/<readable_id> — it should now return a 404
Re-check live in the Django admin when done

### Additional Context

I noticed generateMetadata doesn't check live: true, so it can return metadata for pages that 404; not a blocker but worth a separate ticket.

### DIFF
--- a/frontends/main/src/app/(products)/courses/[readable_id]/page.tsx
+++ b/frontends/main/src/app/(products)/courses/[readable_id]/page.tsx
@@ -1,6 +1,10 @@
 import React from "react"
 import { HydrationBoundary, dehydrate } from "@tanstack/react-query"
-import { safeGenerateMetadata, standardizeMetadata } from "@/common/metadata"
+import {
+  safeGenerateMetadata,
+  standardizeMetadata,
+  MetadataNotFound,
+} from "@/common/metadata"
 import CoursePage from "@/app-pages/ProductPages/CoursePage"
 import { notFound } from "next/navigation"
 import { pagesQueries } from "api/mitxonline-hooks/pages"
@@ -17,17 +21,16 @@ export const generateMetadata = async (
   return safeGenerateMetadata(async () => {
     const queryClient = getQueryClient()
 
-    const coursePages = await queryClient.fetchQuery(
-      pagesQueries.coursePages(readableId),
+    const { results: courses } = await queryClient.fetchQuery(
+      coursesQueries.coursesList({ readable_id: readableId, live: true }),
     )
 
-    if (coursePages.items.length === 0) {
-      notFound()
+    if (courses.length === 0) {
+      throw new MetadataNotFound()
     }
-    const [course] = coursePages.items
+    const [course] = courses
 
-    const image =
-      course.course_details.page.feature_image_src || DEFAULT_RESOURCE_IMG
+    const image = course.page?.feature_image_src || DEFAULT_RESOURCE_IMG
 
     return standardizeMetadata({
       title: course.title,
--- a/frontends/main/src/app/(products)/courses/[readable_id]/page.tsx
+++ b/frontends/main/src/app/(products)/courses/[readable_id]/page.tsx
@@ -49,7 +49,7 @@ const Page: React.FC<PageProps<"/courses/[readable_id]">> = async (props) => {
   const [coursePages, courses] = await Promise.all([
     queryClient.fetchQuery(pagesQueries.coursePages(readableId)),
     queryClient.fetchQuery(
-      coursesQueries.coursesList({ readable_id: readableId }),
+      coursesQueries.coursesList({ readable_id: readableId, live: true }),
     ),
   ])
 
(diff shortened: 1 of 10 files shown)


frontends/main/src/app/(products)/courses/[readable_id]/page.tsx
```diff
@@ -49,7 +49,7 @@ const Page: React.FC<PageProps<"/courses/[readable_id]">> = async (props) => {
   const [coursePages, courses] = await Promise.all([
     queryClient.fetchQuery(pagesQueries.coursePages(readableId)),
     queryClient.fetchQuery(
-      coursesQueries.coursesList({ readable_id: readableId }),
+      coursesQueries.coursesList({ readable_id: readableId, live: true }),
```
Comment (ChristopherChudzicki) [inline frontends/main/src/app/(products)/courses/[readable_id]/page.tsx:55]: > I noticed generateMetadata doesn't check live: true, so it can return metadata for pages that 404; not a blocker but worth a separate ticket.

**Request:** I think we should fix the metadata issue before we mark the issue (https://github.com/mitodl/hq/issues/10597) as closed. *Which will happen automatically right now, since the PR description says "Fixes ...". I think github auto-closes for that text.*

Because this is undesirable:

Image

All the changes you've made in this PR (which look good) would stand, so you could merge this and open. followup. But let's not close https://github.com/mitodl/hq/issues/10597 till ☝️ is resolved.

-------------

### Suggestion / Some Info
1. Swap `pagesQueries.coursePages(readableId)` for coursesQueries.coursesList({ readable_id: readableId, live: true }) in `generateMetadata`.... the latter one has the `live` filter available.
    - This has `course.title` and `course.page.feature_image_src` available
2. Tweak `safeGenerateMetadata` (see below)


Re (2) ... Something I had not realized until you made this PR... and apropos of our metadata discussion this morning... is that calling notFound inside safeGenerateMetadata doesn't really work the way it should:
```ts
export const generateMetadata = params => {
  // read params
  return safeGenerateMetadata(() => {
  // do stuff
  if (condition) notFound()                     <--- This doesn't trigger notFound metadata
  return {...}
  })
)
```

That's because
- `safeGenerateMetadata` does a try/catch, catching Axios 404 errors specifically
- the notFound call above throws ... some other error which NextJS catches internally.
- NextJS does not expose a way for us to catch the error it throws via notFound
    - E.g., the error class is not exported, so we can't check for it via instanceof in the catch block
    - They actually do export this [unstable_rethrow](https://nextjs.org/docs/app/api-reference/functions/unstable_rethrow) but it's marked as unstable. Doesn't seem like something we should use. Annoying.

Pretty disappointing that NextJS doesn't provide a way to catch and rethrow their error other than the unstable_ thing. But you could do this:
```ts
    const { results } = await queryClient.fetchQuery(
      coursesQueries.coursesList({ readable_id: readableId, live: true }),
    )

    if (results.length === 0) {
      throw new MetadataNotFound() // <--- custom error exported next to safeGenerateMetadata
    }
```

ChristopherChudzicki (approved):

frontends/main/src/app/(products)/courses/[readable_id]/page.tsx
Comment (daniellefrappier18) [reply frontends/main/src/app/(products)/courses/[readable_id]/page.tsx:55]: @ChristopherChudzicki I think it makes the most sense to tackle this in this PR. I have implemented those changes:

Case1
FundamentalsofDeepLearning
UniversalAI

Can you give this another re-review when you have a chance?

ChristopherChudzicki (approved):
👍 Works great!

PR #3299, opened 2026-05-06
fix: remove the top border when there is only single item and removed the current episode from the more episode list

ahtesham-quraish:
n/a

- There was edge case that if there is only single episode then the top border was rendering which should not so this is about removing the top border in case single item
- one more change in this PR that the current episode should not be the part of `More from  <podcast>` which was previously part of this list so I removed it from that more episode list.

Before
image

After
image

### How can this be tested?
if you don't have playlist data locally then in frontend env file you should add the following variable `NEXT_PUBLIC_MITOL_API_BASE_URL="https://api.rc.learn.mit.edu"` it will connect with RC learn backend
then we need to enable the flag which is `podcast-detail-page` and then visit the following url `http://open.odl.local:8062/podcast/128673/podcast_episode/128674`


zamanafzal (approved):
LGTM

PR #3306, opened 2026-05-06
fix: change play icon for video collection detail page

ahtesham-quraish:
n/a

Change the play icon on video collection detail page

image

### How can this be tested?
then we need to enable the flag which is `video-playlist-page` and then visit the following url `http://open.odl.local:8062/video-playlist/detail/88444?playlist=88443`


abeglova (approved):
👍

PR #3311, opened 2026-05-07
fix: add iframe for youtube videos playing

ahtesham-quraish:
https://github.com/mitodl/hq/issues/11181#top

- Use Iframe for youtube videos directly instead using the vidoejs library 

image

### How can this be tested?
then we need to enable the flag which is `video-playlist-page` and then visit the following url `http://open.odl.local:8062/video-playlist/detail/113707?playlist=113705`


frontends/main/src/app-pages/VideoPlaylistCollectionPage/YouTubeIframePlayer.tsx
```diff
@@ -0,0 +1,58 @@
...
+  title,
+  ariaLabel,
+  ariaDescribedBy,
+}) => {
+  // Prefer the build-time env var; fall back to the browser's own origin at
+  // runtime (always available in a "use client" component). Omit the param
+  // entirely when neither is available (e.g. during SSR without the env var).
+  const origin =
+    process.env.NEXT_PUBLIC_ORIGIN ||
+    (typeof window !== "undefined" ? window.location.origin : "")
```
ChristopherChudzicki: If NEXT_PUBLIC_ORIGIN is undefined, we have big trouble. We should figure out how to handle TS better for env vars, but I tend to put `invariant(NEXT_PUBLIC_ORIGIN, "NEXT_PUBLIC_ORIGIN must be defined") at the top of the file.

ChristopherChudzicki (changes requested):
@ahtesham-quraish I apologize I didn't get a chance to try this today. I did look over the code a bit. It seems to me that both `VideoDetailPage.tsx` and `VideoSeriesDetailPage.tsx` have a PlayerWrapper that handles:

- loading skeleon
- conditionally displaying videojs vs iframe
- fallbacks for videos without sources 

Am I correct about this? 

I'd suggest instead a shared VideoResourcePlayer component that handles these things:
- both pages can use the same VideoResourcePlayer
- if necessary, both pages can style it differently via styled(...)
- The new VideoDetailPAge test file probably doesn't need to test all the details of the VideoResourcePlayer, then.

Additionally, this PR adds some new code for parsing YouTube ids. Could we instead use the existing `convertToEmbedUrl` function from `frontends/main/src/common/utils.ts:44.`?

frontends/main/src/app-pages/VideoPlaylistCollectionPage/YouTubeIframePlayer.tsx
ahtesham-quraish: Addressed

ChristopherChudzicki (changes requested):
Working well! Two small issues:
- remove `extractYouTubeId` ... It's not longer used now that you're using the function we already had
- need to fix conflicts

ahtesham-quraish:
> * extractYouTubeId

Can you check now please?

ChristopherChudzicki (approved):

PR #3316, opened 2026-05-08
fix: change the video detail page route & load more videos feature on video collection page

ahtesham-quraish:
https://github.com/mitodl/hq/issues/11176

We should change the video detail page URL to be more appropriate. Currently, it's `video-playlist/detail/:id/?playlist=:playlist_id`, It should be now `/video/{videoId}`

- [ ] Desktop screenshots

### How can this be tested?
then we need to enable the flag which is `video-playlist-page` and then visit the following url `http://open.odl.local:8062/video/113707`


frontends/main/src/app-pages/VideoPlaylistCollectionPage/VideoDetailPage.tsx
```diff
@@ -637,7 +637,7 @@ const VideoDetailPage: React.FC<VideoDetailPageProps> = ({
                   shareButtonRef.current as unknown as HTMLDivElement | null
                 }
                 onClose={() => setShareOpen(false)}
-                pageUrl={`${NEXT_PUBLIC_ORIGIN}/video-playlist/${video?.id}?playlist=${playlistId}`}
+                pageUrl={`${NEXT_PUBLIC_ORIGIN}/video/${video?.id}?playlist=${playlistId}`}
```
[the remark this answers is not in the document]
ahtesham-quraish: i think it won't be any issue

frontends/main/src/app-pages/VideoPlaylistCollectionPage/VideoPlaylistCollectionPage.tsx
```diff
@@ -152,6 +178,22 @@ const VideoPlaylistCollectionPage: React.FC<
           </EpisodeListContainer>
         </EpisodeListSection>
       )}
+      {(hasNextPage || itemsLoading) && (
```
daniellefrappier18:
```suggestion
      {hasNextPage && (
```

The issue is that itemsLoading being in the condition at all causes the button to appear before the first page of data arrives. hasNextPage alone is the right signal. It's only true after the first page has loaded and the API has confirmed there's more data to fetch.

frontends/main/src/app/video-playlist/[id]/page.tsx
Comment (daniellefrappier18) [inline frontends/main/src/app/video-playlist/[id]/page.tsx:29]:
```suggestion
        learningResourceQueries.items(playlistId, {
          learning_resource_id: playlistId,
          limit: 1,
        }),
```

This fetches all playlist items to obtain only the first image. No limit is passed, so the API returns the full items list (potentially hundreds of items) just to read items?.[0]?.image?.url. This runs on every metadata request.

daniellefrappier18 (approved):
Looks good but please see my suggestions before merging especially adding a limit the generateMetadata query. Also looks like you have a merge conflict.

frontends/main/src/app/video-playlist/[id]/page.tsx
Comment (ahtesham-quraish) [reply frontends/main/src/app/video-playlist/[id]/page.tsx:29]: Addressed

frontends/main/src/app-pages/VideoPlaylistCollectionPage/VideoPlaylistCollectionPage.tsx
ahtesham-quraish: Addressed

PR #3321, opened 2026-05-08
Show (and count) only published items in userlists/paths

mbertrand:
Closes https://github.com/mitodl/hq/issues/10177

Filters out unpublished resources from user lists and learning paths.

### How can this be tested?
- While logged in as an admin, add a particular resource to both your Favorites list and a learning path.
- View the items in both the favorites list and learning path in the UI, and verify the resource shows up there, and that the counts are accurate.
- In a django shell, run the following:
   ```python
    from learning_resources.utils import *

    lr = LearningResource.objects.get(id=<resource_id>)
    lr.published=False
    lr.save()
    resource_unpublished_actions(lr)
    ```
- Refresh both the Favorites page and learning path page.  Verify that the unpublished resource no longer appears and the count is reduced by 1.
- In the django shell, republish the resource
  ```python
  lr.published=True
  lr.save()
  resource_upserted_actions(lr, False, true)
  ```
- Refresh both the Favorites page and learning path page.   The resource should appear again, and the counts should be back to what they were originally.

### DIFF
--- a/frontends/api/src/generated/v0/api.ts
+++ b/frontends/api/src/generated/v0/api.ts
@@ -2263,7 +2263,7 @@ export interface LearningPath {
    */
   id: number
   /**
-   *
+   * Number of published items in the list.
    * @type {number}
    * @memberof LearningPath
    */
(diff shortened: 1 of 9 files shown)


abeglova (approved):
LGTM

PR #3327, opened 2026-05-11
feat: add video embed page and route group restructure

daniellefrappier18:

Closes https://github.com/mitodl/hq/issues/11123

- Adds a /video/[id]/embed route under a new (embed) route group for embedding individual OCW videos in iframes. The page validates the resource is a playable video (HLS, MP4, or YouTube) before rendering.
- Removes the VideoPlaylistPage feature flag entirely, cleaning it up from  all existing pages that previously gated on it.
- Restructures the app router by introducing a (site) route group that wraps all existing pages with the site header, footer, and global styles, keeping the root layout minimal so the embed route can opt out of the site UI.
- Adds (embed)/not-found.tsx for iframe-appropriate 404s (no home button) and (embed)/error.tsx for non-404 runtime errors, both without site UI
- Adds (site)/[...unmatched]/page.tsx catch-all so unmatched site routes (e.g. /bad-route) correctly resolve to (site)/not-found.tsx with header/footer.
- Adds generateMetadata to the embed page so the iframe has a meaningful <title> and is marked noindex; falls back to "Video" for missing or non-embeddable resources.

- [x] Desktop screenshots
- [x] Mobile width screenshots

Video Embed Page with valid id (not YouTube)
http://open.odl.local:8062/video/115203/embed
Screenshot 2026-05-12 at 9 50 07 AM
Screenshot 2026-05-12 at 9 51 17 AM

Video Embed 404 Page with NON valid id (not YouTube) 
http://open.odl.local:8062/video/000/embed

Screenshot 2026-05-12 at 3 10 04 PM

Video Embed Page with a valid YouTube id
Screenshot 2026-05-12 at 1 12 57 PM

Video Embed Page with non-404 error
I just added  `throw new Error("test error")` temporarily to VideoEmbedPage.tsx to simulate 

`const VideoEmbedPage: React.FC<VideoEmbedPageProps> = ({ videoId }) => {
  throw new Error("test error")   // ← here, remove after testing
  const { data: resource, isLoading } = useLearningResourcesDetail(videoId)
  ...`

Screenshot 2026-05-12 at 2 48 53 PM

Regular non video embed page that doesn't exist 
http://open.odl.local:8062/this-does-not-exist

Screenshot 2026-05-12 at 3 34 46 PM

Existing site route page 
http://open.odl.local:8062/search?q=  
Screenshot 2026-05-12 at 11 34 51 AM

### How can this be tested?

 

- [ ] Navigate to a valid video embed URL (e.g. /video/[id]/embed). The video player renders full-viewport with no with no header/footer and the tab title is that of the video plus the | MIT Learn appendix 
- [ ] Navigate to /video/000/embed (nonexistent ID). The embed 404 page renders with no header/footer or Home button, and the tab title is "Not Found | MIT Learn"
- [ ] Navigate to /this-does-not-exist or another page that will 404. It shows custom 404 page with site header and footer and have the title "Not Found | MIT Learn" 
- [ ] Navigate to any existing site route (e.g. /search) — header and footer still present, no regression
- [ ]  Trigger a non-404 error on an embed page (e.g. temporarily throw new Error("test") at the top of VideoEmbedPage.tsx). The error page renders full-viewport with no header/footer and no Home button, and the tab title is that of the video plus the | MIT Learn appendix


frontends/main/src/app/(site)/video-playlist/detail/[id]/page.tsx
```diff
@@ -0,0 +1,94 @@
...
+import VideoDetailPageRouter from "@/app-pages/VideoPlaylistCollectionPage/VideoDetailPageRouter"
+import { notFound } from "next/navigation"
+import type { VideoResource } from "api/v1"
+
+export const generateMetadata = async (props: PageProps<"/video/[id]">) => {
+  const { id } = await props.params
+  const videoId = Number(id)
+  if (!Number.isInteger(videoId) || videoId <= 0) {
+    notFound()
+  }
```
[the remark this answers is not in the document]
Comment (daniellefrappier18) [reply frontends/main/src/app/(site)/video-playlist/detail/[id]/page.tsx:18]: done

frontends/main/src/app-pages/ErrorPage/ErrorPageTemplate.tsx
```diff
@@ -99,11 +101,13 @@ export const ErrorContent: React.FC<ErrorPageTemplateProps> = ({
       >
         {title}
       </Typography>
-      <Footer>
-        <Button variant="primary" href={HOME} Component="a">
-          Home
-        </Button>
-      </Footer>
+      {!hideHomeButton && (
```
ChristopherChudzicki: **Minor Suggestion:** Minor preference for `showHomeButton` with `default=true`.

In general, I think it's good to avoid double negatives. (`!hide` = show).

frontends/main/src/app/(embed)/video/[id]/embed/page.test.tsx
```diff
@@ -0,0 +1,123 @@
...
+import Page from "./page"
+
+jest.mock("@/app/getQueryClient", () => ({
+  getQueryClient: jest.fn(),
+}))
+
+jest.mock("@tanstack/react-query", () => ({
+  ...jest.requireActual("@tanstack/react-query"),
+  dehydrate: jest.fn().mockReturnValue({}),
+}))
```
Comment (ChristopherChudzicki) [inline frontends/main/src/app/(embed)/video/[id]/embed/page.test.tsx:16]: **Request:** Unless there's a strong reason to mock react-query, I think we should mock the API responses instead. It's closer to reality / production behavior.

frontends/main/src/app/(embed)/video/[id]/embed/page.tsx
```diff
@@ -0,0 +1,91 @@
...
+          resource.content_files?.[0]?.youtube_id,
+        )[0]?.type ?? "",
+      )
+
+    if (!isEmbeddableVideo) {
+      throw new MetadataNotFound()
+    }
+
+    return standardizeMetadata({
+      title: resource?.title ?? "Video",
```
Comment (ChristopherChudzicki) [inline frontends/main/src/app/(embed)/video/[id]/embed/page.tsx]:
```suggestion
      title: resource.title
```
`title` is require, and `resource` has already been verified to exist.

AI does this somethings, and I'm not sure why. The TS types are clear about ☝️

frontends/main/src/app/(embed)/video/[id]/embed/page.tsx
```diff
@@ -0,0 +1,91 @@
...
+      .fetchQuery(learningResourceQueries.detail(videoId))
+      .catch(() => {
+        throw new MetadataNotFound()
+      })
+
+    const isEmbeddableVideo =
+      resource?.resource_type === VideoResourceResourceTypeEnum.Video &&
+      SUPPORTED_TYPES.includes(
+        resolveVideoSources(
+          (resource as VideoResource).video?.streaming_url,
```
Comment (ChristopherChudzicki) [inline frontends/main/src/app/(embed)/video/[id]/embed/page.tsx]:
```suggestion
          (resource).video?.streaming_url,
```
`resource?.resource_type === VideoResourceResourceTypeEnum.Video` refines the type, so no need for a typecast.

frontends/main/src/app/(embed)/video/[id]/embed/page.tsx
```diff
@@ -0,0 +1,91 @@
...
+    videoResource.content_files?.[0]?.youtube_id,
+  )
+
+  if (sources.length === 0 || !SUPPORTED_TYPES.includes(sources[0].type)) {
+    notFound()
+  }
+
+  return (
+    <HydrationBoundary state={dehydrate(queryClient)}>
+      <VideoEmbedPage videoId={videoId} />
```
Comment (ChristopherChudzicki) [inline frontends/main/src/app/(embed)/video/[id]/embed/page.tsx]: What you've done here is exactly what we do elsewhere in the app. Nonetheless,

**Suggestion:** Instead of passing `videoId`, just pass `videoResource`. That will simplify things:
- VideoEmbedPage does not need to re-check that the resource is a video. (I don't think it does currently; but for `VideoEmbedPage` to make sense in isolation, it probably should.)
- VideoEmbedPage clearly has no need for a loading state if we pass videoResource
- VideoEmbedPage tests become simpler. No need to mock the api response, just pass the resource

```suggestion
      <VideoEmbedPage videoResource={videoResource} />
```

**Why don't we do this elsewhere?** Mostly I think it's historical and we should do the way suggested above more offten.
1. Historical: Our usage of react query predates our adoption of nextjs, and before nextjs w/ server components, we could not have done it this way.
2. In some cases, we wanted to retry the query on the client in case there was an intermittent server issue. (I don't think this is a great reason).

**However:**  I would keep HydrationBoundary, etc. We use react query everywhere else and it can make the API call for us via the `fetchQueryOr404` you're using. So no harm. And there are definitely cases where we want to make queries only client-side, where react query is definitely the tool to use. (User-dependent queries, or queries that aren't important enough to delay the initial page load.) So keeping that aspect consistent with the rest of hte app is worthwhile.

frontends/main/src/app/layout.tsx
```diff
@@ -116,14 +112,7 @@ j=d.createElement(s),dl=l!=='dataLayer'?'&l='+l:'';j.async=true;j.src=
...
-          <PageWrapper>
-            <Header />
-            <PageWrapperInner>{children}</PageWrapperInner>
-            <Footer />
-          </PageWrapper>
-        </Providers>
+        <Providers>{children}</Providers>
         {process.env.NEXT_PUBLIC_APPZI_URL ? (
           <Script async src={process.env.NEXT_PUBLIC_APPZI_URL} />
         ) : null}
```
ChristopherChudzicki: **Request:** I this this should go in `(site)/layout.tsx`. It's this thing:

Image

and it will look weird inside iframes.

frontends/main/src/app/error.tsx
```diff
@@ -15,10 +15,6 @@ import FallbackErrorPage from "@/app-pages/ErrorPage/FallbackErrorPage"
 import { ForbiddenError } from "@/common/errors"
 import ForbiddenPage from "@/app-pages/ErrorPage/ForbiddenPage"
 
-export const metadata = {
-  title: "Error",
-}
```
ChristopherChudzicki, line 20: I was confused why this was removed. 

For posterity... `error.tsx` becomes an ErrorBoundary, and error boundaries are always client components. https://nextjs.org/docs/app/getting-started/error-handling#nested-error-boundaries

And client components can't set metadata, so this was useless all along.

_`not-found.tsx` is NOT converted into an error boundary by nextjs, it's explicitly triggered via `notFound` or missng routes on server render, so it's different._

frontends/main/src/app/(embed)/video/[id]/embed/page.tsx
```diff
@@ -0,0 +1,91 @@
...
+import VideoEmbedPage from "@/app-pages/VideoEmbedPage/VideoEmbedPage"
+import { VideoResourceResourceTypeEnum } from "api/v1"
+import type { VideoResource } from "api/v1"
+import {
+  safeGenerateMetadata,
+  standardizeMetadata,
+  MetadataNotFound,
+} from "@/common/metadata"
+
+const SUPPORTED_TYPES = ["application/x-mpegURL", "video/mp4", "video/youtube"]
```
Comment (ChristopherChudzicki) [inline frontends/main/src/app/(embed)/video/[id]/embed/page.tsx]: I think we should remove this check—I realize it may be here because I wrote the issue as excluding youtube; you clarified that, and also asked about DASH. I said DASH wasn't important. 

Here's what I think:
 - VideoResourcePlayer is the real behavior... what works, what doesn't.
 - This route validates the important thing... that the resource is a video. ✅ 
 - Beyond that, it should just delegate to VideoResourcePlayer.
    - If the video player supports that video type, it should play
    - If the video type is not supported, it should error, and that should be the VideoPlayer's responsibility.
 - SUPPORTED_TYPES adds a second compatibility gate that can get out of sync with shared player logic, which would be undesirable.

What I really meant about youtube and dash (and did a poor job communicating) was that we shouldn't go through extra effort here to support it. But we also shouldn't go through extra effort to prevent it. **We should leave that up to the VideoResourcePlayer.**

ChristopherChudzicki (changes requested):
This is working well! I tried out some iframes with the embed page, too:

Image

I left some comments and suggestions. Main requested change is around the appzi link.

frontends/main/src/app-pages/ErrorPage/ErrorPageTemplate.tsx
daniellefrappier18: no problem - I renamed that prop

frontends/main/src/app/(embed)/video/[id]/embed/page.test.tsx
Comment (daniellefrappier18) [reply frontends/main/src/app/(embed)/video/[id]/embed/page.test.tsx:16]: done

frontends/main/src/app/(embed)/video/[id]/embed/page.tsx
Comment (daniellefrappier18) [reply frontends/main/src/app/(embed)/video/[id]/embed/page.tsx]: I should have removed this but forgot after I clarified with you.

frontends/main/src/app/error.tsx
daniellefrappier18: This appears to be dead code so I was just cleaning it up. Is there a reason to keep this that I'm not aware of?

frontends/main/src/app/layout.tsx
daniellefrappier18: Good catch this would have been a problem!

BEFORE 
Screenshot 2026-05-15 at 2 36 09 PM


AFTER
Screenshot 2026-05-15 at 2 36 09 PM

frontends/main/src/app/error.tsx
ChristopherChudzicki: No, you were correct to remove it.

I just meant: "I was confused why it was dead code... the reason it, it should never have been added. `error.tsx` cannot set metadata".

frontends/main/src/app/(embed)/video/[id]/embed/page.tsx
```diff
@@ -0,0 +1,66 @@
...
+    learningResourceQueries.detail(videoId),
+  )
+
+  if (resource.resource_type !== VideoResourceResourceTypeEnum.Video) {
+    notFound()
+  }
+
+  return (
+    <HydrationBoundary state={dehydrate(queryClient)}>
+      <VideoEmbedPage videoResource={resource as VideoResource} />
```
Comment (ChristopherChudzicki) [inline frontends/main/src/app/(embed)/video/[id]/embed/page.tsx]:
```suggestion
      <VideoEmbedPage videoResource={resource} />
```

Comment (ChristopherChudzicki) [reply frontends/main/src/app/(embed)/video/[id]/embed/page.tsx]: Unnecessary cast here. I'm guessing an LLM added this? Need to figure out how to make them do that less. (It's not a big deal, but it is unnecessary noise in code, but casting can also be unsafe.)

frontends/main/src/app-pages/VideoEmbedPage/VideoEmbedPage.tsx
```diff
@@ -0,0 +1,32 @@
...
+  height: "100vh",
+  aspectRatio: "unset",
+})
+
+type VideoEmbedPageProps = {
+  videoResource: VideoResource
+}
+
+const VideoEmbedPage: React.FC<VideoEmbedPageProps> = ({ videoResource }) => {
+  const videoTitleLabel = videoResource.title?.trim() || "Untitled video"
```
ChristopherChudzicki: - We should remove the `?` ... title is required by spec / TS.
-  You could also remove the fallback title. It is not allowed to be blank by our Django model  (https://github.com/mitodl/mit-learn/blob/main/learning_resources/models.py#L469) ... TS can't communicate that, though, so if you want to leave the fallback, feel free.

frontends/main/src/app/layout.tsx
```diff
@@ -116,17 +112,7 @@ j=d.createElement(s),dl=l!=='dataLayer'?'&l='+l:'';j.async=true;j.src=
...
-            <Header />
-            <PageWrapperInner>{children}</PageWrapperInner>
-            <Footer />
-          </PageWrapper>
-        </Providers>
-        {process.env.NEXT_PUBLIC_APPZI_URL ? (
-          <Script async src={process.env.NEXT_PUBLIC_APPZI_URL} />
-        ) : null}
+        <Providers>{children}</Providers>
         {process.env.NEXT_PUBLIC_HUBSPOT_PORTAL_ID ? (
```
ChristopherChudzicki: I apologize for missing this in the earlier review. We should move hubspot script toe `(site)/layout.tsx`. It's for marketing forms, etc, and those wouldn't show up in embedded videos.

~I **think** we should also move the GTM stuff, but let's hold off on that now. I don't understand as well what we use that for.~

Hmm, we may also want to remove GTM/posthog on the root layout, too.

ChristopherChudzicki (changes requested):
Ack, I. missed one thing I think we definitely should move in earlier review https://github.com/mitodl/mit-learn/pull/3327/changes#r3260403444

Could you move the hubspot script, too? Also left a couple minor comments.

After that, I think should be good to go.

frontends/main/src/app/layout.tsx
ChristopherChudzicki: I suggest we remove Posthog and GTM from the root layout, though we could also do this as a more separate, focused, followup PR.

- Posthog and GTM both use cookies / localstorage, which won't work well when embedded in an iframe. Learn becomes a third party site when loaded in a different domain's iframe. E.g., Third-party-within-iframe cookies won't work at all on Safari (which includes all iPhones).
- in the case of GTM, what it uses cookies/localstorage for depends on the tags configured; we don't control that. Right now no tags are configured as far as I can tell.
- Even when the cookies/localstorage do work, the analytics we'd get probably isn't what we want (e.g., pageload would be by loading `https://othersite.com`, not by actively playing our own video.)
- Ebmedded player does not have any of our privacy/consent mechanisms, so shouldn't contain those cookies


Concrete suggestion:
- Move GTM + hubspot to (site)/layout.tsx
- Remove Posthog from base providers tree and layer it separately in `(site)/layout./tsx` (In terms of positioning, the only thing that matters is ConfiguredPostHogProvider is a descendant of QueryClientProvider)
    - if a child (like VideoResourcePlayer) ends up calling `posthog.capture` or something, it will just be a no-op if posthog is initailized

frontends/main/src/app-pages/VideoEmbedPage/VideoEmbedPage.tsx
daniellefrappier18: done

frontends/main/src/app/layout.tsx
daniellefrappier18: Done! Moved GTM and HubSpot to (site)/layout.tsx and removed ConfiguredPostHogProvider from the base providers tree, adding it back in (site)/layout.tsx via a thin SiteProviders client wrapper (needed since (site)/layout.tsx is a server component and ConfiguredPostHogProvider uses hooks).

ChristopherChudzicki (approved):
👍 Looks good!

Could you make sure to merge in main or rebase? Just to double check now new `/app` routes have been added that would need to be pulled into (site). Those would probably not cause conflicts, so we wouldn't detect just by conflict-free.

PR #3330, opened 2026-05-11
dashboard refactor stage 2

gumaerc:
Part of https://github.com/mitodl/hq/issues/11205

This PR executes phase 2 of the dashboard refactoring plan detailed [here](https://gist.github.com/ChristopherChudzicki/d4a0204ba15abc3bdfee33d462fbae78), also now checked into the repo at `frontends/main/src/app-pages/DashboardPage/CoursewareDisplay/dashboardRefactorPlan.md`, to be deleted after refactor is complete.

Phase 2 deduplicates some logic in the B2B and B2C components (`ContractContent` / `EnrollmentDisplay`) surrounding multiple steps that are taken to associate an enrollment / course run with language selection. This process was moved into the view model (`dashboardViewModel.ts`) and the components were updated to use this.

### How can this be tested?
- Ensure you have MITx Online running and connected to Learn as is described in the readme
- Ensure that you have sufficient test data to test all aspects of the Learn dashboard's ability to display a learner's enrolled contracts and individual programs and courses
- Compare how the following pages display with your test data on `main` vs this branch:
  - B2B contract dashboard
  - My Learning (dashboard home page)
    - Course run enrollments
    - Program enrollments
  - Program dashboard page (click View Program on a personal program enrollment)

### DIFF
--- a/frontends/main/src/app-pages/DashboardPage/CoursewareDisplay/model/dashboardViewModel.ts
+++ b/frontends/main/src/app-pages/DashboardPage/CoursewareDisplay/model/dashboardViewModel.ts
@@ -9,10 +9,20 @@
 import type { SimpleSelectOption } from "ol-components"
 import type {
   CourseRunEnrollmentV3,
+  CourseRunLanguageOption,
   CourseRunV2,
   CourseWithCourseRunsSerializerV2,
   V3UserProgramEnrollment,
 } from "@mitodl/mitxonline-api-axios/v2"
+import { selectBestEnrollment } from "../helpers"
+import {
+  getCourseRunForSelectedLanguage,
+  getDistinctLanguageOptions,
+  getEnrollmentForSelectedLanguage,
+  getResolvedRunForSelectedLanguage,
+  getSelectedLanguageOption,
+  selectBestContractEnrollmentForLanguage,
+} from "../languageOptions"
 
 /**
  * A program/contract dashboard's view of a course: every enrollment whose run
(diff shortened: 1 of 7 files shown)


frontends/main/src/app-pages/DashboardPage/CoursewareDisplay/model/dashboardViewModel.ts
```diff
@@ -116,8 +127,237 @@ const groupProgramEnrollmentsByProgramId = (
...
+  )
+
+  additionalOptions.sort((a, b) =>
+    String(a.label).localeCompare(String(b.label)),
+  )
+
+  return [...baseOptions, ...additionalOptions]
+}
+
+const getDistinctDashboardLanguageOptions = (
```
ChristopherChudzicki, line 166: AI's summary of my gut dislike of the implementation of this function:

Heads up — I'd like to swap out `getDistinctDashboardLanguageOptions`. Not because it's wrong, but because answering "what shows up in the picker?" currently means tracing `getDistinctLanguageOptions` → `getEnrollableLanguageOptions` →  `isLanguageOptionEnrollable` in `languageOptions.ts`, plus a parallel local path via `getLanguageOptionFromEnrollment` →  `getNativeLanguageName` — four helpers across two files **for what's effectively union + dedup + sort**.

As I rewrote it a bit, a few things I noticed:
- `getEnrollableLanguageOptions` seems bogus to me. It matches the langauge run ID against `course.courseruns`, but `course.courseruns` generally does not include the language-specific runs. So it almost always returns true. (I do think it successfully filters out the default language options if they are not enrollable.) I dropped it entirely.
- The `enrollment.run` thing we discussed



Here's the commit i made for it: https://github.com/mitodl/mit-learn/commit/7ba9f8bac4aeedbc2bb97d0c39994c17fadc2895 ... I thought an actual commit / diff would be the easiest way to communicate this.

Goal was:
- much more clearly just union + dedupe + sort
- better names
- a few comments

ChristopherChudzicki (changes requested):
Take a look at https://github.com/mitodl/mit-learn/pull/3330/changes#r3228712625 ... What do you think?

I also think `getDashboardLanguageOptions` is useless. The plan doc called it out, but I think it was a mistake. The commit I suggested removes it from the plan, but not from the code.

Can you take a look? I suspect if you accept that commit and remove `getDashboardLanguageOptions`, a bunch of other code will become dead and we can remove it.

In general, I think we need to look closely moving forward and be careful that each phase of the plan actually makes the dashboard more intelligible / easier to reason about.

The logic in `resolveSlotForLanguage` is more complex than i would hope, but it is not new logic just moved, and I don't think we should futz with it too much now.

ChristopherChudzicki (approved):

ChristopherChudzicki:
@gumaerc As discussed yesterday, i pushed an extra two commits. They:
- delete dead code
- add a few docs/ modify plan doc slightly (around success criteria at each step, not overall shape).

ChristopherChudzicki:
@gumaerc As discussed yesterday, i pushed an extra two commits. They:
- delete dead code
- add a few docs/ modify plan doc slightly (around success criteria at each step, not overall shape).

PR #3332, opened 2026-05-12
refactor: make our internally-authored content editing more generic ( First Part)

ahtesham-quraish:
https://github.com/mitodl/hq/issues/11232

**Goal:** Make backend naming `content_type` agnostic and introduce one new content_type. Preserve existing API endpoint at least during cutover.

Success looks like:
- [ ] 1. rename `articles` app and directory to `website_content`
    - renaming an app is nontrivial but not too hard, especially at the low data scale we have for this. Worth doing now.
- [ ] 2. rename `Article` model to `WebsiteContent` with new `content_type` field, restricted to `"news"` and `"article"`. Should default to `"news"`, at least for now.
- [ ] 3. make  `api/v1/website_content/` and `api/v1/articles/` endpoints behave the same, returning any content type with filter. These can be literally the same viewset for now, but we should leave articles in place for a smooth migration of the frontend. (Expand then contract API surface.)
- [ ] 4. Ensure that `news_events/etl` only ingests `content_type=news`.
- [ ] 5. Ensure PR instructions clearly indicate how the app rename should be handled locally and on rc/prod.
- [ ] 6. Any other renames/usage i missed.

- [ ] Desktop screenshots

### How can this be tested?
For testing you need to first take pull of this branch and then run the containers like web and celery and frontend, then you need to visit the following url `http://open.odl.local:8062/news/new`, you first create the article and check if the request is being made to `http://api.open.odl.local:8065/api/v1/website_content/` url, basically, you need to test every corner of the articles CRUD and verify everything looks good. 

You should keep in mind that when article is created then celery task is run which sync the article to news, so It should run to and you will be able to see the all the articles using the following url `http://open.odl.local:8062/news`

### Additional Context
we are NOT renaming the tables, just copying them to a new table. That's fine. the ARticles table on prod only has a dozen or so entries.


website_content/api.py
```diff
@@ -0,0 +1,76 @@
...
+                log.info("Content purge request processed OK.")
+            else:
+                fastly_purge_relative_url.delay(content_url)
+                log.error("Content purge request failed, enqueued for retry.")
+        except Exception:
+            fastly_purge_relative_url.delay(content_url)
+            log.exception("Content purge request failed, enqueued for retry.")
+
+        fastly_purge_website_content_list.delay()
+    else:
```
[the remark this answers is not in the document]
ahtesham-quraish: @ChristopherChudzicki what do you think about this ?

news_events/tasks.py
```diff
@@ -48,55 +48,62 @@ def get_mitpe_events():
...
-def get_articles_news():
-    """Run the Articles News ETL pipeline"""
+def get_website_content_news():
+    """Run the website content news ETL pipeline"""
     pipelines.articles_news_etl()
     clear_views_cache()
 
 
+# Backward-compatible alias so any queued Celery tasks using the old name still work.
+get_articles_news = get_website_content_news
```
daniellefrappier18: The comment says "so any queued Celery tasks using the old name still work," but these aliases don't preserve backward compatibility. They're just Python variable references to the same task object, which is registered under the new name only. The old Celery task names (news_events.tasks.get_articles_news, news_events.tasks.sync_article_to_news) are no longer in the registry. Any messages queued under the old names before deployment would be unroutable.

Change `get_articles_news = get_website_content_news`

to 
```
@app.task(name="news_events.tasks.get_articles_news")
def get_articles_news():
    pipelines.articles_news_etl()
    clear_views_cache()
```

Change `sync_article_to_news = sync_website_content_to_news`

to
```
@app.task(
    name="news_events.tasks.sync_article_to_news",
    bind=True,
    autoretry_for=(Exception,),
    retry_kwargs={"max_retries": 3, "countdown": 5},
)
def sync_article_to_news(self, article_id: int):
    sync_website_content_to_news.apply(args=[article_id], throw=True)
```

website_content/views.py
```diff
@@ -13,103 +13,120 @@
...
-                # If draft=true, return only unpublished articles
                 qs = qs.filter(is_published=False)
-            # Otherwise, return all articles (both published and unpublished)
-            return qs
+        else:
+            qs = qs.filter(is_published=True)
+
+        content_type_param = self.request.query_params.get("content_type")
+        if content_type_param:
+            qs = qs.filter(content_type=content_type_param)
```
daniellefrappier18:
```suggestion
from website_content.constants import CONTENT_TYPE_CHOICES
valid = {k for k, _ in CONTENT_TYPE_CHOICES}
if content_type_param and content_type_param not in valid:
    raise ValidationError({"content_type": f"Must be one of {sorted(valid)}"})
```

Any string is accepted.  ?content_type=typo returns an empty 200 with no indication something is wrong. Given this is a new user-facing filter, callers will silently get zero results rather than a helpful 400.

articles/views.py
```diff
@@ -13,103 +13,120 @@
...
             OpenApiParameter(
                 name="identifier",
                 type=str,
                 location=OpenApiParameter.PATH,
-                description="Article ID (number) or slug (string)",
-                required=True,
+                description="Numeric ID or slug of the content item",
             )
         ],
-        responses={200: RichTextArticleSerializer, 404: OpenApiResponse()},
```
daniellefrappier18, line 112: The old code had responses={200: RichTextArticleSerializer, 404: OpenApiResponse()}. The new code dropped it entirely. This means the OpenAPI spec and generated TypeScript client no longer document that this endpoint can return 404.

Fix: Add it back:

@extend_schema(
    ...
    responses={200: WebsiteContentSerializer, 404: OpenApiResponse()},
)

daniellefrappier18 (commented):
I'm not sure I'm the best person to review this code but took a first pass with Claude.

news_events/etl/articles_news.py
```diff
@@ -1,17 +1,18 @@
...
 from news_events.constants import FeedType
 from news_events.etl import loaders
+from website_content.constants import CONTENT_TYPE_NEWS
+from website_content.models import WebsiteContent
 
 log = logging.getLogger(__name__)
 
 
-def extract_single_article(article: Article) -> dict:
+def extract_single_article(article: WebsiteContent) -> dict:
```
ChristopherChudzicki: `extract_single_website_content` ?

news_events/etl/articles_news.py
```diff
@@ -35,13 +36,15 @@ def transform_single_article(article_data: dict) -> dict:
     return items[0] if items else None
 
 
-def sync_single_article_to_news(article: Article):
+def sync_single_article_to_news(article: WebsiteContent):
```
ChristopherChudzicki: `sync_single_website_content_news_to_news` ?

news_events/etl/articles_news.py
```diff
@@ -35,13 +36,15 @@ def transform_single_article(article_data: dict) -> dict:
...
     """
-    Sync a single published article to the news feed (create or update FeedItem).
+    Sync a single published news content item to the news feed.
+    Only syncs content items with content_type='news'.
     """
-    # Only sync if published
     if not article.is_published:
         return
+    if article.content_type != CONTENT_TYPE_NEWS:
+        return
```
ChristopherChudzicki: **Request:** The filter here is important; we should add a test that checks it.

website_content/models.py
```diff
@@ -7,11 +7,18 @@
 
 from main.models import TimestampedModel
 from profiles.utils import article_image_upload_uri
```
ChristopherChudzicki: 1. `article_image_upload_uri` probably should be renamed.... I'd also change the upload prefix to website_content.
2. Probably also should move it to website_content.utils ... doesn't have anything to do with profiles, right?

website_content/constants.py
```diff
@@ -0,0 +1,11 @@
...
+
+GROUP_STAFF_ARTICLE_EDITORS = "article_editors"
+
+CONTENT_TYPE_NEWS = "news"
+CONTENT_TYPE_ARTICLE = "article"
+
+CONTENT_TYPE_CHOICES = [
+    (CONTENT_TYPE_NEWS, "News"),
+    (CONTENT_TYPE_ARTICLE, "Article"),
+]
```
ChristopherChudzicki: elswhere we tend to use ExtendedEnum. I'd suggest:

```python
from named_enum import ExtendedEnum

class WebsiteContentType(ExtendedEnum):
    """
    Enum for channel types.
    """

    news = "News"
    article = "Article"
```

Then you can:
- replace `CONTENT_TYPE_CHOICES` with `WebsiteContentType.as_tuple`
- replace CONTENT_TYPE_NEWS` with `WebsiteContentType.news`

website_content/models.py
```diff
@@ -26,16 +33,19 @@ class Article(TimestampedModel):
     slug = models.SlugField(max_length=255, unique=True, blank=True, null=True)
     is_published = models.BooleanField(default=False)
     publish_date = models.DateTimeField(null=True, blank=True)
+    content_type = models.CharField(
+        max_length=50,
+        choices=CONTENT_TYPE_CHOICES,
```
ChristopherChudzicki: If you use the extended enum, here's where you'd do `WebsiteContentType.as_tuple()`

website_content/serializers.py
```diff
@@ -37,14 +38,20 @@ class RichTextArticleSerializer(serializers.ModelSerializer):
     title = serializers.CharField(max_length=255)
     author_name = serializers.CharField(required=False, allow_blank=True, default="")
     user = UserSerializer(read_only=True)
+    content_type = serializers.ChoiceField(
+        choices=["news", "article"],
```
ChristopherChudzicki: You should be able to remove `choices` and it will get picked up automatically from the model.

If it doesn't, we should use the constant here.

website_content/serializers.py
```diff
@@ -0,0 +1,72 @@
...
+    """
+    Serializer for WebsiteContent model.
+    """
+
+    created_on = serializers.DateTimeField(read_only=True, required=False)
+    updated_on = serializers.DateTimeField(read_only=True, required=False)
+    publish_date = serializers.DateTimeField(read_only=True, required=False)
+    content = serializers.JSONField(default={})
+    slug = serializers.SlugField(max_length=60, required=False, allow_blank=True)
+    title = serializers.CharField(max_length=255)
```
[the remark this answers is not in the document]
ChristopherChudzicki: @ahtesham-quraish Copilot is correct about this.

website_content/tasks_test.py
```diff
@@ -0,0 +1,95 @@
...
+
+        result = call_fastly_purge_api("/api/v1/news/test/")
+
+        assert result == {"status": "ok", "skipped": True}
+        mock_request.assert_not_called()
+
+
+@pytest.mark.django_db
+class TestFastlyPurgeWebsiteContentList:
+    """Tests for fastly_purge_website_content_list task"""
```
ChristopherChudzicki: this test class doesn't actually call fastly_purge_website_content_list . What's it testing exactly?

Seems like all it's testing is that `main.call_fastly_purge_api` makes a reasonable request? Not sure it's adding any value here.

website_content/api.py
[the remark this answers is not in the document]
ChristopherChudzicki: Copilot's comment makes sense to me. We should only purge news if publishing news

articles/api_test.py
```diff
@@ -1,187 +0,0 @@
...
-    article_published_actions(article=article)
-
-    # Verify logging
-    assert "Triggering article_published plugins for article" in caplog.text
-    assert f"id={article.id}" in caplog.text
-    assert f"title={article.title}" in caplog.text
-
-
-@pytest.mark.django_db
-class TestPurgeArticleOnSave:
```
ChristopherChudzicki, line 104: I don't think these tests got ported to the new app, did they? (Please let me know if I am wrong!)

I think we **should** port them, but not quite as is. Currently they kind of just assert "x called y". 

My suggestion is a new test that asserts on `call_fastly_purge_api` (or request itself) to depend on less of the plumbing. I think that should work fine since our tests use CELERY_ALWAYS_EAGER=True.

And whenever I test something **Doesnt** happy, I try to tie it very explicitly to the thing actually happening, so I like to parametrize in that case
```python
@pytest.mark.parametrize(
    "slug, is_published, expect_purge",
    [
        ("test-article", True, True),
        ("test-article", False, False), # I don't think this can happen now, but makes sense to test
                                        # (i.e., i think we ONLY have slugs after publishing and can't unpublish)
        ("", True, False),          # This also doesn't make much sense, but doesn't hurt, right?
        ("", False, False),
    ],
)
def test_purge_content_on_save(slug, is_published, expect_purge):
    content = WebsiteContentFactory(is_published=False, slug=slug)
    purge_content_on_save(article)
    # Now Assert fastly purged if expected, otherwise assert it was not purged
```

Feel free to bring this back another way, but I do think we should bring it back somehow.

ChristopherChudzicki (changes requested):
This is looking pretty good. I think there are a few renames that got missed, and I left some other comments.

I also think it would be worth updating the PR instructions or noting somewhere that we are NOT renaming the tables, just copying them to a new table. That's fine. the ARticles table on prod only has a dozen or so entries.

news_events/etl/articles_news.py
ahtesham-quraish: addressed

news_events/etl/articles_news.py
ahtesham-quraish: addressed

news_events/etl/articles_news.py
ahtesham-quraish: Done

website_content/models.py
ahtesham-quraish: Addressed

website_content/constants.py
ahtesham-quraish: Done

website_content/models.py
ahtesham-quraish: Done

website_content/serializers.py
ahtesham-quraish: Can you please check it ?

website_content/tasks_test.py
ahtesham-quraish: Removed the tests

website_content/api.py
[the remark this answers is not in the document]
ahtesham-quraish: @ChristopherChudzicki can you please check ?

ChristopherChudzicki (approved):
Looks good 👍 

As mentioned in slack, let's make sure we have some non-admin accounts on RC/prod that do have `article_editors` group membership so we can test that migration.

website_content/views.py
ChristopherChudzicki: @ahtesham-quraish Sorry i missed this in first review. 

We should use a `django_filters` FilterSet to do this filtering, for three reasons:
1. It will add the valiation that @daniellefrappier18 mentioned above
2. It's what the rest of the project uses for filtering
3. It will get us the OpenAPI spec auatomatically, no need for the manual `OpenApiParameter` decorator you added above.

ChristopherChudzicki (changes requested):
Sorry, did come across one other requested change when starting to review the frontend branch https://github.com/mitodl/mit-learn/pull/3332/changes#r3262947501

ahtesham-quraish:
> Sorry, did come across one other requested change when starting to review the frontend branch https://github.com/mitodl/mit-learn/pull/3332/changes#r3262947501

Can you please check now?

ChristopherChudzicki (approved):
👍

PR #3337, opened 2026-05-13
fix open textbooks

abeglova:
None

I changed the content_category for ocw learning materials to singular from plural to match the rest of the content categories but i introduced a bug that prevents  open textbooks from being promoted to ocw learning materials resources since the ocw category is plural

### How can this be tested?
Run 
set CREATE_OCW_LEARNING_MATERIALS=True
docker-compose run web ./manage.py backpopulate_ocw_data --course-name res-2-008-thermodynamics-and-climate-change-summer-2020

Go to http://open.odl.local:8062/search?resource_type_group=learning_material&show_ocw_files=true
You should see "Open Textbook" as a resource type

If you have created ocw learning materials before the regression and have "Open Textbooks" (plural) learning materials in the search index in addition to "Open Textbook" you can get rid of them by running
docker-compose run web ./manage.py  update_ocw_learning_material_resources


mbertrand (approved):
👍

PR #3344, opened 2026-05-14
refactor: make internal editor code more generic

ahtesham-quraish:
https://github.com/mitodl/hq/issues/11232

## Step 2: Frontend editor refactor (Second PR)

Make `page-components/TiptapEditor/` content-type-agnostic. Success looks like:
- [ ] 1. The editor shell, save/publish/draft UI, image upload, schema validation, learning-resource drawer, and toolbar plumbing all live in one place and accept per-type configuration (schema, initial doc, route prefix, theme, extension list).
- [ ] 2. The current news editor becomes a thin composition of the generic shell + a news config module. **No visible behavior changes to `/news` pages.**
- [ ] 3. a new UI under `/articles` that mirrors `/news` for now, with minimal themeing changes. **More changes to come, but start similar.**

- [ ] Desktop screenshots

### How can this be tested?
For testing you need to first take pull of this branch and then run the containers like web and celery and frontend, then you need to visit the following url `http://open.odl.local:8062/news/new`, you first create the article and check if the request is being made to `http://api.open.odl.local:8065/api/v1/website_content/` url, basically, you need to test every corner of the articles CRUD and verify everything looks good. 

You should keep in mind that when article is created then celery task is run which sync the article to news, so It should run to and you will be able to see the all the articles using the following urls `http://open.odl.local:8062/news`
`http://open.odl.local:8062/website_content/article/new`


ChristopherChudzicki (changes requested):
@ahtesham-quraish I haven't actually tried creation/editing flows on this branch yet, but some feedback on the code is below.

### TipTap Editor Organization
Overall, I think we need to strive to better separate news + article. Currently, we've created the `/TiptapEditor/contentTypes` directory with article and news, but:
- `newsExtensions` = everything
- `articleExtensions` = just alias of news

We should think about how extensions group:

| Tier | Criterion | Where | Examples |
|---|---|---|---|
| **1** | Every content type needs it (now & likely future) **and** no per-type config | `extensions/baseExtensions.ts` — one factory | `StarterKit`, `Placeholder`, `Heading`, `Image*`, `Divider`, `MediaEmbed*`, `LearningResource*`, … |
| **2** | Shared, **but** takes per-type config | `extensions/node/*`, composed into each type's list | `Banner` (breadcrumb, background, dark/light), `Byline` (share-URL prefix — hardcoded `/news/` today, `ArticleByLineInfoBar.tsx:120`) |
| **3** | Specific to one content type | `contentTypes/<type>/` | `Document` expression, seed doc, the per-type config object |

Tier-1 test: constructible from only the `(uploadHandler, setUploadError)` runtime deps, no type-specific arg. Don't pre-split it for hypothetical future exclusions. Net effect: each `create<Type>Extensions` collapses to `[ <Type>Document, ...createBaseExtensions(deps), Banner, Byline ]` + seed + config.

All the generic extensions should go somewhere common, something like, maybe inside `/TiptapEditor/extensions/baseExtensions.ts`.

Other notes:
1. `ArticleByLineInfoBar` this should be renamed; its name dates from the olden days when we called news pages were called articles. I think it could just be `ByLineInfoBar`? 
2. Re `GenericEditor`: Minor, but Copilot suggested calling this  `WebsiteContentEditor`... it's not truly generic, it is specific to website content. So i think that's a good name.
3. `ArticleProvider` / `useArticle` look like naming debt; `WebsiteContentProvider` / `useWebsiteContent` would better match the actual `WebsiteContent` payload.

### Regarding NextJS Routes

You added a bunch of new NextJS routes... e.g., `/articles`, `/articles/new`, etc. If we continue with that pattern, every time we add a new `content_type`, we will need to add a whole bunch of new route files.

I see two groups of routes:
- **User Facing:** `/news`, `/news/:id`, `/articles`, `/articles/:id` ... these are user-facing; adding new route files seems reasonabl
- **Internal for editing:** the `/new`, `/[id]/edit`, `/drafts` routes.

I think we should make the editing routes generic:  `/website_content/[type]/new`, `/website_content/[type]/[id-or-slug]/edit`, `/website_content/drafts?content_type=whatever` routes for editing call content types.

That way, when we add new content types, we don't need to add new editing routes.


### An issue with redirect.

Did you try the new `/articles`, `/articles/new`, etc routes that you added? These currently redirect to `/news`, due to our nextjs config. (From https://github.com/mitodl/hq/issues/10324)

The website_content routes suggested above avoid the issue, at least for editing URLs.

We'll need to check in with Ferdi about whether or not we can remove the redirect. My recollection is that we added it because things were published at the old `/articles/` URL before we were expecting. I'll check data tomorrow. It could be that if no one is visiting that page, we can get rid of it.

ahtesham-quraish:
@ChristopherChudzicki I have applied the styles for the articles page you should check it too please.

image

ChristopherChudzicki (changes requested):
@ahtesham-quraish The extension separation is much cleaner / clearer now. Thanks!

I didn't get too far with manually testing the flows, but I am seeing some code issues still. I can revisit manual testing first thing in the morning.

1. `next.config.js` redirects ... We can remove these. Per https://github.com/mitodl/hq/issues/11365, we are ok to remove them, and will coordiante with Devops to remove the fastly redirects, too.
2. **Naming:** Lots of things still reference article where they shouldn't:
    - **`ArticleContext.tsx`**: This file should be WebsiteContentContext.tsx, right? Let's do a real rename rather than alias + deprecated flag. Just rename it. It's simpler. Diff is a little larger, but less confusion. 
    - Same comment for everything in the `TiptapEditor/extensions/node/ArticleByLineInfoBar` directory.
    - **`GenericEditor`**: Similarly, statements like `import { WebsiteContentEditor } from "../../core/GenericEditor"` are confusing. Don't deprecate + alias. Just rename it.
        - Doubly so for this file. It's completely new.
    - **`ArticleEditor.tsx`**: There are two ArticleEditor.tsx files. This is confusing. 
        - As far as I can tell, `TiptapEditor/ArticleEditor.tsx` is really the News editor, and exists only for the sake of its sibling, `ArticleEditor.happydom.test.tsx` which, similarly, is really testing the newseditor.
            - At the very least, let's get rid of `TiptapEditor/ArticleEditor.tsx` and move the happydom test to be a sibling of NewsEditor.
    - **`TiptapEditor/useArticleSchema.ts`**: This seems unused. Just get rid of it?
3. ByLine ... I'm noticing the the articles byline shows a Share URL pointing to news, see screenshot below. It doesn't seem to be visible in the published url, just on the editing view? We should have the correct url though.
4. **Editing URLS**: Good that /articles/new, etc, are thin wrappers of the website content versiosn. IMO, thogh, we should just drop `/articles/new` and `/news/new` entirely. May want to confirm with @Ferdi , though.

Image

ChristopherChudzicki:
@ahtesham-quraish One more note: I think this branch needs a rebase against your other branch, or against main after that is merged.

that should fix the failing openapi thing.

ChristopherChudzicki (changes requested):
I'm getting lost lately with so many comments, so I'm going to put mine all in one comment, hopefully with enough context / links.

--------------

As discussed on zoom, still seeing a lot of naming issues. I know you're starting to address some of them already. Also a few other bugs.

**Re Naming:** We can't use `article` where we mean generic website content or news. `article` is specifically the `content_type="article"` from now one. This should apply everywhere. Otherwise, we end up with weird things like `app-pages/Articles` (which is really news) and `app-pages/UserArticles` which is really articles.

### Naming
1. **Hooks**: discussed on zoom; `frontends/api/src/hooks/articles` should be `website_content` now. I don't think they ever need to mention article or news explicitly.
2. **`app-pages/Articles` and `**app-pages/UserArticles`** These are abot "news" and "article" content types, respectively, right? They should be named that way. Additionally, I think both of these directories should have much fewer files (See next item).
3. **Suggested File Structure:**  This PR has separate ArticleDetailPage and NewsDetailPage files (named differently, per Item 2 above). But they are nearly identical: They just render the editor in read-only mode. And that should continue into the future. Any visual difference in the pages should be handled by the tiptap rendering. So we should be able to get rid of both ArticleDetailPage and NewsDetailPage in favor of WebsiteContentDetail:
    ```
    app-pages/articles/
        ArticleListingPage.tsx
    app-pages/news/
        NewsListingPage.tsx
    app-pages/website_content/
        WebsiteContentDetail.tsx                (suggested as new, to replace
                                                ArticleDetailPage, etc)
        WebsiteContentDraftListingPage.ts       (exists)
        WebsiteContentEditPage.tsx              (exists)
        WebsiteContentNewPage.tsx               (exists)
        + test files
    ```

    IMO Makes sense to keep the two separate route directories `app/news/...` and `app/articles...`, though. 
4. **Dead Code:** `ArticleNewPage`, `ArticleEditPage`, `UserArticleEditPage`, `UserArticleNewPage` ... aren't these all dead code now that `WebsiteContentEditPage` etc exist?
    - :warning: `ArticleNewPage.test.tsx` (existing) should probably be converted into `WebsiteContentNewPage.test.tsx` before deleting the former.
5. **urls.ts**: This includes things like `ARTICLES_LISTING=/news`. Could you double check this file?
6. **`TiptapEditor/index.ts:2`** exports `ArticleEditor as UserArticleEditor` but nothing uses this. Let's delete the export.
7. **Styling Comment:** Copilot's comment https://github.com/mitodl/mit-learn/pull/3344#discussion_r3272860496 is accurate and worth fixing; besides the fragile selector, should replace `abc` with a useful class name.
8. :warning::warning: **Can't create articles**: When try creating a new article at http://open.odl.local:8062/website_content/article/new, the `content_type` is `"news"`.
9. **Listing Page Filter:** See https://github.com/mitodl/mit-learn/pull/3344#discussion_r3272860655
10. **byLineViewer type:** See https://github.com/mitodl/mit-learn/pull/3344#discussion_r3272860440
11. **Remove Dead Props:** ArticleEditorProps in `TiptapEditor/contentTypes/article/ArticleEditor.tsx` marks several things as deprecated, but they are never used. Just delete them. Same for `/contentTypes/news/NewsEditor.tsx`
12. **Remove Deprecated ArticleByLineInBannerViewer**: At https://github.com/mitodl/mit-learn/pull/3344/changes#diff-189cba8943939cc1aa7937abb3d5f8f9e592f1777749d828f954f2abe05e0c04R43 There's only one callsite, just remove the deprecated thing.
13. **Permission.ArticleEditor:** Either rename this and update the backend serializer, or if you want to keep backend untouched for this PR, at least leave a comment that the permission is actually for all website_content; I'd be ok with teither for now.


> [!NOTE]
> Would suggest doing a few copilot local reviews making sure everything is addressed... a few of Copilot's relevant comments were not addressed, so a local copilot review helps make sure.

ChristopherChudzicki:
> 7. **Styling Comment:** Copilot's comment https://github.com/mitodl/mit-learn/pull/3344#discussion_r3272860496 is accurate and worth fixing; besides the fragile selector, should replace `abc` with a useful class name.
> 8. ... fixed ✅ 
> 9. ... fixed ✅ 
> 10. **byLineViewer type:** See https://github.com/mitodl/mit-learn/pull/3344#discussion_r3272860440

I don't think these got addressed.

I had copilot run a quick review locally, and it caught some more things:

- Lingering `UserArticleCard` in ArticleListingPage, plus `USER_ARTICLES_*  constants in `urls.ts`. 
    - note: These can just be article now, right? You renamed the directory, but still using "User Article", which isn't a great name since most users can't create these things. 
- common/articleUtils.ts still caries article name, but it's website_content agnostic now
- `ArticleViewer.test.tsx` file + `describe("ArticleViewer")` block still article-named while testing NewsEditor.

frontends/main/src/app-pages/UserArticles/UserArticleListingPage.tsx
```diff
@@ -0,0 +1,177 @@
...
+      </Card.Footer>
+    </ArticleCardWrapper>
+  )
+}
+
+const UserArticleListingPage: React.FC = () => {
+  const [page, setPage] = useState(1)
+  const scrollRef = useRef<HTMLDivElement>(null)
+
+  const { data: articles, isLoading } = useArticleList({
```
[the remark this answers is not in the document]
ahtesham-quraish: This has been refactored and this file is removed

frontends/main/src/app-pages/WebsiteContent/WebsiteContentDetail.tsx
```diff
@@ -20,14 +21,15 @@ const Spinner = styled(LoadingSpinner)({
...
-export const ArticleDetailPage = ({
+const WebsiteContentDetail = ({
   articleId,
   learningResourceIds = [],
 }: {
   articleId: string
   learningResourceIds?: number[]
 }) => {
-  const { data: article, isLoading } = useArticleDetailRetrieve(articleId)
+  const { data: article, isLoading } =
```
ChristopherChudzicki: This should be `content`, right? Not `article`. Same for the prop `articleId` above... should be `contentId`.

frontends/main/src/app-pages/WebsiteContent/WebsiteContentDetail.tsx
```diff
@@ -43,8 +45,14 @@ export const ArticleDetailPage = ({
   return (
     <PageContainer>
       <LearningResourceProvider resourceIds={learningResourceIds}>
-        <ArticleEditor article={article} readOnly />
+        {article.content_type === "article" ? (
+          <ArticleEditor article={article} readOnly />
+        ) : (
+          <NewsEditor newsItem={article} readOnly />
+        )}
```
ChristopherChudzicki: We will be adding more content types, so a dictionary lookup would be more scalable here than conditionals.

Alternatively, rename WebsiteContentEditor to BaseWebsiteContentEditor and have WebsiteContentEditor itself switch based on content_type.

frontends/main/src/app/(site)/articles/[slugOrId]/page.tsx
```diff
@@ -0,0 +1,96 @@
...
+
+  const queryClient = getQueryClient()
+
+  return safeGenerateMetadata(async () => {
+    const article = await queryClient.fetchQuery(
+      websiteContentQueries.websiteContentDetailRetrieve(slugOrId),
+    )
+
+    const description = extractArticleDescription(article)
+    const leadImage = extractImageMetadata(article)
```
Comment (ChristopherChudzicki) [inline frontends/main/src/app/(site)/articles/[slugOrId]/page.tsx]: `extractImageMetadata` is copied between both article/ and /news pages. I think we should put this in `@/common/website_content.ts`. You could do the same for `extractArticleDescription`, though that more conceivably differs between content types (even though it does not right now).

Additionally, I'd suggest renaming `article` to content, and adding a check
```ts
if (content.content_type != "article") notFound()
```
and a similar check on the news page.

frontends/main/src/page-components/TiptapEditor/contentTypes/article/ArticleEditor.tsx
```diff
@@ -0,0 +1,95 @@
...
+        href="/website_content/drafts?content_type=article"
+        size="small"
+      >
+        Drafts
+      </ButtonLink>
+      <ButtonLink variant="primary" href={editUrl} size="small">
+        Edit
+      </ButtonLink>
+    </>
+  ) : null
```
ChristopherChudzicki: `toolbarSlot` is a function purely of `content_type`. 

If you give `WebsiteContentEditor` a `content_type: WebsiteContentContentTypeEnum` prop, you could construct it there very easily rather than having both ArticleEditor and NewsEditor construct it.

  - "Drafts" link → `websiteContentDraftsView(contentType)`
  - "Edit" link → `websiteContentEditView(contentType, slugOrId)`

frontends/main/src/page-components/TiptapEditor/TiptapEditor.tsx
```diff
@@ -82,6 +82,9 @@ const Container = styled.div<{
         },
       }
     : {}),
+  "&& .tiptap.ProseMirror.tiptap-viewer > :nth-child(2)": {
+    paddingTop: "36px",
+  },
```
ChristopherChudzicki: ⚠️ This is a visual regression on News pages:

Image

IMO:
- if we actually need this, it should be scoped to articles only.
- But maybe we can just remove it ? (Do we have designs for this page?)

ChristopherChudzicki (changes requested):
This is looking much cleaner. Some comments below, including one styling regression for published news articles.

There is also some complication around the banner / byline difference sbetween news + article that I hadn't appreciated, and which I'll comment about in a followup.

1. http://open.odl.local:8062/website_content/drafts/ with no query param seems to show "Articles" title but display all website content.
    - I think the simplest thing for now would be to make the default behave like content_type=news. What do you think?
2. A few lingering renames
    - WebsiteContentContext still has its field named article?: WebsiteContent, and the core WebsiteContentEditor prop is still article
  (NewsEditor renamed its own prop but passes article={newsItem}).
    - Some lingering UserArticle references (in comments)
    - WebsiteContentDetail uses `articleId` where it should use `contentId`. Some internal article naming to WebsiteContentDetail, see https://github.com/mitodl/mit-learn/pull/3344/changes#r3289665347
3. `/articles/` should be read-only, like "news", right?
    - ⚠️  Per discussion in slack, I lean toward just getting rid of this route for now, until we get designs from Bilal.
4. https://github.com/mitodl/mit-learn/pull/3344/changes#r3289681768
5. Metadata... Let's share extractImageMetadata + add a content_type check https://github.com/mitodl/mit-learn/pull/3344/changes#r3289724607
6. :warning: Styling regression vs main, see https://github.com/mitodl/mit-learn/pull/3344/changes#r3290090895
7. Banner breadcrumb text, see https://github.com/mitodl/mit-learn/pull/3344/changes#r3290389595
8. **Please see https://github.com/mitodl/mit-learn/pull/3344#issuecomment-4522454404**. 

### New Routes
The new routes make sense to me, but I think we should (A) remove `/articles/new`, `/articles/draft`, and `/articles/[slugOrId]/edit` and (B) double check with Ferdi if we actually want `/articles/` route at all yet / if it should be feature flagged.
```
/website_content/[type]/new
/website_content/[type]/[idOrSlug]/edit
/website_content/drafts?content_type=...

NEW ARTICLE ROUTES:
PUBLIC
    /articles                     <-- Let's check if we want this
    /articles/[slugOrId]
PRIVATE:
    /articles/[slugOrId]/draft
    /articles/new                 <-- suggest removing
    /articles/draft               <-- suggest removing
    /articles/[slugOrId]/edit     <-- suggest removing

EXISTING NEWS ROUTES
PUBLIC
    /news
    /news/[slugOrId]
PRIVATE
/news/[slugOrId]/draft
    /news/new                 <-- unnecessary, but redirects for backward compat
    /news/draft               <-- unnecessary, but redirects for backward compat
    /news/[slugOrId]/edit     <-- unnecessary, but redirects for backward compat
```

frontends/main/src/page-components/TiptapEditor/extensions/node/Banner/ArticleBannerNode.tsx
```diff
@@ -0,0 +1,215 @@
...
+    <ReactNodeViewContentProvider content={children}>
+      <StyledNodeViewWrapper as="div">
+        <>
+          <BreadcrumbBar>
+            <BreadcrumContainer className="breadcrum">
+              <Breadcrumbs
+                variant="dark"
+                ancestors={[
+                  { href: "/", label: "Home" },
+                  { href: "/articles", label: "MIT Learn Articles" },
```
ChristopherChudzicki: Just saw discussion in slack... This should be just "Articles" not "MIT Learn Articles"

ChristopherChudzicki:
@ahtesham-quraish I took a closer look at the banner and byline nodes today. Details below, along with an example commit I think you should apply to this branch. (But please read this and the commit carefully to see if you agree.)

----

The byline is more different than it needs to be between article and news: News has a single render path for the byline, article has two, which leads to some bugs. More broadly, the article editor has differences between read-only and editing mode that it should not have.

Here's a more detailed summary:

- **News**:
    - banner node renders BannerWrapper
    - byline node renders ByLineInfoBar (which includes metadata from the article as a whole)
- **Articles:** For articles, the behavior is divergent between editing mode and read-only mode:
    - The banner node renders as:
        - Editing mode: ArticleBannerWrapper
        - read-only mode: a differently-implemented ArticleBannerViewer, which includes the byline
    - The byline node renders as:
        - Editing Mode: ByLineInfoBar
        - Read-only: Null

This leads to divergence between the editing and read-only views for articles. Concretely:
1. visual differences (breadcrumbs are different + byline is different)
2. functional differences (the ArticleBannerViewer didn't implement the share button)

We should unify the two views:
- Both editing and read-only should render ArticleBannerWrapper
    - technically editing vs viewing still wraps differently (`ReactNodeViewRenderer` vs `ReactNodeViewContentProvider` wrapper) but our custom UI is the same.
- Both editing and read-only render ByLineInfoBar, same for news vs article
    - account for different styles by CSS

I took this approach in https://github.com/mitodl/mit-learn/commit/35f6bf906fa6eeb75b1d581bff6fead7605f4a5b (branch `ahtesham/tiptap-front-branch-cc-suggestions`)

### Styling
You'll see now "news" and "article" use exactly the same ByLineInfoBar node. They are just styled slightly differently via a classname on WebsiteContentEditor. In general, where UI differs between article vs news vs other content types but is structurally the same (author - reading time - date - share) we should account for the differences just with CSS. (*Aside: I would say that if they are structurally very similar, but slightly different, then we should at least raise with designers (Bilal) whether they can be made structurally identical.*)

Additionally, you'll see in the commit above that all styling differences for ByLine between article and news live in ArticleEditor.tsx. This is better than putting them at the node level...As we add more content types, individual nodes should not need to know about the content type, at least if the differences are style.

**Note:** I also dropped the backgroundColor prop on WebsiteContentEditor in the commit above. No need for a prop, can just be css via styled(...).

frontends/main/src/app-pages/WebsiteContent/WebsiteContentDetail.tsx
ahtesham-quraish: Done

frontends/main/src/app/(site)/articles/[slugOrId]/page.tsx
Comment (ahtesham-quraish) [reply frontends/main/src/app/(site)/articles/[slugOrId]/page.tsx]: Addressed

frontends/main/src/page-components/TiptapEditor/TiptapEditor.tsx
ahtesham-quraish: Addressed

frontends/main/src/page-components/TiptapEditor/extensions/node/Banner/ArticleBannerNode.tsx
ahtesham-quraish: Addressed

ahtesham-quraish: Addressed

frontends/main/src/app-pages/WebsiteContent/WebsiteContentDetail.tsx
ahtesham-quraish: addressed

frontends/main/src/app-pages/WebsiteContent/WebsiteContentDetail.tsx
ahtesham-quraish: addressed

frontends/main/src/page-components/TiptapEditor/TiptapEditor.tsx
ahtesham-quraish: addressed

ChristopherChudzicki (changes requested):
I think we are very close to merging this. I still see a few unaddressed issues and some dead code/props:

1. `WebsiteContentEditor` has two styling props that should just be CSS customizable via ArticleEditor or NewsEditor:
    - `backgroundColor`: This is unnecessary now that `ArticleEditor` sets background color in `StyledWebsiteContentEditor`
    - `applyViewerTopSpacing`: It would be better to handle this in `ArticleEditor`'s `StyledWebsiteContentEditor`. If necessary, add a well-named CSS class to the relevant child node so it can be targeted by CSS. Other notes here:

        - I think this only affects viewing mode? Maybe we don't want it at all?
        
        - :warning: **Please do not use `nth-child(2)`**.  GitHub Copilot's automated review flagged this selector as brittle, and I've raised it once or twice, too. Let's not eintroduce it. (Github also made a bad suggestion at one point to use a class `.abc`; use a well-named class if we need this at all).

            You can see it's brittle because I am pretty sure it's targeting a different element than you intend:
            Screenshot 2026-05-26 at 8 19 49 AM

2. Also on `WebsiteContentEditor`, the `bylineViewer` prop is unused. It should be deleted.
3. Banner breadcrumb text, see https://github.com/mitodl/mit-learn/pull/3344/changes#r3290389595; Still not addressed
4. Breadcrum Color (New): Apologies for not catching this one in earlier reviews. The banner breadcrumb color is wrong for the article title itself, "Lorem Ipsum Whatever" in the screenshot below. Currently, `frontends/main/src/page-components/TiptapEditor/extensions/node/Banner/ArticleBannerNode.tsx` is using `variant="dark"`; that's intended for breadcrumbs on a dark background (like news pages). The article pages should use `variant="light"`.

    Note: This is an a11y contrast issue on a public page. There are also some contrast issues in editing mode, but those are not as critical, since that's not a public page.

    Screenshot 2026-05-26 at 8 15 53 AM

5. Article breadcrumbs still say "MIT Learn Articles"

frontends/main/src/page-components/TiptapEditor/extensions/node/Banner/ArticleBannerNode.tsx
ChristopherChudzicki: I don't think this was addressed. It's still there: https://github.com/mitodl/mit-learn/pull/3344/changes#diff-ba9b9ab6c55b2f2070fe3bec9567350b5398ca99a8fcfddfe667b5b4a1760057R140

ahtesham-quraish:
I have addressed the feedback can you take a look again at this?

ChristopherChudzicki (approved):
👍 Looks good.

Only request is `bylineViewer` prop from `TipTapViewer`. Nothing uses it, and nothing should. No need for a re-review

frontends/main/src/app-pages/WebsiteContent/WebsiteContentNewPage.tsx
```diff
@@ -0,0 +1,73 @@
...
+
+const WebsiteContentNewPage: React.FC<WebsiteContentNewPageProps> = ({
+  type,
+}) => {
+  const router = useRouter()
+  const Editor = EDITORS[type]
+  const viewUrl = PUBLISHED_VIEW_URL[type]
+
+  if (!Editor || !viewUrl) {
+    notFound()
```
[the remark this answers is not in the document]
ChristopherChudzicki: BTW, this comment is wrong; `notFound` in a client component works fine, it's just a. "soft 404" not real 404 html response.

PR #3345, opened 2026-05-15
fix: link the podcast episode to its detail page and fix the video count on collection page

ahtesham-quraish:
n/a

- Podcast episode having hover and it should link to its detail page instead of playing it
- The video count on video collection page should be total of all the video count which are part of that particular playlist
- Some spaces issues are fixed which happens in case there is single episode in podcast or no episode at all

https://github.com/user-attachments/assets/cb45c494-81a3-4bbc-b3d3-1836c19986b7

### How can this be tested?
if you don't have playlist data locally then in frontend env file you should add the following variable `NEXT_PUBLIC_MITOL_API_BASE_URL="https://api.rc.learn.mit.edu"` it will connect with RC learn backend
then we need to enable the flag which is `podcast-detail-page` and then visit the following url `http://open.odl.local:8062/podcast/detail/15261?podcast=15260`

### DIFF
--- a/frontends/main/src/app-pages/PodcastPage/PodcastDetailPage.tsx
+++ b/frontends/main/src/app-pages/PodcastPage/PodcastDetailPage.tsx
@@ -122,8 +122,10 @@ const HeaderTextContent = styled.div({
 
 /* ── Episodes list ── */
 
-const EpisodesSection = styled.div(({ theme }) => ({
-  padding: "0 48px",
+const EpisodesSection = styled("div", {
+  shouldForwardProp: (prop) => prop !== "hasMoreEpisodes",
+})<{ hasMoreEpisodes?: boolean }>(({ theme, hasMoreEpisodes }) => ({
+  padding: hasMoreEpisodes ? "0 48px" : "0 48px 40px 48px",
   [theme.breakpoints.down("sm")]: {
     padding: "0 0 48px",
   },
@@ -150,7 +152,6 @@ const EpisodesHeading = styled(Typography)(({ theme }) => ({
 }))
 
 const EpisodeList = styled.div({
-  listStyle: "none",
   margin: 0,
   padding: 0,
   display: "grid",
@@ -183,6 +184,10 @@ const EpisodeRow = styled("div", {
   "&:hover": {
     backgroundColor: theme.custom.colors.lightGray1,
     cursor: "pointer",
+  },
+  "&:focus-visible": {
+    outline: `2px solid ${theme.custom.colors.red}`,
+    outlineOffset: "-2px",
   },
   "&:hover .episode-title, &:focus-visible .episode-title": {
     color: theme.custom.colors.red,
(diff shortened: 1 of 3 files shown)


frontends/main/src/app-pages/PodcastPage/PodcastDetailPage.tsx
```diff
@@ -340,7 +354,15 @@ export const EpisodeItem: React.FC<EpisodeItemProps> = ({
...
     <EpisodeRow
-      onClick={() => (isPlaying ? onPauseClick?.() : onPlayClick(episode))}
+      onClick={handleRowNavigate}
+      onKeyDown={(e) => {
+        if (e.key === "Enter" || e.key === " ") {
+          e.preventDefault()
+          handleRowNavigate()
+        }
+      }}
+      role="link"
```
daniellefrappier18: role="link" semantically implies an href attribute (like an anchor). Screen readers will announce it as a link but there's no associated URL in the accessible tree, and AT-users can't copy/open-in-new-tab.

Instead render EpisodeRow as an <a> tag passing href directly, and remove the manual role, tabIndex, onClick, and onKeyDown, the browser handles all of that natively.

frontends/main/src/app-pages/PodcastPage/PodcastDetailPage.tsx
```diff
@@ -566,22 +597,27 @@ export const PodcastDetailPage: React.FC<PodcastDetailPageProps> = ({
...
-                    key={episode.id}
-                    episode={episode}
-                    onPlayClick={handlePlayClick}
-                    onPauseClick={() => playerRef.current?.pause()}
-                    isPlaying={
-                      playingEpisode?.id === episode.id && isAudioPlaying
-                    }
-                    isPlayable={Boolean(getEpisodeAudioUrl(episode))}
-                  />
+                  <div key={episode.id} role="listitem">
```
daniellefrappier18: EpisodeList is now a div role="list", and each item is wrapped in `<div role="listitem">` containing `<EpisodeItem>` which itself renders EpisodeRow (another div). This adds an extra DOM node with no purpose. 

if you change:
```
{episodes.map((episode) => (
  <div key={episode.id} role="listitem">
    <EpisodeItem
      episode={episode}
      href={...}
      ...
    />
  </div>
))}
```

to 
```
{episodes.map((episode) => (
  <EpisodeItem
    key={episode.id}
    episode={episode}
    href={...}
    ...
  />
))}
```

and add `role="listitem"` to `EpisodeRow` that eliminates the extra div wrapper with no semantic purpose

daniellefrappier18 (changes requested):
This looks good but I think we should fix the role="link" on a div without href which is an invalid ARIA pattern before merging

frontends/main/src/app-pages/PodcastPage/PodcastDetailPage.tsx
ahtesham-quraish: Addressed

frontends/main/src/app-pages/PodcastPage/PodcastDetailPage.tsx
ahtesham-quraish: Addressed

daniellefrappier18 (approved):
LGTM

PR #3351, opened 2026-05-18
Granian reload for local dev

rhysyngsun:

N/A

This enables granian's reload feature for localdev.

### How can this be tested?

Run the app, edit a pytjon source file, you should see it reload with this log message:
```
[INFO] Changes detected, reloading workers..
```


gumaerc (approved):
👍

PR #3359, opened 2026-05-20
Fix eslint config to ignore node_modules

rhysyngsun:

N/A

Fixes eslint config so pre-commit doesn't traverse and load tsconfigs from node_modules.

### How can this be tested?

You should be able to run `pre-commit run eslint -a` and it should pass and run much faster.

### DIFF
--- a/frontends/.eslintrc.js
+++ b/frontends/.eslintrc.js
@@ -19,7 +19,7 @@ module.exports = {
   settings: {
     "import/resolver": {
       typescript: {
-    project: ["tsconfig.json", "*/tsconfig.json"]
+        project: ["tsconfig.json", "*/tsconfig.json"],
       },
     },
     "jsx-a11y": {
--- a/frontends/.eslintrc.js
+++ b/frontends/.eslintrc.js
@@ -19,7 +19,7 @@ module.exports = {
   settings: {
     "import/resolver": {
       typescript: {
-        project: "**/tsconfig.json",
+        project: "*/tsconfig.json",
       },
     },
     "jsx-a11y": {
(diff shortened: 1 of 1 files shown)


frontends/.eslintrc.js
```diff
@@ -19,7 +19,7 @@ module.exports = {
   settings: {
     "import/resolver": {
       typescript: {
-        project: "**/tsconfig.json",
+        project: "*/tsconfig.json",
```
ChristopherChudzicki:
```suggestion
        project: ["tsconfig.json", "*/tsconfig.json"]
```

I was surprised that
```sh
pre-commit run eslint --files  frontends/jest-shared-setup.ts
```
works on your branch. It does, but it uses the wrong tsconfig. Apparently ESLint just falls back to trying to t check the file with some other tsconfig it found.

If you run:
```sh
DEBUG='eslint-import-resolver-typescript' yarn workspace frontends exec -- eslint jest-shared-setup.ts 2>&1
```
then you'll see lines like:
>  eslint-import-resolver-typescript no tsconfig matched /Users/cchudzicki/dev/mit-learn/frontends/jest-shared-setup.ts with /Users/cchudzicki/dev/mit-learn/frontends/api/tsconfig.json /Users/cchudzicki/dev/mit-learn/frontends/main/tsconfig.json /Users/cchudzicki/dev/mit-learn/frontends/ol-components/tsconfig.json /Users/cchudzicki/dev/mit-learn/frontends/ol-test-utilities/tsconfig.json /Users/cchudzicki/dev/mit-learn/frontends/ol-utilities/tsconfig.json , **trying from the the nearest one +0ms**

With the suggested version above, it gets the right file.

ChristopherChudzicki (changes requested):

ChristopherChudzicki (approved):

PR #3361, opened 2026-05-20
feat: migrate video shorts to learning resources API

daniellefrappier18:

Closes https://github.com/mitodl/hq/issues/11036

This branch migrates Video Shorts from the old v0 API shape to the v1 learning-resources API shape. The modal now uses VideoResource and plays video.streaming_url instead of the old direct MP4 URL.

- [ ] Desktop screenshots

https://github.com/user-attachments/assets/e708ffbe-3db0-4a71-b7f2-242d61b84555

### How can this be tested?

**Automated tests**

- [ ] Run the updated unit tests to verify the component renders correctly with the new VideoResource shape

**Manual testing** 

1. Navigate to the homepage and confirm the Video Shorts carousel section loads. Thumbnail images should appear.
2. Click a thumbnail to open the modal. The video should begin playing automatically (muted).
3. Scroll/swipe through multiple shorts in the modal and verify each video loads and plays from the HLS streaming_url.
4. Toggle the mute/unmute button and confirm audio state changes correctly.
5. Verify the modal closes cleanly via the close button.
6. Confirm thumbnail images display correctly (sourced from video.image.url instead of the old thumbnail_small_url).
7. Resize to mobile width (< 900px) and verify the carousel and modal render at full width without layout issues.


frontends/api/src/hooks/videoShorts/index.ts
```diff
@@ -12,3 +13,19 @@ export const useVideoShortsList = () => {
...
+    queryKey: ["video_shorts", "learning_resources"],
+    queryFn: async () => {
+      const { data } =
+        await learningResourcesSearchApi.learningResourcesSearchRetrieve({
+          resource_category: ["Video Short"],
+          limit: 50,
+        })
+      return data.results.filter(
+        (r): r is VideoResource => r.resource_type === "video",
+      )
```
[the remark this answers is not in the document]
daniellefrappier18: This comment seems overly cautious. I think this is fine

daniellefrappier18:
For a solution to the videos appearing "soft" for a brief moment on startup due to adaptive HLS rendition selection @mbertrand has created the following [issue.](https://github.com/mitodl/hq/issues/11396)

Matt implemented ^ and this has solved the issue.

frontends/main/src/app-pages/HomePage/VideoShortsSection.tsx
```diff
@@ -80,7 +79,7 @@ const ImagePlaceholder = styled.div(({ theme }) => ({
 }))
 
 const VideoShortsSection = () => {
-  const { data, isLoading } = useVideoShortsList()
+  const { data, isLoading } = useVideoShortsLearningResources()
```
ChristopherChudzicki: This comment is long-winded, it boils down to a few things:
1. use the existing `learningResourceQueries.search` query
    - After this, I expect you could remove the whole `api/hooks/video_shorts` directory
3. I think we should add `sortby: "new"`
    - The default search is pretty similar when there's no `q` filter (search query text), but IMO best to explicitly use `new`, that's closest to old `-published_at`, I think.

------

Rather than adding a new hook `useVideoShortsLearningResources`, I'd suggest re-using the existing `learningResourceQueries.search` query. Something like below.

Also:
- filter on `resource_type` at the API level rather than afterwards. (It probably doesn't matter much since you were filtering on "Video Short" category anyway)
    - Generally filtering serverside is better...if you request 50 results and filter client-side, you might end up with fewer. But, as I said, it shouldn't matter here. Still good to do.

I'd suggest using
```ts
  const { data, isLoading } = useQuery({
    ...learningResourceQueries.search({
      resource_category: ["Video Short"],
      resource_type: [ResourceTypeEnum.Video],
      limit: 50,
      sortBy: "new"
    }),
    select: (data) =>
      data?.result.results.filter(
        (r) => r.resource_type === ResourceTypeEnum.Video,
      ),
  })
```
Two notes about the `select` i suggested:
1. This query returns `{ results, time } ` after #3186 ... it's the only query that does this, for the reasons mentioned in that pr.
2. The filter inside `select` is redundant. I kept it just to make TS happy.

Also
- generally nowadays we prefer `useQuery(queryOptions)` rather than defining custom query hooks. 
- In older versions of `@tanstack/react-query`, defining custom query hooks was the best way, but the [queryOptions helper](https://tanstack.com/query/latest/docs/framework/react/guides/query-options) they added in v5 eliminated the need for custom query hooks. (Custom mutation hooks still big helpful.)

ChristopherChudzicki (changes requested):
Looks good.. One request!

I did not notice any degradation of quality in first few moments. Maybe that's @mbertrand 's recent change.

frontends/main/src/app-pages/HomePage/VideoShortsSection.tsx
daniellefrappier18: Thanks! I followed the pattern from the existing `useVideoShortsList` hook when writing `useVideoShortsLearningResources`, but I can see now that using `useQuery(queryOptions)` is the cleaner approach. I'll keep that in mind going forward.

frontends/main/src/app-pages/HomePage/VideoShortsModal.tsx
```diff
@@ -5,10 +5,11 @@ import { CarouselV2Vertical } from "ol-components/CarouselV2Vertical"
 import { RiCloseLine, RiVolumeMuteLine, RiVolumeUpLine } from "@remixicon/react"
 import { ActionButton } from "@mitodl/smoot-design"
 import { useWindowDimensions } from "ol-utilities"
-import type { VideoShort } from "api/v0"
+import type { VideoResource } from "api/v1"
 import MITOpenLearningLogo from "@/public/images/mit-open-learning-logo.svg"
 
-const NEXT_PUBLIC_ORIGIN = process.env.NEXT_PUBLIC_ORIGIN
+const MODAL_VIDEO_CHROME_OFFSET = 60
```
ChristopherChudzicki: Could we change this variable name? I've noticed AI often calls UI, especially video and podcast UI, "chrome". With zero connection to the browser. I find it confusing. Maybe others would too?

**Suggestion:** One of these maybe MODAL_UI_OFFSET, MODAL_CONTROLS_OFFSET, or MODAL_VERTICAL_PADDING.

frontends/main/src/app-pages/HomePage/VideoShortsSection.tsx
```diff
@@ -80,7 +81,18 @@ const ImagePlaceholder = styled.div(({ theme }) => ({
 }))
 
 const VideoShortsSection = () => {
-  const { data, isLoading } = useVideoShortsList()
+  const { data, isLoading } = useQuery({
+    ...learningResourceQueries.search({
+      resource_category: ["Video Short"],
+      resource_type: [ResourceTypeEnum.Video],
+      limit: 50,
```
ChristopherChudzicki, line 88: I'm not sure how many we were fetching before, but 50 seems like a lot. I don't think there's a big perf issue, though, since we only mount (I think) 2 video tags at a time and most of the cards are rendered off screen.

ChristopherChudzicki (changes requested):
Changes look good!

I did however notice a bigger issue. The new m3u8 videos don't work in firefox.

They DO work in firefox on the video detail page, e.g., https://learn.mit.edu/video/115603?playlist=115525

That page uses VideoResourcePlayer, which uses VideoJS and supports the m3u8 streaming. I'm not sure we should use VideoResourcePlayer wholesale in the carousel, but it may have useful pieces.

frontends/main/src/app-pages/HomePage/VideoShortsSection.tsx
daniellefrappier18: Yea it was 50 before too: https://github.com/mitodl/mit-learn/blob/main/frontends/api/src/hooks/videoShorts/index.ts what are do you think the limit should be?

frontends/main/src/app-pages/HomePage/VideoShortsModal.tsx
daniellefrappier18: done :)

daniellefrappier18:
> Changes look good!
> 
> I did however notice a bigger issue. The new m3u8 videos don't work in firefox.
> 
> They DO work in firefox on the video detail page, e.g., https://learn.mit.edu/video/115603?playlist=115525
> 
> That page uses VideoResourcePlayer, which uses VideoJS and supports the m3u8 streaming. I'm not sure we should use VideoResourcePlayer wholesale in the carousel, but it may have useful pieces.

I went with adding hls.js because Video.js owns its own <video> element internally, so I wouldn't have access to the raw element ref. That would've meant rewriting all the play/pause/mute logic to go through the Video.js Player API instead. hls.js just attaches to the existing element, so none of that needed to change.

frontends/main/package.json
```diff
@@ -47,6 +47,7 @@
     "api": "workspace:*",
     "classnames": "^2.5.1",
     "formik": "^2.4.6",
+    "hls.js": "^1.6.16",
```
ChristopherChudzicki: *Added to support the HLS streaming, which Firefox doesn't support natively.*

@daniellefrappier18 I'd like to request that we try doing this with VideoJS[^1] instead of hls.js. The reasons are:
- we use it elsewhere anyway
- I believe `hls.js` is reasonably large dep; Looking at my node_modules dir, its dist files are:
    - hls.min.js (full, what you get): 543 KB raw → ~164 KB gzip (~138 KB over brotli)
    - hls.light.min.js: 354 KB raw → ~109 KB gzip

    That's not exactly what would get bundled by NextJS necessarily (since those are pre-built dists) but I suspect close.

[^1]: A third option would be to use `@videojs/http-streaming` directly. It's already in our repo and supports HLS, but it's in our repo as a transitive dependency of VideoJS. We could of course install it ourselves, but there's no guarantee it doesn't get out of sync with VideoJS itself, and then we'd end up with two copies. So I'd rec just using VideoJS itself.

So using VideoJS saves us a reasonably large chunk of KB. Here I think we should try to avoid that since VideoJS will handle these HLS videos fine (we know it does on the video pages).

You mentioned in the office you added hls.js because you needed a ref for play/pause/etc. I think you should be able to do that fine with VideoJS (see below).
```tsx
<VideoJSPlayer
   onReady={player => {someRef.current = player}}
/>

// then:
player.play()
player.pause()
player.paused() // return a boolean
// also player.ended(), .currentTime(), .duration(), .readyState()
player.on("error", onError)
```

**Note:** I know this isn't a one-line swap. This is an actual refactor... the shorts use a bare <video> for the custom no-controls UI, so I think you'd run VideoJS with controls: false and add loop/muted/playsinline props. Some scope creep here, but our goal was loading the HLS videos, and avoiding introduction of a new, large dep would be good.

ChristopherChudzicki (changes requested):
See what you think of I think https://github.com/mitodl/mit-learn/pull/3361/changes#r3307086373 ... I'm hoping we can do this without introducing hls.js.   No fundamental objection to that library in particular, but it would be nice to use the dependencies we already have.

frontends/main/package.json
daniellefrappier18: @ChristopherChudzicki I went ahead and did the VideoJS refactor, it turned out to be cleaner than expected. I've swapped out hls.js entirely and am now using VideoJsPlayer with an onReady callback to store the player ref, which gives us play/pause/mute/error handling through the VideoJS API. The lockfile is updated to drop hls.js as a dep.

frontends/main/src/app-pages/HomePage/VideoShortsModal.tsx
```diff
@@ -108,17 +116,22 @@ const Placeholder = styled.div(({ theme }) => ({
...
-
-  const isReady = videoElement.readyState >= 2
-
-  const hasDuration = videoElement.duration > 0
+const playerPaused = (player: Player): boolean => player.paused() !== false
+const playerEnded = (player: Player): boolean => player.ended() !== false
+const playerCurrentTime = (player: Player): number => player.currentTime() ?? 0
+const playerReadyState = (player: Player): number => player.readyState() ?? 0
+const playerDuration = (player: Player): number => player.duration() ?? 0
+const playerMuted = (player: Player): boolean => player.muted() ?? true
```
ChristopherChudzicki: Some of these have unnecessary fallbacks.


```suggestion
const playerPaused = (player: Player): boolean => player.paused()
const playerEnded = (player: Player): boolean => player.ended()
const playerCurrentTime = (player: Player): number => player.currentTime() ?? 0
const playerReadyState = (player: Player): number => player.readyState()
const playerDuration = (player: Player): number => player.duration() ?? 0
const playerMuted = (player: Player): boolean => player.muted() ?? true
```
In general, I'd be inclined to just use the player methods directly, e.g., `player.muted(!player.muted())` when toggling. I guess the methods that are both getter and setter return undefined when used as a setter.

I guess for some things, like `playerCurrentTime(player) > 0`, using the method directly would be annoying from a TS perspective.

ChristopherChudzicki (approved):
👍 Looks Good!

Noticed a bunch of accessibility issues, but they are all pre-existing. https://github.com/mitodl/hq/issues/11527

PR #3362, opened 2026-05-20
feat: add PostHog tracking for onboarding CTA interactions

rachellougee:

https://github.com/mitodl/hq/issues/10916#issuecomment-4499456757

adds PostHog tracking for onboarding CTA interactions

image

### How can this be tested?

- Log in at http://od.odl.local:8062
- Navigate directly to http://od.odl.local:8062/onboarding
- Click the Next button
- In PostHog, verify that the `cta_clicked event` is captured for the page http://open.odl.local:8062/onboarding

image

- Confirm the event contains the following data attribute
image

### DIFF
--- a/frontends/main/src/app-pages/OnboardingPage/OnboardingPage.tsx
+++ b/frontends/main/src/app-pages/OnboardingPage/OnboardingPage.tsx
@@ -166,18 +166,18 @@ const OnboardingPage: React.FC = () => {
     initialValues: initialFormData ?? ProfileSchema.getDefault(),
     validationSchema: ProfileSchema,
     onSubmit: async (values) => {
+      if (formik.dirty) {
+        await mutateAsync({
+          ...values,
+          topic_interests: values.topic_interests.map((id) => parseInt(id)),
+        })
+      }
       const label = activeStep < NUM_STEPS - 1 ? "Next" : "Finish"
       if (process.env.NEXT_PUBLIC_POSTHOG_API_KEY) {
         posthog.capture(PostHogEvents.CallToActionClicked, {
           label,
           step: activeStep + 1,
           location: "onboarding",
-        })
-      }
-      if (formik.dirty) {
-        await mutateAsync({
-          ...values,
-          topic_interests: values.topic_interests.map((id) => parseInt(id)),
         })
       }
       if (activeStep < NUM_STEPS - 1) {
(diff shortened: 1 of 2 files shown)


frontends/main/src/app-pages/OnboardingPage/OnboardingPage.tsx
```diff
@@ -163,6 +166,14 @@ const OnboardingPage: React.FC = () => {
     initialValues: initialFormData ?? ProfileSchema.getDefault(),
     validationSchema: ProfileSchema,
     onSubmit: async (values) => {
+      const label = activeStep < NUM_STEPS - 1 ? "Next" : "Finish"
+      if (process.env.NEXT_PUBLIC_POSTHOG_API_KEY) {
+        posthog.capture(PostHogEvents.CallToActionClicked, {
```
daniellefrappier18, line 177: posthog.capture fires before `mutateAsync`, so if the mutation throws and error the event is already recorded even though the step wasn't actually completed. Quick fix just move this after:

```
      if (formik.dirty) {
        await mutateAsync({
          ...values,
          topic_interests: values.topic_interests.map((id) => parseInt(id)),
        })
      }
```

frontends/main/src/app-pages/OnboardingPage/OnboardingPage.tsx
```diff
@@ -163,6 +166,14 @@ const OnboardingPage: React.FC = () => {
     initialValues: initialFormData ?? ProfileSchema.getDefault(),
     validationSchema: ProfileSchema,
     onSubmit: async (values) => {
+      const label = activeStep < NUM_STEPS - 1 ? "Next" : "Finish"
+      if (process.env.NEXT_PUBLIC_POSTHOG_API_KEY) {
```
daniellefrappier18, line 176: I'm not sure you need this guard here. posthog.capture is safe to call unconditionally.

daniellefrappier18 (approved):
See my comment about timing but otherwise looks good.

frontends/main/src/app-pages/OnboardingPage/OnboardingPage.tsx
rachellougee: I kept it here because Copilot pointed out in my first PR that this PostHog capture isn’t guarded by `NEXT_PUBLIC_POSTHOG_API_KEY`, while most other captures in frontends/main check that env var before calling posthog.capture.”. I don’t think that guard is necessary. 

https://github.com/mitodl/mit-learn/pull/3218#discussion_r3093805163

PR #3363, opened 2026-05-20
Task rate limits

mbertrand:
Closes https://github.com/mitodl/hq/issues/11394

Adds a `cooldown_task` decorator that makes use of redis lock keys to prevent a task from running more than once every `n` seconds (3600 for `get_mit_edx_data`, 900 for `get_xpro_data`, `get_oll_data`, and `get_mitxonline_data`).

### How can this be tested?
- `docker compose up`
- Run `./manage.py backpopulate_mitxonline_data`
- Try running it again afterward.  You should see this in the logs:

  ```bash
  Skipping ratelimit:learning_resources.tasks.get_mitxonline_data: rate-limited (900s)
  ```

  And the command should finish quickly:
  ```bash
  Population of MITX Online data finished, took 0.041112 seconds
  Task skipped (rate-limited). See celery logs for details.
  ```
- Try again with `--force`.  This time it should process normally:
  ```
  Population of MITX Online data finished, took 77.92632 seconds
  Populated 268 resources. See celery logs for details.
  ```

### DIFF
--- a/learning_resources/management/commands/backpopulate_xpro_data.py
+++ b/learning_resources/management/commands/backpopulate_xpro_data.py
@@ -39,7 +39,7 @@ class Command(ConfirmDeleteMixin, TestResourceConfigurationMixin, BaseCommand):
             ):
                 resource_delete_actions(learning_resource)
         else:
-            task = get_xpro_data.delay()
+            task = get_xpro_data.delay(_cooldown_force=True)
             self.stdout.write(f"Started task {task} to get xpro course data")
             self.stdout.write("Waiting on task...")
             start = now_in_utc()
--- a/main/decorators.py
+++ b/main/decorators.py
@@ -1,0 +1,94 @@
+"""main decorators"""
+
+import inspect
+import logging
+from collections.abc import Callable
+from functools import wraps
+
+from celery import current_task, states
+from celery.exceptions import Reject
+from django.core.cache import caches
+
+log = logging.getLogger(__name__)
+
+KEY_PREFIX = "cooldown"
+
+
+def cooldown_task(
+    wait_time: int,
+    key: str | None = None,
+    key_func: Callable[..., str] | None = None,
+):
+    """
+    Drop calls made within `wait_time` seconds of the previous invocation.
+
+    The lock is acquired before the wrapped function runs and is not released
+    on exception — failures count against the cooldown to prevent retry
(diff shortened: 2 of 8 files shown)


main/decorators.py
```diff
@@ -0,0 +1,85 @@
...
+        @wraps(func)
+        def wrapper(*args, **kwargs):
+            force = kwargs.pop("_cooldown_force", False)
+            lock_key = _key_for(*args, **kwargs)
+            if force:
+                log.info("Force-overriding cooldown for %s", lock_key)
+                caches["redis"].set(lock_key, "1", timeout=wait_time)
+            elif not caches["redis"].add(lock_key, "1", timeout=wait_time):
+                log.info("Skipping %s: cooldown active (%ss)", lock_key, wait_time)
+                return None
```
rhysyngsun, line 85: I'm going to sleep on it overnight but I'm a little concerned that letting the task succeed is going to cause confusion later when tasks appear to be successful but they aren't actually getting run.

rhysyngsun (commented):
Generally looking good, I'll test tomorrow

main/decorators.py
rhysyngsun: I think we'll just leave it as-is and this should come up in the logs if anyone investigates.

mbertrand: I think I might have just come up with a way to make it more obvious in the celery monitoring dashboard - raise `Reject` so the status is `REJECTED` instead of `SUCCEEDED` - I think that should not raise a sentry error (which is what this is intended to avoid).  And the mgmt commands should probably override the cooldown by default without needing to supply `--force` - what do you think of that idea?

mbertrand: I went ahead and pushed those changes in a new commit, can revert if you prefer the original approach.

rhysyngsun (approved):
LGTM

learning_resources/management/commands/backpopulate_xpro_data.py
```diff
@@ -39,7 +39,7 @@ def handle(self, *args, **options):  # noqa: ARG002
             ):
                 resource_delete_actions(learning_resource)
         else:
-            task = get_xpro_data.delay()
+            task = get_xpro_data.delay(_cooldown_force=True)
```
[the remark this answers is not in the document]
mbertrand: No TypeError raised, it works

PR #3364, opened 2026-05-20
feat(nextjs): standalone output, deterministic builds, and scoped Fastly purge

blarghmatey:
N/A

Addresses the root causes of missing-chunk 404 errors in the MIT Learn Next.js deployment by removing the EFS/blue-green architecture and adding runtime env injection so a single Docker image works across all environments.

**1. Scoped Fastly cache invalidation**
Adds a `Surrogate-Key: html-pages` response header to HTML routes and sitemaps. The Concourse pipeline now calls `/service/b3/purge/html-pages` instead of `purge_all` — so immutable `/_next/static/` content-addressed chunks are never evicted during a deploy.

**2. Deterministic builds**
- `generateBuildId` returns `NEXT_PUBLIC_VERSION || GIT_REF || 'dev'` — stabilises the build manifest filename across identical builds.
- Webpack `output.filename` / `output.chunkFilename` overridden to use `[contenthash]` instead of `[chunkhash]`, making chunk filenames depend only on file content rather than module graph ordering.

**3. Standalone Docker image (replaces EFS + Kubernetes build Job)**
- Adds `output: 'standalone'` to `next.config.js` so Next.js emits a self-contained server bundle.
- Removes the `build_skip_yarn` Docker stage; adds a new slim `runner` stage that copies `.next/standalone/`, `.next/static/`, and `public/` from the build stage into a clean `node:24-alpine` image.
- Adds `ARG GIT_REF` / `ENV GIT_REF` to the build stage so the git SHA passed by Concourse as `BUILD_ARG_GIT_REF` is available to `next build`.
- The build is now fully baked into the image at CI time — no EFS volume or Kubernetes Job needed at deploy time.

**4. Runtime env injection via PublicEnvScript + env() helper**
Fixes a critical issue with standalone builds: webpack's `DefinePlugin` inlines all `process.env.NEXT_PUBLIC_*` references as literal strings at build time. Since `yarn build` runs in CI without per-environment values, all `NEXT_PUBLIC_*` vars would be empty strings in the bundle — even though Kubernetes sets the correct values.

The solution decouples env var reads from the build:

- **`src/env.ts`** — `env("NEXT_PUBLIC_FOO")` helper. In the browser, reads `window.__ENV[key]` (set before any JS loads). On the server, uses dynamic bracket access (`process.env[key]`) which is **not** replaced by DefinePlugin. Falls back to `process.env[key]` in the browser so existing `vi.stubEnv()` / direct `process.env` assignments in tests continue to work without modification.

- **`src/app/components/PublicEnvScript.tsx`** — a Next.js Server Component that calls `connection()` (opts the route out of SSG so env is read at request time) and renders a synchronous inline `<script>window.__ENV={...}</script>` in `<head>` with all `NEXT_PUBLIC_*` env vars serialized as HTML-safe JSON. Because it's a plain `<script>` tag (not `next/script`), it executes synchronously before any JS module — including Sentry's `instrumentation-client.ts`.

- **56 source files** migrated: all `process.env.NEXT_PUBLIC_*` references replaced with `env()` calls; no test files changed.

- **PostHog bootstrap feature flags** (`NEXT_PUBLIC_FEATURE_*`) are now computed at runtime from `window.__ENV` inside `useEffect` in `ConfiguredPostHogProvider`, rather than being baked via `processFeatureFlags()` in `next.config.js`.

- **Server startup validation**: `next.config.js` skips `validateEnv()` during Docker builds (`NEXT_BUILD_CI=1`). Runtime validation was moved to `instrumentation-node.ts` (runs via `instrumentation.ts register()` at server startup) — if required env vars are missing, the pod exits with code 1 and crash-loops to make the misconfiguration immediately visible.

**Companion infrastructure PR:** mitodl/ol-infrastructure#4652 — removes the blue/green EFS deployment and Kubernetes build Job from Pulumi. **These two PRs must be merged and deployed together.**

### How can this be tested?

**Standalone image + env injection:**
```bash
docker build -f frontends/main/Dockerfile.web --target runner -t mit-learn-nextjs:test .
docker run --rm -p 3000:3000 \
  -e NEXT_PUBLIC_ORIGIN=http://localhost:3000 \
  -e NEXT_PUBLIC_MITOL_API_BASE_URL=https://api.example.com \
  -e NEXT_PUBLIC_SITE_NAME="MIT Learn" \
  -e NEXT_PUBLIC_MITOL_SUPPORT_EMAIL=mitlearn-support@mit.edu \
  -e NEXT_PUBLIC_CSRF_COOKIE_NAME=csrftoken \
  mit-learn-nextjs:test
```
- Visit http://localhost:3000 — app should load with correct env values
- `curl -s http://localhost:3000/ | grep 'window.__ENV'` — should show the injected env vars in the HTML `<head>`
- Start the container **without** required env vars — pod should exit 1 immediately with a clear error message

**Fastly surrogate key:**
```bash
curl -I http://localhost:3000/
# → Surrogate-Key: html-pages

curl -I http://localhost:3000/_next/static/chunks/main.js
# → no Surrogate-Key header
```

**Deterministic builds:**
Build twice from the same source; confirm chunk filenames are identical.

### Additional Context

**Deployment coordination:** this PR and mitodl/ol-infrastructure#4652 must be deployed as a single `pulumi up` — deploying the new Dockerfile without removing the Kubernetes build Job (or vice versa) will break the app.

**One-time post-deploy:** the first deployment after merging will have no existing `html-pages`-tagged objects in Fastly (nothing was tagged before). A one-time manual `purge_all` should be run immediately after the first deploy to ensure stale pages are cleared, after which the scoped purge takes over.


blarghmatey:
@copilot address the test failures

frontends/main/Dockerfile.web
gumaerc: Not new in this PR, but there are several repeated env var assignments in this file. `NEXT_PUBLIC_SENTRY_DSN`, `NEXT_PUBLIC_SENTRY_ENV`, `NEXT_PUBLIC_VERSION`, `NEXT_PUBLIC_SENTRY_PROFILES_SAMPLE_RATE`, and `NEXT_PUBLIC_SENTRY_TRACES_SAMPLE_RATE` each appear twice.

frontends/main/src/instrumentation-node.ts
```diff
@@ -23,11 +23,36 @@ import {
 } from "./otel-utils"
 import { parseSampleRate } from "./sentry-utils"
 
+// Validate required NEXT_PUBLIC_* env vars at server startup. These are
+// injected by Kubernetes (not baked at build time), so they must be present
+// when the server starts. validateEnv() is skipped in next.config.js during
+// Docker builds (NEXT_BUILD_CI=1) since standalone mode does not re-run
+// next.config.js at startup.
+const REQUIRED_ENV_VARS = [
```
gumaerc: `validateEnv` calls out these vars explicitly, making them required in local dev. Might it be worth using some kind of shared constant to keep these in sync?

gumaerc (approved):
This worked for me in local dev, and I can approve the infrastructure PR but figure someone from devops should also look at that? Anyway, I only noticed a few minor things, one of them not related to this PR and the other just kind of a defensive thing:

frontends/main/Dockerfile.web
ChristopherChudzicki: @blarghmatey I think all these build args are irrelevant now. I thought I opened an issue to get rid of them, but I can't find it now.

Not sure if it makes sense to do here.

blarghmatey: I think that's a sensible thing to do in this PR. I'll add it in before I merge.

frontends/main/src/instrumentation-node.ts
blarghmatey: @gumaerc can you give a bit more detail on what you're thinking here? I'm happy to work it into this PR. I'm also happy to let it follow on in a separate PR if you already have a plan in mind.

frontends/main/next.config.js
```diff
@@ -1,7 +1,13 @@
...
-validateEnv()
+// In CI Docker builds (NEXT_BUILD_CI=1), NEXT_PUBLIC_* vars are not available
+// at build time — they are injected at runtime via PublicEnvScript. Skip
+// build-time validation; validateEnv() runs at server startup instead (see
+// src/instrumentation-node.ts).
+if (!process.env.NEXT_BUILD_CI) {
+  validateEnv()
+}
 
 const NEXT_PUBLIC_OPTIMIZE_IMAGES = Boolean(
```
ChristopherChudzicki: **Request / Blocking Question:**  I think `NEXT_PUBLIC_OPTIMIZE_IMAGES` might get baked in as `true`, which is contrary to the current productions setting.

I think all `next.config.js` values need to be available at build time, so this would also affect `NEXT_CACHE_S_MAXAGE_SECONDS`. 

Still testing a bit locally... but wanted to mention this. If we need to hardcode those values, that's probably OK. But eventually it would be nice to have Fastly doing the image optimization, and that may need build-time env vars in the `image` conflict slot?

ChristopherChudzicki (changes requested):

frontends/main/next.config.js
ChristopherChudzicki: @blarghmatey I failed to find explicit docs about this, but it's clear from testing that env vars referenced in `next.config.js` are baked in at build time. e.g., running
```
docker run -d --rm --name mln-test -p 8062:3000 \
      --add-host=api.open.odl.local:host-gateway \
      -e NEXT_PUBLIC_ORIGIN=$NEXT_PUBLIC_ORIGIN \
      -e NEXT_PUBLIC_MITOL_API_BASE_URL=$NEXT_PUBLIC_MITOL_API_BASE_URL \
      -e NEXT_PUBLIC_SITE_NAME=$NEXT_PUBLIC_SITE_NAME \
      -e NEXT_PUBLIC_MITOL_SUPPORT_EMAIL=$NEXT_PUBLIC_MITOL_SUPPORT_EMAIL \
      -e NEXT_PUBLIC_CSRF_COOKIE_NAME=$NEXT_PUBLIC_CSRF_COOKIE_NAME \
      -e NEXT_PUBLIC_OPTIMIZE_IMAGES=false \
      mit-learn-nextjs:test
```
on my local does NOT turn image optimization off it was built with `NEXT_PUBLIC_OPTIMIZE_IMAGES` on.

More generally, next.config.js uses two env vars right now:

|                          | CI                         | QA       | Production |
| ------------------------ | -------------------------- | -------- | ---------- |
| `cache_s_maxage_seconds` | (unset → default **1800**) | **900**  | **7200**   |
| `optimize_images`        | **true**                   | **true** | **false**  |



- cache_s_maxage_seconds we can move to middleware.
- optimize_images we could just turn off. We have it off on production and IMO we should [use fastly anyway](https://github.com/mitodl/hq/discussions/10595#discussioncomment-16547908).

ChristopherChudzicki: BTW, from my comment on https://github.com/mitodl/hq/discussions/10595#discussioncomment-16547908:

> NextJS's image optimization has a loader option that should let us switch easily to using Fastly as the backend for image optimization. Using loader, NextJS would continue generating things like srcset and quality for us, but we could offload the actual optimizing to Fastly.

Briefly investigating with Claude, it seems like the loader URL (e.g., `mgproxy.odl.mit.edu/?url=....`) could be customized per env—it's a separate file, not baked into the config.

blarghmatey:
Addressed the CHANGES_REQUESTED feedback (rebased on main + new commit `bcfc095`):

**1. `NEXT_PUBLIC_OPTIMIZE_IMAGES` baked at build time (blocking)**
Removed the env-var gate entirely. Image optimization is now unconditionally disabled (`unoptimized: true`) — production already had it off, and the long-term plan is Fastly-based image transforms. This removes the risk of the CI-default `true` silently winning in production.

**2. `NEXT_CACHE_S_MAXAGE_SECONDS` baked at build time (blocking)**
Added `src/middleware.ts`. Cache-Control (`s-maxage=...`) is now set at request time by Next.js middleware running on the Edge Runtime, where `process.env` is the live Kubernetes value. The `Surrogate-Key: html-pages` header stays in `next.config.js` (static value — fine to bake in). The middleware mirrors the same URL exclusion logic (static-file extensions + `/healthcheck`) and also explicitly includes sitemaps despite their `.xml` extension.

**3. Stale `NEXT_PUBLIC_DEFAULT_SEARCH_*` ENV lines in Dockerfile**
Removed the five `NEXT_PUBLIC_DEFAULT_SEARCH_*` ENV declarations from the `base` build stage. These are runtime vars now covered by the PublicEnvScript / `window.__ENV` mechanism — they'll be injected by Kubernetes, not baked into the image.

ChristopherChudzicki (changes requested):
@blarghmatey I'm seeing a lot of references to `process.env` variables still existing, even outside of test files.

What's slightly tricky is that many of the usage is outside the `main/` workspace where your new env helper lives. These seem generally to fall in three categories:

1. **API client configuration:** E.g., `frontends/api/src/clients.ts` (and other files) set base URLs from an env var and other client config like CSRF cookie names.
2. **Embedly:** `frontends/ol-components/src/components/EmbedlyCard/EmbedlyCard.tsx` and another file both use embedly URL.
3. `frontends/ol-utilities/src/images/backgroundImages.ts` does some stuff related to image optimization and background images.

I think the best thing to do is... I'll open a separate PR to resolve each of these. I don't thing anything outside of `main` should be using env vars (except for `process.env.NODE_ENV`). 

Roughly my plan is:
1. **API client configuration:** expose a `configureClients` func or something that `main` can call.
2. Delete embedly entirely, we don't use it.
3. Delete `backgroundImages.ts` (we have image optimization off) or possibly move it to `main`.

ChristopherChudzicki:
OK, I took this over from @blarghmatey to address a few issues. Mostly lingering `NEXT_PUBLIC` env vars. After merging
- https://github.com/mitodl/mit-learn/pull/3380
- https://github.com/mitodl/mit-learn/pull/3377

plus the changes in this PR, all **direct reads** of `NEXT_PUBLIC` env vars are gone, with a few exceptions:
- `NEXT_PUBLIC_VERSION`, which we guarantee to exist and be stable across environments
- `NEXT_PUBLIC_ORIGIN` read once in `next.config.js`, but **only for dev mode**
    - this is annoying, but required after https://github.com/mitodl/mit-learn/pull/3399, which we merged to avoid https://github.com/vercel/next.js/issues/90898
- and references in test files (perhaps we should update those, but not here).

Additionally, a linting rule now bans direct reads of `NEXT_PUBLIC` vars.

See https://github.com/mitodl/mit-learn/pull/3364/commits/f0d876e1f2a6923a557e23bc926374bbba239050 for a brief summary... All my changes are in one commit with a summary message.

--------

To review updates, I suggest:

1. Start local dev docker stack like normal, then `docker compose stop watch`.
2. `docker build -f frontends/main/Dockerfile.web --target runner -t mit-learn-nextjs:test .`
3. Then: (This should fail with a useful error message...not all missing required env vars)
    ```sh
    docker run --rm --name mln-test -p 8062:3000 \
                --add-host=api.open.odl.local:host-gateway \
                -e NEXT_PUBLIC_MITOL_SUPPORT_EMAIL=$NEXT_PUBLIC_MITOL_SUPPORT_EMAIL \
                -e NEXT_PUBLIC_CSRF_COOKIE_NAME=$NEXT_PUBLIC_CSRF_COOKIE_NAME \
                -e NEXT_PUBLIC_MITX_ONLINE_LEGACY_BASE_URL=$NEXT_PUBLIC_MITX_ONLINE_LEGACY_BASE_URL \
                -e NEXT_PUBLIC_MITX_ONLINE_BASE_URL=$NEXT_PUBLIC_MITX_ONLINE_BASE_URL \
                -e NEXT_PUBLIC_MITX_ONLINE_CSRF_COOKIE_NAME=$NEXT_PUBLIC_MITX_ONLINE_CSRF_COOKIE_NAME \
                mit-learn-nextjs:test
    ```
4. Then (and whatever other env vars you want). This should work, and should start the site. Try it out. Maybe also check `window.__ENV`.
    ```sh
    docker run --rm --name mln-test -p 8062:3000 \
                --add-host=api.open.odl.local:host-gateway \
                -e NEXT_PUBLIC_ORIGIN=$NEXT_PUBLIC_ORIGIN \
                -e NEXT_PUBLIC_MITOL_API_BASE_URL=$NEXT_PUBLIC_MITOL_API_BASE_URL \
                -e NEXT_PUBLIC_SITE_NAME=$NEXT_PUBLIC_SITE_NAME \
                -e NEXT_PUBLIC_MITOL_SUPPORT_EMAIL=$NEXT_PUBLIC_MITOL_SUPPORT_EMAIL \
                -e NEXT_PUBLIC_CSRF_COOKIE_NAME=$NEXT_PUBLIC_CSRF_COOKIE_NAME \
                -e NEXT_PUBLIC_MITX_ONLINE_LEGACY_BASE_URL=$NEXT_PUBLIC_MITX_ONLINE_LEGACY_BASE_URL \
                -e NEXT_PUBLIC_MITX_ONLINE_BASE_URL=$NEXT_PUBLIC_MITX_ONLINE_BASE_URL \
                -e NEXT_PUBLIC_MITX_ONLINE_CSRF_COOKIE_NAME=$NEXT_PUBLIC_MITX_ONLINE_CSRF_COOKIE_NAME \
                mit-learn-nextjs:test
    ```
5. Repeat step 4, but with `NEXT_PUBLIC_MITOL_API_BASE_URL=https://api.learn.mit.edu` ... smoke test we can pass different env vars to the same build.
6. Verify that `s-maxage` headers are present on server-rendered HTML routes but NOT static files.
7. I guess you could also try out the linting rule and try using a NEXT_PUBLIC_... var somewhere directly.

frontends/main/next.config.js
```diff
@@ -72,26 +64,32 @@ const nextConfig = {
       {
         source: "/sitemaps/:path*.xml",
         headers: [
-          {
-            key: "Cache-Control",
-            value: PAGE_CACHE_CONTROL,
-          },
+          // Tag sitemaps with "html-pages" so Fastly can purge them on deploy
+          // without also purging immutable /_next/static/ chunks.
+          // Cache-Control is set at runtime in src/middleware.ts so that
```
gumaerc: Minor nit:

In this file on lines 69 and 89, the comments reference middeware.ts which was renamed to proxy.ts as part of the NextJS upgrade.

gumaerc (approved):
👍 

This worked well for me and I was able to verify the test procedure you left in your last comment. Besides the comment nitpick I left below, the only other thing notable that AI called out was in `AiRecommendationBotDrawer.tsx` and `AiChatSyllabusSlideDown.tsx`. It mentions that the env vars `NEXT_PUBLIC_LEARN_AI_RECOMMENDATION_ENDPOINT` and `NEXT_PUBLIC_LEARN_AI_SYLLABUS_ENDPOINT` are marked as optional. If the feature flag is enabled and these vars are unset, it could be making requests to `undefined`.

ChristopherChudzicki (approved):

PR #3371, opened 2026-05-26
remove authentication requirement from hybrid search 

shanbady:

Closes https://github.com/mitodl/hq/issues/11337

This PR removes the authentication check for the vector_search=True parameter on the search page. This isn't really a feature that needs to require authentication to view it and it was causing confusion for folks passing links around and testing hybrid search.

### How can this be tested?
1. checkout main
2. make sure you have learning resources embedded locally
3. login as an admin user and visit the[ search page](open.odl.local:8062/search) - search for something
4. enable the "hybrid search" via the admin checkbox - note that it adds a vector_search=true get parameter and vector search results are loaded 
5. copy the url and logout - re-visit the url and see that it does not load the vector search results (you can double confirm via web inspector to see the endpoint it attempts to load)
6. checkout this branch and attempt to load the url as an anonymous user and note that it loads the vector search results


abeglova (approved):
lgtm

PR #3374, opened 2026-05-27
Send task-rejected event when a celery task is manually rejected, so Leek picks it up

mbertrand:
Follow-up to https://github.com/mitodl/mit-learn/pull/3363 

The changes made in https://github.com/mitodl/mit-learn/pull/3363 are correctly causing tasks to be rejected within the cooldown period (`task.status` == `REJECTED`) but Leek is not picking up that status change, so it is shown as stuck in "STARTED" status.

This PR modifies the decorator to send a "task-rejected" event, which in theory should let Leek pick up the status change to REJECTED.

### How can this be tested?
- Start up containers.
- Run this in a django shell:
```python
import threading
import time

from learning_resources.tasks import get_mitxonline_data
from main.celery import app
from main.decorators import KEY_PREFIX

TASK = get_mitxonline_data
LOCK_KEY = f"{KEY_PREFIX}:{TASK.name}"
QUEUE = "edx_content"
MAX_WAIT = 30 

events = []
seen_rejected = threading.Event()


def handler(event):
    if not event.get("type", "").startswith("task-"):
        return
    events.append(event)
    if event["type"] == "task-rejected":
        seen_rejected.set()


def listen():
    with app.connection() as conn:
        recv = app.events.Receiver(conn, handlers={"*": handler})
        recv.capture(limit=None, timeout=None, wakeup=True)


listener = threading.Thread(target=listen, daemon=True)
listener.start()
time.sleep(3)  # let the receiver bind its queue before we dispatch

caches["redis"].set(LOCK_KEY, "1", timeout=900)
result = TASK.apply_async(queue=QUEUE)
print(f"dispatched {TASK.name} id={result.id} queue={QUEUE}")

seen_rejected.wait(timeout=MAX_WAIT)
caches["redis"].delete(LOCK_KEY)

mine = [e for e in events if e.get("uuid") == result.id]
print("events for this run:")
for e in mine:
    print(f"  {e['type']:<14} requeue={e.get('requeue')}")

print("\nresult backend state:", result.state)
rejected = [e for e in mine if e["type"] == "task-rejected"]
if rejected and rejected[0].get("requeue") is False:
    print("PASS: task-rejected event emitted (Leek will show REJECTED)")
else:
    print("FAIL: no task-rejected event seen on the stream")
```

### DIFF
--- a/main/constants.py
+++ b/main/constants.py
@@ -40,6 +40,7 @@ ALLOWED_HTML_TAGS = {
     "ul",
 }
 ALLOWED_HTML_ATTRIBUTES = {}
+TASK_REJECTED = "task-rejected"
 
 
 class PostHogEvents(Enum):
--- a/main/decorators.py
+++ b/main/decorators.py
@@ -9,6 +9,8 @@ from celery import current_task, states
 from celery.exceptions import Reject
 from django.core.cache import caches
 
+from main.constants import TASK_REJECTED
+
 log = logging.getLogger(__name__)
 
 KEY_PREFIX = "cooldown"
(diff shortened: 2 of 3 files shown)


shanbady (approved):
The test script appears to work - i tried getting a dockerized leek setup to test locally but it might be more expedient to deploy

PR #3377, opened 2026-05-27
chore: Remove embedly

ChristopherChudzicki:
None

This PR removes embedly from Learn, both frontend and backend.
- **Frontend:**  Until somewhat recently, the frontend was using it to display videos in the resource drawer. It was removed when we stopped showing videos there, and instead just link to the video pages. HOWEVER: Even if did want videos in the drawer, embedly was unnecessary. We could just have embedded the youtube iframe directly.
- **Backend:** Removed the embedly app. This was an link-preview/oembed proxy that we used in Open Discussions. The app never even registered in INSTALLED_APPS for Learn.

None.

### How can this be tested?
1. Site should still work.
2. Tests should pass. 
3. Searching "Embedly" should not give any meaningful results in the codebase. (There's a hit in an old rfc and the release file).


abeglova (approved):
lgtm

PR #3378, opened 2026-05-27
Cut down on repeated APISIX user DB changes and logins

mbertrand:
Related: mitodl/hq#11506

Updates ApisixUserMiddleware so requests avoid unnecessary User/Profile saves or logins when APISIX-provided identity data has not changed.

This keeps APISIX/Keycloak/Django identity data synchronized by still comparing the APISIX-controlled user and profile fields on each request, while only saving the User, syncing the Profile, or calling login() when the current request actually requires it.

### How can this be tested?

1. Run the app locally with the normal APISIX/Keycloak auth stack enabled.
2. Sign in as an existing Keycloak/APISIX user that already has a matching Django User. Confirm the app signs in normally, the user menu/profile identity is correct, and there are no logout loops.
3. Capture the Django User/Profile values after the first successful sign-in, especially User `last_login`, User `updated_on`, and Profile `updated_at`. Refresh the app and visit several authenticated and user-aware pages, such as search/results, a resource detail page, dashboard/profile, and a saved/favorites flow. Re-check those timestamps and confirm they do not change just from browsing with unchanged APISIX identity data.
4. While still signed in as that existing user, confirm normal API-backed pages continue to work and the user remains authenticated across additional page loads.
5. Sign out completely. Confirm the app returns to an anonymous state and authenticated UI/actions are no longer available.
6. Sign in as a brand-new Keycloak/APISIX user that does not yet exist in Django. Confirm a Django User is created, the Profile exists with the expected APISIX profile values, first-time user setup still runs, and the user can browse the same set of pages without errors.
7. Sign out, then sign back in as the original existing user. Confirm the Django session switches cleanly to the APISIX user and the visible account identity is correct.
8. If convenient in the local Keycloak admin, change an APISIX-controlled claim for the existing user, such as first name, last name, or email. Sign in or refresh with the updated claim and confirm the corresponding Django User/Profile field updates exactly when the header value changes.

### DIFF
--- a/main/middleware/apisix_user.py
+++ b/main/middleware/apisix_user.py
@@ -90,7 +90,9 @@ def decode_apisix_headers(request, header, model=settings.AUTH_USER_MODEL):
     }
 
 
-def get_user_from_apisix_headers(request, decoded_headers, original_header):
+def get_user_from_apisix_headers(  # noqa: C901
+    request, decoded_headers, original_header
+):
     """
     Get a user based on the APISIX headers, create user/profile if needed.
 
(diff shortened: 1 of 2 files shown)


rhysyngsun (approved):
LGTM

PR #3379, opened 2026-05-27
contract variants UI

gumaerc:
Closes https://github.com/mitodl/hq/issues/11487

Uses the new `variant_options` field on `ContractPage` to generate a "variant picker" on B2B contract pages with one or more variants. This variant can contain any combination of:

- Language
- Industry
- Length

Switching variants switches out the runs associated with each dashboard card, just like the language picker functionality that has now been removed. The variant functionality supplants it, and it is only focused on B2B. Therefore, after the language picker was stripped out, the variant picker was not added to the B2C program dashboard.

Changes:

**Core feature**
- New `VariantPicker` component — radio-button card group rendered once per contract; hidden when fewer than 2 variants are available. Includes `role="radiogroup"` / `aria-labelledby` accessibility semantics and a "Viewing:" summary strip.
- New `useDashboardVariantPicker` hook — manages selected-variant state, defaults to the `default_variant: true` entry, and resets gracefully when options change.
- `useContractDashboardData` — reads `contract.variant_options` (from the `ContractPage` API response) to drive the picker. Fires a lazy `courseVariantRuns` second-phase query only when a non-default variant is selected; maps results into a `variantRunsByCourseId` lookup consumed by `buildCourseEntry`.
- `dashboardViewModel` — adds `selectVariantRunForCourse` (best-matching run selection with language/industry/length filtering and date tiebreaking), `buildVariantKey`, and `buildVariantLabel`. Removes the now-unused `getDistinctContractVariantOptions`.

**B2C program dashboard cleanup**
- `useProgramDashboardData` — removes dead language-picker state (`availableLanguages`, `selectedLanguageKey`, `setSelectedLanguageKey`) that was computed but never rendered after the language picker was removed from `ProgramEnrollmentDisplay`.

**Test infrastructure**
- Bumps `@mitodl/mitxonline-api-axios` to `2026.5.28`, which adds `variant_options: SupportedVariant[]` to `ContractPage` and `course_id: number` to `BaseCourseRun`.
- Adds `variant_options: []` to the contract factory; adds `course_id` to run factories to fix type errors from the new required field.
- Adds `courseVariantRuns` URL helper and `courseVariantRunsList` query option.
- New unit tests: `useDashboardVariantPicker.test.ts` (9 tests), `selectVariantRunForCourse` / `buildVariantKey` / `buildVariantLabel` in `dashboardViewModel.test.ts` (18 tests).
- Refactors 4 variant-picker integration tests in `ContractContent.test.tsx` to set `variant_options` on the contract mock instead of `possible_variant_sets` on course mocks.

image
image

### How can this be tested?
Requires a local mitxonline instance running the code from mitxonline PR #3618 (now merged to main).

1. Ensure the `mitxonline` backend is running with the merged variant options changes
2. Follow the instructions in this PR to create proper test data in `mitxonline`: https://github.com/mitodl/mitxonline/pull/3618
3. Load the B2B dashboard for that contract — the variant picker should appear if there are 2+ variant options
4. Select a non-default variant — course run titles should update to reflect the selected variant
5. When you enroll in a variant, you should be redirected to the courseware URL associated with that variant
6. With only one variant option on the contract, confirm the picker does not appear


frontends/main/src/app-pages/DashboardPage/CoursewareDisplay/VariantPicker.tsx
```diff
@@ -0,0 +1,177 @@
...
+      {variantOptions.map((variant) => {
+        const value = buildVariantKey(variant)
+        const label = buildVariantLabel(variant)
+        return (
+          <VariantCard
+            key={value}
+            name={`variant-option-${value}`}
+            value={value}
+            title={label}
+            certEligible={variant.default_variant}
```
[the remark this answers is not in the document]
gumaerc: This is indeed the intent

frontends/main/src/app-pages/DashboardPage/CoursewareDisplay/VariantPicker.tsx
```diff
@@ -0,0 +1,194 @@
...
+        name={name}
+        value={value}
+        checked={selected}
+        onChange={onChange}
+      />
+      <Typography variant="subtitle1">{title}</Typography>
+      <Stack
+        direction="row"
+        alignItems="center"
+        visibility={certEligible ? "visible" : "hidden"}
```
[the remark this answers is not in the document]
gumaerc: This is not true, the visibility prop works just fine.

frontends/main/src/app-pages/DashboardPage/CoursewareDisplay/model/dashboardViewModel.ts
```diff
@@ -857,6 +910,101 @@ const getRenderableContractCollections = (
...
+ * Matching is exact on every non-empty field of the selected variant
+ * (language, industry, length).  If the course has no run for that combination
+ * the function returns `null`, signalling to the caller that it should fall
+ * back to the course's `next_run_id`-based default.
+ *
+ * Among matching runs, the one with the nearest *upcoming* start date is
+ * preferred; if all start dates are in the past the most-recent past run wins;
+ * runs with no start date are last.
+ */
+const selectVariantRunForCourse = (
```
ChristopherChudzicki, line 901: The docstring says:
> Given the list of runs the API returned for one course (which may include both the course's default-variant runs AND the requested-variant runs), return the single best run that matches the selected variant combination.

I don't really think this is accurate.

The current behavior of this function is:
1. Given runs like `[(en, H, ""), (de, H, ""), (de, H, S), (de, F, ""), (de, "", ""), (fr, H, "")]`
2. then selections produce:
    - selected `(de, "", "")` filters to `(de, H, ""), (de, F, ""), (de, "", ""), (de, H, S)`
    - selected `(de, H, "")` filters to `(de, H, ""), (de, H, S)

In other words, it treats `""` as a sort of wildcard.

Two issues:
- I don't think it's what we want. Industry = "" means "general", not wildcard, so we shouldn't match anything other than "" exactly.
- IMO this doesn't match the docstring. E.g., if there was no healthcare version for geman, then `(de, "", "")` would best match the selection of `(de, H, "")`, but it would be filtered out.

It turns out not to matter at all because `api/v3/courses/variant_runs/` filters exactly anyway. So if you do `api/v3/courses/variant_runs/?language=de`, then it matches `(de, "", "")` exactly. 

**Still, a request:** IMO, we should:
- switch to exact match
- fix the docstring

Additionally, I'd recommend extracting the filter func to something like 
```ts
const runMatchesVariant = (variant: SupportedVariant) => (run: BaseCourseRun) =>
  variant.language === (run.language ?? "") &&
  variant.variant_industry === (run.variant_industry ?? "") &&
  variant.variant_length === (run.variant_length ?? "")

// then:
const matching = runs.filter(runMatchesVariant(selectedVariant))
```

frontends/main/src/app-pages/DashboardPage/CoursewareDisplay/model/dashboardViewModel.ts
```diff
@@ -857,6 +910,101 @@ const getRenderableContractCollections = (
...
+  S: "Short",
+  F: "Full",
+}
+
+const buildVariantKey = (variant: SupportedVariant): string =>
+  `language:${variant.language ?? ""}|industry:${variant.variant_industry ?? ""}|length:${variant.variant_length ?? ""}`
+
+const buildVariantLabel = (variant: SupportedVariant): string => {
+  const langLabel = variant.language
+    ? getNativeLanguageName(variant.language.replace("_", "-"))
```
ChristopherChudzicki:
```suggestion
    ? getNativeLanguageName(getLanguageCodeFromText(variant.language))
```

- lowercases
- replaces more underscores (probably useless)
- links to the docstring on getNativeLanguageName explaining why it exists (django vs js language format quirk)

frontends/main/src/app-pages/DashboardPage/CoursewareDisplay/model/dashboardViewModel.ts
```diff
@@ -577,35 +580,85 @@ type RequirementSection = {
...
  *
  * `enrollments` is stored uncollapsed on the entry — the full list, never
  * filtered to the displayed choice. The `displayedEnrollment`/`displayedRun`
  * pair is derived by `resolveCourseEntryForLanguage` for the legacy card UI.
+ *
+ * When `opts.variantRun` is provided (variant picker path):
+ *  - non-null: use that run as `displayedRun`; look up the full `CourseRunV2` by id
+ *  - null: this course has no variant run for the selection — fall back to
+ *    `next_run_id`-based resolution (same as empty `selectedVariantKey`)
+ *  - undefined (default): use the existing language-based resolution path
```
ChristopherChudzicki: I'm confused about these three options. `variantRun` not null is very clear, but the `null` vs `undefined` distinction is confusing. 

There are three big paths through `buildCourseEntry`:
1. Path 1, `variantRun` truthy
2. Path 2, `variantRun` null
3. Path 3, variantRun undfeined

**The null and undefined paths should be combined.** The two paths both call `resolveCourseEntryForLanguage` with exactly the same arguments, with one exception:
- Path 2: calls `resolveCourseEntryForLanguage` with  `selectedLanguageKey` hardcoded to `""`
- Path 3 calls `resolveCourseEntryForLanguage` with `selectedLanguageKey=selectedVariantKey`, where selectedVariantKey is piped from arg of `buildCourseEntry`...But `buildCourseEntry` is called in exactly two scenarios: 
    - via `buildRequirementSections`, whose callsite passes `selectedVariantKey = ""`.... so this is the same as Path 2
    - via `useContractDashboardData` ... this callsite either:
        - passes variantRun, in which case this is really Path 1
        - or  variantRun is null/undefined. But this is exactly correlated in `useContractDashboardData` with `selectedVariantKey=""`, so this is the same as Path 2.

Which is all to say:
- In `buildCourseEntry`, paths 2 and 3 are safe to combine. Both Paths always send `selectedVariantKey=""` to resolveCourseEntryForLanguage
- And in fact, ALL callsites of `resolveCourseEntryForLanguage` use `selectedLanguageKey=""`; the arg can be deleted and `resolveCourseEntryForLanguage` can be simplified.

In fact, I generally expect `resolveCourseEntryForLanguage` is no longer necessary... We can't get rid of it entirely, but it should become `resolveDisplayedRunAndEnrollment(course, enrollments, {contractId})`. or something—its job is to resolve the default displayed run when NOT constrained by variants.

ChristopherChudzicki (commented):
As discussed in slack, a bunch of language stuff still in here that we should get rid of. like `useDashboardLanguagePicker`. Left some other feedback that I think we should simplify. Main thing is the language stuff.

In general, working well.

non-issues for frontend:
- filter out `active: false`, but that will be handled on the backend as discussed in slack

frontends/main/src/app-pages/DashboardPage/CoursewareDisplay/hooks/useContractDashboardData.ts
```diff
@@ -85,20 +86,57 @@ const useContractDashboardData = (
...
+  // Empty when the default (Original) variant is active.
+  //
+  // The API returns all runs for a course that match either the selected
+  // variant OR the course's default variant (so the array can hold multiple
+  // sessions and/or a mix of variant types).  We therefore filter to runs
+  // that explicitly match every non-empty field of the selected variant and
+  // pick the best session among those.  Courses with no matching run are
+  // absent from the map, which causes buildCourseEntry to fall back to
+  // next_run_id via the null variantRun path.
+  const variantRunsByCourseId = React.useMemo<
```
ChristopherChudzicki, line 120: **Bug Here:** This picks a run without knowledge of user's enrollments.

Roughly, I think it should work like this:
1. `useContractDashboardData` should collect the variant runs:
    ```ts
    const variantRunsByCourseId:  Record<number,  BaseCourseRun[]> = React.useMemo(/* etc*/)
    ```
2. `resolveCourseEntryForLanguage` ... renamed to `resolveDisplayedRunAndEnrollment` pending https://github.com/mitodl/mit-learn/pull/3379/changes#r3326430955 ... Thinking something similar to this?
    ```ts
    /**
     * Pick course run and enrollment to display.
     * Options:
     *  - contractId, if supplied, filters runs/enrollments to that contract
     *  - variant: if supplied, filters runs/enrollments to that variant
     *  - variantCandidateRuns: must be supplied if variant is supplied;
     *    the list of runs to consider when looking for a variant if not enrolled
     *    already
     */
    resolveDisplayedRunAndEnrollment(course, enrollments, {
        contractId,
        variant?: SupportedVariant, 
        variantCandidateRuns?: BaseCourseRun[],
    })
    ```

frontends/main/src/app-pages/DashboardPage/CoursewareDisplay/model/dashboardViewModel.ts
```diff
@@ -857,6 +910,101 @@ const getRenderableContractCollections = (
...
+      const aMs = a.start_date ? new Date(a.start_date).getTime() : null
+      const bMs = b.start_date ? new Date(b.start_date).getTime() : null
+      const aFuture = aMs !== null && aMs >= now
+      const bFuture = bMs !== null && bMs >= now
+      if (aFuture && bFuture) return aMs! - bMs! // nearest upcoming first
+      if (aFuture) return -1
+      if (bFuture) return 1
+      if (aMs !== null && bMs !== null) return bMs - aMs // most-recent past first
+      if (aMs !== null) return -1
+      if (bMs !== null) return 1
```
ChristopherChudzicki: In the past, we've always used `next_run_id` as the first choice when user isn't enrolled. We can't do that here because I assume it points specifically to the default.

But this sorting isn't what we want... It prioritizes future runs. (Probably AI assumed `next_run_id` was in the future always, which it isn't.... Great names.)

The logic should probably be.... sort enrollable first, then descending by start date.

ChristopherChudzicki (changes requested):
Noticed two more things:

frontends/main/src/app-pages/DashboardPage/CoursewareDisplay/model/dashboardViewModel.ts
```diff
@@ -355,156 +345,174 @@ const assembleHomeCardList = ({
...
     enrollments,
-    selectedLanguageOption,
-    selectedRun,
   )
-  const displayedEnrollment = selectedLanguageKey
-    ? selectedLanguageEnrollment
-    : pickDisplayedEnrollmentForLegacyDashboard(course, enrollments)
 
-  const displayedRun = getResolvedRunForSelectedLanguage(
+  const displayedRun = synthesizeDisplayedRun(
```
ChristopherChudzicki: This function has some issues, imo.

1. This doesn't make sense to me. There are two variables `matchedEnrollment` and `displayedEnrollment` that differ (matched enrollment looks like the enrollment for the default run?) and the calculation of displayedRun is based on matchedEnrollment. This seems wrong.
2. Enrollments are selected differently for variant vs non-variant path.  E.g., `variantEnrollment` uses the **first** enrollment matching that variant, whereas other branches use `selectBestEnrollment`.

I think it should look more like below. I'm a little unclear if `synthesizeDisplayedRun` is necessary at all anymore?

do all filtering up front (unfortunately not quite all, since displayedRun still branches on variants):

```ts
const filterEnrollmentsForContract = (contractId?: number) => {
  if (typeof contractId !== "number") return isNonContractEnrollment
  return (enrollment: CourseRunEnrollmentV3) =>
    enrollment.b2b_contract_id === contractId
}

const resolveDisplayedRunAndEnrollment = (
  course: CourseWithCourseRunsSerializerV2,
  enrollments: CourseRunEnrollmentV3[],
  opts?: {
    contractId?: number
    variant?: SupportedVariant
    variantCandidateRuns?: BaseCourseRun[]
  },
): {
  displayedEnrollment: CourseRunEnrollmentV3 | null
  displayedRun: BaseCourseRun | null
} => {
  const variantFilter = opts?.variant
    ? runMatchesVariant(opts.variant)
    : () => true
  const relevantEnrollments = enrollments
    .filter(filterEnrollmentsForContract(opts?.contractId))
    .filter((e) => variantFilter(e.run))

  const displayedEnrollment = selectBestEnrollment(course, relevantEnrollments)
  if (displayedEnrollment) {
    const displayedRun =
      course.courseruns.find((r) => r.id === displayedEnrollment.run.id) ??
      displayedEnrollment.run
    return { displayedEnrollment, displayedRun }
  }

  const defaultRun =
    getBestRun(course, {
      enrollableOnly: true,
      contractId: opts?.contractId,
    }) ?? null
  const displayedRun = opts?.variant
    ? (selectVariantRunForCourse(
        opts.variantCandidateRuns ?? [],
        opts.variant,
      ) ?? defaultRun)
    : defaultRun
  return { displayedEnrollment: null, displayedRun }
}
```

frontends/main/src/app-pages/DashboardPage/CoursewareDisplay/hooks/useContractDashboardData.ts
```diff
@@ -155,15 +181,13 @@ const useContractDashboardData = (
...
-          enrollmentsByCourseId[course.id] ?? [],
-          selectedLanguageKey,
-          {
-            availableLanguages: languageOptions,
-            contractId: contract.id,
-          },
-        ),
+        buildCourseEntry(course, enrollmentsByCourseId[course.id] ?? [], {
+          contractId: contract.id,
+          variant: isDefaultVariantSelection ? undefined : selectedVariant!,
```
ChristopherChudzicki, line 186:
```suggestion
          variant: isDefaultVariantSelection ? undefined : selectedVariant,
```

ChristopherChudzicki (approved):
👍 I left one comment about a ts non-null assertion we can clean up. Otherwise, looks good to me!

PR #3381, opened 2026-05-28
podcast etl cleanup

abeglova:
None

This pr makes small cleanups to the podcast. 
1) It makes website optionally since podcasts in search now link to the new learn podcast pages and we recently added podcasts that do no have their own websites
2) Catches  ConnectionResetError when pulling podcast rss data so we don't have a sentry error. This error is temporary

https://mit-office-of-digital-learning.sentry.io/issues/7452956445/?project=4506260054540288&query=is%3Aunresolved%20issue.priority%3A%5Bhigh%2C%20medium%5D%20podcast&referrer=issue-stream&sort=date

### How can this be tested?
with OPEN_PODCAST_DATA_BRANCH not set or set to the default run
```
from learning_resources.etl.pipelines import podcast_etl
podcast_etl()
```

you should see warning about missing topics and not being able to load glimpse podcast but the pipeline should finish

set OPEN_PODCAST_DATA_BRANCH=  ab/cleanup

run

```
from learning_resources.etl.pipelines import podcast_etl
podcast_etl()
```

again.
You should no longer see the warnings

### DIFF
--- a/learning_resources/etl/podcast.py
+++ b/learning_resources/etl/podcast.py
@@ -67,11 +67,8 @@ def validate_podcast_config(podcast_config):
         errors.append("Podcast data should be a dict")
         return errors
 
-    for required_key in ["rss_url", "website"]:
-        if required_key not in podcast_config:
-            errors.append(  # noqa: PERF401
-                f"Required key '{required_key}' is not present"
-            )
+    if "rss_url" not in podcast_config:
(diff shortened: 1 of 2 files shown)


mbertrand (approved):
LGTM, but would  be good to add a little extra unit test coverage though for the changed lines (like one that raises a `ConnectionResetError` for example)

PR #3383, opened 2026-05-28
refactor API mocks: Use Axios Adapter

ChristopherChudzicki:
Facilitates
- https://github.com/mitodl/mit-learn/pull/3380
- https://github.com/mitodl/mit-learn/pull/3364

This PR changes a bit how we mock API calls in tests:
1. **We no longer mock `axios` during tests.** Instead, we use an adapter to avoid making actual network requests. `setMockResponse` seeds data for the adapter. [^1]
    - We do *patch* axios via `jest.mock` during test setup to ensure that the adapter is always used.
2. **Changes `makeRequest` call signature.** This is mostly unrelated to the above change, but we needed to replace many `axios.request` test assertions to `makeRequest` anyway. I changed the `makeRequest` signature to be an object instead of positional arguments. This facilitates partial matching via `expect.objectContaining(...)`

**There is no change in how we actually use the mocks, apart from (2) above.**

**Motivation:** As part of https://github.com/mitodl/mit-learn/pull/3380, which is required for https://github.com/mitodl/mit-learn/pull/3364, `baseUrl` configuration (e.g., `https://api.learn.mit.edu/mitxonline` for the mitxonline api calls) moves from the client to the axios instance. This broke the way we had been asserting things.

This change makes our tests do a bit less mocking—real axios used—so are now less fragile to that sort of detail.

### How can this be tested?
Tests should pass.

**Note:** I also tested this by running `/scripts/test/jest-repeat.sh -- ''`, to run all tests 20 times. (Frontend tests, anyway)

### Why is the diff so big?

Mostly because:
1. There are more comments now
2. Roughly 30 `expect(axios.request).toHaveBeenCalledWith` calls were changed to `expect(makeRequest).toHaveBeenCalledWith`, and the new object form gets split across 4 or 5 lines, so ~100 new lines from that.

### DIFF
--- a/frontends/api/package.json
+++ b/frontends/api/package.json
@@ -12,6 +12,7 @@
     "./constants": "./src/common/constants.ts",
     "./ssr/*": "./src/ssr/*.ts",
     "./test-utils/factories": "./src/test-utils/factories/index.ts",
+    "./test-utils/mockAxios": "./src/test-utils/mockAxios.ts",
     "./test-utils": "./src/test-utils/index.ts",
     "./mitxonline-hooks/*": "./src/mitxonline/hooks/*/index.ts",
     "./mitxonline-test-utils": "./src/mitxonline/test-utils/index.ts"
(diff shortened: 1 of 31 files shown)


daniellefrappier18 (approved):
LGTM

PR #3384, opened 2026-05-28
Remove video_shorts app (phase 1)

mbertrand:
Related to mitodl/hq#11469

This is phase 1 of removing the legacy `video_shorts` app now that video shorts are ingested as learning resources.

- Adds a Django migration to delete the `VideoShort` model/table.
- Removes the legacy `video_shorts` API, webhook, backpopulate command, serializers, factories, and app-specific tests.
- Leaves `video_shorts` installed as a migration-only stub so deployed environments can discover and run the table-drop migration.
- Leaves a no-op Celery task shim for the old S3 deletion task path in case any queued tasks still reference it.
- Regenerates OpenAPI, frontend API clients, and load-testing clients so removed endpoints disappear from generated artifacts.

### How can this be tested?
- Confirm the migration  `video_shorts.0003_delete_videoshort` runs successfully.
- Confirm `/api/v0/video_shorts/` and `/api/v1/webhooks/video_shorts/` are absent from the regenerated OpenAPI specs.

### Additional Context
This intentionally does not remove the final `video_shorts` app package or migrations yet. Phase 2 should remove the remaining installed app stub only after `video_shorts.0003_delete_videoshort` has been applied.


shanbady (approved):
👍

PR #3385, opened 2026-05-28
Remove video_shorts app (phase 2)

mbertrand:
Depends on #3384
Closes mitodl/hq#11469

Phase 2 of removing the retired `video_shorts` Django app.

This removes the final migration-only app hook after Phase 1:

- removes `video_shorts` from `INSTALLED_APPS`
- deletes the remaining `video_shorts` package files
- deletes the historical `video_shorts` migrations and Phase 1 `DeleteModel` migration
- removes the no-op task shim that was kept only for Phase 1 compatibility

### How can this be tested?
Tests should pass


shanbady (approved):
LGTM (after deploying https://github.com/mitodl/mit-learn/pull/3384)

PR #3386, opened 2026-05-28
hybrid search sorting and dense-only cutoff score fix

shanbady:

Closes https://github.com/mitodl/hq/issues/11327

This PR also resolves an issue with broken sorting - this is for a specific path that we already treat in a special way to work around some qdrant api limitations - when we do hybrid search + q parameter + min_score

#### separate min scores for hybrid search vs dense only
When using dense only search (hybrid_search=false GET parameter) the min score is now set to a different and more appropriate threshold since we are not dealing with rescored hybrid+sparse results

#### Fix for hybrid search sorting when there is a query present
Qdrant does not support OrderBy queries when using a score_threshold (exception is thrown on the instance). To get around this we need to manually sort by the sortby field after fetching the results.

#### Testing https://github.com/mitodl/hq/issues/11327
1. currently on main/deployed envs hitting the [learning-resources vector search endpoint](https://api.learn.mit.edu/api/v0/vector_learning_resources_search/?q=ai) (without the hybrid_search=true parameter) returns (almost) all learning resources. This is due to a min score specific to hybrid search being applied to dense only results.
2. checkout this branch. make sure you have learning resources embedded. hit the same endpoint and the resulting list should look more reasonable. you can also adjust the score manually via the score_cutoff GET parameter (try score_cutoff=0.3)

#### Testing the sorting fix
1. on production login as an admin user
2. visit the search page. enable "vector hybrid search" from the admin options and search for something.
3. attempt to use the "sort by" drop down and note it does not work.
4. checkout this branch. make sure you have resources embedded and see that the sorting functionality works

### Additional Context
Due to qdrant's hard limitation around not being able to execute OrderBy queries that have a score_threshold, we must [sort the results manually](https://github.com/mitodl/mit-learn/pull/3376/changes#diff-07dd609f11fdcea7fbd8a4958cae0902c188992a60a40db1c17676b6e16d3562R511) after fetching it from qdrant (for this very specific branch where we are hitting the learning resource endpoint + have a query string + score_cutoff).


vector_search/views.py
```diff
@@ -461,21 +495,26 @@ async def async_vector_search(  # noqa: PLR0913
...
+                score_cutoff=normalized_score,
                 hybrid_search=hybrid_search,
             )
+            if order_by:
+                # in addition to fetching all results unpaginated,
+                # manually apply the sorting since
+                # Qdrant does not support sorting with score cutoffs
+                hits = sorted(
+                    hits,
+                    key=lambda x: _sort_key(x, order_by),
```
mbertrand: Descending sort is broken: `_sort_key` never negates the value and `sorted()` has no `reverse=`, so `-views` returns ascending order and `None` bubbles to the top. The new test `test_vector_search_sortby_with_score_cutoff_manually_sorted` asserts `[200,100,50]` but gets `[50,100,200]` — this is the red `python-tests`. Suggest sorting ascending then `reverse=` for descending, encoding None-placement independently of direction so missing values stay last either way.

For example, http://open.odl.local:8065/api/v0/vector_learning_resources_search/?q=ai&sortby=-created_on&hybrid_search=true  and http://open.odl.local:8065/api/v0/vector_learning_resources_search/?q=ai&sortby=created_on&hybrid_search=true both return results in ascending order

vector_search/views.py
```diff
@@ -461,21 +495,26 @@ async def async_vector_search(  # noqa: PLR0913
         *,
         hybrid_search: bool = False,
     ):
-        if (
-            query_string
-            and type(score_cutoff) is float
-            and score_cutoff >= VECTOR_SEARCH_MIN_SCORE
-        ):
+        normalized_score = _normalize_score_cutoff(score_cutoff, hybrid_search)
+        if query_string and normalized_score is not None:
```
mbertrand, line 497: `_normalize_score_cutoff` clamps with `max(value, min)` and only returns `None` for non-numeric input. Since the serializer defaults `score_cutoff` to `0.0`, `normalized_score` is never `None`, so this branch now fires for **every** non-empty query — fetching `VECTOR_SEARCH_PAGE_MAX_LIMIT` (200) hits unpaginated, ignoring the caller's `limit`/`offset`, and using `_async_vector_resource_counts` instead of the concurrent count path. That's a ~20× per-request regression on all vector search, not just explicit-cutoff requests. Gate this on an *explicitly supplied* cutoff (e.g. serializer `default=None`, enter only when the caller actually passed one), and slice back to `offset:offset+limit`.

learning_resources/serializers.py
```diff
@@ -1227,7 +1228,7 @@ class Meta:
             "resources",
             "etl_source",
             "view_count",
-            *COMMON_IGNORED_FIELDS,
+            "updated_on",
```
mbertrand, line 1231: This now serializes `created_on` at runtime, but `@extend_schema_serializer(exclude_fields=COMMON_IGNORED_FIELDS)` still hides it from the schema — so the API returns a field that's in neither `v0.yaml` nor the generated TS `LearningResource`.

vector_search/views.py
```diff
@@ -548,14 +587,21 @@ async def get(self, request):
             limit = request_data.data.get("limit", 10)
             offset = request_data.data.get("offset", 0)
             order_by = request_data.data.get("sortby")
+            score_cutoff = request_data.data.get(
+                "score_cutoff",
+                HYBRID_VECTOR_SEARCH_MIN_SCORE
+                if hybrid_search
+                else DENSE_VECTOR_SEARCH_MIN_SCORE,
```
mbertrand: This `.get()` default is dead code: the validated serializer always populates `score_cutoff` (default `0.0`), so the fallback never runs and the real clamping happens later in `_normalize_score_cutoff`. Misleading — drop it or consolidate the mode-aware defaulting in one place.

vector_search/views_test.py
```diff
@@ -407,24 +411,25 @@ def test_vector_search_sortby_parameter(mocker, client, query_string, hybrid_sea
...
             hybrid_search=hybrid_search,
         )
     )
 
     if query_string:
         call_kwargs = mock_qdrant.query_points.mock_calls[0].kwargs
-        assert isinstance(call_kwargs["query"], models.OrderByQuery)
-        assert call_kwargs["query"].order_by.key == "views"
-        assert call_kwargs["query"].order_by.direction == models.Direction.DESC
+        if not hybrid_search:
```
mbertrand: With `hybrid_search=True` this `if not hybrid_search:` guard means **no assertions run** for the hybrid parametrization — the hybrid+sortby path is nominally covered but verifies nothing. Add an explicit assertion for the hybrid case (e.g. no `OrderByQuery` passed and results come back manually sorted).

vector_search/views.py
```diff
@@ -130,9 +164,9 @@ async def _build_search_params(  # noqa: PLR0913
             ),
             "limit": limit,
         }
-
-        if type(score_cutoff) is float and score_cutoff >= VECTOR_SEARCH_MIN_SCORE:
-            search_params["score_threshold"] = score_cutoff
+        normalized_score = _normalize_score_cutoff(score_cutoff, hybrid_search)
+        if normalized_score is not None:
+            search_params["score_threshold"] = normalized_score
```
mbertrand, line 167: This always-set `score_threshold` (default `0.0` clamps up to the mode min) breaks the **dense + `sortby`** path. The dense branch below (`if order_by:`, ~line 209) lacks the guard the hybrid branch has at line 194, so it emits an `OrderByQuery` alongside the threshold. Qdrant rejects the combination and raises an unhandled `AioRpcError` (`StatusCode.INVALID_ARGUMENT`: `Bad request: Can't use score_threshold with an order_by query.`) → **500**.

Repro: `GET /api/v0/vector_learning_resources_search/?q=ai&sortby=created_on` (dense, no explicit `score_cutoff`). Because the default cutoff is always clamped to a non-None minimum, every dense + `sortby` request 500s.

Mirror the hybrid guard on the dense branch so the manual Python sort handles ordering:
```python
if order_by and "score_threshold" not in search_params:
```
Worth a test for the dense + `order_by` + threshold combo — the current tests mock `_resource_vector_hits` and never exercise `_build_search_params` here, so it slipped past CI.

mbertrand (changes requested):
Noticed a few issues, added some comments from Claude too, also it looks like the OpenAPI spec needs to be regenerated.

learning_resources/serializers.py
shanbady: This was intentional - what i found to happen is that simply adding this field required changing every frontend factory in addition to fixing tests and other scaffholding (pretty large diff). I was thinking of handling this stand-alone as a follow up if that makes sense.

mbertrand: makes sense

vector_search/views.py
```diff
@@ -461,21 +495,29 @@ async def async_vector_search(  # noqa: PLR0913
...
                 query_string,
                 params,
                 order_by=order_by,
                 limit=settings.VECTOR_SEARCH_PAGE_MAX_LIMIT,
                 offset=0,
                 search_collection=search_collection,
-                score_cutoff=score_cutoff,
+                score_cutoff=normalized_score,
                 hybrid_search=hybrid_search,
             )
```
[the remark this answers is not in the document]
shanbady: this is resolved now and the default for the sortby param is None. adjusted the tests to account for the change

vector_search/views.py
shanbady: fixed

vector_search/views.py
shanbady: vector learning resources is expected to always hit this branch if a query is specified since we will always have a min-clamped cutoff score. - the score cutoff parameter only exists on the learning resource serializer (its been intentionally omitted from the contentfile serializer) - in addition to that since we do clamp to the minimums, we will always only get a small handful of results.

vector_search/views.py
shanbady: fixed. changed it to request_data.data.get("score_cutoff")

vector_search/views.py
shanbady: Added a new test to excercise _build_search_params and the different conditions

mbertrand (approved):
👍

PR #3388, opened 2026-05-29
Improve channels API performance, remove all channels CRUD code

mbertrand:
Related to https://github.com/mitodl/hq/issues/9157

Improves the performance of the channels API in two ways:

- Removes most N+1 queries when serializing channels. A new `ChannelQuerySet.with_detail_relations()` applies the full `prefetch_related`/`select_related`/`annotate` chain the `ChannelSerializer` needs, so the list and detail endpoints render with a constant number of queries instead of scaling with the number of lists/sub-channels per channel.
- Caches channel responses  for all users 
- Removes all backend CRUD code for channels and some leftover frontend CRUD code (no longer needed).

Also adds a k6 load test for the channel endpoints used to measure the before/after improvement.

### How can this be tested?
**Automated tests**
```
docker compose run --rm web uv run pytest channels/ main/utils_test.py
```
Covers the prefetch query counts, the moderator/staff cache-bypass behavior, and the key-prefixed cache invalidation.

**Load test (k6) — measuring the perf improvement**

1. Seed channels with enough lists/sub-channels to make the N+1 difference measurable (idempotent — re-running replaces the existing `loadtest-*` channels):
   ```python
   docker compose exec -T web python manage.py shell <<'PY'
   from channels.factories import ChannelFactory, ChannelListFactory, SubChannelFactory
   from channels.models import Channel
   from learning_resources.factories import LearningPathFactory

   NUM_CHANNELS = 10
   LISTS_PER_CHANNEL = 12
   SUBCHANNELS_PER_CHANNEL = 8

   existing = Channel.objects.filter(name__startswith="loadtest-")
   print(f"deleting {existing.count()} existing loadtest channels")
   existing.delete()

   for i in range(NUM_CHANNELS):
       name = f"loadtest-{i}"
       channel = ChannelFactory.create(
           name=name,
           title=f"Load Test Channel {i}",
           is_topic=True,
           published=True,
           featured_list=LearningPathFactory.create().learning_resource,
       )
       ChannelListFactory.create_batch(LISTS_PER_CHANNEL, channel=channel)
       SubChannelFactory.create_batch(SUBCHANNELS_PER_CHANNEL, parent_channel=channel)
       print(f"  {name}: {channel.lists.count()} lists, "
             f"{channel.sub_channels.count()} sub_channels, "
             f"featured_list={channel.featured_list_id}")

   print(f"seeded {NUM_CHANNELS} rich loadtest channels")
   PY
   ```
2. Run the channels load test against your backend. `BACKEND_BASE_URL=http://learn.odl.local:8061` hits Django directly (no APISIX gateway required). The Docker helper mounts `load_testing/` at `/app` (note: `scripts/k6.sh` already supplies the `run` subcommand, so don't pass it yourself):
   ```
   ./scripts/k6.sh /app/channels.compare.ts \
     -e BACKEND_BASE_URL=http://learn.odl.local:8061 \
     -e RUN_LABEL=after -e VUS=15 -e DURATION=3m
     -e CHANNEL_NAME_PREFIX=loadtest-  
   ```
   The summary prints p50/p95/p99 per endpoint, and full results are written to `load_testing/results/channels..json`.


channels/views.py
```diff
@@ -125,31 +170,65 @@ def get_serializer_class(self):
     def perform_destroy(self, instance):
```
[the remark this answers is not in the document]
mbertrand: pathway type is dead/obsolete, probably should just be removed.  There are 0 pathways on production and likely always will be.

channels/views.py
```diff
@@ -201,30 +282,36 @@ def delete(self, request, *args, **kwargs):  # noqa: ARG002
...
+        if detail_relation:
+            queryset = queryset.select_related(detail_relation)
+        return queryset
 
-    @method_decorator(
-        cache_page_for_all_users(
-            settings.REDIS_VIEW_CACHE_DURATION, cache="redis", key_prefix="channels"
-        )
-    )
+    @method_decorator(cache_page_for_all_users(cache="redis", key_prefix="channels"))
```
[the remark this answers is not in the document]
mbertrand: The counts view seems like a special case, I prefer to not touch it for now

channels/models.py
```diff
@@ -132,6 +161,16 @@ def channel_url(self) -> str | None:
...
 
+    @property
+    def ordered_lists(self) -> list[LearningResource]:
+        """The channel's LearningPaths, ordered by position.
+
+        Resolves through the prefetched ``lists`` relation when available
+        (see ``ChannelQuerySet.with_detail_relations``) so serializing a
+        channel does not trigger an extra query per list.
+        """
+        return [channel_list.channel_list for channel_list in self.lists.all()]
```
abeglova, line 177: should this be ordered by position?

channels/models.py
```diff
@@ -47,6 +47,35 @@ def annotate_channel_url(self):
...
+        """
+        Apply the prefetch/select_related/annotate chain needed to serialize
+        a full ChannelSerializer without N+1 queries.
+        """
+        return (
+            self.prefetch_related(
+                Prefetch(
+                    "lists",
+                    queryset=ChannelList.objects.prefetch_related(
+                        "channel_list", "channel__lists", "channel__featured_list"
```
abeglova: i think channel__lists and channel__featured_list might not be used in the serializer?  Those are the lists for the parent channel, right?

channels/models.py
```diff
@@ -47,6 +47,35 @@ def annotate_channel_url(self):
...
+                    queryset=Channel.objects.annotate_channel_url(),
+                ),
+            )
+            .annotate_channel_url()
+            .select_related(
+                "featured_list",
+                "topic_detail",
+                "department_detail",
+                "unit_detail",
+                "pathway_detail",
```
abeglova, line 80: Would be nice to pull just the relevant one for the channel type when making an api call for one channel but probably it's not a big deal to pull them since the queries are small when the relevant detail is null

channels/serializers.py
```diff
@@ -292,6 +291,76 @@ def to_representation(self, instance):
         return serializer_cls(instance=instance, context=self.context).data
 
 
+def upsert_channel_lists(instance: Channel, validated_data: dict) -> None:
```
abeglova: I think all the channel editing functionality has been removed from the current UI. As far as i can tell the only way to currently edit a channel through the ui is through the admin dashboard. I also wasn't able to test through the api because i wasn't sure how to do a post as a user with the correct permissions

abeglova: But maybe i just didn't find it

channels/views.py
```diff
@@ -27,10 +28,71 @@
...
 from main.permissions import AnonymousAccessReadonlyPermission
-from main.utils import cache_page_for_all_users
+from main.utils import cache_page_for_all_users, clear_views_cache
 
 log = logging.getLogger(__name__)
 
+MODERATED_CHANNEL_IDS_REQUEST_ATTR = "_mit_learn_moderated_channel_ids"
+
+
+def get_moderated_channel_ids(request: Request) -> set[int]:
```
abeglova: The moderator channel cache bypass seems complicated. Is it necessary? Seems like the channels shouldn't be changed that often and also there are not currently channel moderators that are not staff

abeglova (changes requested):

channels/models.py
mbertrand: `ChannelList.Meta1` sets ordering = ["position"], so self.lists.all() is position-ordered even on the non-prefetched path.

`with_detail_relations` prefetches lists with an explicit .order_by("position") now, so the prefetched path is too.

I'll add a unit test to make sure though

channels/models.py
mbertrand: The `list` endpoint for all channels would need all of them I think,  inclined to think it's not worth creating different queries here for a channel detail view vs a list view.  On the other hand not sure if the list view is actually used for anything anyway.

abeglova (approved):
lgtm once you remove
[‎docs/superpowers/specs/2026-06-03-channel-readonly-design.md](https://github.com/mitodl/mit-learn/pull/3388/changes#diff-e58921dfdc564bcd4cba97e27a205712324aa05f4a08e9bf4463664ec9380502)

PR #3390, opened 2026-06-01
fix: render issue date on certificate pages

ahtesham-quraish:
https://github.com/mitodl/hq/issues/10711#issuecomment-4481397774

- Displayed the issue date on certificate pages where as validFrom date is removed because it was not coming in api response for multiple courses and programs

image

### How can this be tested?
- If you have mitxonline running locally and certificates set up, navigate to http://open.odl.local:8062/certificate/<"course"|"program">/<certificate_id>/pdf

- If not, grab a response from e.g. https://api.mitxonline.mit.edu/api/v2/course_certificates/19d376df-63e3-4924-81ad-722fc8638c57/ and return it in the certificateQueries.courseCertificatesRetrieve() hook queryFn and load http://open.odl.local:8062/certificate/course/19d376df-63e3-4924-81ad-722fc8638c57/


frontends/main/src/app-pages/CertificatePage/CertificatePage.tsx
```diff
@@ -590,23 +590,25 @@ const Certificate = ({
...
+  }
 }) => {
   const title = certificate.course_run.course.title
 
   const userName = certificate.user.name
 
   const { displayType } = getCertificateInfo()
   const signatories = certificate.certificate_page.signatory_items
-  const validFrom = certificate?.verifiable_credential_json?.validFrom
+  const issueDate = certificate?.issue_date
```
daniellefrappier18, line 601: What if there is no certificate.issue_date are we okay with this being blank?

daniellefrappier18 (approved):
Look good I was able to verify the test case. I did just have one question you should take a look at before merging. 

Image

frontends/main/src/app-pages/CertificatePage/CertificatePage.tsx
ahtesham-quraish: IMO there won't be any case where the issue date will be blank because this PR tells us that issue date will be there.
https://github.com/mitodl/mitxonline/pull/3587

PR #3397, opened 2026-06-02
feat: article listing page design implementation

ahtesham-quraish:

https://github.com/mitodl/hq/issues/11580

As a site visitor, I want to browse a paginated list of MIT articles, so I can discover and read the latest insights and discoveries from MIT.

### Desktop:
image

### Mobile:
image

### How can this be tested?
For testing we need to just create few articles in draft/publish state and visit the following url `/articles` here you would see the list of articles


frontends/main/src/app-pages/Articles/ArticleListingPage.tsx
```diff
@@ -1,182 +1,540 @@
...
+              {item.title}
+            </Link>
+          </StoryTitle>
+          {articleContent.paragraph && (
+            <StorySummary
+              dangerouslySetInnerHTML={{
+                __html: linkifyText(articleContent.paragraph),
+              }}
+            />
+          )}
```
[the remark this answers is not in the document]
ahtesham-quraish: I have changed the implementation check now. we need to retain the html because we can have some formatting on the sub-title

frontends/main/src/app-pages/Articles/ArticleListingPage.tsx
```diff
@@ -1,182 +1,541 @@
...
+                      ))}
+                    </Grid2>
+                  </GridContainer>
+                ) : null}
+              </AboveMdOnly>
+            </>
+          )}
+        </Container>
+
+        {!isLoading && gridStories.length > 0 && (
```
daniellefrappier18: I believe it's a UX best practice is to only show pagination when there is more than one page of results because displaying it unconditionally adds unnecessary UI noise and can confuse users when there's nothing to navigate. Currently, the pagination component renders regardless of the total number of pages.
```suggestion
        {!isLoading && gridStories.length > 0 && totalPages > 1 && (
```

frontends/main/src/common/websiteContentUtils.ts
```diff
@@ -1,3 +1,118 @@
...
+
+function extractText(nodes: ProseMirrorNode[] | undefined): string {
+  if (!nodes) return ""
+  return nodes
+    .filter((n) => n.type === "text" && typeof n.text === "string")
+    .map((n) => n.text as string)
+    .join("")
+}
+
+function nodesToHtml(nodes: ProseMirrorNode[] | undefined): string {
```
daniellefrappier18, line 48: `nodesToHtml` puts `mark.attrs.href `directly into an `<a>` tag after HTML-escaping, but `escapeHtml` doesn't strip javascript: URIs. Low risk since only article editors can create content, but easy to fix — suggest validating the protocol before emitting the href: `const safeHref = /^https?:\/\//i.test(href) ? escapeHtml(href) : "#"`.

frontends/main/src/app-pages/Articles/ArticleListingPage.tsx
```diff
@@ -1,182 +1,541 @@
...
+    </StoryCard>
   )
 }
 
 const ArticleListingPage: React.FC = () => {
-  const [page, setPage] = useState(1)
-  const scrollRef = useRef<HTMLDivElement>(null)
-  const canCreateArticle = useUserHasPermission(Permission.ArticleEditor)
+  const [searchParams, setSearchParams] = useSearchParams()
+  const parsedPage = Number.parseInt(searchParams.get("page") ?? "1", 10)
```
daniellefrappier18, line 418: If a user navigates to ?page=999 but only 2 pages of articles exist, the API returns results: [] (offset past the end) while count is still the full count. The component renders the "There are no news to display at this time" empty state, which is wrong because articles do exist

Suggested fix: After `totalPages` is known, redirect/reset to page 1 if the current page is out of bounds:
```
const totalPages = articles?.count ? getLastPage(articles.count) : 0

// If the page is out of bounds once we know the total, snap back to 1
React.useEffect(() => {
  if (!isLoading && totalPages > 0 && page > totalPages) {
    setSearchParams((current) => {
      const copy = new URLSearchParams(current)
      copy.delete("page")
      return copy
    })
  }
}, [isLoading, page, totalPages])
```

frontends/main/src/common/websiteContentUtils.ts
ahtesham-quraish: Addressed

frontends/main/src/app-pages/Articles/ArticleListingPage.tsx
```diff
@@ -1,182 +1,552 @@
...
+  ${theme.breakpoints.down("sm")} {
+    ${{ ...theme.typography.body2 }}
+  }
+`
 
+const BreadcrumbBar = styled.div(({ theme }) => ({
+  position: "relative",
+  left: "50%",
+  right: "50%",
+  marginLeft: "-50vw",
```
daniellefrappier18: The "escape container" technique uses width: 100vw with left: 50% / marginLeft: -50vw. On Windows and other systems where scrollbars occupy layout space. 100vw is wider than the usable page width. This creates a horizontal scrollbar on every page that uses this component.

Suggested Fix: Add overflow-x: clip on a suitable ancestor, or use `calc(100vw - (100vw - 100%))` as the width to account for scrollbar width:


```
width: "calc(100vw - (100vw - 100%))",
marginLeft: "calc(-50vw + (100vw - 100%) / 2)",
```

Image

frontends/main/src/app-pages/Articles/ArticleListingPage.tsx
```diff
@@ -1,182 +1,552 @@
...
+        <RegularStoryTitleWrapper>
+          <StoryTitle>
+            <Link href={articleView(item.slug ?? String(item.id))}>
+              {item.title}
+            </Link>
+          </StoryTitle>
+          {articleContent.paragraph && (
+            <StorySummary
+              dangerouslySetInnerHTML={{
+                __html: linkifyText(articleContent.paragraph),
```
daniellefrappier18: `articleContent.paragraph` is already HTML (produced by `nodesToHtml` in `websiteContentUtils.ts`). It contains tags like `<strong>`, `<em>`, etc. When linkifyText runs its regex `/(https?:\/\/[^\s]+)/g` on that HTML string, the pattern `[^\s]+` doesn't stop at `<`, so a URL at the end of an inline element gets matched together with the closing tag and will result in `<a href="https://example.com</strong>">`



```suggestion
                __html: articleContent.paragraph,
```

And remove the linkifyText import on line 23 if it's no longer used elsewhere in the file.

daniellefrappier18 (approved):
Sorry two things I missed the first time, after that you should be good.

PR #3405, opened 2026-06-02
add mynumbers game

abeglova:
https://github.com/mitodl/hq/issues/11688

This pr adds the password game/ hacksnack from 
https://www.dropbox.com/scl/fi/om7jj7nvyd0i3p003thjg/password_game.zip?rlkey=frntioog3h6pyvl2s4obdh1r1&st=ojchbpgl&dl=0 as game in learn behind a feature flag. The library is here 
https://github.com/mitodl/hacksnack/pull/1

Screenshot 2026-06-17 at 7 42 06 PM

### How can this be tested?
go to http://open.odl.local:8062/games/hacksnack you should get a 404

enable "hacksnack" as flag in posthog then go back to 
go to http://open.odl.local:8062/games/hacksnack

the game should load and work correctly

If you want the answers to see how the app responds to correct solutions you can get them in 
assets/puzzle_picker_list.json in https://github.com/mitodl/hacksnack/pull/1

### DIFF
--- a/frontends/main/package.json
+++ b/frontends/main/package.json
@@ -14,6 +14,7 @@
     "@emotion/cache": "^11.13.1",
     "@emotion/styled": "^11.11.0",
     "@floating-ui/react": "^0.27.16",
+    "@mitodl/arithmix": "^0.2.2",
     "@mitodl/course-search-utils": "^3.5.2",
     "@mitodl/mitxonline-api-axios": "2026.6.10",
     "@mitodl/smoot-design": "^6.27.0",
--- a/frontends/main/src/app/(site)/games/arithmix/[[...rest]]/ArithmixClient.tsx
+++ b/frontends/main/src/app/(site)/games/arithmix/[[...rest]]/ArithmixClient.tsx
@@ -1,0 +1,24 @@
+"use client"
+
+import React from "react"
+import dynamic from "next/dynamic"
+import ArithmixFlagGate from "./ArithmixFlagGate"
+
+// The mynumbers package accesses `document` at module-evaluation time, so it
+// must be loaded client-side only (no SSR).
+const Arithmix = dynamic(
+  () => import("@mitodl/arithmix").then((mod) => mod.Arithmix),
+  {
+    ssr: false,
+  },
+)
+
+const ArithmixClient: React.FC = () => {
+  return (
+    <ArithmixFlagGate>
+      <Arithmix basename="/games/arithmix" />
+    </ArithmixFlagGate>
+  )
+}
+
+export default ArithmixClient
--- a/frontends/main/src/app/(site)/games/arithmix/[[...rest]]/ArithmixClient.test.tsx
+++ b/frontends/main/src/app/(site)/games/arithmix/[[...rest]]/ArithmixClient.test.tsx
@@ -1,0 +1,95 @@
+import React from "react"
+import { renderWithProviders, screen } from "@/test-utils"
+import { useFeatureFlagEnabled } from "posthog-js/react"
+import { notFound } from "next/navigation"
+import { allowConsoleErrors } from "ol-test-utilities"
+import { useFeatureFlagsLoaded } from "@/common/useFeatureFlagsLoaded"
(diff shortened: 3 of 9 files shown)


frontends/main/package.json
```diff
@@ -54,6 +54,7 @@
     "lodash": "^4.17.21",
     "lodash.throttle": "^4.1.1",
     "moment": "^2.30.1",
+    "mynumbers": "https://github.com/mitodl/mynumbers.git#ab/make-it-react",
```
abeglova: this will be updated once https://github.com/mitodl/mynumbers/pull/1/ is approved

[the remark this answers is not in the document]
abeglova: The plan is to push the library to npm once https://github.com/mitodl/mynumbers/pull/1 is approved

frontends/main/src/app/(site)/games/arithmix/[[...rest]]/ArithmixClient.tsx
```diff
@@ -0,0 +1,30 @@
...
+// The mynumbers package accesses `document` at module-evaluation time, so it
+// must be loaded client-side only (no SSR).
+const Arithmix = dynamic(
+  () => import("mynumbers").then((mod) => mod.Arithmix),
+  {
+    ssr: false,
+  },
+)
+
+const ArithmixClient: React.FC = () => {
```
Comment (ahtesham-quraish) [inline frontends/main/src/app/(site)/games/arithmix/[[...rest]]/ArithmixClient.tsx:16]: **Suggestion:** this feature-flag gating logic is now duplicated in both `ArithmixClient` and `ExplainerClient`.

It may be worth extracting a small shared helper/component for:

- reading `FeatureFlags.Arithmix`
- waiting for flags to load
- returning `notFound()` when disabled

That would reduce duplication and make future Arithmix routes less error-prone.

frontends/main/src/app/(site)/arithmix/explainer/ExplainerClient.test.tsx
```diff
@@ -0,0 +1,52 @@
...
+}))
+
+jest.mock("posthog-js/react")
+
+const mockUseFeatureFlagEnabled = jest.mocked(useFeatureFlagEnabled)
+const mockUseFeatureFlagsLoaded = jest.mocked(useFeatureFlagsLoaded)
+const mockNotFound = jest.mocked(notFound)
+
+describe("ExplainerClient", () => {
+  test("renders the ExplainerPage when the feature flag is enabled", async () => {
```
ahtesham-quraish: **Suggestion:** since this component depends on a dynamic import of an external package, can we add a test for the import-failure path as well?

Right now we cover enabled/loading/disabled flag states, but not what happens if `mynumbers` fails to load or exports an unexpected shape. That feels like an especially relevant edge case for a git-sourced dependency.

ahtesham-quraish (changes requested):
Overall looks fine just few comments

frontends/main/src/app/(site)/arithmix/explainer/ExplainerClient.test.tsx
abeglova: I added code to return a 404 if mynumbers fails to load.  It will not be a git sourced dependancy long term. The plan is to push the library to npm once https://github.com/mitodl/mynumbers/pull/1 is approved,

ahtesham-quraish (changes requested):
Overall looks good now following comment needs attention 

- The gate's behavior is tested indirectly through ArithmixClient.test.tsx and ExplainerClient.test.tsx. But since ArithmixFlagGate is now the shared, authoritative piece of logic for all current and future Arithmix routes, a dedicated ArithmixFlagGate.test.tsx would make the contract explicit and protect against regressions if client tests are refactored. Not a blocker, but worth calling out.

ahtesham-quraish (approved):
LGTM!

ChristopherChudzicki:
@abeglova I pushed the commit with the [optional catch-all route group](https://nextjs.org/docs/pages/building-your-application/routing/dynamic-routes#optional-catch-all-segments) that we discussed on zoom.

PR #3408, opened 2026-06-03
hide courseware cards in b2b contracts when variant selected if no variant run

gumaerc:
n/a

This PR sets up the frontend to completely hide courseware cards in the situation where the user has selected a variant but there are no variant runs available for that course.

### How can this be tested?
Requires a local mitxonline instance running the code from mitxonline PR #3618 (now merged to main).

1. Ensure the `mitxonline` backend is running with the merged variant options changes
2. Follow the instructions in this PR to create proper test data in `mitxonline`, but make sure that for one of the courses you do *not* create a variant run in one of the variants: https://github.com/mitodl/mitxonline/pull/3618
4. Load the B2B dashboard for that contract — the variant picker should appear if there are 2+ variant options
5. Select the variant with the missing course run and verify that the card is hidden

### DIFF
--- a/frontends/main/src/app-pages/DashboardPage/CoursewareDisplay/hooks/useContractDashboardData.ts
+++ b/frontends/main/src/app-pages/DashboardPage/CoursewareDisplay/hooks/useContractDashboardData.ts
@@ -19,7 +19,6 @@ import type {
 } from "@mitodl/mitxonline-api-axios/v2"
 import {
   buildCourseEntry,
-  entryMatchesVariant,
   getCollectionFirstCoursesInDisplayOrder,
   getProgramCoursesInContractOrder,
   getRenderableContractCollections,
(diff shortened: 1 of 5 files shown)


frontends/main/src/app-pages/DashboardPage/CoursewareDisplay/hooks/useContractDashboardData.ts
```diff
@@ -131,6 +131,15 @@ const useContractDashboardData = (
     return map
   }, [isDefaultVariantSelection, variantRunsQuery.data])
 
+  // True while a non-default variant is selected but the lazy variant-runs
+  // query has not yet resolved (status = 'pending', i.e. no data yet).  Using
+  // isPending (not isLoading) ensures this is true on the very first render
+  // after the picker changes — before React Query has started the fetch and set
+  // isFetching.  While loading, we skip passing the variant to buildCourseEntry
+  // so that entries are not filtered out before the run data arrives.
+  const variantRunsLoading =
```
ChristopherChudzicki, line 140: IMO we should add a test to `dashboardViewModel.test.ts` about the null return case.

frontends/main/src/app-pages/DashboardPage/CoursewareDisplay/model/dashboardViewModel.test.ts
```diff
@@ -768,7 +768,7 @@ describe("dashboardViewModel", () => {
         courseruns: [run],
         next_run_id: run.id,
       })
-      const entry = buildCourseEntry(course, [], {})
+      const entry = buildCourseEntry(course, [], {})!
```
ChristopherChudzicki, line 771: Not worth changing right now, but my inclination is always to use an `invariant(...)` guard for narrowing rather than non-null assertions

ChristopherChudzicki (approved):
👍

PR #3413, opened 2026-06-03
make healthcheck route nocache

ChristopherChudzicki:
None, followup to #3364 

Makes https://learn.mit.edu/healthcheck have `Cache-Control: no store`

### How can this be tested?
View http://open.odl.local:8062/healthcheck and check that it has the Cache-Control: no-store header in dev tools.


abeglova (approved):
LGTM but please update the description

PR #3416, opened 2026-06-03
adding fix for topic page search

shanbady:

Related/Follow fix for https://github.com/mitodl/hq/issues/11167

This PR ensures that the hybrid search component preserves/passes along pre-existing filters when performing a search with a query (in this case the pre-existing filter associated with each channel/topic page)

### How can this be tested?
1. on the current rc - open web inspector and go to a topic page. https://rc.learn.mit.edu/c/topic/business-analytics
2. note the vector search endpoint that gets hit which by default includes the topic=Business+Analytics query param
3. search for something and note that the vector endpoint does not have the topic filter/param
4. on this branch perform the same and note that the "topic" parameter is carried over when searching


abeglova (approved):
lgtm

PR #3418, opened 2026-06-04
add edx_module_id back into the vector content file serializer

shanbady:

Closes https://github.com/mitodl/hq/issues/11628

this [commit](https://github.com/mitodl/mit-learn/commit/15d76830d) had inadvertadly removed the edx_module_id as a parameter from the vector contentfiles request serializer assuming it was not used as a filter parameter however it is what the video bot uses to filter results.

### How can this be tested?
1. checkout this branch
2. make sure you have contentfiles local and also embedded
3. load /api/v0/vector_content_files_search?file_extension=.srt and find a record and copy the edx_module_id
4. you should be able to filter the vector contentfiles search endpoint by that module id (after url encoding it) via the edx_module_id parameter

### Additional Context
This is currently causing the video bots to fail silently (or give odd responses) since it does not filter the chunks by edx_module_id


mbertrand (approved):
👍

PR #3419, opened 2026-06-04
fix: remove event section if no event data available

ahtesham-quraish:
https://github.com/mitodl/hq/issues/11614

 - Hide the event section if no event data is available 

### Before
image

### After
image

### How can this be tested?
For testing you need to just return 0 event in apis response and you will see the News section will stretch to whole available space and event section will be hide


ChristopherChudzicki (approved):
Working well!

ChristopherChudzicki:
@ahtesham-quraish I am going to merge and release this so we can get it out before the weekend.

Your PR is the only one not in main right now, so will be a tiny / safe release late thursday.

PR #3422, opened 2026-06-04
add s-maxage=0, fastly ignores no-store

ChristopherChudzicki:
Followup to #3413 

Gives healthcheck route `s-maxage=0`. Turns out fastly ignores `no-store`:

>  Directives ignored by Fastly include, but are not limited to, Cache-Control: no-cache, Cache-Control: no-store, and Cache-Control: must-revalidate.
> – https://www.fastly.com/documentation/guides/full-site-delivery/caching/about-cache-control-headers/

### How can this be tested?
Visit http://open.odl.local:8062/healthcheck and check the header.


abeglova (approved):
lgtm

PR #3424, opened 2026-06-04
Sort variants + Variant Picker UI Updates

ChristopherChudzicki:
None

This PR:
- Updates the variant picker UI a bit, see screenshot
- sorts variants by (language, industry, length), but ensures the group matching default is shown first
- fixes a test that was failing on main
- Adds a focus ring when `focus-visible` to distinguish between selected vs selected+focused-via-keyboard. (Only affects keyboard users).

Screenshot 2026-06-04 at 1 00 32 PM

Keyboard-only focus ring:
Screenshot 2026-06-04 at 2 32 56 PM

### How can this be tested?
1. View a contract with variants set up. It should look like above, and default should be sorted to first card.


jkachel (approved):
LGTM - worked well via mouse or keyboard nav. 👍

PR #3427, opened 2026-06-04
make selected vs selected+focus-visible different

ChristopherChudzicki:
Tiny followup to https://github.com/mitodl/mit-learn/pull/3424

This is a commit i forgot to push on https://github.com/mitodl/mit-learn/pull/3424 🤦 .

It makes the keyboard-only focus ring on variant picker match the rest of the site.

Keyboard-only focus ring:
Screenshot 2026-06-04 at 2 32 56 PM

### How can this be tested?
1. View a contract with variants, tab to the variant picker. Should see blue ring.

#### Additional Notes
No rush on this, going to wait till @gumaerc is back.


gumaerc (approved):
👍 LGTM. I didn't see a focus style in Figma which is why I didn't implement this originally, but good call.

PR #3428, opened 2026-06-04
boost programs index in search results

abeglova:
https://github.com/mitodl/mit-learn/issues/3429

This pr boosts the scores of programs in search results

### How can this be tested?
Run 
docker compose run --rm web python manage.py backpopulate_mitxonline_data
go to 
http://open.odl.local:8062/search?q=supply+chain

The SCM program should be the first result

set 
SEARCH_PROGRAM_INDEX_BOOST=1 in your environment

go to 
http://open.odl.local:8062/search?q=supply+chain

the  SCM program will no longer be your first result


mbertrand (approved):
LGTM, copilot's nitpick on the noqa formatting seems valid even if trivial; the other comment about `SEARCH_PROGRAM_INDEX_BOOST` I think might be safe to ignore.

PR #3430, opened 2026-06-05
fix: add link to the creation page of article and news

ahtesham-quraish:
N/A

 - Add link to creation of article and news pages.

Screenshot 2026-06-05 at 4 50 14 PM

image

image

### How can this be tested?
For testing you need to go to following urls and see the button added for article and news creation pages
`/articles`
`/news`
and we have added the links inside the UserMenu as well

### DIFF
--- a/frontends/main/src/app-pages/Articles/ArticleListingPage.tsx
+++ b/frontends/main/src/app-pages/Articles/ArticleListingPage.tsx
@@ -17,7 +17,6 @@ import {
   Breadcrumbs,
 } from "ol-components"
 import Link from "next/link"
-import { useRouter } from "next-nprogress-bar"
 import { RiArrowLeftLine, RiArrowRightLine } from "@remixicon/react"
 import type { WebsiteContent } from "api/v1"
 import { LocalDate } from "ol-utilities"
@@ -25,7 +24,7 @@ import { useWebsiteContentList } from "api/hooks/website_content"
 import { extractArticleContent } from "@/common/websiteContentUtils"
 import { articleView, websiteContentCreateView } from "@/common/urls"
 import { Permission, useUserHasPermission } from "api/hooks/user"
-import { Button } from "@mitodl/smoot-design"
+import { ButtonLink } from "@mitodl/smoot-design"
 
 const PAGE_SIZE = 10
 const MAX_PAGE = 50
@@ -306,7 +305,7 @@ const BannerSection = styled.div`
   background: ${theme.custom.colors.white};
   border-bottom: 1px solid ${theme.custom.colors.lightGray2};
 `
-const NewArticleLink = styled(Button)`
+const NewArticleLink = styled(ButtonLink)`
   display: flex;
   justify-content: end;
 `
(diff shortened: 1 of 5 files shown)


frontends/main/src/app-pages/Articles/ArticleListingPage.tsx
```diff
@@ -454,11 +457,16 @@ const ArticleListingPage: React.FC = () => {
...
+            {isArticleEditor && (
+              <NewArticleLink
+                variant="primary"
+                onClick={() => router.push(websiteContentCreateView("article"))}
+              >
+                <Typography variant="body1" color="white">
+                  New Article
+                </Typography>
+              </NewArticleLink>
+            )}
```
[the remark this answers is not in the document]
ahtesham-quraish: We don't need to keep the description

frontends/main/src/app-pages/Articles/ArticleListingPage.tsx
```diff
@@ -299,6 +306,10 @@ const BannerSection = styled.div`
   background: ${theme.custom.colors.white};
   border-bottom: 1px solid ${theme.custom.colors.lightGray2};
 `
+const NewArticleLink = styled(Button)`
```
daniellefrappier18: Both `NewArticleLink` components are `styled(Button`) with `onClick={() => router.push(...)}`. This is semantically incorrect for navigation: screen readers announce `<button>` as an interactive action trigger, not a link. Right-click → "Open in new tab" won't work. The library already exports ButtonLink specifically for this use case.

Fix: Replace `styled(Button)` with `styled(ButtonLink)` and drop the `onClick/useRouter` entirely. ButtonLink accepts href and renders an `<a>` element:


```
const NewArticleLink = styled(ButtonLink)``
// ...
<NewArticleLink variant="primary" href={websiteContentCreateView("article")}>
  New Article
</NewArticleLink>
```
This also makes useRouter unnecessary in ArticleListingPage, which can be removed.

frontends/main/src/app-pages/News/NewsListingPage.test.tsx
```diff
@@ -23,6 +23,7 @@ describe("NewsListingPage", () => {
   beforeEach(() => {
     mockedUseFeatureFlagEnabled.mockReturnValue(true)
     mockedUseFeatureFlagsLoaded.mockReturnValue(true)
+    setMockResponse.get(urls.userMe.get(), { is_authenticated: false })
```
daniellefrappier18, line 26: The only change is adding a baseline mock for the unauthenticated case. There are no tests verifying the "Add news" / "New Article" button appears when is_article_editor: true, or is hidden when false. The UserMenu.tsx changes are also untested for the new "Article" and "News" menu items.

Suggested Fix: Add two cases — one with is_article_editor: true asserting the button is visible, one without asserting it's absent.

daniellefrappier18 (commented):
Looks good. I was able to verify the changes. I left two small things that probably need a look before merging.

Image
Image
Image
Image

daniellefrappier18 (approved):
Looks good

PR #3432, opened 2026-06-05
fix podcasts type

ChristopherChudzicki:
None

Fixes `learning_resource.podcast_episode.podcasts` type. It was incorrectly `str[]` but should be `number[]`

### How can this be tested?
1. Check that the API results actually show `number[]` for the `podcast_episode.podcasts` property at https://api.learn.mit.edu/api/v1/podcast_episodes/
2. And now the spec does, too.

### Additional Context

> [!NOTE]
> oasdiff will classify this as a breaking change. It's not. It's a spec bugfix.

### DIFF
--- a/frontends/main/src/page-components/LearningResourceDrawer/LearningResourceDrawer.tsx
+++ b/frontends/main/src/page-components/LearningResourceDrawer/LearningResourceDrawer.tsx
@@ -197,7 +197,7 @@ const DrawerContent: React.FC<{
       bottomCarousels.push(
         itemsCarousel(
           "Other Episodes in this Podcast",
-          parseInt(resource.data.podcast_episode.podcasts[0]),
+          resource.data.podcast_episode.podcasts[0],
           resourceId,
         ),
       )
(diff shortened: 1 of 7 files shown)


frontends/main/src/page-components/LearningResourceDrawer/LearningResourceDrawer.tsx
```diff
@@ -194,7 +194,7 @@ const DrawerContent: React.FC<{
     bottomCarousels.push(
       itemsCarousel(
         "Other Episodes in this Podcast",
-        parseInt(resource.data.podcast_episode.podcasts[0]),
+        resource.data.podcast_episode.podcasts[0],
```
ChristopherChudzicki: `parseInt` here was unnecessary because `podcasts[0]` was already an integer.

zamanafzal (approved):
Great catch on the podcasts type fix! 👍

While looking at this, I noticed there's a sibling bug in [VideoResourceSerializer.get_playlists](https://github.com/mitodl/mit-learn/blob/main/learning_resources/serializers.py#L1484-L1486) — same pattern: parent_id is an integer FK, but the return type is annotated as list[str]. 

Should we cover it in a follow-up with the same treatment?

PR #3433, opened 2026-06-08
fix: pick cover image of video if no image is inserted in article content

ahtesham-quraish:
https://github.com/mitodl/hq/issues/11666

As a content editor, I want news articles to always have a cover image in the news feed in case video is embedded, so that listings and social previews display correctly even when I haven't explicitly set a featured image.

### How can this be tested?
For this you just need to embed the video using the following urls 
`https://video.odl.mit.edu/videos/753ca8bda51d43c99a41e0922a82e1fc/embed/`
`https://rc.learn.mit.edu/video/121583/embed`
after that you should publish the article and visit the following urls and see if the image is there against that particular news or article
`/news`
`/articles`

### Backpopulation plan

New cover_image field is auto-derived on save() going forward. Existing records need a one-time backfill.

Why not a data migration: Django's overridden save() isn't called by migrations, and calling it via RunPython would fire Fastly purge signals and slug-generation logic on every row unnecessarily.

The management command calls extract_image_from_content directly and writes only the cover_image column via queryset.update(), so no signals fire and no slug logic runs.


news_events/etl/articles_news.py
```diff
@@ -114,10 +116,13 @@ def transform_items(articles_data: list[dict]) -> list[dict]:
...
         content_text = extract_text_from_content(content_json)
 
-        # Extract first image from content
-        image_data = extract_image_from_content(content_json)
+        cover_image_url = article.get("cover_image", "")
+        image_data = (
+            {"url": cover_image_url, "alt": "", "description": ""}
+            if cover_image_url
+            else None
+        )
```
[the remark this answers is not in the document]
ahtesham-quraish: We are using the fallback when we create the cover-image field so we don't need it

frontends/main/src/page-components/TiptapEditor/extensions/node/ByLineInfoBar/ByLineInfoBar.tsx
```diff
@@ -34,8 +34,9 @@ const InnerContainer = styled(Container, {
   shouldForwardProp: (prop) => prop !== "noAuthor",
 })<{ noAuthor?: boolean }>(({ noAuthor }) => ({
   display: "flex",
-  justifyContent: noAuthor ? "flex-end" : "space-between",
+  justifyContent: noAuthor ? "space-between" : "space-between",
   alignItems: "center",
```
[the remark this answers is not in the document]
ahtesham-quraish: this is intended change

frontends/main/src/page-components/TiptapEditor/extensions/node/ByLineInfoBar/ByLineInfoBar.tsx
```diff
@@ -126,38 +127,38 @@ export const ByLineInfoBarContent = ({
...
-                  })
-                : isEditable
-                  ? null
-                  : "Draft"}
-            </InfoText>
-          </InfoContainer>
-        )}
+      <InnerContainer noAuthor={!isEditable}>
+        <InfoContainer>
+          {isEditable ? (
```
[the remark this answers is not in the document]
ahtesham-quraish: This is fine

frontends/main/src/page-components/TiptapEditor/extensions/node/MediaEmbed/MediaEmbedNode.ts
```diff
@@ -47,6 +47,15 @@ export const MediaEmbedNode = Node.create({
             ? { "data-mit-learn-video-id": String(attrs.mitLearnVideoId) }
             : {},
       },
+      coverImageUrl: {
```
ChristopherChudzicki: This adds a `coverImageUrl` to mediaEmbedNode, right? Do we need to expand the node's data? Can't `extract_image_from_content` get the cover image based on the node's `mitLearnVideoId: 1234` ?

In general, nice to avoid expanding the data we store on the nodes. (E.g., how do we backpopualte it?)

website_content/utils.py
```diff
@@ -1,10 +1,206 @@
...
+    def traverse_for_image(node):  # noqa: C901, PLR0911, PLR0912
+        """Recursively traverse JSON structure to find first image"""
+        if not node:
+            return None
+
+        # Handle dict nodes
+        if isinstance(node, dict):
+            # Check if this node is an image node (common patterns)
+
+            # ProseMirror imageWithCaption (your format)
```
ChristopherChudzicki: "(your format)" 

?

website_content/utils.py
```diff
@@ -1,10 +1,206 @@
...
+            # ProseMirror image node
+            if node.get("type") == "image":
+                attrs = node.get("attrs", {})
+                if attrs.get("src"):
+                    return {
+                        "url": attrs["src"],
+                        "alt": attrs.get("alt", attrs.get("title", "")),
+                        "description": attrs.get("alt", attrs.get("title", "")),
+                    }
+                # EditorJS image block
```
ChristopherChudzicki: What is EditorJS?

website_content/utils.py
```diff
@@ -1,10 +1,206 @@
...
+    Args:
+        content_json (dict): The JSON content from Article
+
+    Returns:
+        dict | None: Image data dict with url, alt, description or None
+    """
+    if not content_json:
+        return None
+
+    def traverse_for_image(node):  # noqa: C901, PLR0911, PLR0912
```
ChristopherChudzicki: In `traverse_for_image`, I see a lot of conditional branches:

1. `if node.get("type") == "imageWithCaption"` ... this is what our editor uses
2. `if node.get("type") == "image"`, which has two sub-branches
    - A)  sub-branch for ProseMirror
    - B)  sub-branch for EditorJS (I think?) That's a completely different WYSIWYG editor.
3. `if "image" in node or "img" in node:` ... Doesn't happen in TipTap, right? Nodes are always `{ type, attrs }` ? 
4. `if node.get("url") and ...` Again, this doesn't happen in TipTap nodes

Of those four branches, only the first ever actual matches, right? The rest are dead code?

I'd get rid of at least 2B,  3 and 4. I can see an argument for keeping 2A. At least it's prosemirror/tiptap schema.

**Suggestion:** I think you should add a module-level TypedDict akin to
```python
class ProseMirrorNode(TypedDict):  
    type: str
    attrs: NotRequired[dict[str, Any]]     # may be absent
    content: NotRequired[list["ProseMirrorNode"]]  # may be absent (leaf nodes)
```

This would help document the structure we are after. It doesn't encode the whole schema/typing (☝️ leaves `attrs` as `Any`) but it would help readers understand.

website_content/utils.py
```diff
@@ -1,10 +1,206 @@
...
+                for ext in [".jpg", ".jpeg", ".png", ".gif", ".webp", ".svg"]
+            ):
+                return {
+                    "url": node["url"],
+                    "alt": node.get("alt", ""),
+                    "description": node.get("description", node.get("caption", "")),
+                }
+
+            # Traverse nested dict values
+            for value in node.values():
```
ChristopherChudzicki: I don't think we should recurse into `node.values()`.

TipTap nodes are always of the form `{ type, attr, content }` (Please double check this. That's my recollection).

I think we should only recurse into `content` ?

website_content/utils.py
```diff
@@ -1,10 +1,206 @@
...
+        if cover:
+            return {"url": cover, "alt": "", "description": ""}
+        src = attrs.get("src", "")
+        if src:
+            parsed = urlparse(src)
+            if (
+                parsed.scheme in {"http", "https"}
+                and parsed.hostname == "video.odl.mit.edu"
+            ):
+                thumbnail_url = fetch_odl_video_thumbnail(src)
```
ChristopherChudzicki: **Request** I think you should drop support for "infer cover image from video.odl.mit.edu embed pages" unless it's absolutely required. I don't think it is, though.

The whole reason we added the learn-internal video embed route was to stop using the OVS (video.odl.mit.edu) route. And for the Learn route, https://learn.mit.edu/video/embed/116608, you know the resource ID, so you can get the image from the resource directly.

Plus:
- it requires an external network request to OVS **during a db save**. This seems hazardous.
- it grabs the image by parsing the `var SETTINGS = ...` thing. That's an internal detail to OVS. We should not rely on it.

website_content/models.py
```diff
@@ -58,6 +62,11 @@ def save(self, *args, **kwargs):
                 counter += 1
 
         self.slug = slug
+
+        if not self.cover_image:
+            image_data = extract_image_from_content(self.content)
+            self.cover_image = image_data.get("url", "") if image_data else ""
```
ChristopherChudzicki: This only sets the cover image once, and never again, right? I think we want it to update if author changes the first image, or if they remove the image.

Also, should add some tests to cover this model save() behavior.

website_content/migrations/0005_add_cover_image_to_websitecontent.py
```diff
@@ -0,0 +1,17 @@
+# Generated by Django 4.2.30 on 2026-06-08 09:17
+
+from django.db import migrations, models
```
ChristopherChudzicki, line 3: We should have a backpopulate ~Calling `.save()` on every object should be fine for the migration, I think. (Particularly after removing the OVS fetch.)~ Apparently not, I guess the overriden method isn't available in migrations

Still, we should have a plan for backpopulating before merging. Even just a shell script would be OK. Could include it in your PR description.

frontends/main/src/page-components/TiptapEditor/extensions/node/ByLineInfoBar/ByLineInfoBar.tsx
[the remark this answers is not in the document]
ChristopherChudzicki: Worth a comment on the PR description. It's unrelated to everything.

But also... the ternary is unnecessary now. It should just be `space-between`.

frontends/main/src/common/websiteContentUtils.ts
```diff
@@ -106,7 +106,13 @@ export function extractArticleContent(
             typeof imageAttrs.caption === "string" ? imageAttrs.caption : null,
         }
       : null
-
+  if (image === null && article?.cover_image) {
```
ChristopherChudzicki: Shouldn't `article?.cover_image` be the source of truth? I.e., use that as the first priority, then fall back to the other thing?

I would _prefer_ to rid of the `const imageNode = topLevel.find((n) => n.type === "imageWithCaption")` that tries to infer it. Trust the backend data. However, we may want to do that as a followup instead of as part of this PR, waiting until after we've backpopulated the data.

website_content/serializers.py
```diff
@@ -36,6 +36,7 @@ class WebsiteContentSerializer(serializers.ModelSerializer):
     content = serializers.JSONField(default=dict)
     slug = serializers.SlugField(max_length=60, required=False, allow_blank=True)
     title = serializers.CharField(max_length=255)
+    cover_image = serializers.URLField(max_length=2083, allow_blank=True, default="")
```
ChristopherChudzicki:
```suggestion
    cover_image = serializers.URLField(max_length=2083, allow_blank=True, default="", read_only=True)
```

ChristopherChudzicki (changes requested):
Left some feedback.

One of my biggest comments is I don't think we should support inferring cover images from the OVS embed route `https://video.odl.mit.edu/video/uuid/embed`. The way it's done is fragile, and—I realize we've handled it manually two or three times now—but we added the internal learn embed route as a replacement for that OVS route. So let's just support the internal learn route.

website_content/utils.py
ahtesham-quraish: removed

website_content/utils.py
ahtesham-quraish: it was old fallback block which is not needed now so I removed it

website_content/utils.py
```diff
@@ -1,10 +1,110 @@
...
+def _image_from_media_embed(attrs: dict) -> dict | None:
+    mit_learn_id = attrs.get("mitLearnVideoId")
+    if not mit_learn_id:
+        return None
+
+    from learning_resources.models import LearningResource
+
+    try:
+        resource = LearningResource.objects.get(id=mit_learn_id)
+        url = resource.video and resource.video.cover_image_url
```
ChristopherChudzicki: Seems rare, but if the `resource` isn't a video, `resource.video` will throw:

```python
c = LearningResource.objects.filter(resource_type='course').first()
c.video # throws RelatedObjectDoesNotExist
```

Could do:
```python
video = Video.objects.filter(learning_resource_id=mit_learn_id).first()
if video is None:
    log.warning(
        "mediaEmbed references mitLearnVideoId=%s but no Video was found",
        mit_learn_id,
    )
    return None

if video.cover_image_url:
    return {"url": video.cover_image_url, "alt": "", "description": ""}
```

ChristopherChudzicki (approved):
Looks good! Thanks for making all those changes.

I noticed one small new issue—probably won't show up in practice, but would be nice to handle. After that, should be good to merge. I don't think it needs a re-review.

PR #3436, opened 2026-06-08
remove duplicate resource task

shanbady:

Closes https://github.com/mitodl/hq/issues/10835

This PR removes a celery task that we added to remove instances of learning resources that had duplicate resource ids and were unpublished (this had previously led to missing embeddings). Since we now key off of platform+readable id, this is no longer an issue (see https://github.com/mitodl/hq/issues/10219)

### How can this be tested?
There are no functional changes. Tests should pass


abeglova (approved):
👍

PR #3442, opened 2026-06-08
feat: redemption_status migration + Assign Seats CSV flow

daniellefrappier18:

Closes https://github.com/mitodl/hq/issues/11449

Updates the B2B contract admin page across two areas:

**1. redemption_status migration**

Adopts the new `redemption_status` shape (`"unassigned" | "assigned" | "redeemed"`) introduced in [mitodl/mitxonline#3625](https://github.com/mitodl/mitxonline/pull/3625), replacing the old boolean `is_redeemed` field. Updates the hand-rolled `ContractCode` type and factory, wires all previously stubbed display columns (`assigned_to`, `assigned_on`, `last_sent`), fixes the tab filter and row action menu to branch on `redemption_status`, and derives summary stats from the full codes array (including unassigned). Unassigned codes are excluded from visible table rows.

**2. Assign Seats CSV flow**

`AssignSeatsSection` now supports both manual email entry (comma or newline delimited) and `.csv` file upload. Includes client-side email validation, a confirmation modal (`AssignSeatsConfirmModal`) that previews parsed addresses before submission, graceful CSV read error handling, and debounced `aria-live` announcements for screen reader accessibility. Reusable `parseEmails` / CSV utilities are extracted to `ol-utilities/src/string/utils.ts`.

> Note: The actual seat assignment write action is not wired in this PR — the MITx Online API endpoint is still read-only. That is scoped to [mitodl/hq#11450](https://github.com/mitodl/hq/issues/11450).

### Screen Recordings (if appropriate):

Seat Assignments Table Screen Recoding 

https://github.com/user-attachments/assets/682db964-75b2-4b14-947c-fed603cac3fb

Assign Seats Manually Screen Recoding 

https://github.com/user-attachments/assets/625aea4d-3ff3-4f4c-beb9-e255eb11b3d8

Assign Seat via CSV Recording 

https://github.com/user-attachments/assets/c204a9b8-5ba3-4cb7-b80a-e16040280acf

NVDA Screen Recoding 

https://github.com/user-attachments/assets/f13e55d2-c561-44c3-a7e1-8573475f67c9

### How can this be tested?

**redemption_status migration**

1. Navigate to a contract admin page with a mix of unassigned, assigned, and redeemed codes.
2. Confirm the four stat blocks (Total purchased, Unassigned, Pending claim, Redeemed) reflect accurate counts derived from the full codes array.
3. On the All tab, verify no unassigned codes appear in the table rows.
4. Switch to the Pending claim tab - only pending claim should appear. 
5. Switch to the Redeemed tab - only redeemed codes should appear.
6. Type a learner's email in the search box and confirm it filters the table.
7. Type a learner's name in the search box and confirm it also filters the table (search covers both `assigned_to` and `assigned_name`).
8. Verify the Assigned to, Assigned on, and Last sent columns show real values instead of `—` stubs.
9. Open the row action menu on a pending (assigned) row - Change assigned email, Resend claim email, Copy claim link and Release seat are present. Only copy link is not disabled. 
10. Open the row action menu on a redeemed row - only Uninvite should appear and should be disabled.

**Assign Seats CSV flow**

1. Type a comma-separated list of valid emails manually and confirm they are accepted and allows the Assign Seats button to go from disabled to enabled.
2. Type a newline-separated list of valid emails manually and confirm they are accepted and allows the Assign Seats button to go from disabled to enabled.
3. Enter one or more invalid email addresses and confirm per-address error messaging appears and the email turns red.
4. Enter one more duplicated email addresses and confirm the duplicated emails were removed in the dialog.
5. Enter some emails manually in the text area. Then click "import from CSV" emails that we previously in the text area should be cleared. 
6. Upload .csv file with all valid emails and confirm the emails are parsed correct and confirms in the dialog. 
7. Upload ALL malformed or unreadable emails and confirm a error appears and dialog does not open. 
8. Upload a mix of valid and malformed or unreadable emails and confirm the correct emails are parsed and the incorrect are listed in the dialog.
9. Upload a mix of valid and some duplicated emails and confirm the correct amount emails are parsed and the duplicated one are removed. 
10. Upload a mix of valid, duplicated, and malformed emails  and confirm the correct amount emails are parsed and the duplicated one are removed and the incorrect are listed in the dialog.
9. With valid emails entered, click to proceed and confirm the confirmation modal opens showing the parsed address list.
11. In the confirmation modal, confirm the submit button is present but the write action is stubbed (no API call fires yet).
12. Confirm screen reader announcements (e.g. error states, status changes) are present and not rapid-firing on keystroke.

 ### Additional Context


frontends/main/src/app-pages/ContractAdminPage/AssignSeatsSection.tsx
```diff
@@ -61,40 +236,99 @@ const AssignSeatsSection: React.FC = () => {
...
+        <ButtonWrapper>
+          <Button
+            variant="primary"
+            disabled={!canSubmit}
+            onClick={() => {
+              // TODO: implement submit handler when API is available
+            }}
+          >
+            Assign Seats
+          </Button>
```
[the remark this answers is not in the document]
daniellefrappier18: This is fine for now

frontends/main/src/app-pages/ContractAdminPage/AssignSeatsSection.tsx
```diff
@@ -46,12 +64,216 @@ const DisabledLink = styled.span(({ theme }) => ({
...
+  // Overlay is always visible when there is content. Tokens are colored as soon
+  // as a delimiter follows them (commit-on-delimiter); on blur all tokens commit.
+  const tokens = useMemo(
+    () => tokenizeInput(emailInput, !focused),
+    [emailInput, focused],
+  )
+  const showOverlay = hasEmails
+
+  const announcement = hasEmails
+    ? `${validCount} valid email${validCount !== 1 ? "s" : ""}${invalidCount > 0 ? `, ${invalidCount} invalid` : ""}`
```
dsubak: Possibly a dumb question, but would `pluralize` work here too for the word email?

frontends/ol-utilities/src/string/utils.ts
```diff
@@ -1,3 +1,104 @@
...
+
+/**
+ * Extract email addresses from pre-parsed CSV rows from PapaParse.
+ * Scans all columns per row for the first value containing "@".
+ *
+ * Silently skips rows with no "@" (header rows, name columns, empty fields).
+ * Values containing "@" that fail email validation are collected in `invalid`.
+ * Duplicate emails (case-insensitive) within the rows are removed and counted.
+ */
+export const extractEmailsFromCsvRows = (
```
dsubak, line 43: I did notice one inconsistency in the messaging between assigning seats via direct text input and via CSV upload. It looks like malformed emails w/o an `@` symbol are indicated if you input the text directly, but if you upload a CSV containing the same contents it shows duplicates removed but not invalid lines removed. See screencaps below

Image
Image

Contents of file were:
```
ljnwvrljnvwljn@foo.bar
aljnflqj
ljnwvrljnvwljn@foo.bar
jknwvkjnwekvjnwkjenvk
```

That behavior is consistent w/ the behavior in the docstring above (and frankly, this seems fairly minor to me anyway since we get the correct set of valid stuff indicated to users) but did want to flag it in case folks felt that'd be confusing.

dsubak (approved):
High level, functionality looks good! I left one comment about some inconsistent messaging that might be worth consideration, but the core uses worked as described.

frontends/main/src/app-pages/ContractAdminPage/AssignSeatsSection.tsx
daniellefrappier18: swapped to use pluralize util

frontends/ol-utilities/src/string/utils.ts
daniellefrappier18: good catch - I added the manual step step where I didn't surface the skipped emails but then discussed it with Steve and decided to add it for the CSV update and forgot to go back and fix this.

frontends/api/src/mitxonline/hooks/organizations/queries.ts
ChristopherChudzicki: The API client now provides `ManagerEnrollmentCode`. (Maybe by a PR you or @dsubak made?)

There is one difference: Your ContractCode has a string union for ` redemption_status: "unassigned" | "assigned" | "redeemed"`, whereas the client-provided has `ManagerEnrollmentCode`. I assume your string union is (A) correct and (B) useful. If that's true, would be nice to update the client.

In the meantime... Minor, but:
1. You could use `type ContractCode = ManagerEnrollmentCode & { redemption_status: "unassigned" | "assigned" | "redeemed" }`
2. You can get rid of the `as unknown` here:
    - the typecasting is slightly safer without the `as unknown`. It's still not safe, but TS has some restrictions / sanity checks[^1] on `A as B`, whereas `A as unknown as B` is "Shutup TS, I know what I'm doing". 

[^1]: Technically: A as B requires that B is either a subtype or a supertype of B, as opposed to something completely different. So `{ a: string }` as `{ b: string }` will fail completely.

frontends/main/src/app-pages/ContractAdminPage/AssignSeatsSection.tsx
```diff
@@ -46,12 +65,220 @@ const DisabledLink = styled.span(({ theme }) => ({
...
 
+const ValidationBadge = styled.div(({ theme }) => ({
+  display: "inline-flex",
+  gap: "24px",
+  backgroundColor: theme.custom.colors.lightGray1,
+  border: `1px solid ${theme.custom.colors.lightGray2}`,
+  borderRadius: "4px",
+  padding: "8px 16px",
+  ...theme.typography.body3,
+  fontWeight: theme.typography.fontWeightMedium as number,
```
ChristopherChudzicki: If I'm looking at the right thing in Figma, the style for this is `subtitle3`

https://www.figma.com/design/mSDdN8giMcncBwNgwcVJL9/UX-Wireframes?node-id=16566-456048&m=dev

frontends/api/src/mitxonline/hooks/organizations/queries.ts
daniellefrappier18: no problem, I addressed both

frontends/main/src/app-pages/ContractAdminPage/AssignSeatsSection.tsx
daniellefrappier18: good catch

frontends/main/src/app-pages/ContractAdminPage/AssignSeatsSection.tsx
```diff
@@ -61,46 +288,136 @@ const AssignSeatsSection: React.FC = () => {
...
+        <Stack
+          flex={1}
+          gap="8px"
+          minWidth={0}
+          alignItems="flex-start"
+          width="100%"
+        >
+          <EmailInputRoot $focused={focused}>
+            {showOverlay && (
+              <EmailHighlightLayer
```
ChristopherChudzicki, line 325: I noticed that invalid emails don't get highlighted if they are at the end AND have ac omma:

Image

they aren't highlighted as invalid, but they are counted (see 1 invalid).

Not highlighting makes sense... would be annoying while user typing. But maybe should highlight if last AND input not focused?

frontends/main/src/app-pages/ContractAdminPage/RowActionMenu.tsx
```diff
@@ -107,35 +107,36 @@ const RowActionMenu: React.FC<RowActionMenuProps> = ({ code }) => {
...
+        copied ? (
+          <CopiedMenuItem key="copy-link" disabled>
+            Link copied to clipboard
+          </CopiedMenuItem>
+        ) : (
+          <ActionMenuItem key="copy-link" onClick={handleCopyClaimLink}>
+            Copy claim link
+          </ActionMenuItem>
+        ),
+        <Divider key="divider" />,
```
ChristopherChudzicki: Strictly speaking this is invalid HTML as-is (Direct children of `ul` should always be `li`). Suggest:
```suggestion
        <Divider component="li" role="separator" key="divider" />
```

frontends/main/src/app-pages/ContractAdminPage/AssignSeatsSection.test.tsx
```diff
@@ -25,47 +64,250 @@ describe("AssignSeatsSection", () => {
...
 
-  test("disabled pseudo-links are in the tab order", () => {
+  test("download sample CSV pseudo-link is in the tab order and disabled", () => {
     renderWithTheme(<AssignSeatsSection />)
 
-    // Both DisabledLinks get aria-label="Coming soon" from the enclosing Tooltip
-    const pseudoLinks = screen.getAllByRole("button", { name: "Coming soon" })
-    expect(pseudoLinks).toHaveLength(2)
+    // Only the download link is still disabled — import from CSV is now active
+    const downloadLink = screen.getByRole("button", { name: "Coming soon" })
```
ChristopherChudzicki: if you put `describeChild` on the tooltips, you can keep using their real name as the label, and Coming Soon won't leak into the tests.

ChristopherChudzicki (changes requested):
I left a minor request about a `<Divider />`. Also a small issue with the highlighting.

The following is a somewhat bigger issue and may need input from @steven-hatch , unsure. The "Assign Seats" dialog doesn't seem to do any validation regarding the number of seats remaining. I think it should.

Image

frontends/main/src/app-pages/ContractAdminPage/AssignSeatsSection.tsx
daniellefrappier18: Good catch! Fixed the last token now commits (and highlights) if it has a preceding delimiter

daniellefrappier18:
@ChristopherChudzicki @steven-hatch 
This is currently the state of the over capacity flow:
Manual entry - An inline error ("Only N unassigned seats available.") appears below the email input and the Assign Seats button is disabled, so the user never reaches the confirmation dialog.
CSV upload - The confirmation dialog opens but the Send and Cancel buttons are replaced with a single "Review Learner List" button. Clicking it closes the dialog so the user can trim their list before trying again.
Screenshot 2026-06-15 at 3 49 03 PM

Screenshot 2026-06-15 at 3 49 16 PM

daniellefrappier18:
@ChristopherChudzicki @steven-hatch 

I have updated the language on over-capacity dialog to the following:

Screenshot 2026-06-16 at 12 46 03 PM

ChristopherChudzicki (approved):
👍  I think we discussed just rejecting CSVs with bad entries, but if we go that route it can always be a followup.

Image

PR #3443, opened 2026-06-08
dashboard refactor phase 7a

gumaerc:
https://github.com/mitodl/hq/issues/11205

Phase 7a of the dashboard refactor. Builds `CoursewareCard` — a unified replacement for `DashboardCard` and `ModuleCard` — and begins migrating callsites to it.

**Model changes**
- Moves `DashboardType` and `DashboardResource` from `DashboardCard.tsx` into `model/dashboardViewModel.ts` (the durable home for dashboard types). `DashboardCard.tsx` re-exports them for backward compat during migration.
- Absorbs all of `helpers.ts` into `dashboardViewModel.ts`. `helpers.ts` is now a thin re-export shim kept alive only for legacy card imports; it will be deleted in the final PR of this phase.

**New component: `CoursewareCard`**
- Accepts `entry: DashboardCourseEntry` for course arms. Dispatches internally on `entry.displayedEnrollment !== null` (enrolled) vs `null` (unenrolled).
- `moduleRow` variant replaces `ModuleCard` usage in `ProgramAsCourseCard`. Reads `useVerifiedEnrollment` and `parentProgramReadableIds` from `entry.ancestorContext`.
- Re-exports `CoursewareCardRoot` (for `OrganizationCards` migration in the next PR).
- Delegates internally to the legacy `DashboardCard` / `ModuleCard` implementations; card rendering logic moves in the deletion PR.

**Callsite migrations**
- `ProgramAsCourseCard`: module rows now use `CoursewareCard variant="moduleRow"` + `buildCourseEntry` instead of `selectBestEnrollment` + `ModuleCard`.
- `ProgramEnrollmentDisplay`: course arm uses `CoursewareCard entry={item.entry}` directly; the program-enrollment arm still uses `DashboardCard` and is marked for migration in the next PR.

**Not in this PR:** `HomeEnrollmentsDisplay`, `ContractContent`, `OrganizationCards`, and the legacy card deletions.

### How can this be tested?
- Ensure that you have MITx Online and Learn connected as described in the readme with sufficient test data, which in this case is a B2B contract with several programs and program collections. Also make sure your user is enrolled in said contract.
- Browse to the dashboard and open the contract. Ensure that everything looks the same as it does when running Learn on `main`


frontends/main/src/app-pages/DashboardPage/CoursewareDisplay/CoursewareCard.tsx
```diff
@@ -0,0 +1,123 @@
...
+
+type CoursewareCardBaseProps = {
+  entry: DashboardCourseEntry
+  className?: string
+  Component?: React.ElementType
+}
+
+/** Props for the default course / enrollment card display. */
+type CoursewareCardDefaultProps = CoursewareCardBaseProps & {
+  variant?: "default" | "stacked"
```
ChristopherChudzicki: maybe we should call this `layout` or `layoutVariant` since we use variant for courseruns now?

We do use variant widely for ui stuff, `<Button variant="primary" />`, but seems potentially more confusing in this context.

gumaerc:
FYI there is some stuff in here working around an issue with the AskTIM feature flag that looks unrelated, and that's because it is. I'll strip that stuff out tomorrow, but the failures I was seeing locally are related to https://github.com/mitodl/mit-learn/pull/3455

frontends/main/src/app-pages/DashboardPage/CoursewareDisplay/ProgramAsCourseCard.tsx
```diff
@@ -582,28 +580,27 @@ const ProgramAsCourseCard: React.FC<ProgramAsCourseCardProps> = ({
       </ProgramCardSubHeader>
       <ProgramCardBody>
         {displayedModuleCourses.map((course) => {
-          const bestEnrollment = selectBestEnrollment(
+          const entry = buildCourseEntry(
```
ChristopherChudzicki, line 583: **Behavior Change:** (Desirable, I think).
- Currently on prod, b2b enrollments in a ProgramAsCourse's child courseruns leak into the program dashboard. You can see this easily with UAI. (Note: Only for program-as-course child runs, not for the standalone verticals)
- With this change: They no longer do. That's because `buildCourseEntry` filters out b2b enrollments unless you pass the b2b contract id

However it's still leaking into the program dashboard in the completion count. See screenshot below. OTOH, you did complete the course, and the backend counts this toward your program cert completion

Image

Not 100% sure what the desired behavior is here. We may want to check with Peter/Linda.

frontends/main/src/app-pages/DashboardPage/CoursewareDisplay/CoursewareCard.tsx
```diff
@@ -0,0 +1,123 @@
...
+
+  const adapted = adaptCourseEntryToLegacyDashboardCardProps(entry)
+
+  return (
+    <DashboardCard
+      resource={adapted.resource}
+      selectedCourseRun={adapted.selectedCourseRun}
+      contractId={adapted.contractId}
+      programEnrollment={adapted.programEnrollment}
+      buttonHref={adapted.buttonHref}
```
ChristopherChudzicki: **Suggestion:** Here or in a followup, I'm like 99% sure we can get rid of `buttonHref`:

1. Cards have title action + button action
2. Originally, title action was a link that went to product page and button action was "enroll or go to courseware"
    - But we wanted different behavior on the b2c and b2b pages, so we introduced buttonHref (and maybe a titleAction or titleHref prop) so that parents could override
3. We changed title action and button action to be identical. (I think this happened in #2935)

Basically:
1. Right now, ALL cards construct their links internally
2. But we allow passing a `buttonHref` from parent

But in practice, (2) is never used. But (2) makes it hard to tell where the URL is coming from, so we should remove (2).

ChristopherChudzicki (approved):
👍 
- One suggestion (variant layout naming)
- One simplification for here or next PR
- One comment/question that I am not 100% sure what to do about, but seems like an improvement from current behavior. See https://github.com/mitodl/mit-learn/pull/3443/changes#r3387981510

frontends/main/src/app-pages/DashboardPage/CoursewareDisplay/ProgramAsCourseCard.tsx
gumaerc: We discussed this today, and the consensus is that for now the b2c program dashboard should behave in a consistent manner and hide b2b enrollments. We will revisit this later when we start implementing the work surrounding multiple enrollment display.

PR #3447, opened 2026-06-09
feat: adopt mitodl/ol-python-base:3.12 and add BuildKit cache mounts

blarghmatey:
N/A

Adopts the new shared Python base image and adds BuildKit cache mounts to speed up cold-build and `uv.lock`-bump rebuilds in the Tilt/K3d local-dev environment.

**Changes:**
- `FROM mitodl/ol-python-base:3.12` replaces the four duplicated substrate stages (apt core, uv binary, mitodl user, /opt/venv env vars). The shared base is defined in https://github.com/mitodl/ol-infrastructure/tree/main/dockerfiles/ol-python-base.
- App-specific apt (libcairo2-dev, poppler-utils, default-jre, chromium, chromium-driver) kept inline with `--mount=type=cache` for apt so downloaded packages are not baked into the image layer.
- `--mount=type=cache,target=/opt/uv-cache,uid=1000,gid=1000` on `uv sync` so `uv.lock`-bump rebuilds reuse the download cache across Tilt builds rather than re-downloading all wheels.
- `apt.txt` trimmed to extras only; the common-core packages it previously listed are now in the base image.
- Multi-stage structure preserved (base → deps → final).

**Image size:** 4.36 GB → 3.71 GB (−650 MB), which cuts the ~119s registry push proportionally.

### How can this be tested?
After https://github.com/mitodl/ol-infrastructure/pull/4750 merges and the Concourse pipeline publishes the base image:
```bash
# From the mit-learn repo root
DOCKER_BUILDKIT=1 docker build -t mitodl/mit-learn-app:test .
docker run --rm mitodl/mit-learn-app:test python --version
# Confirm uv sync cache reuse: touch uv.lock and rebuild — second build should
# skip all wheel downloads.
```
In the local-dev Tilt environment: `tilt up` mit-learn, confirm live_update still syncs Python source and triggers granian reload.

### Additional Context
Part of a multi-repo series to eliminate duplicated Dockerfile substrate and speed up Tilt/K3d cold builds. Baseline: cold `tilt up` for mit-learn spent ~163s building and ~119s pushing its 4.36 GB image to the local k3d registry at `localhost:5001`. The shared base image (~590 MB) plus per-app layer slimming reduces both dimensions.

Merge order: ol-infrastructure PR first → Concourse pipeline registered → this PR can be merged.


blarghmatey (commented):
Addressing the bot review comments:

- **`UV_CACHE_DIR` not set**: no change needed. The base image (`mitodl/ol-python-base:3.12`) sets `ENV UV_CACHE_DIR=/opt/uv-cache` and creates `/opt/uv-cache` owned by `mitodl`, so the BuildKit cache mount is fully effective as written.
- **`COPY certs/*.crt` fails when no `.crt` files present**: fixed — changed to `COPY --chmod=644 certs/ /usr/local/share/ca-certificates/` so the build succeeds in clean checkouts where only `certs/.keep` exists.
- **`apt.txt` unused**: fixed — `apt.txt` has been deleted; the inline Dockerfile RUN is now the single source of truth for app-specific apt packages.

feoh (approved):

PR #3448, opened 2026-06-09
Fix error pages after swap to runtime env vars

ChristopherChudzicki:

- Closes https://github.com/mitodl/hq/issues/11679

This PR changes how we provide env variables on the frontend. Originally, we baked env variables into the NextJS assets at build time (the usual approach). Following https://github.com/mitodl/mit-learn/pull/3364, we switched to runtime env vars. The mechanism for runtime env vars was:
1. Runtime env vars are always available on server—normal NextJS behavior.
2. By NextJS SSR, the root `layout.tsx` render function included a `<script>window.__ENV= { /* env values*/ }` script into document head.
3. The build-time static assets included a `env(...)` helper that read from the SSR-injected script tag's window assignment.

This works well for normal NextJS pages. Cool.

**However:** It did not work for many error pages. When an error is thrown server-side (unintentionally, or intentionally via `notFound()`):
1. NextJS starts server-rendering  HTML from render functions (like `layout.tsx`) before error
2. The error is thrown
3. The server-rendered HTML is thrown away, the error page renders client side

That resulted in the server-rendered `<script>window.__ENV={...}</script>` tag being thrown away. (The env vars are still encoded in the HTML , but not as a regular script tag, and not as a way that is available to the static assets on load: they're included as a RSC payload that you can see [here](https://gist.github.com/ChristopherChudzicki/398da195b2a8458e2dd5f788e864a112#file-response-does-not-exist-2026-06-09-L1047).)

**What this does:** This PR moves the env var delivery mechanism out of a layout function and into a `<meta name="x-public-env" value="env values here">` tag.

This is included whether the NextJS server responds 200 or whether an error is thrown during request handling.

### How to test this
I recommend you test this using a production build.

1. To orient yourself, visit `courses/does-not-exist` on production or on `main`. The page is blank. Uh-oh.
2. `git checkout cc/runtime-env-on-error-pages~1` t checkout the penultimate commit on this branch. This commit includes some debugging code to help you throw server errors at runtime. (The final commit just removes that code.)
3.  Build the docker image: `docker build -f frontends/main/Dockerfile.web --target runner -t mit-learn-nextjs:test .`
4. Then run the docker image, see script below. First with `PAGE_ERROR` and `GLOBAL_ERROR` env vars false.
    - View server-rendered HTML in network tab. You should see a `<meta x-public-env>` tag.
    - Page should load, everything should work.
5. Visit `courses/does-not-exist`. You should get the regular 404 page and the header should correctly reflect whether you are logged in or not. If you're logged out, the login link should be correct.
6. Restart container with `PAGE_ERROR=true`. This will make the homepage return a 500. You should see the correct error page, `window.__ENV` should be defined, and the header should correctly reflect your auth status.
7. Restart the container with `GLOBAL_ERROR=true`. This emulates an error in our root layout (extremely rare), so you won't see a header/footer (those come from root layout). But it should render, and you should see `window.__ENV` defined.

Running the container (set `PAGE_ERROR` and `GLOBAL_ERROR` as desired) ... Will need to replace `$NEXT_PUBLIC_ORIGIN` if they aren't in your host env
```sh
# Then run with your local env vars:
echo 'docker run --rm --name mln-test -p 8062:3000 \
    --add-host=api.open.odl.local:host-gateway \
    -e NEXT_PUBLIC_ORIGIN=$NEXT_PUBLIC_ORIGIN \
    -e NEXT_PUBLIC_MITOL_API_BASE_URL=$NEXT_PUBLIC_MITOL_API_BASE_URL \
    -e NEXT_PUBLIC_SITE_NAME=$NEXT_PUBLIC_SITE_NAME \
    -e NEXT_PUBLIC_MITOL_SUPPORT_EMAIL=$NEXT_PUBLIC_MITOL_SUPPORT_EMAIL \
    -e NEXT_PUBLIC_CSRF_COOKIE_NAME=$NEXT_PUBLIC_CSRF_COOKIE_NAME \
    -e NEXT_PUBLIC_MITX_ONLINE_LEGACY_BASE_URL=$NEXT_PUBLIC_MITX_ONLINE_LEGACY_BASE_URL \
    -e NEXT_PUBLIC_MITX_ONLINE_BASE_URL=https://api.learn.mit.edu/mitxonline \
    -e NEXT_PUBLIC_MITX_ONLINE_CSRF_COOKIE_NAME=$NEXT_PUBLIC_MITX_ONLINE_CSRF_COOKIE_NAME \
    -e NEXT_CACHE_S_MAXAGE_SECONDS=7200 \
    -e NEXT_PUBLIC_POSTHOG_API_KEY=phc_k26arhw0Wafe79ZDBTqeN3v4Y7dzNS7os9naQPxJhr5 \
    -e NEXT_PUBLIC_MITOL_AXIOS_WITH_CREDENTIALS=$NEXT_PUBLIC_MITOL_AXIOS_WITH_CREDENTIALS \
    -e PAGE_ERROR=false \
    -e GLOBAL_ERROR=false \
    mit-learn-nextjs:test'
```

### DIFF
--- a/frontends/main/src/app/components/PublicEnvScript.tsx
+++ b/frontends/main/src/app/components/PublicEnvScript.tsx
@@ -18,19 +18,16 @@
  */
 import React from "react"
 import { connection } from "next/server"
+import { publicEnvObject } from "@/env"
 
 export async function PublicEnvScript() {
   // `connection()` opts this route out of static prerendering so that
   // process.env is read fresh on every request (not baked at build time).
   await connection()
 
-  const publicEnv = Object.fromEntries(
-    Object.entries(process.env).filter(([k]) => k.startsWith("NEXT_PUBLIC_")),
-  )
-
   // Escape `<` to prevent a value like `</script><script>...` from breaking
   // out of the script tag.
-  const json = JSON.stringify(publicEnv).replace(/</g, "\\u003c")
+  const json = JSON.stringify(publicEnvObject()).replace(/</g, "\\u003c")
 
   return (
     <script dangerouslySetInnerHTML={{ __html: `window.__ENV=${json};` }} />
--- a/frontends/main/src/app/layout.tsx
+++ b/frontends/main/src/app/layout.tsx
@@ -1,17 +1,29 @@
 import React from "react"
 import Providers from "./providers"
 import { PublicEnvScript } from "./components/PublicEnvScript"
-import { env } from "@/env"
+import { env, publicEnvObject } from "@/env"
 
 import "./GlobalStyles"
+import { Metadata } from "next"
 
 const NEXT_PUBLIC_ORIGIN = env("NEXT_PUBLIC_ORIGIN")
 
 /**
- * As part of the root layout, this metadata object is site-wide defaults
+ * Site-wide metadata defaults plus an x-public-env meta tag carrying all
+ * NEXT_PUBLIC_* values as JSON. NOTES:
+ *  1. This delivery mechanism for NEXT_PUBLIC_* facilitates runtime env vars (not buildtime)
+ *  2. Delivery via metadata works even when errors are thrown server-side,
+ *     in which case React SSR is aborted and rendering falls back to a client-side render
+ *  3. Available to synchronous scripts below <meta x-public-env> or async scripts
+ *     above it, as long as this tag is delivered in the same HTML chunk as the async
+ *     scripts above it. (In practice, this is true for metadata + HTML content
+ *     down to the first suspense boundary.)
  */
-export const metadata = {
+export const metadata: Metadata = {
   metadataBase: NEXT_PUBLIC_ORIGIN ? new URL(NEXT_PUBLIC_ORIGIN) : null,
+  // raw JSON is fine: the attribute is HTML-escaped when rendered and
+  // getAttribute() returns it decoded, so JSON.parse round-trips.
+  other: { "x-public-env": JSON.stringify(publicEnvObject()) },
 }
 
 /**
--- a/frontends/main/src/app/layout.tsx
+++ b/frontends/main/src/app/layout.tsx
@@ -41,6 +41,9 @@ export default function RootLayout({
 }: Readonly<{
   children: React.ReactNode
 }>) {
+  if (process.env.GLOBAL_ERROR === "true") {
+    throw new Error("This is a test error on root layout")
+  }
   return (
     <html lang="en">
       <head>
--- a/frontends/main/src/app/layout.tsx
+++ b/frontends/main/src/app/layout.tsx
@@ -41,9 +41,6 @@ export default function RootLayout({
 }: Readonly<{
   children: React.ReactNode
 }>) {
-  if (process.env.GLOBAL_ERROR === "true") {
-    throw new Error("This is a test error on root layout")
-  }
   return (
     <html lang="en">
       <head>
--- a/frontends/main/src/env.ts
+++ b/frontends/main/src/env.ts
@@ -45,6 +45,23 @@ type RequiredPublicEnvVar = Extract<
   `NEXT_PUBLIC_${string}`
 >
 
+// Populate window.__ENV from the x-public-env <meta> on first read. Error/
+// not-found pages are a client-rendered shell where PublicEnvScript never runs,
+// so window.__ENV is otherwise unset there. Shared by env() and fullEnv().
+const bootstrapClientEnv = (): void => {
+  if (window.__ENV) return
+  const content = document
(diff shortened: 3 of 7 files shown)


frontends/main/src/page-components/ConfiguredPostHogProvider/ConfiguredPostHogProvider.tsx
```diff
@@ -3,22 +3,20 @@ import posthog from "posthog-js"
...
  */
 const getBootstrapFeatureFlags = (): Record<
   string,
   boolean | string
 > | null => {
   if (typeof window === "undefined") return null
-  const envSource = window.__ENV ?? {}
-  const prefix = envSource["NEXT_PUBLIC_POSTHOG_FEATURE_PREFIX"] ?? "FEATURE_"
+  const prefix = env("NEXT_PUBLIC_POSTHOG_FEATURE_PREFIX") ?? "FEATURE_"
+  const envSource = fullEnv()
```
ChristopherChudzicki, line 19: This change is related to the rest only insofar as I don't think we should be reading directly from `__ENV`

frontends/main/src/app/components/PublicEnvScript.tsx
```diff
@@ -18,19 +18,16 @@
...
   await connection()
 
-  const publicEnv = Object.fromEntries(
-    Object.entries(process.env).filter(([k]) => k.startsWith("NEXT_PUBLIC_")),
-  )
-
   // Escape `<` to prevent a value like `</script><script>...` from breaking
   // out of the script tag.
-  const json = JSON.stringify(publicEnv).replace(/</g, "\\u003c")
+  const json = JSON.stringify(publicEnvObject()).replace(/</g, "\\u003c")
```
ChristopherChudzicki, line 30: I am pretty confident we could safely get rid of PublicEnvScript script all together with the env vars in the `meta` tag. However, I'd rather leave that for a separate PR.

`publicEnvObject()` at least ensures the meta tag and the script tag have exactly the same content.

frontends/main/src/env.ts
```diff
@@ -45,6 +45,23 @@ type RequiredPublicEnvVar = Extract<
...
+    .querySelector('meta[name="x-public-env"]')
+    ?.getAttribute("content")
+  if (content) {
+    try {
+      window.__ENV = JSON.parse(content)
+    } catch {
+      /* malformed; leave unset and fall through to process.env */
+    }
+  }
+}
```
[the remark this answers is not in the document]
ChristopherChudzicki: This seems fine. If the values are malformed, we are pretty hosed.

frontends/main/src/app/layout.tsx
```diff
@@ -1,17 +1,27 @@
...
- * As part of the root layout, this metadata object is site-wide defaults
+ * This includes site-wide metadata defaults including x-public-env meta tag
+ * carrying all NEXT_PUBLIC_* values as JSON. NOTES:
+ *  1. This delivery mechanism for NEXT_PUBLIC_* facilitaites runtime env vars (not buildtime)
+ *  2. Delivery via metadata works even when errors are thrown server-side side,
+ *     in which case React SSR is aborted and rendering falls back to a client-side
+ *  3. Available to synchronous scripts below <meta x-public-env> or async scripts
+ *     above it, as long as this tag is delivered in the same HTML chunk as the async
+ *     scripts above it. (In practice, this is true for metadata + HTML content
+ *     down to the first suspense boundary.)
```
ChristopherChudzicki, line 20: This caveat is true even in our current approach that is working on prod. On production, if you inspect the server HTML, you will see:
```html
<html>
  <head>
    <script async src="_next/whatever-chunk1.js>
    <script async src="_next/whatever-chunk2.js>
    <!-- etc -->
    <script> <!-- synchronous script -->
    window.__ENV={ /* env vars */ }
    </script>
  </head>
</html>
```
The env vars are only availabe to chunk1 and chunk2 at module level provided that the async scripts above it and the synchronous setting script at delivered within the same HTML chunk.

frontends/main/src/app/layout.tsx
```diff
@@ -1,17 +1,27 @@
...
  */
 export const metadata = {
   metadataBase: NEXT_PUBLIC_ORIGIN ? new URL(NEXT_PUBLIC_ORIGIN) : null,
+  // raw JSON is fine; react escapes the attribute and getAttribute unescapes it.
```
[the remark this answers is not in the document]
ChristopherChudzicki: I made the comment more generic, but it actually is react that does this; nextjs renders via react <meta>

gumaerc (approved):
👍

PR #3455, opened 2026-06-09
clear out env vars before running jest tests

gumaerc:
Closes https://github.com/mitodl/hq/issues/11718

This PR fixes an issue where env vars from your local env leak into Jest test runners. It adds a script to `setupFiles` that clears out the env aside from `NODE_ENV` and `FAKER_SEED` before running the tests.

### How can this be tested?
- Spin up Learn on this branch
- Run the tests with `docker compose exec watch yarn test`
- The tests should pass
- If you want, go back to main and do the same thing and tests should fail if you have something like Posthog configured that puts env vars in the frontend, because mocking feature flags stops working


frontends/jest-clear-next-public-env.ts
```diff
@@ -0,0 +1,20 @@
...
+ *   NODE_ENV  — used by React, libraries, and Jest internals.
+ *   FAKER_SEED — allows reproducible test runs via the shared setup.
+ */
+const preserved: NodeJS.ProcessEnv = {
+  NODE_ENV: process.env.NODE_ENV,
+}
+if (process.env.FAKER_SEED !== undefined) {
+  preserved.FAKER_SEED = process.env.FAKER_SEED
+}
+process.env = preserved
```
ChristopherChudzicki: In recent versions of Node, `process.env` is actually not a regular object, it's a Proxy (fancy getter/setters, basically). That's how, for example, string coercion happens:
```js
process.env.BOOLEAN_FLAG = true
console.log(process.env.BOOLEAN_FLAG) // "true", a string
```

So I think a whitelist of allowed vars is OK, but let's delete the others, rather than re-assigning the whole object. We should also expand the whitelist:

```ts
const PRESERVE = new Set([
  // Node stuff
  "NODE_ENV",
  "TZ",
  "PATH",
  "HOME",
  "TMPDIR",
  // Jest stuff
  "CI",
  "JEST_WORKER_ID",
  "FORCE_COLOR",
  "NO_COLOR",
  // Custom
  "FAKER_SEED",
]);

for (const key of Object.keys(process.env)) {
  if (!PRESERVE.has(key)) delete process.env[key];
}
```

ChristopherChudzicki (changes requested):
This is better than what we have now: prevents host env from leaking into tests. ✔️ 

_Aside: I'm not sure this is related to recent env var architecture changes. I believe host env vars have always leaked in this way. The recent env var PR may have removed some posthog override for tests, though._ 

This approach does seem a little fragile to me, because there are some host env vars that are actually relevant (`TZ`, `PATH`, `TMPDIR`, `NODE_ENV`,...) Did we get them all? I dunno. Will we need to whitelist new ones in the future? I dunno.


-------------

I think a better approach for the tests would be:
1. Change app to forbid `process.env` access across the board.
    - We currently disallow it in app code for `NEXT_PUBLIC_`, but not for serverside env vars like `process.env.OTEL_RESOURCE_ATTRIBUTES` and not for `NEXT_PUBLIC_` vars inside tests.
2. Tests use real `process.env`, but mock `env()`. Mock can have defaults set in the setup file and be overridden per-test as necessary.  

☝️ Seems more like what our django pytest tests do: I'm not 100% sure, but I imagine `os.environ` is real, but `django.conf.settings` is mocked.

**Edit:** Claude pointed out that your current approach protects us from third-party libraries consuming envs, which the above suggestion does not. 🤷 Anyway, the current approach is good enough for now!

--------------


**Request:** Not requesting that we do ☝️ right now, but:
1. Copilot's naming comment seems worth changing
2. My proxy / whitelist comment

ChristopherChudzicki (approved):

PR #3456, opened 2026-06-09
feat: implement PostHog tracking for video shorts interactions

rachellougee:

https://github.com/mitodl/hq/issues/10916

adds PostHog tracking for video shorts interactions
- video_shorts_opened — fires when a thumbnail is clicked. Captures the video ID, title, and its position in the carousel.
- video_short_viewed — fires each time the user scrolls to a new video. Captures video ID, title, session position, and (for non-first videos) how long they spent on the previous video and its total duration.
- video_shorts_closed — fires when the modal is closed. Captures total session duration and how many videos were viewed.

### How can this be tested?

frontend.local.env
```
NEXT_PUBLIC_FEATURE_VIDEO_SHORTS=True
```
backend.local.env 
```
OVS_API_BASE_URL=https://video.odl.mit.edu/
```

Run `docker compose run --rm web python manage.py backpopulate_ovs_data` to populate some video shorts locally

Click on one of these video shorts on the homepage, and verified these events are captured

video_shorts_opened
Pasted Graphic 4

video_shorts_viewed
Pasted Graphic 3

video_shorts_closed
Pasted Graphic 5

### DIFF
--- a/frontends/main/src/app-pages/HomePage/VideoShortsModal.tsx
+++ b/frontends/main/src/app-pages/HomePage/VideoShortsModal.tsx
@@ -257,7 +257,6 @@ const VideoShortsModal = ({
   const muteButtonRef = useRef<HTMLButtonElement>(null)
   const sessionStartedAtRef = useRef<number>(Date.now())
   const currentVideoStartedAtRef = useRef<number>(Date.now())
-  const isInitialSlideRef = useRef<boolean>(true)
   const sessionPositionRef = useRef<number>(0)
   const viewedIndicesRef = useRef<Set<number>>(new Set([startIndex]))
   const selectedIndexRef = useRef<number | null>(startIndex)
@@ -325,16 +324,14 @@ const VideoShortsModal = ({
       setAnnouncement(
         `${inView[0] + 1} of ${videoData.length}: ${videoData[inView[0]].title}`,
       )
-      const isInitial = isInitialSlideRef.current
-
       const prevIndex = selectedIndexRef.current
       const prevPlayer =
         prevIndex !== null ? playersRef.current[prevIndex] : null
-      let videoDurationMs: number | undefined
-      if (!isInitial && prevPlayer) {
+      let prevVideoDurationMs: number | undefined
+      if (prevPlayer) {
         const duration = prevPlayer.duration()
         if (duration && isFinite(duration) && duration > 0) {
-          videoDurationMs = Math.round(duration * 1000)
+          prevVideoDurationMs = Math.round(duration * 1000)
         }
       }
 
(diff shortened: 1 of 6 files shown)


frontends/main/src/app-pages/HomePage/VideoShortsModal.tsx
```diff
@@ -268,22 +289,117 @@ const VideoShortsModal = ({
...
+      window.removeEventListener("pagehide", handlePageHide)
+      document.removeEventListener("visibilitychange", handleVisibilityChange)
+    }
+  }, [captureSessionEnd])
+
+  // Analytics only; intentionally separate from the playback logic below so it
+  // can be added/removed without touching existing behavior. The prevIndex
+  // guard also makes it safe against CarouselV2Vertical firing the same event
+  // on duplicate listeners.
+  const trackSlideViewed = (newIndex: number) => {
```
daniellefrappier18: trackSlideViewed has two issues worth addressing:

It doesn't check sessionEndedRef.current, so if the user scrolls and closes the modal almost simultaneously, Embla's RAF-queued slidesInView event can fire after captureSessionEnd has already run — sending a video_short_viewed event after video_shorts_closed in PostHog.

It's a plain function recreated on every render, which causes onSlidesInView (which calls it) to also be a new reference each render — triggering CarouselV2Vertical's effect to remove and re-add the Embla listener on every state change. In the brief RAF gap between cleanup and re-subscription, a slidesInView event could be dropped and analytics lost.

Neither is a blocker, but both are easy fixes:


```
const trackSlideViewed = useCallback((newIndex: number) => {
  if (sessionEndedRef.current) return
  // ... rest of function
}, [videoData, startIndex, capture])
```

frontends/main/src/app-pages/HomePage/VideoShortsModal.tsx
```diff
@@ -268,22 +289,117 @@ const VideoShortsModal = ({
...
+        timeOnVideoMs: Date.now() - currentVideoStartedAtRef.current,
+        ...(videoDurationMs !== undefined ? { videoDurationMs } : {}),
+      })
+      viewedIndicesRef.current.add(prevIndex)
+    }
+    selectedIndexRef.current = newIndex
+    currentVideoStartedAtRef.current = Date.now()
+  }
 
   const onSlidesInView = (inView: number[]) => {
```
daniellefrappier18: onSlidesInView is a plain function recreated on every render. Since CarouselV2Vertical now correctly includes it in its effect dep array, every state change in VideoShortsModal (play/mute toggle etc.) causes Embla's listener to be removed and re-added. Wrapping in useCallback would prevent the churn:


```
const onSlidesInView = useCallback((inView: number[]) => {
  // ... rest of function
}, [trackSlideViewed, videoData])
```

daniellefrappier18 (approved):
Looks good I was able to verify that the events are properly captured. 

Image

There is two minor performance issues but not blockers.

PR #3459, opened 2026-06-10
fix: Run Typegen before Typecheck

ChristopherChudzicki:
None

NextJS generate some types based on our files, and typechecking won't pass unless those types are generated. For humans, this isn't usually a problem, because the dev server generates them. However for LLMs, they don't have the dev server running. So now:

- `yarn typecheck` now build nextjs types BEFORE running `typecheck`.
- I removed the `yarn build` step in `ci.yaml`. This is the how we were generating NextJS types prior to typechecking before.
    - typegen is faster
    - `yarn build` was redundant, since we still build the production docker image in the `build-nextjs-container` step of `ci.yml`

### How can this be tested?
1. `yarn typecheck` should pass
2. CI should pass

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -107,21 +107,6 @@ jobs:
       - name: Lints
         run: yarn run lint-check
 
-      - name: Build Next.js frontend
-        run: yarn workspace main build
-        env:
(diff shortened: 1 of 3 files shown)


gumaerc (approved):
👍

PR #3461, opened 2026-06-10
Add course enrollment gtm

cp-at-mit:
Adds some utility functions and triggers for course/program enroll/unenroll tags in GTM.

### How can this be tested?
If testing using the Learn GTM container, set the following in your `frontend.local.env` file:
```
GTM_TRACKING_ID=GTM-NMJ2CT6T
GTM_AUTH=Rl-qxbJnD78rPEq1OFQLow
GTM_PREVIEW=env-3
GTM_COOKIES_WIN=x
```
1. A the current time, you will need to unpause any tags you are trying to test (course enroll/unenroll)
2. Click the "Preview" button on this page: https://tagmanager.google.com/?authuser=1#/container/accounts/6351372910/containers/250259136/workspaces/11/variables
3. Enroll into a course or program and verify that the page you see in GTM indicates that the tag did get triggered.


frontends/main/src/common/analytics/gtm.ts
```diff
@@ -0,0 +1,61 @@
+type DataLayerPayload = Record<string, unknown>
+
+declare global {
+  interface Window {
+    dataLayer?: DataLayerPayload[]
+  }
+}
+
+const pushGtmEvent = (event: string, payload: DataLayerPayload = {}) => {
```
dsubak, line 9: I wonder if the whole thing ought to be wrapped in a try/catch while we're still putting everything in place.

Function seems pretty safe as is, so consider this one dealer's choice as to whether or not you want to consider it.

dsubak (approved):
I had a hard time actually getting an enrollment done on my local learn but I _was_ able to do some synthetic testing by injecting calls into spots I could trigger - using Google's tag assistant and pointing it at learn.odl.local, I was able to see that things were properly tracked in the data layer correctly. This ought to be a pretty safe first cut to get these events into GTM

Image

PR #3469, opened 2026-06-11
dashboard refactor phase 7c

gumaerc:
https://github.com/mitodl/hq/issues/11205

This PR migrates `ContractContent` to use the new `CoursewareCard` display component. The `noun` optional string property was added to support how the B2B contract dashboard pages override the noun to say "Module" instead of "Course." Really this just sets the stage for the final part of the conversion.

### How can this be tested?
- Ensure that you have MITx Online and Learn connected as described in the readme with sufficient test data, which in this case is a B2B contract with several programs and program collections. Also make sure your user is enrolled in said contract.
- Browse to the dashboard and open the contract. Ensure that everything looks the same as it does when running Learn on `main`

### DIFF
--- a/frontends/main/src/app-pages/DashboardPage/CoursewareDisplay/CoursewareCard.tsx
+++ b/frontends/main/src/app-pages/DashboardPage/CoursewareDisplay/CoursewareCard.tsx
@@ -131,10 +131,7 @@ const CoursewareCard: React.FC<CoursewareCardProps> = (props) => {
 
   return (
     <DashboardCard
-      resource={adapted.resource}
-      selectedCourseRun={adapted.selectedCourseRun}
-      contractId={adapted.contractId}
-      programEnrollment={adapted.programEnrollment}
+      {...adapted}
       showNotComplete={showNotComplete}
       offerUpgrade={offerUpgrade}
       isLoading={isLoading}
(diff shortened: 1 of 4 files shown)


frontends/main/src/app-pages/DashboardPage/CoursewareDisplay/CoursewareCard.tsx
```diff
@@ -140,6 +141,7 @@ const CoursewareCard: React.FC<CoursewareCardProps> = (props) => {
       onUpgradeError={onUpgradeError}
       contextMenuItems={contextMenuItems}
       variant={layout}
+      noun={props.noun}
```
ChristopherChudzicki, line 141: Kind of missed this in 7b I guess, but above it should be `parentProgramReadableIds={adapted.parentProgramReadableIds}`.

But better would be to just spread the props. That's the point of the adapter.

This would be an actual bug on the b2b pages, I think. I guess no test caught it. Could add a test in ContractContent.test.tsx, something like: course row with no enrollment ->  click "Enroll," assert createB2bEnrollment receives program_id = the program's readable_id.

ChristopherChudzicki (changes requested):

ChristopherChudzicki (approved):
👍

PR #3471, opened 2026-06-12
fix: add article model to Django Admin

ahtesham-quraish:
https://github.com/mitodl/hq/issues/10359

It would be nice if the Article model showed up in Django Admin, for when/if we need to handle things there.

Screenshot 2026-06-12 at 4 09 22 PM

### How can this be tested?
Just visit the django admin page and see the tables for `Website contents` and `Website content image uploads` locally.

### DIFF
--- a/website_content/admin.py
+++ b/website_content/admin.py
@@ -1,1 +1,33 @@
-# Register your models here.
+"""Admin configuration for website_content app."""
+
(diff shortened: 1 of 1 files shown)


daniellefrappier18 (approved):
LGTM

PR #3472, opened 2026-06-12
feat: add title to learning resource summary endpoint

ChristopherChudzicki:

Supports readable URLs:

- https://github.com/mitodl/hq/issues/11210

Adds `title` to response of `GET /api/v1/learning_resources/summary/`

This is split out of the upcoming readable-URLs frontend PR, which builds human-readable (slugged) sitemap URLs from summary listings. Landing and deploying the backend half first means the summary endpoint already serves titles when the frontend feature arrives — otherwise sitemap URLs would temporarily degrade to their slug-less form.

(Split out so the backend requirement can be deployed first)

### How can this be tested?

`GET /api/v1/learning_resources/summary/` — each result should now include `title` alongside `id`, `last_modified`, and `url`.

### DIFF
--- a/frontends/api/src/generated/v1/api.ts
+++ b/frontends/api/src/generated/v1/api.ts
@@ -4960,6 +4960,12 @@ export interface LearningResourceSummary {
    * @memberof LearningResourceSummary
    */
   url?: string | null
+  /**
+   *
+   * @type {string}
+   * @memberof LearningResourceSummary
+   */
+  title: string
 }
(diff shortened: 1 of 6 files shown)


mbertrand (approved):
LGTM 👍 

Copilot's comment on adding a sort field does seem like a nice thing to have, but not really required by the issue, so leaving it up to you whether to add one here or save it for a future enhancement.

PR #3473, opened 2026-06-12
dashboard refactor phase 7d

gumaerc:
Closes https://github.com/mitodl/hq/issues/11205

This PR is the final step of the dashboard code refactoring work we've been doing. In this PR:

- `DashboardCard` and `ModuleCard` both deleted
- New piecemeal presentational components created in their place for each use case:
  - `CardShared.tsx` - Shared visual components between all dashboard cards
  - `UnenrolledCourseCard.tsx` - Handles the "unenrolled" state aka you have a course run but no enrollment
  - `EnrolledCourseCard.tsx` - Handles the "enrolled" state where you have an enrollment and the run tied to it
  - `ProgramEnrollmentCard.tsx` - Handles program enrollments
- All relevant `DashboardCard` tests for each scenario migrated to their appropriate test file
- "Noun override" functionality removed; instead of showing "View Course" on B2C and "View Module" in B2B, buttons now simply say the verb aka "View" everywhere based on the latest designs, aside from programs which still have "View Program."
- `ModuleCard` was ported over as an alternate layout option in `UnenenrolledCourseCard` / `EnrolledCourseCard`

image
image

### How can this be tested?
- Ensure your instance of MIT Learn is connected to a local instance of MITx Online as described in the readme
- You should have sufficient test data to test the entire dashboard. That means a user in multiple B2B contracts, at least one with variants, B2C course and program enrollments, etc.
- The entire dashboard needs to be smoke tested and compared with the latest designs.


ChristopherChudzicki (changes requested):
This looks great.

I have a couple questions / comment:
1. UI change I was not expecting, see below
    - ⚠️ I know you discussed this with steve some so maybe my expectation is wrong
2. I think we should delete dashboardPlan.md now ... It's too technical / code focused to be useful going forward.
3. See minor comments / suggestions from AI

**Edit:** To be clear... (1) is the requested change, but I think (2) and 3c are worth doing, at least, too.

--------------

### 1 Unexpected UI Change 
I noticed this difference...

Image
Image

### 2 dashboardPlan
Yeah, I think we delete dashboard plan

### 3 Small AI comments

(C) at least seems worth addressing

- A) **Verified-program redirect-URL coverage gap.** `useEnrollmentHandler`'s `selectedCoursewareUrl` branch (~`:99-104`) — which redirects a verified-program enroll to the *selected* run's `courseware_url` — has no test asserting `window.location.href`. The verified-program test asserts the POST endpoint only. The legacy suite asserted the redirect target; restore that assertion.
- B) **Stale comment + dead scaffolding.** `ContractContent.test.tsx:35-36` claims the b2bEnroll client "does not yet accept a request body … program_id never reaches the network layer (see the skipped DashboardCard test)" — false on both counts now (`@mitodl/mitxonline-api-axios@2026.6.10` ships `B2BEnrollRequestRequest`; `program_id` is type-checked end-to-end and asserted at the network layer in `DashboardDialogs.test.tsx`; the referenced skipped test no longer exists). The `program_id?: string` type extension in `frontends/api/src/mitxonline/hooks/enrollment/index.ts:18` is correspondingly redundant.
- 3. **Naming collision.** `ProgramEnrollmentDisplay.tsx:16` defines a local `const ProgramEnrollmentCard = styled(CoursewareCard)` that shadows the real `ProgramEnrollmentCard` component — misleading now that this is the canonical program-enrollment path. Consider renaming to `StyledCoursewareCard`.
```

ChristopherChudzicki (approved):

PR #3474, opened 2026-06-15
fix: suppress 'Add some text' placeholder inside the list and task items in Tiptap editor

ahtesham-quraish:
https://github.com/mitodl/hq/issues/11790

- As a content editor, I want list and task list items to appear clean without a placeholder label, so the editor doesn't show distracting "Add some text" hints inside every bullet or task item.
- Links in subheading should be visible properly which are not currently.

### Before
image

### After
image

## Links before 
image

## Links after
image

### How can this be tested?
- You just need to visit the following url `/website_content/news/new` and then play around the list controls and see if there is no placeholder when list is inserted

### DIFF
--- a/frontends/main/src/page-components/TiptapEditor/extensions/baseExtensions.ts
+++ b/frontends/main/src/page-components/TiptapEditor/extensions/baseExtensions.ts
@@ -55,7 +55,7 @@ export const createBaseExtensions = (
     showOnlyCurrent: false,
     includeChildren: true,
     placeholder: ({ node, editor }): string => {
-      let parentNode: typeof node | null = null
+      let parentNode: ProseMirrorNode | null = null
 
       editor.state.doc.descendants((n: ProseMirrorNode) => {
         n.forEach((childNode: ProseMirrorNode) => {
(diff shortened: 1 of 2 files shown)


frontends/main/src/page-components/TiptapEditor/extensions/baseExtensions.ts
```diff
@@ -70,8 +70,13 @@ export const createBaseExtensions = (
       })
 
       if (parentNode) {
+        const pn = parentNode as ProseMirrorNode
+        if (pn.type.name === "listItem" || pn.type.name === "taskItem") {
+          return ""
+        }
+
```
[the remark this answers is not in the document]
ahtesham-quraish: We  have used the `parentNode.type.name` but it does not work out so we need to get around it

frontends/main/src/page-components/TiptapEditor/extensions/baseExtensions.ts
```diff
@@ -70,8 +70,13 @@ export const createBaseExtensions = (
       })
 
       if (parentNode) {
+        const pn = parentNode as ProseMirrorNode
```
daniellefrappier18, line 73: `parentNode` is already typed as `ProseMirrorNode | null`, and the `if (parentNode)` guard narrows it to `ProseMirrorNode`. Creating const `pn = parentNode as ProseMirrorNode` adds a redundant re-cast.

daniellefrappier18 (approved):
Looks good. I left one very minor optional comment

frontends/main/src/page-components/TiptapEditor/extensions/baseExtensions.ts
ahtesham-quraish: I already seen that but it does not work out that is why i went the other way

PR #3477, opened 2026-06-15
fix for remove_run_content_files task

shanbady:

Closes  https://github.com/mitodl/hq/issues/11799

This PR fixes an issue where the celery task to remove unpublished run contentfiles from qdrant is a no-op

### How can this be tested?
1. checkout main
2. make sure you have some resources embedded locally
3. find a learning resource run that has embedded contentfiles (find a run id  that doesnt have special characters so you can search for it in the qdrant dashboard)
4. go to your [local qdrant contentfiles collection](http://localhost:6333/dashboard#/collections/resource_embeddings.content_files) and search for run_readable_id:<your run id> - verify points exist for it
5. in django shell run the task that is supposed to remove embeddings for runs (replace run_id with the pk of the run):
```python
from vector_search.tasks import remove_run_content_files
remove_run_content_files.apply([run_id])
```
6. visit the [qdrant dashboard](http://localhost:6333/dashboard#/collections/resource_embeddings.content_files) and search for the run_id again - note that points still exist in qdrant
7. checkout this branch and perform the same - note that the points are removed

### DIFF
--- a/main/celery.py
+++ b/main/celery.py
@@ -20,6 +20,7 @@ app.conf.task_routes = {
     "vector_search.tasks.generate_embeddings": {"queue": "embeddings"},
     "vector_search.tasks.embed_run_content_files": {"queue": "embeddings"},
     "vector_search.tasks.remove_run_content_files": {"queue": "embeddings"},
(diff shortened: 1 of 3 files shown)


mbertrand (approved):
LGTM 👍

PR #3478, opened 2026-06-15
refactor(CoursewareCard): dispatch on `kind`, make `layout` density-only

ChristopherChudzicki:
On base, `layout` has four values: `default | compact | moduleRow | program`. This mixes some concepts:
- `default | compact` is related to layout. Keep these.
- `program` is about what sort of data the card is receiving
- `moduleRow` is about the context in which the card is renering (child of ProgramAsCourse).

This switches to `layout + kind`, where kind ~ data and layout ~ how card appears.

Drop `moduleRow` entirely: It was just `compact`


gumaerc (approved):
👍

PR #3480, opened 2026-06-16
feat: implement podcast embed page

ahtesham-quraish:
https://github.com/mitodl/hq/issues/11789

- Implement the podcast episode embed page

### Desktop
image

### Mobile 
image

### How can this be tested?
For testing you need to follow this url `http://open.odl.local:8062/podcast/embed/[id]` and see episode embedding

### DIFF
--- a/frontends/main/src/app-pages/PodcastPage/PodcastEmbedPlayer.tsx
+++ b/frontends/main/src/app-pages/PodcastPage/PodcastEmbedPlayer.tsx
@@ -1,0 +1,403 @@
+"use client"
+
+import React, { useRef, useState, useEffect, useCallback } from "react"
+import { styled, Typography, LoadingSpinner } from "ol-components"
+import {
+  RiPlayCircleLine,
+  RiPauseCircleLine,
+  RiReplay10Line,
+  RiForward30Line,
+} from "@remixicon/react"
+import type { LearningResource } from "api/v1"
+
+// ─── Helpers ──────────────────────────────────────────────────────────────────
+
+const SPEED_OPTIONS = [0.75, 1, 1.25, 1.5, 2]
+
+const formatTime = (seconds: number): string => {
+  const m = Math.floor(seconds / 60)
+  const s = Math.floor(seconds % 60)
+  return `${String(m).padStart(2, "0")}:${String(s).padStart(2, "0")}`
+}
+
+function getAudioUrl(resource: LearningResource): string {
+  if (resource.resource_type !== "podcast_episode") return ""
+  return (
+    resource.podcast_episode?.audio_url ??
+    resource.podcast_episode?.episode_link ??
+    ""
+  )
+}
+
+// ─── Styled components ────────────────────────────────────────────────────────
+
+const Shell = styled.div(({ theme }) => ({
+  width: "100vw",
+  height: "100vh",
+  display: "flex",
+  alignItems: "center",
+  justifyContent: "center",
+  backgroundColor: theme.custom.colors.white,
+  padding: "24px",
+  boxSizing: "border-box",
+}))
+
+const PlayerCard = styled.div(({ theme }) => ({
+  display: "grid",
+  gridTemplateColumns: "auto 1fr",
+  gridTemplateAreas: '"art info" "art controls" "art progress"',
+  gap: "12px 24px",
+  width: "100%",
+  maxWidth: "80%",
+  alignItems: "start",
+  [theme.breakpoints.down("sm")]: {
+    gridTemplateColumns: "1fr",
+    gridTemplateAreas: '"art" "info" "controls" "progress"',
+  },
+}))
+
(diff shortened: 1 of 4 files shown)


frontends/main/src/app-pages/PodcastPage/PodcastEmbedPlayer.tsx
```diff
@@ -0,0 +1,403 @@
...
+  const m = Math.floor(seconds / 60)
+  const s = Math.floor(seconds % 60)
+  return `${String(m).padStart(2, "0")}:${String(s).padStart(2, "0")}`
+}
+
+function getAudioUrl(resource: LearningResource): string {
+  if (resource.resource_type !== "podcast_episode") return ""
+  return (
+    resource.podcast_episode?.audio_url ??
+    resource.podcast_episode?.episode_link ??
```
daniellefrappier18: `episode_link ` is a podcast web page URL - not a direct audio stream. When audio_url is null, this gets set as the <audio> element's src, the browser fails to load it, onError fires, and the button silently reverts to Play with no error message to the user.

frontends/main/src/app-pages/PodcastPage/PodcastEmbedPlayer.tsx
```diff
@@ -0,0 +1,403 @@
...
+          </PlayPauseButton>
+
+          <IconButton
+            onClick={() => handleSkip(30)}
+            aria-label="Forward 30 seconds"
+          >
+            <RiForward30Line />
+          </IconButton>
+
+          <SpeedButton onClick={handleSpeedCycle} aria-label="Playback speed">
```
daniellefrappier18: aria-label="Playback speed" is static. Screen reader users hear "Playback speed, button" but get no indication whether the current rate is 1x, 1.25x, etc. The visible text conveys this but the accessible name overrides it.

Suggested Fix:

`aria-label={`Playback speed: ${SPEED_OPTIONS[speedIndex]}x`}`

frontends/main/src/app-pages/PodcastPage/PodcastEmbedPlayer.tsx
```diff
@@ -0,0 +1,403 @@
...
+  RiReplay10Line,
+  RiForward30Line,
+} from "@remixicon/react"
+import type { LearningResource } from "api/v1"
+
+// ─── Helpers ──────────────────────────────────────────────────────────────────
+
+const SPEED_OPTIONS = [0.75, 1, 1.25, 1.5, 2]
+
+const formatTime = (seconds: number): string => {
```
daniellefrappier18, line 17: formatTime breaks for episodes ≥ 60 minutes,` formatTime(3661)` returns "61:01" instead of "1:01:01"

Suggested Fix:

const formatTime = (seconds: number): string => {
  const h = Math.floor(seconds / 3600)
  const m = Math.floor((seconds % 3600) / 60)
  const s = Math.floor(seconds % 60)
  const mm = String(m).padStart(2, "0")
  const ss = String(s).padStart(2, "0")
  return h > 0 ? `${h}:${mm}:${ss}` : `${mm}:${ss}`
}

daniellefrappier18 (changes requested):

frontends/main/src/app-pages/PodcastPage/PodcastEmbedPlayer.tsx
```diff
@@ -0,0 +1,406 @@
...
+          <TimeLabel variant="body3">{formatTime(currentTime)}</TimeLabel>
+          <ProgressRange
+            type="range"
+            min={0}
+            max={duration || 1}
+            value={currentTime}
+            step={1}
+            percent={percent}
+            aria-label="Seek"
+            onChange={(e) => {
```
daniellefrappier18, line 394: Sorry I missed this before but with a type="range" input with aria-label="Seek" with no aria-valuetext a screen reader will announce the raw value (currentTime in seconds, e.g. "90") instead of a meaningful time like "1:30". 

Suggested fix: Add `aria-valuetext={formatTime(currentTime)}`

frontends/main/src/app-pages/PodcastPage/PodcastEmbedPlayer.tsx
```diff
@@ -0,0 +1,406 @@
...
+  alignSelf: "center",
+  [theme.breakpoints.down("sm")]: {
+    width: "100%",
+    height: "auto",
+  },
+}))
+
+const CoverArtPlaceholder = styled.div(({ theme }) => ({
+  gridArea: "art",
+  width: "120px",
```
daniellefrappier18: Sorry another thing I missed the first time. Should CoverArt and CoverArtPlaceholder be the same size.

daniellefrappier18 (approved):
Two minor things to look at before merging but otherwise looks good.

PR #3482, opened 2026-06-16
Fix bug in vector resource metadata updates

mbertrand:
Closes https://github.com/mitodl/hq/issues/11786

Refreshes a resource's full Qdrant payload (topics, etc.) when metadata changes without the embedding text changing. The old skip-path update projected through a filter-param map keyed on the wrong names (`topic` vs `topics`), so those fields silently never got written.

### How can this be tested?
- Pick an already-embedded resource, note its `id`, change its topics in admin, and save.
-  Reprocess the existing point: `docker compose run --rm web python manage.py generate_embeddings --resource-ids <id> --overwrite`
- Confirm the payload topics now match:
   ```python
   from vector_search.utils import qdrant_client, vector_point_id, vector_point_key
   from learning_resources_search.serializers import serialize_bulk_learning_resources
   from vector_search.constants import RESOURCES_COLLECTION_NAME

   doc = next(iter(serialize_bulk_learning_resources([<id>])))
   pid = vector_point_id(vector_point_key(doc))
   print(qdrant_client().retrieve(RESOURCES_COLLECTION_NAME, ids=[pid])[0].payload["topics"])
   ```

With QDRANT_ENABLE_INDEXING_PLUGIN_HOOKS enabled (as it is on RC and prod), every ETL create/update re-embeds on a text change and refreshes the payload otherwise, so changes should reconcile automatically.

### DIFF
--- a/vector_search/utils.py
+++ b/vector_search/utils.py
@@ -483,12 +483,14 @@ def _process_resource_embeddings(serialized_resources):
 
 
 def update_learning_resource_payload(serialized_document):
-    points = [vector_point_id(vector_point_key(serialized_document))]
-    _set_payload(
-        points,
-        serialized_document,
-        param_map=QDRANT_RESOURCE_PARAM_MAP,
+    """
+    Refresh a resource's Qdrant payload without re-embedding.
+    """
+    point_id = vector_point_id(vector_point_key(serialized_document))
+    qdrant_client().overwrite_payload(
         collection_name=RESOURCES_COLLECTION_NAME,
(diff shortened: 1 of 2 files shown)


shanbady (changes requested):
works as intended - just left a minor note to restrict this to only indexed payload fields asa per qdrant's recommendation

shanbady (approved):
removed my comment about mapping with the payload fields that are indexed since i dont think topics cleanly maps (which i think was what caused the original issue)

PR #3491, opened 2026-06-18
feat: embed posdcast and podcast episode inside the editor

ahtesham-quraish:
https://github.com/mitodl/hq/issues/11789
As a content editor, I want to paste a podcast or podcast episode URL into the Tiptap editor and have it automatically embed the correct UI, so I can enrich articles with podcast content without manually constructing embeds.

image

### How can this be tested?
For testing you need to insert these urls into article or news 
`https://rc.learn.mit.edu/podcast/136070`
`https://rc.learn.mit.edu/podcast/136070/podcast_episode/137175`
`https://rc.learn.mit.edu/podcast/embed/137277` 
you can replace the ids as per your local setup, if you are using RC then these ids will work fine in urls


zamanafzal (changes requested):
LGTM Overall. I have tested the changes and found that when I try to edit the article, and it has a podcast, then it plays two audio tracks for the same podcast at the same time, and there is no way to close the second audio track and it only closes on refreshing the page. i've created a loom video where one track is audible but audio icon can be seen on tab while the track is paused. 

https://www.loom.com/share/a62135980d0c430aa924d1f9c30b25a7 

Reproducing Steps:

Please add a podcast in an article and publish, and when you try to edit, it will run two tracks automatically at the same time.

ahtesham-quraish:
> LGTM Overall. I have tested the changes and found that when I try to edit the article, and it has a podcast, then it plays two audio tracks for the same podcast at the same time, and there is no way to close the second audio track and it only closes on refreshing the page. i've created a loom video where one track is audible but audio icon can be seen on tab while the track is paused.
> 
> https://www.loom.com/share/a62135980d0c430aa924d1f9c30b25a7
> 
> Reproducing Steps:
> 
> Please add a podcast in an article and publish, and when you try to edit, it will run two tracks automatically at the same time.

@zamanafzal I have fixed the issue and one more thing that I have implemented is that when we embed the podcast during that time the podcast player should not be auto by default because it does not make sense to play auto so I have made changes for that

frontends/main/src/app-pages/PodcastPage/PodcastEmbedPlayer.tsx
```diff
@@ -258,15 +269,14 @@ const PodcastEmbedPlayer: React.FC<PodcastEmbedPlayerProps> = ({
...
+    setIsBuffering(false)
 
     if (!hasAudioSource) return
 
     const audio = audioRef.current
     if (!audio) return
     audio.load()
     audio.playbackRate = SPEED_OPTIONS[speedIndexRef.current]
-    void startPlayback()
   }, [audioUrl, hasAudioSource, startPlayback])
```
zamanafzal: nit: I think `startPlayback` isn't called in this effect anymore — we should drop it from the deps so it's just [audioUrl, hasAudioSource].

zamanafzal (approved):
Left a suggestion. Overall LGTM.

PR #3493, opened 2026-06-18
feat: wire up B2B seat management write actions

daniellefrappier18:

Closes https://github.com/mitodl/hq/issues/11450

#### B2B seat management write actions 
Wires up the Contract Admin seat-management UI to the mitxonline B2B write APIs. Previously these actions were stubbed/disabled ("Coming soon"); they are now functional. 

**Known limitation**

- **Uninvite** (redeemed-row action) currently fails: the B2B revoke endpoint returns **409** for already-redeemed codes.

⚠️ Note: Everything in this section is not part of the B2B feature work. These are unavoidable type changes as a result of regenerating the API client at the newer version: @mitodl/mitxonline-api-axios 2026.6.10 → 2026.6.17

This bump pulls in mitxonline [#3615](https://github.com/mitodl/mitxonline/pull/3615) ("Fix some N+1 errors around programs on courses API"), which reshaped how programs are serialized on the courses API. 

1. programs on course serializers is now a non-nullable array
CourseWithCourseRunsSerializerV2.programs and V2Course.programs changed from Program[] | null to Program[] now it serializes programs as a list (empty when there are none) rather than null. Updated the course and v2Course test factories to default programs: [] instead of programs: null.

2. display_mode moved onto the program serializers
BaseProgram now carries a display_mode field, so the baseProgram factory was updated to include display_mode: null by default.

3. BaseProgramDisplayMode → V2ProgramDisplayMode
The display-mode enum was renamed as it moved to the V2 program serializer. Updated the import and the isProgramAsCourse signature in dashboardViewModel.ts.

4. Certificate retrieve param cert_uuid → uuid
The bump also renamed this query param. Updated CertificatePage.tsx, the certificate page/PDF routes, and the certificate queries to match.

- [ ] Desktop screenshots
https://github.com/user-attachments/assets/dc9839f4-9fc2-496e-87de-6794678d1ddc

Email Sent:
Screenshot 2026-06-18 at 10 22 31 AM

Copy Claim link:
As daniellerichard1@gmail.com
Screenshot 2026-06-18 at 11 28 18 AM

As the admin after link is claimed:
Screenshot 2026-06-18 at 11 29 41 AM

### How can this be tested?

**Verify email send through the browser**

1. Keep docker compose logs -f celery running.
2. In the B2B contract admin, do a assign or a reassign on a pending seat with a new email.
3.  In the celery logs, the email prints addressed to the new email 

**Copy claim link**
Setup: You need a B2B contract with at least one pending (assigned but not yet redeemed) seat. Navigate to the Contract Admin page for that contract and find the seat table.

1. On a pending seat row, click the ⋮ "More actions" button
2. Click Copy claim link.
3. Verify the menu item swaps to a disabled "Link copied to clipboard" state (it no longer says "Copy claim link").
4. Paste the clipboard contents somewhere (e.g. the URL bar or a text field). It should be: 
`<current origin>/enrollmentcode/<code>` e.g. http://open.odl.local:8062/enrollmentcode/ABC123 — using the same code shown for that row.
5. Log in as the learner the seat was assigned to (the email on that row).
6.  With that learner logged in, open the copied link in a fresh tab. It should land on the enrollment-code claim/attach view for that code 
7. Then log back in as your manager user and verify that the user now has the status of "redeemed" 

**Assign / reassign / remind / revoke**

1. On the Contract Admin page, use the Assign Seats box to assign one or more emails. Confirm the inline result Alert reports success.
2. On a pending row's ⋮ menu: use **Change assigned email** (validates the new email, then reassigns), **Resend claim email**, and **Release seat** (confirmation dialog). Each should succeed and refresh the seat table.
3. ⚠️ Known limitation: **Uninvite** on a redeemed row currently 409s (see Known limitation above)

### DIFF
--- a/frontends/api/src/mitxonline/hooks/organizations/index.ts
+++ b/frontends/api/src/mitxonline/hooks/organizations/index.ts
@@ -7,7 +7,11 @@ import {
   B2bApiB2bManagerOrganizationsContractsCodesRemindCreateRequest,
   B2bApiB2bManagerOrganizationsContractsCodesRevokeDestroyRequest,
 } from "@mitodl/mitxonline-api-axios/v2"
-import { organizationQueries, managerOrganizationQueries } from "./queries"
+import {
+  organizationQueries,
+  managerOrganizationQueries,
+  managerOrganizationKeys,
+} from "./queries"
 
 const useB2BAttachMutation = (opts: B2bApiB2bAttachCreateRequest) => {
   const queryClient = useQueryClient()
@@ -33,15 +37,12 @@ const useBulkAssignSeats = () => {
     mutationFn: (
       opts: B2bApiB2bManagerOrganizationsContractsCodesBulkAssignCreateRequest,
     ) => b2bApi.b2bManagerOrganizationsContractsCodesBulkAssignCreate(opts),
-    onSettled: () => {
+    onSettled: (_data, _err, vars) => {
       queryClient.invalidateQueries({
-        queryKey: [
-          "mitxonline",
-          "manager",
-          "organizations",
-          "contracts",
-          "codes",
-        ],
+        queryKey: managerOrganizationKeys.contractCodes({
+          id: vars.id,
+          parent_lookup_organization: vars.parent_lookup_organization,
+        }),
       })
     },
   })
@@ -54,15 +55,12 @@ const useRemindCode = () => {
     mutationFn: (
       opts: B2bApiB2bManagerOrganizationsContractsCodesRemindCreateRequest,
     ) => b2bApi.b2bManagerOrganizationsContractsCodesRemindCreate(opts),
-    onSettled: () => {
+    onSettled: (_data, _err, vars) => {
       queryClient.invalidateQueries({
-        queryKey: [
-          "mitxonline",
-          "manager",
-          "organizations",
-          "contracts",
-          "codes",
-        ],
+        queryKey: managerOrganizationKeys.contractCodes({
+          id: vars.id,
+          parent_lookup_organization: vars.parent_lookup_organization,
+        }),
       })
     },
   })
@@ -75,15 +73,12 @@ const useRevokeCode = () => {
     mutationFn: (
       opts: B2bApiB2bManagerOrganizationsContractsCodesRevokeDestroyRequest,
     ) => b2bApi.b2bManagerOrganizationsContractsCodesRevokeDestroy(opts),
-    onSettled: () => {
+    onSettled: (_data, _err, vars) => {
       queryClient.invalidateQueries({
-        queryKey: [
-          "mitxonline",
-          "manager",
-          "organizations",
-          "contracts",
-          "codes",
-        ],
+        queryKey: managerOrganizationKeys.contractCodes({
+          id: vars.id,
+          parent_lookup_organization: vars.parent_lookup_organization,
+        }),
       })
     },
   })
@@ -100,15 +95,12 @@ const useReassignCode = () => {
     mutationFn: (
       opts: B2bApiB2bManagerOrganizationsContractsCodesReassignUpdateRequest,
     ) => b2bApi.b2bManagerOrganizationsContractsCodesReassignUpdate(opts),
-    onSettled: () => {
+    onSettled: (_data, _err, vars) => {
       queryClient.invalidateQueries({
-        queryKey: [
-          "mitxonline",
-          "manager",
-          "organizations",
-          "contracts",
-          "codes",
-        ],
+        queryKey: managerOrganizationKeys.contractCodes({
+          id: vars.id,
+          parent_lookup_organization: vars.parent_lookup_organization,
+        }),
       })
     },
   })
--- a/frontends/api/src/mitxonline/hooks/organizations/index.ts
+++ b/frontends/api/src/mitxonline/hooks/organizations/index.ts
@@ -7,11 +7,7 @@ import {
   B2bApiB2bManagerOrganizationsContractsCodesRemindCreateRequest,
   B2bApiB2bManagerOrganizationsContractsCodesRevokeDestroyRequest,
 } from "@mitodl/mitxonline-api-axios/v2"
-import {
-  organizationQueries,
-  managerOrganizationQueries,
-  managerOrganizationKeys,
-} from "./queries"
+import { organizationQueries, managerOrganizationQueries } from "./queries"
 
 const useB2BAttachMutation = (opts: B2bApiB2bAttachCreateRequest) => {
   const queryClient = useQueryClient()
@@ -37,7 +33,7 @@ const useBulkAssignSeats = () => {
     mutationFn: (
       opts: B2bApiB2bManagerOrganizationsContractsCodesBulkAssignCreateRequest,
     ) => b2bApi.b2bManagerOrganizationsContractsCodesBulkAssignCreate(opts),
-    onSettled: (_data, _err, vars) => {
+    onSettled: () => {
       queryClient.invalidateQueries({
         queryKey: [
           "mitxonline",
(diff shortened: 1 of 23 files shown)


frontends/main/src/app-pages/ContractAdminPage/RowActionMenu.tsx
```diff
@@ -78,6 +109,43 @@ const RowActionMenu: React.FC<RowActionMenuProps> = ({ code }) => {
...
+  const handleResend = async () => {
+    handleClose()
+    try {
+      await remind.mutateAsync({
+        code: code.code,
+        id: contractId,
+        parent_lookup_organization: orgId,
+      })
+      onResult(
+        `Claim email resent${code.assigned_to ? ` to ${code.assigned_to}` : ""}.`,
```
dsubak: Is this because we don't disable the button for reassign even if assigned_to is empty?

My recollection is that we don't explicitly guard against the redemption record not having an assigned_to email or sending emails for already redeemed codes, but that's easy and I'm thinking I ought to add both at minimum.

daniellefrappier18: Good catch. You're right we weren't explicitly guarding.

revoke_code and reassign_code both guard with if assignment_record.user or assignment_record.redeemed_on: → 409, but send_reminder_for_code_assignment doesn't.  It only 404s on a missing record, then queues the email regardless. Two cases slip through:

- Redeemed: it'll re-send to an already-claimed code (no 409 in its {200, 404} schema). Implement the same check as in revoke/reassign.
- Blank assigned_email: since status is derived from redeemed_on/user and not the email, a record with an empty assigned_email still reads as "assigned" and has a record so it doesn't 404, and remind would queue an email to an empty address. That's the case my frontend guard catches, but the endpoint should probably bail too if there's no recipient.

Could you add both so my frontend guard isn't the only thing holding the line?

dsubak (approved):
Functionality looked good in testing. I was able to reassign and revoke seats which hadn't been redeemed (and got nice error messages for those that had been). 

Reassign and bulk assign did kick out emails (validated via local SMTP server, see photo below) and I wasn't able to submit any invalid emails on either bulk assign or the single reassign row action. Emails were properly deduped in bulk assignment as well. Claim links still look correctly formed. 

Code seemed reasonable to me - I left one small comment, but definitely don't consider my approval sufficient w/r/t code structure - I'll leave that to folks who've done more frontend work here!

Image

Image

dsubak:
`Log in as the learner the seat was assigned to (the email on that row).`

As a small aside, there's no enforcement that the assigned email is the same as the person redeeming it. Details on that decision can be found at https://github.com/mitodl/mitxonline/pull/3625#discussion_r3359726497

frontends/main/src/app-pages/ContractAdminPage/RowActionMenu.tsx
dsubak: Yup, that makes sense to me and I agree that we should have em on the backend. I'll get that put together shortly!

frontends/api/src/mitxonline/hooks/organizations/index.ts
```diff
@@ -16,5 +26,93 @@ const useB2BAttachMutation = (opts: B2bApiB2bAttachCreateRequest) => {
...
+    mutationFn: (
+      opts: B2bApiB2bManagerOrganizationsContractsCodesBulkAssignCreateRequest,
+    ) => b2bApi.b2bManagerOrganizationsContractsCodesBulkAssignCreate(opts),
+    onSettled: (_data, _err, vars) => {
+      queryClient.invalidateQueries({
+        queryKey: managerOrganizationKeys.contractCodes({
+          id: vars.id,
+          parent_lookup_organization: vars.parent_lookup_organization,
+        }),
+      })
```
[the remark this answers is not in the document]
ChristopherChudzicki: @daniellefrappier18 I believe the code you originally had was correct.

When you bulk assign, you bulk assign within a single (org, contract) pair. That's what the `opts` specify. No need to invalidate other org/contracts the user has fetched. (This certainly works currently though—invalidating too much is definitely better than invalidating too little.)

Whether or not you revert to the old version, I do think it's worth it to use a query key function ("key factory"), lest this hard-coded invalidation ever get out of sync with what the query actually uses. If you wanted to keep it how it is, you could as below and use `contractCodesRoot()`, but like I said, I think what you had originally is fine
```ts
  contractCodesRoot: () =>
    ["mitxonline", "manager", "organizations", "contracts", "codes"] as const,
  contractCodes: (
    opts: B2bApiB2bManagerOrganizationsContractsCodesListRequest,
  ) => [...managerOrganizationKeys.contractCodesRoot(), opts] as const,
```

**Separately:** Just as an FYI: React Query query keys are hierararararachical and invalidation matching is partial, so:
```ts
const key1 = ["a", 5, "b", { x: 1} ] 
const key2 = ["a", 5, "b", { x: 1, y: 2}, "c" ]
const key3 = ["a", 5, "b", { x: 10, y: 2} ]
const key4 = ["a", 2, "b", { x: 1, y: 2} ]
const key5 = ["w", 5, "b", { x: 1, y: 2} ]
```
- hierarrararchical:
    - invalidating `["a"]` invalidates keys 1, 2, 3, 4
    - invalidating ["a", 5"] invalidates keys 1, 2, 3 only
- partial matches on objects
    - invalidating `["a", 5, "b", { x: 1 }]` invalidates keys 1 and 2, but not 3 or 4 or 5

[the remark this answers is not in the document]
ChristopherChudzicki: @daniellefrappier18 BTW, part of the reason for my long-ish explainish here is that Copilot has this backwards. It said:
> , so invalidating with a different opts object containing only {id, parent_lookup_organization} will likely not match and can leave the codes table stale after bulk-assign.

That is **wrong**. If you invalidate using a key that just has too few objects, it **will** work, but will probably invalidate other stuff, too.

frontends/main/src/app-pages/ContractAdminPage/AssignSeatsSection.tsx
```diff
@@ -184,29 +186,51 @@ const tokenizeInput = (input: string, allCommitted: boolean) => {
...
+  orgId,
+  contractId,
   availableSeats,
 }) => {
   const [emailInput, setEmailInput] = useState("")
   const [focused, setFocused] = useState(false)
   const [csvReadError, setCsvReadError] = useState(false)
   const [csvNoValid, setCsvNoValid] = useState(false)
   const [modalData, setModalData] = useState<ModalData | null>(null)
+  const [result, setResult] = useState<AssignResult | null>(null)
```
ChristopherChudzicki, line 228: @daniellefrappier18 This state exists to store the data returned by the API. It works and is pretty clear.

That said mentioning: you don't actually need this `useState`:  `useBulkAssignSeats` is just a custom wrapper around `useMutation`, and `useMutation` stores the data for you already. Let me rename `bulkAssign` to `mutation` for a moment, I.e., `const mutation = useBulkAssignSeats()`. Then:
- `mutation.mutateAsync` => async call to api, you already use it
- `mutation.data` is the data returned by `mutateAsync`, if it returned
- `mutation.error` is the error thrown by `mutateAsync`, if it threw
- `mutation.reset` resets the mutation state (`data`, `error`, and `status`)
- Every time you call `mutateAsync` or `mutate`, the state is reset automatically.

So if you want, you could ditch `result, setResult`, put a pure function helper outside the comppnent, and derive the alert stuff from mutation:
```ts
// Pure function, outside component
const getAlertInfo = ({data, isError}: {
  data: BulkAssignResult | null
  isError: boolean
}) => {
  if (!data && !isError) return null
  if (isError) { // API call errored
    return {
      severity: "error" as const,
      message: "Something went wrong assigning seats. Please try again.",
    }
  }
  if (data) {
    // API returned successfully, may include error data
    // etc
  }
}

// Then in component:
const alertInfo = getAlertInfo(bulkAssign)
```


you'd still need to manually call `mutation.reset()` on the alert dismissal, though.

frontends/api/src/mitxonline/hooks/organizations/index.ts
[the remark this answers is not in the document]
daniellefrappier18: Thanks for the detailed explanation that clears it up :) I've reverted to the original precise invalidation scoped to the mutated (org, contract) pair, and applied your contractCodesRoot() suggestion so the key prefix is defined once and contractCodes() composes from it.

ChristopherChudzicki (approved):

PR #3502, opened 2026-06-18
remove langchain-openai

abeglova:
None

renovate has a pr to update langchain-openai: https://github.com/mitodl/mit-learn/pull/3497

I believe we can remove it instead. We don't call openai through langchain-openai but through litellm

### How can this be tested?
tests should pass
Verify that learn can access open ai. 
set OPENAI_API_KEY

Run

```
from  learning_resources.models import *
from learning_resources.content_summarizer import ContentSummarizer

ContentSummarizerConfiguration.objects.get_or_create(
	allowed_content_types=["file"],
	allowed_extensions=[".srt", ".vtt"],
	is_active=True,
	llm_model="gpt-4o-mini",
	defaults={
		"platform": LearningResourcePlatform.objects.get(code="mitxonline")
	}
)
cf = LearningResource.objects.filter(require_summaries=True).first().runs.first().content_files.filter(file_extension=".srt").first()

```
verify that cf.content is not blank

Then run
```
summarizer.summarize_content_files_by_ids([cf.id], overwrite=True)
```

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -56,7 +56,6 @@ dependencies = [
     "jedi>=0.19.0,<0.20",
     "langchain>=1.3.9,<1.4",
     "langchain-experimental>=0.4,<0.5",
-    "langchain-openai>=0.3.2,<0.4",
     "litellm==1.84.0",
     "llama-index>=0.14.0,<0.15",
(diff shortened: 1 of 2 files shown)


shanbady (approved):
LGTM 👍

PR #3503, opened 2026-06-18
Ingest contentfile archives for all runs (two-tier indexing)

mbertrand:
Closes mitodl/hq#11872

Ingests and embeds **every** run's contentfiles into Qdrant, not just the best published run.

- **Qdrant** (powers the VideoGPT bot, TutorBot,  `vector_content_files_search` endpoint): should now hold contentfiles from all runs of published/test-mode courses, so any run's content is findable.    
- **OpenSearch** (resource search): unchanged — should still only contain contentfiles from the best published run (or any published run of a `test_mode` course), as children or nested documents of resources.  

In this branch, all runs, both published and unpublished, of a published/`test_mode` course are loaded into the DB and embedded into Qdrant.  Unpublishing a run removes that run's contentfiles from OpenSearch without flipping `ContentFile.published`, keeping it in Qdrant and the REST API.  Ditto for an published course that is later unpublished.  Only deleting a course/run will remove its contentfiles from Qdrant.

Opensearch behavior should be unaffected (still only contentfiles from best published run should be indexed, and deindexed if that course/run is unpublished).

### How can this be tested?
Pick a mitxonline course that has **both published and unpublished runs**, then backpopulate its files.  A good candidate  is `course-v1:MITxT+15.356.2x` (3 runs with about 67 content files each).

```
./manage.py backpopulate_mitxonline_files --resource-ids <resource_id> --overwrite
```

Then verify:
- **Qdrant / `vector_content_file_search`** contains  results from *all* runs of the course (published and unpublished):
http://open.odl.local:8065/api/v0/vector_content_files_search/?run_readable_id=course-v1%3AMITxT%2B15.356.2x%2B2T2023
http://open.odl.local:8065/api/v0/vector_content_files_search/?run_readable_id=course-v1%3AMITxT%2B15.356.2x%2B3T2025
http://open.odl.local:8065/api/v0/vector_content_files_search/?run_readable_id=course-v1%3AMITxT%2B15.356.2x%2B3T2026

- But it returns only results from the best run if `resource_id` is a parameter without `run_readable_id` or `edx_module_id`:
- http://open.odl.local:8065/api/v0/vector_content_files_search/?resource_readable_id=course-v1%3AMITxT%2B15.356.2x

- **OpenSearch / `learning_resources_search` and `content_files_search** still returns contentfiles only from the best published run (plus any "runless" content file(s) for the resource):
http://open.odl.local:8065/api/v1/content_file_search/?resource_id=<resource_id for course-v1:MITxT+15.356.2x>

- Unpublishing one run drops it from OpenSearch but keeps it in Qdrant; unpublishing the whole course works the same.

### DIFF
--- a/learning_resources/etl/constants.py
+++ b/learning_resources/etl/constants.py
@@ -95,6 +95,15 @@ class ETLSource(ExtendedEnum):
     ovs = "ovs"
 
 
+QDRANT_RETAINED_SOURCES = (
+    ETLSource.mit_edx.value,
+    ETLSource.mitxonline.value,
+    ETLSource.xpro.value,
+    ETLSource.oll.value,
+    ETLSource.canvas.value,
+)
+
+
 class CourseNumberType(Enum):
     """Enum of course number types"""
 
--- a/learning_resources/etl/edx_shared.py
+++ b/learning_resources/etl/edx_shared.py
@@ -106,6 +106,9 @@ def build_run_lookup(
     for run in runs:
         normalized = normalize_run_id(etl_source, run.run_id)
         lookup.setdefault(normalized, []).append(run)
+        if etl_source == ETLSource.oll.name:
+            # OLL archive filenames omit the MITx+ prefix the run_ids carry
+            lookup.setdefault(normalized.removeprefix("mitx."), []).append(run)
     return lookup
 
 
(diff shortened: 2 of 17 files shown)


shanbady (approved):
changes here look good and function as needed. 

some lingering concerns i had (but not something to address here):

- how many new records will qdrant have to handle? (we are close to 5 million points on the contentfiles collection) - it might make sense to switch to 1024 for chunk size instead of 512 if this becomes an issue (to reduce the size of the collection)
- performance concerns related to the above 
- growing complexity around all the conditionals for what gets ingested/visible etc

shanbady (approved):
looks good - just minor nit with the name "was_published" - maybe "previously_published" might be more clear

PR #3504, opened 2026-06-18
add card type and end date

gumaerc:
https://github.com/mitodl/hq/issues/11878

This PR adds the "card type" indicator to both Course and Program cards. Since we removed this text from the CTA buttons, leaving only the verb, we now indicate what type of resource the card represents with this new indicator in the top left of the card.

This PR also adds an "end date" display to the left of where the ceritficate upgrade / view link appears on the full size cards and to the left of the CTA button on compact cards. If the date is in the future it says "Ends in x days." If the date is in the past, it says "Ended x days ago." Additionally, the logic surrounding when the certificate upgrade link was adjusted. Previously, the certificate upgrade link would show on course runs that had ended. This PR ensures that it does not appear if the course has ended.

image
image
image
image

#### Card type indicator
1. Navigate to the dashboard (home or a program page that shows module rows).
2. Verify each course card shows **"Course"** in small gray text in the top-left corner above the title.
3. Verify program cards show **"Program"** in the same position.

#### Start date display — enrolled course (full/default layout)
1. Find an enrolled course with a **future** start date. Verify the card shows **"Starts in X days"** ("Starts tomorrow" if one day away) to the left of the certificate/upgrade link row, below the title.

#### End date display — enrolled course (full/default layout)
1. Find an enrolled course with a **future** end date. Verify the card shows **"Ends in X days"** ("Ends tomorrow" if one day away) to the left of the certificate/upgrade link row, below the title.
2. Find an enrolled course with a **past** end date. Verify the card shows **"Ended X days ago"** in the same position.
3. Find an enrolled course with **no end date**. Verify no end date text appears.

#### End date display — compact layout (program module rows)
1. Open a program that shows module rows (compact cards).
2. Find a module (enrolled or unenrolled) with a future end date. Verify **"Ends in X days"** appears **to the left of the CTA button** (Continue / Start), not below the title.
3. Find a module with a past end date. Verify **"Ended X days ago"** appears in the same position.

#### Date popover
Click any of the date indicators and you should see a popover. Verify that it displays the date information like the Figma and Google Doc describe:
https://www.figma.com/design/mSDdN8giMcncBwNgwcVJL9/UX-Wireframes?node-id=15731-427278&m=dev
https://docs.google.com/document/d/1YLVVvzH7CX5dsIxcSAgcNJJ_VJBF69pVc1ScJAbUWXg/edit?tab=t.0

#### Certificate upgrade link gating
1. Find an enrolled course where the course run has **ended**. Verify the **"Add a certificate"** upgrade link does **not** appear.
2. Find an enrolled course where the **upgrade deadline** has passed (but the course itself is still running). Verify the upgrade link does **not** appear.
3. Find an enrolled course that is active and upgradable. Verify the upgrade link **does** appear as before.


daniellefrappier18 (commented):
Probably not bound to come in a real situation but `formatCalendarDays()` returns with no fallback to an actual calendar date for large values.  So for longer dates we could end up with something like this:

Image

daniellefrappier18 (commented):
When `layout="compact"` and both CourseDateSummary and the upgrade banner are present, the right-side row gets crowded enough to squeeze the title into two lines. 

Does the spec account for this combination? 

Image

gumaerc:
> Probably not bound to come in a real situation but `formatCalendarDays()` returns with no fallback to an actual calendar date for large values. So for longer dates we could end up with something like this:
> 
> Image

@steven-hatch and I spoke about this and the rough consensus right now is that after ~3 months it should switch over to a calendar date. It totally makes sense from a readability standpoint; I can't really picture how far away 5000 days actually is. The main issues as I see them are internationalization and character count. 3 months is roughly 90 days (based on one semester) so we'd be replacing "Ends in 90 days" with "Ends on MM/DD/YYYY" with variance depending on country. Basically, the text is potentially just a bit longer but not by much. The bigger issue really is date formatting but `Intl` should be able to handle that.

frontends/main/src/app-pages/DashboardPage/CoursewareDisplay/EnrolledCourseCard.tsx
```diff
@@ -125,40 +130,48 @@ export const EnrolledCourseCard = ({
...
-      {certificateLink ? (
-        <SubtitleLink href={certificateLink} layout={layout}>
-          <RiAwardLine size="16px" />
-          {isCompact ? "Certificate" : "View Certificate"}
-        </SubtitleLink>
+  const endDateAndCertSection = (
+    <Stack direction="row" alignItems="center">
+      {courseDateText}
+      {courseDateText !== null && (showUpgradeBanner || !!certificateLink) ? (
+        <Separator />
```
daniellefrappier18: courseDateText holds a JSX element, not a primitive value, so courseDateText !== null is always true. This means the Separator renders whenever showUpgradeBanner || !!certificateLink is true,even when CourseDateSummary produces no visible output (like below).  The fix is to replace the JSX null-check with hasCourseDateText, a boolean derived from the underlying date data that mirrors the conditions under which getCourseDateText would return a non-null value.

Image

frontends/main/src/app-pages/DashboardPage/CoursewareDisplay/CardShared.tsx
```diff
@@ -183,79 +169,145 @@ const CoursewareButtonLink = styled(ButtonLink)(({ theme, variant }) => ({
...
+            size="small"
+            onClick={() => setPopoverAnchorEl(null)}
+          >
+            Got it!
+          </DatePopoverDismissButton>
+        </DatePopoverContent>
+      </Popover>
+      <DatePopoverTrigger
+        color="black"
+        onClick={(event) => setPopoverAnchorEl(event.currentTarget)}
```
daniellefrappier18, line 300: DatePopoverTrigger is `styled(Link)` with no `href.` A Link component without href typically renders as a non-interactive element (span/div), making it unreachable by keyboard (Tab) and incapable of being activated(no Enter/Space). It's also missing aria-expanded, aria-haspopup, and aria-controls to communicate the popover relationship to assistive technology. The popover itself has no role="dialog" or aria-label.

Suggested Fix: Use a `<button>` element instead of a `styled(Link)` and add the necessary ARIA attributes:


```
<DatePopoverTrigger
  component="button"
  aria-expanded={!!popoverAnchorEl}
  aria-haspopup="dialog"
  onClick={(event) => setPopoverAnchorEl(event.currentTarget)}
>
  {triggerText}
</DatePopoverTrigger>
```

frontends/main/src/app-pages/DashboardPage/CoursewareDisplay/ProgramEnrollmentCard.tsx
```diff
@@ -66,11 +67,11 @@ export const ProgramEnrollmentCard = ({
           View Certificate
         </SubtitleLink>
       ) : null}
-    </>
+    </Stack>
   )
   const buttonSection = (
     <ButtonLink size="small" variant="primary" href={programView(program.id)}>
-      View Program
+      View
```
daniellefrappier18, line 79: For sighted users I think there is enough context with the card that View Program to View is fine but for a screenreader user they would just hear "View" and I don't think this alone would describe what the control does. 

Suggested Fix: add aria-label={View program: ${program.title}} to the button.

daniellefrappier18 (approved):
I don't see anything blocking here. I called out some minor a11y issues and another minor UI edge case.

PR #3505, opened 2026-06-19
Improvements and fixes for issues in the vector search app

shanbady:

Closes https://github.com/mitodl/hq/issues/11822

This PR fixes a set of issues claude had discovered after I asked for an audit:
1. embed_new_learning_resources queues duplicate tasks.
2. vector_search/constants.py (line 40) maps level to runs[].level[].code, but vector_search/constants.py (line 71) indexes
3. tune_collection runs inside every embed
4. Hybrid prefetch nesting likely caps candidates at 10 per branch.
5. `client.get_collection(collection_name=collection_name)` executing in loop (leads to unecessary requests per chunk to qdrant)

#### issue 1
1. checkout main 
2. run the following in django shell which simulates queuing up  embed_new_learning_resources tasks:
```python
import datetime
from unittest.mock import patch

from django.test import override_settings

from learning_resources.factories import LearningResourceFactory
from learning_resources_search.constants import COURSE_TYPE, PROGRAM_TYPE
from vector_search.tasks import embed_new_learning_resources

with override_settings(QDRANT_EMBEDDINGS_TASK_LOOKBACK_WINDOW=60, QDRANT_CHUNK_SIZE=100):
    resources = (
        LearningResourceFactory.create_batch(3, resource_type=COURSE_TYPE, published=True)
        + LearningResourceFactory.create_batch(2, resource_type=PROGRAM_TYPE, published=True)
    )

    for resource in resources:
        resource.created_on = now_in_utc() - datetime.timedelta(minutes=1)
        resource.save(update_fields=["created_on"])

    with (
        patch("vector_search.tasks.generate_embeddings") as generate_embeddings,
        patch("vector_search.tasks.celery.group", side_effect=lambda tasks: list(tasks)),
        patch.object(embed_new_learning_resources, "replace", side_effect=lambda task_group: task_group),
    ):
        embed_new_learning_resources.run()

        print(generate_embeddings.si.call_count)
        print([call.args for call in generate_embeddings.si.call_args_list])

    LearningResource.objects.filter(id__in=[resource.id for resource in resources]).delete()
```
3. note how many tasks are started
4. checkout this branch and execute the same and note that far fewer tasks are started

#### issue 2
1. checkout this branch
2. delete your local learning resources qdrant collection and re-embed learning resources via ```python manage.py generate_embeddings --courses --skip-contentfiles```
3. visit your local qdrant dashoard and check the collection info for the learning resources collection. you should see "runs[].level[].code" as an indexed payload field and that there are points associated with it.

#### issue 3
for this issue there is not much to locally test - reference [this change](https://github.com/mitodl/mit-learn/pull/3505/changes#diff-35a3042773b7652baf679eb2fe743e78debcf410d4ed6c72196fac1d9f52d8b8R189) which simply moves the tune_collection call to a scheduled celery task instead of in the hot path of embeddings

### issue 4
This issue is also a bit tricky to visually test locally. just verify that [hybrid search on the topic pages](http://open.odl.local:8062/c/topic/biology) behave as expected (assuming you have embedded learning resources locally)

## Addition Details
also included is this [small tweak](https://github.com/mitodl/mit-learn/pull/3505/changes#diff-89a385763e73d2532fc11a069c2fcba173bd13b6f98fd229eaf62981f150d1a4R1340) to the score decay function after some manual testing on our deployed environments


vector_search/utils.py
```diff
@@ -204,6 +204,27 @@ def create_qdrant_collections(force_recreate):
     update_qdrant_indexes()
 
 
+def tune_qdrant_collections():
```
mbertrand, line 207: Unit test?

mbertrand (approved):
LGTM, could use a unit test for the new task but that's it

PR #3508, opened 2026-06-22
fix(dashboard): enroll in the displayed variant run, not getBestRun's…

ChristopherChudzicki:
none

Fixes a contract bug where contract page was displaying the variant card but enrolling people in the non-variant (aka default variant) run.

### How can this be tested?
1. Set up a contract with variants
2. Try enrolling in default variant vs non-default variant. It should enroll you correctly.
3. The enrollment button should be correctly disabled if a run is not enrollable.


gumaerc (approved):
👍 

Tested this locally with a clone of the Novant contract and I was successfully enrolled in the variant run.

PR #3514, opened 2026-06-23
remove semantic chunker and dependance on langchain-experimental

shanbady:

Closes  https://github.com/mitodl/hq/issues/11945

This PR removes the Semantic chunking functionality which is currently unused and also depends on the [deprecated langchain-experimental library](https://github.com/langchain-ai/langchain-experimental/issues/87). the SemanticChunker is also not something that appears to be ported to a now supported library.

### How can this be tested?
Tests should pass. Also verify nothing else depends or imports from the langchain-experimental library


mbertrand (approved):
👍

PR #3512, opened 2026-06-23
Program certificate title from CMS "Certificate Title" (behind flag)

zamanafzal:
### What are the relevant tickets?
https://github.com/mitodl/hq/issues/11907

### Description (What does it do?)

- Program certificates on MIT Learn now show the CMS certificate page's `product_name` ("Certificate Title") as the displayed title, instead of the program title.
- Gated behind the **`cms-certificate-title`** PostHog flag. Flag **off** (default) or empty `product_name` → falls back to the program title (today's behaviour).
- **Course** certificates are unchanged. **PDF + SEO metadata** intentionally stay on the program title for now (they run server-side where we don't yet evaluate PostHog) — gated in a follow-up.
- Adds a `getCertificateTitle(productName, fallbackTitle)` helper + unit tests, and flag on/off + fallback coverage. The title is resolved once in `CertificatePage` and passed to the sub-components (printed cert, page heading, share text, download filename all use the same value).

> **Rollout note:** keep the flag **off in production** until #11938 (server-side gating of the PDF + share/SEO metadata) lands; otherwise an enabled flag shows the CMS title on the page and share text but the program title in the downloaded PDF and the link-preview card.

### Screenshots (if appropriate):
With Flag On

Screenshot 2026-06-23 at 5 08 45 PM

With Flag OFF
Screenshot 2026-06-23 at 5 08 05 PM

Screenshot 2026-06-23 at 5 06 42 PM

### How can this be tested?

**Need (a) a **program** certificate whose CMS "Certificate Title" differs from its program title, and (b) enable the cms-certificate-title.

1. **Get a certificate whose CMS "Certificate Title" ≠ program title** — either use an existing one (e.g. a UAI program cert; confirm the value in the CMS → Certificate page), **or create one** (set a distinct "Certificate Title" in the CMS + publish, then issue a cert for that program via MITx Online Django admin — no custom scripts).

3. **Toggle the cms-certificate-title flag
4. **Verify** — flag ON → CMS title; OFF → program title; course unchanged; PDF/metadata still program title.

PR #3518, opened 2026-06-24
Use CMS Certificate Title for program LinkedIn "Add to Profile"

zamanafzal:

Follow-up to the certificate-title work ([hq#11907](https://github.com/mitodl/hq/issues/11907), tracked in [hq#11938](https://github.com/mitodl/hq/issues/11938)).

### Description

The LinkedIn **"Add to Profile"** link for a **program** certificate currently sends the **program title** as the credential `name`. This changes it to the CMS **"Certificate Title"** (`CertificatePage.product_name`), so the certification a learner adds to LinkedIn matches the title printed on the certificate and shown on the certificate page. Falls back to the program title when `product_name` is unset.

Reuses the `getCertificateTitle` helper (added in #3512) — the same resolver the certificate page heading, PDF, and metadata already use — so the LinkedIn name stays consistent with the displayed title.

- **Programs only.** The course branch of `getCertificateLinkedInUrl` is unchanged.
- This is the **unsigned** LinkedIn path, used when the viewer is not the cert owner, or when the certificate has no verifiable credential. When the owner has a signed verifiable credential, the LinkedIn name comes from the VC via `getVerifiableCredentialLinkedInURL` — which mitxonline [#3698](https://github.com/mitodl/mitxonline/pull/3698) updates to carry the same Certificate Title. Both paths therefore converge on the Certificate Title.

## Change

`frontends/main/src/common/certificateUtils.ts` — `getCertificateLinkedInUrl`, program branch:

```ts
getCertificateTitle(
  (certificateData as V2ProgramCertificate).certificate_page?.product_name,
  (certificateData as V2ProgramCertificate).program.title,
)
```

## How to test

Automated:

```
yarn test frontends/main/src/common/certificateUtils.test.ts
```

New `getCertificateLinkedInUrl` tests assert:
- program cert with a CMS `product_name` distinct from `program.title` → LinkedIn `name` param is the `product_name`;
- empty `product_name` → falls back to `program.title`;
- course cert → still uses the course title.

Manual:
1. Pick a **program certificate** whose CMS "Certificate Title" (`product_name`) differs from the program title (e.g. program title `… MicroMasters`, Certificate Title `… MicroMasters Certificate`).
2. Open its certificate page at `/certificate/program/<uuid>` and confirm the heading shows the **Certificate Title**.
3. Click **Share → LinkedIn / Add to profile**.
4. Inspect the LinkedIn "Add to Profile" URL (address bar, or right-click the option → copy link): the **`name=`** param should be the **Certificate Title**, not the bare program title.
5. _(Most convincing)_ Change the "Certificate Title" in the MITx Online Wagtail admin to a distinct value, publish, reload the page, and confirm the LinkedIn `name=` follows the new title — proving it reads the CMS field live.

### DIFF
--- a/frontends/main/src/common/certificateUtils.test.ts
+++ b/frontends/main/src/common/certificateUtils.test.ts
@@ -1,9 +1,12 @@
 import {
+  CertificateType,
   getCertificateBadgeLines,
   getCertificateBadgeTypography,
   getCertificateInfo,
+  getCertificateLinkedInUrl,
   getCertificateTitle,
 } from "./certificateUtils"
+import { factories } from "api/test-utils"
 
 describe("getCertificateInfo", () => {
   it("returns default certificate label when no program type is provided", () => {
@@ -146,5 +149,46 @@ describe("getCertificateBadgeTypography", () => {
     expect(getCertificateBadgeTypography("  MicroMasters®  ")).toEqual(
       micromastersTypography,
     )
+  })
+})
+
+describe("getCertificateLinkedInUrl", () => {
+  const pageUrl = "https://example.com/certificate/program/abc"
+
+  it("uses the CMS product name for program certificates", () => {
+    const certificate = factories.mitxonline.programCertificate()
+    certificate.certificate_page.product_name = "Universal AI"
+    certificate.program.title = "Fundamentals of Programming and ML"
+
+    const url = new URL(
+      getCertificateLinkedInUrl(CertificateType.Program, certificate, pageUrl),
+    )
+
+    expect(url.searchParams.get("name")).toBe("Universal AI")
+  })
+
+  it("falls back to the program title when product name is empty", () => {
+    const certificate = factories.mitxonline.programCertificate()
+    certificate.certificate_page.product_name = ""
+    certificate.program.title = "Fundamentals of Programming and ML"
+
+    const url = new URL(
+      getCertificateLinkedInUrl(CertificateType.Program, certificate, pageUrl),
+    )
+
+    expect(url.searchParams.get("name")).toBe(
+      "Fundamentals of Programming and ML",
+    )
+  })
+
+  it("uses the course title for course certificates", () => {
+    const certificate = factories.mitxonline.courseCertificate()
+    certificate.course_run.course.title = "Intro to Python"
+
+    const url = new URL(
+      getCertificateLinkedInUrl(CertificateType.Course, certificate, pageUrl),
+    )
+
+    expect(url.searchParams.get("name")).toBe("Intro to Python")
   })
 })
--- a/frontends/main/src/common/certificateUtils.ts
+++ b/frontends/main/src/common/certificateUtils.ts
@@ -253,14 +253,17 @@ export const getCertificateLinkedInUrl = (
   certificateData: V2ProgramCertificate | V2CourseRunCertificate,
   pageUrl: string,
 ): string => {
-  const credentialName =
-    certificateType === CertificateType.Course
-      ? (certificateData as V2CourseRunCertificate).course_run.course.title
-      : getCertificateTitle(
-          (certificateData as V2ProgramCertificate).certificate_page
-            ?.product_name,
-          (certificateData as V2ProgramCertificate).program.title,
-        )
+  let credentialName: string
+  if (certificateType === CertificateType.Course) {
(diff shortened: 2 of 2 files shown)


frontends/main/src/common/certificateUtils.ts
```diff
@@ -256,7 +256,11 @@ export const getCertificateLinkedInUrl = (
   const credentialName =
     certificateType === CertificateType.Course
       ? (certificateData as V2CourseRunCertificate).course_run.course.title
-      : (certificateData as V2ProgramCertificate).program.title
+      : getCertificateTitle(
+          (certificateData as V2ProgramCertificate).certificate_page
+            ?.product_name,
+          (certificateData as V2ProgramCertificate).program.title,
+        )
```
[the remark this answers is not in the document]
ahtesham-quraish: This resolves `certificate_page.product_name` for the LinkedIn credential name unconditionally, but the page heading, `SharePopover` title, and PDF download filename all gate `product_name` behind the `CmsCertificateTitle` flag. When the flag is **off** and a `product_name` is set, the page and PDF show the program title while the LinkedIn "Add to Profile" `name` shows the Certificate Title — so the two diverge in exactly the scenario the flag is meant to control.

Could we route this through the same flag-aware value instead? `CertificatePage.tsx` already computes the resolved `title` (the `useCmsTitle` branch), so the cleanest option is probably to pass that `title` into `getCertificateLinkedInUrl` and use it directly, keeping page, PDF, and LinkedIn on one source of truth:

```ts
const credentialName =
  certificateType === CertificateType.Course
    ? (certificateData as V2CourseRunCertificate).course_run.course.title
    : resolvedTitle // flag-aware title passed in from the caller
```

Alternatively, gate the `product_name` argument behind `CmsCertificateTitle` at the call site. Either way, the LinkedIn name should follow the same flag as the displayed title.

frontends/main/src/common/certificateUtils.test.ts
```diff
@@ -148,3 +151,44 @@ describe("getCertificateBadgeTypography", () => {
...
+      getCertificateLinkedInUrl(CertificateType.Program, certificate, pageUrl),
+    )
+
+    expect(url.searchParams.get("name")).toBe(
+      "Fundamentals of Programming and ML",
+    )
+  })
+
+  it("uses the course title for course certificates", () => {
+    const certificate = factories.mitxonline.courseCertificate()
```
ahtesham-quraish, line 185: the new tests never exercise the CmsCertificateTitle flag being off. Once the flag gap here https://github.com/mitodl/mit-learn/pull/3518/changes#r3503883608 is addressed, add a case asserting the LinkedIn name falls back to program.title when the flag is disabled (mirroring the existing page-level flag behavior), so the divergence can't silently regress.

frontends/main/src/common/certificateUtils.ts
```diff
@@ -256,7 +256,11 @@ export const getCertificateLinkedInUrl = (
   const credentialName =
     certificateType === CertificateType.Course
       ? (certificateData as V2CourseRunCertificate).course_run.course.title
-      : (certificateData as V2ProgramCertificate).program.title
+      : getCertificateTitle(
```
ahtesham-quraish: the repeated (certificateData as V2ProgramCertificate) casts are a bit noisy; consider narrowing once to a local const program = certificateData as V2ProgramCertificate at the top of the program branch and reusing it for both certificate_page?.product_name and program.title for readability.

ahtesham-quraish (changes requested):
Looks good and Left few comments

frontends/main/src/common/certificateUtils.ts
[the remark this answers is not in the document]
zamanafzal: @ahtesham-quraish, your understanding is wrong here. The flag was removed as part of this [PR](https://github.com/mitodl/mit-learn/pull/3517/changes) and no longer exists.

frontends/main/src/common/certificateUtils.test.ts
zamanafzal: @ahtesham-quraish There is no gap as the feature flag no longer exist.

ahtesham-quraish: I see let me mark this closed

ahtesham-quraish (approved):

PR #3522, opened 2026-06-25
Revert "Retry embeddings tasks on Qdrant gRPC DEADLINE_EXCEEDED (#3520)"

mbertrand:
This reverts a previous commit that rertried grpc errors and raise after final retry.  Reverting because the `embed_run_content_files` task creates a chain of `generate_embeddings` tasks and if one fails, the entire rest of the chain is halted.   Will figure out an alternative approach later.

Related  to https://github.com/mitodl/hq/issues/11987

Just reverts the commit

### How can this be tested?
Tests should pass

### DIFF
--- a/vector_search/tasks.py
+++ b/vector_search/tasks.py
@@ -2,7 +2,6 @@ import datetime
 import logging
 
 import celery
-import grpc
 import sentry_sdk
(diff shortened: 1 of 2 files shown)


shanbady (approved):
just a revert. approved

PR #3525, opened 2026-06-25
Canvas-AI: resolve orphaned files

shanbady:

Closes https://github.com/mitodl/hq/issues/11940

This PR resolves an issue for the canvas etl where we incorrectly skip files that: 
- are uploaded with no explicit "published"/visible state, are not changed or modified (so it never shows up in the files_meta.xml) and are not embedded on any page or module (so will not show up in modules_meta.xml)

### How can this be tested?
- Check out main
- restart celery, fetch canvas course [2198](https://canvas.mit.edu/courses/2198) course via:`./manage.py backpopulate_canvas_courses --canvas-ids 2198`
- once done, view the contentfiles that were created in django shell:
```python
list(LearningResource.objects.get(readable_id__icontains="2198").runs.first().content_files.all().values_list("source_path",flat=True))
```
- make note of the list of files - "21G.515-Syllabus-Fall2025.docx" which is a file visible to students is missing from the list
- checkout this branch - restart celery and do the same - note that there are more files and 21G.515-Syllabus-Fall2025.docx is in the list

### DIFF
--- a/learning_resources/etl/canvas_test.py
+++ b/learning_resources/etl/canvas_test.py
@@ -18,6 +18,7 @@ from learning_resources.etl.canvas_utils import (
     _compact_element,
     get_published_items,
     is_file_published,
+    parse_canvas_files,
     parse_canvas_settings,
     parse_files_meta,
     parse_module_meta,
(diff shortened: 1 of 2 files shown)


abeglova (approved):
LGTM

PR #3526, opened 2026-06-25
feat(vector_search): embeddings deadline resilience (retry + finalize tail)

mbertrand:
## What are the relevant tickets?

Closes mitodl/hq#11987

## Description

Makes content-file embedding resilient to Qdrant gRPC `DEADLINE_EXCEEDED` errors that occur when many runs embed at once (e.g. `backpopulate_mitxonline_files`) and stop the parent task getting stuck in STARTED.

- **Run every chunk, then fail loud.** `generate_embeddings` owns its retry (jittered exponential backoff, `max_retries=3`). On a *keyed* path, exhausted/non-transient failures are recorded to a per-invocation counter in the shared `redis` cache and swallowed so the chain continues; a `finalize_embeddings` tail then fails the parent (→ FAILED, never stuck) if any chunk failed. Generic callers (no `failure_key`) propagate exactly as before.
- **Cut per-chunk Qdrant load.** `create_qdrant_collections` now runs once per worker process (was once per ~10-file chunk).
- **Defer indexing during bulk writes.** Optimizer `indexing_threshold` ratio 0.4 → 0.8.

> [!NOTE]
> [A related PR](https://github.com/mitodl/ol-infrastructure/pull/4827) to try to  reduce the DEADLINE_EXCEEDED errors — a dedicated, concurrency-bounded `celery-embeddings` worker — will be in an ol-infrastructure PR. This PR is the in-app resilience + load reduction; it makes deadlines non-fatal but does not by itself bound write concurrency.

## How can this be tested?

Reproduce the original failure and confirm the new behavior on a worker with `embeddings` queue:

1. Embed a run's content files: `embed_run_content_files.delay(<run_id>)` (or trigger a content-file load).
2. **Stuck-parent fix:** force a chunk to fail —  temporarily `raise` in `embed_learning_resources`. Confirm the remaining chunks still run, the parent task resolves to **FAILURE** (not stuck in STARTED) in Flower/the result backend, and the failure shows in Sentry.
3. **Counter cleanup:** check `redis` has no leftover `embed_errors:<task_id>` key after `finalize_embeddings` runs.
4. **Per-process guard:** with debug logging, confirm `create_qdrant_collections` is called once per worker process across many chunks, not per chunk.
5. Happy path: a clean run finishes SUCCESS with all points embedded.


shanbady (approved):
LGTM

PR #3530, opened 2026-06-26
feat: server-side pagination and search for contract codes

daniellefrappier18:
#### Summary

- Migrates contract codes list from client-side to server-side with pagination and debounced search
- Adds a dedicated `contractDetail` API query for per-status seat counts (total, unassigned, assigned, redeemed) instead of deriving them by filtering the full codes list client-side
- Implements the CSV export button 
- Removes the "Uninvite" action for redeemed codes — redeemed rows now show a disabled button with no menu

### Screen Recording (if appropriate):

https://github.com/user-attachments/assets/3a9ca69c-7b78-4c06-a567-4e3856baf857

Using NVDA

https://github.com/user-attachments/assets/799756a6-f534-4658-a4db-ee595f92c8b3

### How can this be tested?

- [ ] Load the Contract Admin page — verify the four stat blocks Total purchased, Unassigned, Pending claim, Redeemed
- [ ] Type in the search box — verify results filter server-side after a short debounce. Screen reader announcement reflects the correct total count after results load
- [ ] Clear the search — verify the full list is restored
- [ ] Tab between All / Pending claim / Redeemed — verify tab filtering works on the current page of results
- [ ] Paginate through results — verify the correct page loads and the footer shows the right page info
- [ ] Click "Export CSV" — verify a CSV downloads with all assigned/redeemed codes 
- [ ] Confirm redeemed rows show a disabled action button with no menu

### DIFF
--- a/frontends/api/package.json
+++ b/frontends/api/package.json
@@ -16,6 +16,7 @@
     "./test-utils/mockAxios": "./src/test-utils/mockAxios.ts",
     "./test-utils": "./src/test-utils/index.ts",
     "./mitxonline-hooks/*": "./src/mitxonline/hooks/*/index.ts",
+    "./mitxonline-clients": "./src/mitxonline/clients.ts",
     "./mitxonline-test-utils": "./src/mitxonline/test-utils/index.ts"
   },
   "peerDependencies": {
@@ -31,7 +32,7 @@
     "ol-test-utilities": "0.0.0"
   },
   "dependencies": {
-    "@mitodl/mitxonline-api-axios": "2026.6.17",
+    "@mitodl/mitxonline-api-axios": "https://github.com/mitodl/mitxonline-api-clients/raw/refs/heads/dansubak/202606_paginated_codes_means_paginated_problems-client-7af6a2c/src/typescript/mitxonline-api-axios/package.tgz",
     "@tanstack/react-query": "^5.66.0",
     "axios": "^1.12.2",
     "tiny-invariant": "^1.3.3"
--- a/frontends/api/package.json
+++ b/frontends/api/package.json
@@ -32,7 +32,7 @@
     "ol-test-utilities": "0.0.0"
   },
   "dependencies": {
-    "@mitodl/mitxonline-api-axios": "https://github.com/mitodl/mitxonline-api-clients/raw/refs/heads/dansubak/202606_paginated_codes_means_paginated_problems-client-7dbc25d/src/typescript/mitxonline-api-axios/package.tgz",
+    "@mitodl/mitxonline-api-axios": "https://github.com/mitodl/mitxonline-api-clients/raw/refs/heads/dansubak/202606_paginated_codes-client-443be46/src/typescript/mitxonline-api-axios/package.tgz",
     "@tanstack/react-query": "^5.66.0",
     "axios": "^1.12.2",
     "tiny-invariant": "^1.3.3"
--- a/frontends/api/package.json
+++ b/frontends/api/package.json
@@ -31,7 +31,7 @@
     "ol-test-utilities": "0.0.0"
   },
   "dependencies": {
-    "@mitodl/mitxonline-api-axios": "https://github.com/mitodl/mitxonline-api-clients/raw/refs/heads/dansubak/202606_paginated_codes-client-443be46/src/typescript/mitxonline-api-axios/package.tgz",
+    "@mitodl/mitxonline-api-axios": "2026.6.30-1",
     "@tanstack/react-query": "^5.66.0",
     "axios": "^1.12.2",
     "tiny-invariant": "^1.3.3"
--- a/frontends/api/package.json
+++ b/frontends/api/package.json
@@ -32,7 +32,7 @@
     "ol-test-utilities": "0.0.0"
   },
   "dependencies": {
-    "@mitodl/mitxonline-api-axios": "https://github.com/mitodl/mitxonline-api-clients/raw/refs/heads/dansubak/202606_paginated_codes_means_paginated_problems-client-7af6a2c/src/typescript/mitxonline-api-axios/package.tgz",
+    "@mitodl/mitxonline-api-axios": "https://github.com/mitodl/mitxonline-api-clients/raw/refs/heads/dansubak/202606_paginated_codes_means_paginated_problems-client-7dbc25d/src/typescript/mitxonline-api-axios/package.tgz",
     "@tanstack/react-query": "^5.66.0",
     "axios": "^1.12.2",
     "tiny-invariant": "^1.3.3"
--- a/frontends/api/package.json
+++ b/frontends/api/package.json
@@ -16,7 +16,6 @@
     "./test-utils/mockAxios": "./src/test-utils/mockAxios.ts",
     "./test-utils": "./src/test-utils/index.ts",
     "./mitxonline-hooks/*": "./src/mitxonline/hooks/*/index.ts",
-    "./mitxonline-clients": "./src/mitxonline/clients.ts",
     "./mitxonline-test-utils": "./src/mitxonline/test-utils/index.ts"
   },
   "peerDependencies": {
--- a/frontends/api/src/mitxonline/hooks/organizations/queries.ts
+++ b/frontends/api/src/mitxonline/hooks/organizations/queries.ts
@@ -4,6 +4,8 @@ import {
   OrganizationPage,
   ManagerContractDetail,
   ManagerEnrollmentCode,
+  PaginatedManagerEnrollmentCodeList,
+  PaginatedOrganizationPageList,
   B2bApiB2bOrganizationsRetrieveRequest,
   B2bApiB2bManagerOrganizationsContractsRetrieveRequest,
   B2bApiB2bManagerOrganizationsContractsCodesListRequest,
@@ -11,6 +13,13 @@ import {
 
 type ContractCode = ManagerEnrollmentCode & {
   redemption_status: "unassigned" | "assigned" | "redeemed"
+}
+
+type PaginatedContractCodes = Omit<
+  PaginatedManagerEnrollmentCodeList,
+  "results"
+> & {
+  results: ContractCode[]
 }
 
 const organizationKeys = {
@@ -47,9 +56,20 @@ const managerOrganizationKeys = {
     ] as const,
   contractCodesRoot: () =>
     ["mitxonline", "manager", "organizations", "contracts", "codes"] as const,
+  // Prefix key used for invalidation — matches all pages/search states for one contract
+  contractCodesForContract: (id: number, orgId: number) =>
+    [...managerOrganizationKeys.contractCodesRoot(), id, orgId] as const,
   contractCodes: (
     opts: B2bApiB2bManagerOrganizationsContractsCodesListRequest,
-  ) => [...managerOrganizationKeys.contractCodesRoot(), opts] as const,
+  ) =>
+    [
+      ...managerOrganizationKeys.contractCodesForContract(
+        opts.id,
+        opts.parent_lookup_organization,
+      ),
+      opts.page,
+      opts.search_term,
+    ] as const,
 }
 
 const managerOrganizationQueries = {
@@ -57,7 +77,12 @@ const managerOrganizationQueries = {
     queryOptions({
       queryKey: managerOrganizationKeys.list(),
       queryFn: async (): Promise<OrganizationPage[]> =>
-        b2bApi.b2bManagerOrganizationsList().then((res) => res.data),
+        b2bApi.b2bManagerOrganizationsList().then((res) => {
+          const data = res.data as
+            | PaginatedOrganizationPageList
+            | OrganizationPage[]
+          return Array.isArray(data) ? data : data.results
+        }),
     }),
   managerContractDetail: (
     opts: B2bApiB2bManagerOrganizationsContractsRetrieveRequest,
@@ -74,10 +99,10 @@ const managerOrganizationQueries = {
   ) =>
     queryOptions({
       queryKey: managerOrganizationKeys.contractCodes(opts),
-      queryFn: async (): Promise<ContractCode[]> =>
+      queryFn: async (): Promise<PaginatedContractCodes> =>
         b2bApi
           .b2bManagerOrganizationsContractsCodesList(opts)
-          .then((res) => res.data as ContractCode[]),
+          .then((res) => res.data as PaginatedContractCodes),
     }),
 }
 
@@ -88,4 +113,4 @@ export {
   managerOrganizationKeys,
 }
 
-export type { ContractCode }
+export type { ContractCode, PaginatedContractCodes }
--- a/frontends/api/src/mitxonline/hooks/organizations/queries.ts
+++ b/frontends/api/src/mitxonline/hooks/organizations/queries.ts
@@ -69,6 +69,7 @@ const managerOrganizationKeys = {
       ),
       opts.page,
       opts.search_term,
+      opts.status,
     ] as const,
 }
 
--- a/frontends/api/src/mitxonline/hooks/organizations/queries.ts
+++ b/frontends/api/src/mitxonline/hooks/organizations/queries.ts
@@ -56,22 +56,15 @@ const managerOrganizationKeys = {
     ] as const,
   contractCodesRoot: () =>
     ["mitxonline", "manager", "organizations", "contracts", "codes"] as const,
-  // Prefix key used for invalidation — matches all pages/search states for one contract
-  contractCodesForContract: (id: number, orgId: number) =>
-    [...managerOrganizationKeys.contractCodesRoot(), id, orgId] as const,
   contractCodes: (
     opts: B2bApiB2bManagerOrganizationsContractsCodesListRequest,
-  ) =>
-    [
-      ...managerOrganizationKeys.contractCodesForContract(
-        opts.id,
-        opts.parent_lookup_organization,
-      ),
-      opts.page,
-      opts.page_size,
-      opts.search_term,
-      opts.status,
-    ] as const,
+  ) => [...managerOrganizationKeys.contractCodesRoot(), opts] as const,
+  // Prefix key used for invalidation — matches all pages/search states for one contract
+  contractCodesForContract: (id: number, orgId: number) =>
+    managerOrganizationKeys.contractCodes({
+      id,
+      parent_lookup_organization: orgId,
+    }),
 }
 
 const managerOrganizationQueries = {
--- a/frontends/api/src/mitxonline/hooks/organizations/queries.ts
+++ b/frontends/api/src/mitxonline/hooks/organizations/queries.ts
@@ -3,24 +3,11 @@ import { b2bApi } from "../../clients"
 import {
   OrganizationPage,
   ManagerContractDetail,
-  ManagerEnrollmentCode,
   PaginatedManagerEnrollmentCodeList,
-  PaginatedOrganizationPageList,
   B2bApiB2bOrganizationsRetrieveRequest,
   B2bApiB2bManagerOrganizationsContractsRetrieveRequest,
   B2bApiB2bManagerOrganizationsContractsCodesListRequest,
 } from "@mitodl/mitxonline-api-axios/v2"
-
-type ContractCode = ManagerEnrollmentCode & {
-  redemption_status: "unassigned" | "assigned" | "redeemed"
-}
-
-type PaginatedContractCodes = Omit<
-  PaginatedManagerEnrollmentCodeList,
-  "results"
-> & {
-  results: ContractCode[]
-}
 
 const organizationKeys = {
   root: ["mitxonline", "organizations"],
@@ -72,12 +59,7 @@ const managerOrganizationQueries = {
     queryOptions({
       queryKey: managerOrganizationKeys.list(),
       queryFn: async (): Promise<OrganizationPage[]> =>
-        b2bApi.b2bManagerOrganizationsList().then((res) => {
-          const data = res.data as
-            | PaginatedOrganizationPageList
-            | OrganizationPage[]
-          return Array.isArray(data) ? data : data.results
-        }),
+        b2bApi.b2bManagerOrganizationsList().then((res) => res.data.results),
     }),
   managerContractDetail: (
     opts: B2bApiB2bManagerOrganizationsContractsRetrieveRequest,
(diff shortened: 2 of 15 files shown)


frontends/main/package.json
```diff
@@ -16,7 +16,7 @@
     "@floating-ui/react": "^0.27.16",
     "@mitodl/arithmix": "^0.2.2",
     "@mitodl/course-search-utils": "^3.5.2",
-    "@mitodl/mitxonline-api-axios": "2026.6.17",
+    "@mitodl/mitxonline-api-axios": "https://github.com/mitodl/mitxonline-api-clients/raw/refs/heads/dansubak/202606_paginated_codes_means_paginated_problems-client-7af6a2c/src/typescript/mitxonline-api-axios/package.tgz",
```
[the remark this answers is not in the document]
daniellefrappier18: this is fine for now it will not get merged this way.

frontends/api/package.json
```diff
@@ -31,7 +32,7 @@
     "ol-test-utilities": "0.0.0"
   },
   "dependencies": {
-    "@mitodl/mitxonline-api-axios": "2026.6.17",
+    "@mitodl/mitxonline-api-axios": "https://github.com/mitodl/mitxonline-api-clients/raw/refs/heads/dansubak/202606_paginated_codes_means_paginated_problems-client-7af6a2c/src/typescript/mitxonline-api-axios/package.tgz",
```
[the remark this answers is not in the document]
daniellefrappier18: same as above

frontends/main/src/app-pages/ContractAdminPage/AssignSeatsSection.tsx
```diff
@@ -546,11 +540,10 @@ const AssignSeatsSection: React.FC<AssignSeatsSectionProps> = ({
...
         </ActiveLink>
-        <Tooltip title="Coming soon" describeChild>
-          <DisabledLink role="button" aria-disabled="true" tabIndex={0}>
-            (download sample CSV)
-          </DisabledLink>
-        </Tooltip>
+        {/* TODO: replace href with a real sample CSV once we have one */}
+        <Link href="#" color="red" rel="noopener noreferrer" target="_blank">
+          (download sample CSV)
+        </Link>
```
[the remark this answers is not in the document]
daniellefrappier18: this is fine for now, I'll fix before merging

frontends/main/src/app-pages/ContractAdminPage/RowActionMenu.tsx
```diff
@@ -137,15 +137,9 @@ const RowActionMenu: React.FC<RowActionMenuProps> = ({
...
-      const status = (err as AxiosError)?.response?.status
-      onResult(
-        status === 409
-          ? "This code has already been redeemed and cannot be revoked."
-          : "Could not complete the action. Please try again.",
-        "error",
-      )
+      onResult("Seat released.", "success")
+    } catch {
+      onResult("Could not release the seat. Please try again.", "error")
```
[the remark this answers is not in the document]
daniellefrappier18: fixed

frontends/main/src/app-pages/ContractAdminPage/ContractAdminPage.tsx
```diff
@@ -461,14 +498,51 @@ const ContractAdminPageInternal: React.FC<ContractAdminPageInternalProps> = ({
...
-      setSearchAnnouncement(`${count} result${count !== 1 ? "s" : ""}`)
-    }, 0)
+  const handleExportCsv = async () => {
+    if (!totalPurchased) return
+    try {
+      const response = await b2bApi.b2bManagerOrganizationsContractsCodesList({
+        id: contract.id,
+        parent_lookup_organization: org.id,
+        page_size: totalPurchased,
+      })
```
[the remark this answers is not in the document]
daniellefrappier18: added tests

frontends/main/src/app-pages/ContractAdminPage/ContractAdminPage.tsx
```diff
@@ -461,14 +498,51 @@ const ContractAdminPageInternal: React.FC<ContractAdminPageInternalProps> = ({
...
+      const rows = response.data.results
```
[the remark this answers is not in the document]
daniellefrappier18: fixed

frontends/main/src/app-pages/ContractAdminPage/ContractAdminPage.tsx
```diff
@@ -379,10 +409,35 @@ const ContractAdminPageInternal: React.FC<ContractAdminPageInternalProps> = ({
...
       id: contract?.id ?? 0,
       parent_lookup_organization: org?.id ?? 0,
+      page,
+      search_term: debouncedSearchQuery || undefined,
+      status:
+        statusFilter === "redeemed"
+          ? "redeemed"
+          : statusFilter === "pending"
+            ? "assigned"
+            : undefined,
```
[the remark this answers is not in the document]
daniellefrappier18: fixed description

frontends/main/src/app-pages/ContractAdminPage/AssignSeatsSection.tsx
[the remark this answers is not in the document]
daniellefrappier18: this is fine for now I will replace this before merging

frontends/api/src/mitxonline/hooks/organizations/queries.ts
```diff
@@ -47,17 +56,34 @@ const managerOrganizationKeys = {
...
+  ) =>
+    [
+      ...managerOrganizationKeys.contractCodesForContract(
+        opts.id,
+        opts.parent_lookup_organization,
+      ),
+      opts.page,
+      opts.search_term,
+      opts.status,
+    ] as const,
```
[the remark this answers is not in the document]
daniellefrappier18: this is a non-issue. The only call to managerContractCodes in the app never passes page_size and the CSV export calls b2bApi directly, bypassing React Query's cache entirely. But it's one line so I'll add it anyways.

frontends/main/src/app-pages/ContractAdminPage/ContractAdminPage.tsx
```diff
@@ -461,14 +493,71 @@ const ContractAdminPageInternal: React.FC<ContractAdminPageInternalProps> = ({
...
+        "Status",
+        "Assigned on",
+        "Redeemed on",
+        "Last sent",
+      ])
+      const dataRows = rows.map((c) =>
+        buildCsvRow([
+          c.assigned_to,
+          c.redeemed_by,
+          c.redemption_status === "redeemed" ? "Redeemed" : "Pending claim",
```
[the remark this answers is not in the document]
daniellefrappier18: fixed

frontends/main/src/app-pages/ContractAdminPage/ContractAdminPage.tsx
```diff
@@ -415,44 +470,21 @@ const ContractAdminPageInternal: React.FC<ContractAdminPageInternalProps> = ({
...
-    page * PAGE_SIZE,
-  )
+  const totalCount = codes?.count ?? 0
+  // Derive total pages from the response: if there's a next page we're on a
+  // full page so results.length equals the backend page size; otherwise this
+  // is the last page so totalPages equals the current page number.
+  const totalPages =
+    codes?.next && pageResults.length > 0
+      ? Math.ceil(totalCount / pageResults.length)
+      : page
```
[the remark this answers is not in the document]
daniellefrappier18: The current logic is correct as-is.

frontends/main/src/app-pages/ContractAdminPage/ContractAdminPage.tsx
```diff
@@ -370,6 +390,16 @@ const ContractAdminPageInternal: React.FC<ContractAdminPageInternalProps> = ({
...
   const contract = org?.contracts.find((c) => c.slug === contractSlug)
 
+  const { data: contractDetail, isLoading: isLoadingContractDetail } = useQuery(
+    {
+      ...managerOrganizationQueries.managerContractDetail({
+        id: contract?.id ?? 0,
+        parent_lookup_organization: org?.id ?? 0,
+      }),
+      enabled: !!org && !!contract,
+    },
```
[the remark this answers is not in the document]
daniellefrappier18: The contractDetail query now has the same error handling pattern as the other two queries — 401/403 shows "Access denied", anything else shows the generic error page.

frontends/main/src/app-pages/ContractAdminPage/AssignSeatsSection.test.tsx
```diff
@@ -68,22 +73,31 @@ describe("AssignSeatsSection", () => {
...
+        orgId={1}
+        contractId={2}
+        availableSeats={50}
+        isLoadingSeats={false}
+      />,
     )
 
-    // Only the download link is still disabled — import from CSV is now active
-    const downloadLink = screen.getByRole("button", {
+    // The download link is a real anchor — in the tab order and not disabled
```
ChristopherChudzicki: The old comment was worthwhile. The new comment is not, IMO. (Of course it's a real anchor tag. Why wouldn't it be?) 

I find LLMs somewhat often write comments that only make sense if you know the history of the file, and otherwise look kind of weird.

frontends/main/src/app-pages/ContractAdminPage/ContractAdminPage.tsx
```diff
@@ -415,44 +481,21 @@ const ContractAdminPageInternal: React.FC<ContractAdminPageInternalProps> = ({
...
-    page * PAGE_SIZE,
-  )
+  const totalCount = codes?.count ?? 0
+  // Derive total pages from the response: if there's a next page we're on a
+  // full page so results.length equals the backend page size; otherwise this
+  // is the last page so totalPages equals the current page number.
+  const totalPages =
+    codes?.next && pageResults.length > 0
+      ? Math.ceil(totalCount / pageResults.length)
+      : page
```
ChristopherChudzicki: There's a tiny bug here, pagination display flickers:

https://github.com/user-attachments/assets/549c0aca-251b-4216-8ead-5baea1768f57

reason is that `codes.next` and `codes.count`  is undefined while the results are loading, so it falls back to `page`.

**Suggestion:** Do two things:
1. Pass an explicit `page_size` to the query. Then calculating total pages is very simple, just `Math.ceil(codes.count / page_size)` , and doesn't depend on `next` at all (which could be null).
    - Bonus: This means the frontend controls the page size. Since that really is a UI concern.
2. Add `placeholderData: keepPreviousData` to the query. This means that while fetching new data, the query keeps the old one, and `codes.next + codes.count` won't be undefined.
    - See https://tanstack.com/query/v5/docs/framework/react/guides/paginated-queries#better-paginated-queries-with-placeholderdata

You could do (2) alone. That would 95% fix the issue above, though you'd still get flicker when moving from the last page to a previous page (if the previous page wasn't loaded yet).

frontends/api/src/mitxonline/hooks/organizations/queries.ts
```diff
@@ -47,17 +56,35 @@ const managerOrganizationKeys = {
...
 const managerOrganizationQueries = {
   managerOrganizationsList: () =>
     queryOptions({
       queryKey: managerOrganizationKeys.list(),
       queryFn: async (): Promise<OrganizationPage[]> =>
-        b2bApi.b2bManagerOrganizationsList().then((res) => res.data),
+        b2bApi.b2bManagerOrganizationsList().then((res) => {
+          const data = res.data as
+            | PaginatedOrganizationPageList
+            | OrganizationPage[]
```
ChristopherChudzicki: Do we need this? It's always paginated once @dsubak PR goes out. This works with both the paginated and non-paginated versions, but the rest of the PR only works paginated, right?

frontends/api/src/mitxonline/hooks/organizations/queries.ts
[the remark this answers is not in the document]
ChristopherChudzicki: @daniellefrappier18 I'd recommend against this change overall. Keep the original version:
```ts
 contractCodes: (
    opts: B2bApiB2bManagerOrganizationsContractsCodesListRequest,
  ) => [...managerOrganizationKeys.contractCodesRoot(), opts] as const
```
If you also want to keep `contractCodesForContract`, make it:
```ts
contractCodesForContract: (id: number, orgId: number) => contractCodes({  id, parent_lookup_id: orgId })
```

React query's key matching (e.g., for invalidation) matches objects partially: `["a", { id: 2, parent_lookup_organization: 3 } ]` will match `["a", { id: 2, parent_lookup_organization: 3, page_size: whatever, filter: another-thing } ]`.

**AND:** The new form has the disadvantage that if the backend adds parameters, we could easily forget to add the parameters on the frontend.

-----

> ...The only call to managerContractCodes in the app never passes page_size ... 

But I think it should 😄

frontends/main/src/app-pages/ContractAdminPage/ContractAdminPage.tsx
```diff
@@ -461,14 +504,73 @@ const ContractAdminPageInternal: React.FC<ContractAdminPageInternalProps> = ({
...
     setSearchQuery(e.target.value)
-    setPage(1)
   }
 
-  const announceSearchCount = (count: number) => {
-    setSearchAnnouncement("")
-    setTimeout(() => {
-      setSearchAnnouncement(`${count} result${count !== 1 ? "s" : ""}`)
-    }, 0)
+  const handleExportCsv = async () => {
```
ChristopherChudzicki, line 515: Two comments:
1. I had thought the plan was to build the CSV on the backend. Is that still the long term plan, or did we change that? IMO, backend generation is still the better long-term plan:
    - with the backend implementation, I'm imagining a single link `<a href="/mitxonline/api/v0/b2b/manager/organizations/{org_id}/contracts/{contract_id}/codes/export.csv" download> Export CSV </a>` ... very simple on the frontend side, and Python has good builtin tools for building the CSV
    - We discussed an org with several thousand members...9k was the number I remember. This would be 18 serial frontend queries. That's not fast, plus any one of them could fail for transient network issues.
2. If you leave it here, I would lean toward doing `queryClient.fetchQuery(managerOrganizationQueries.managerContractCodes(...params))` rather than using `b2bApi` client directly.
    - This is the only direct use of an axios API client outside the react-query layer in the entire codebase, so it's a bit of a pattern change
    - going through react query has some advantages, like retrying with exponential backoff (up to max 3 retries) if a query fails due to transient network errors

@dsubak what would you think of shifting the CSV generation to the backend? [Alternatively, I think I missed a meeting on the admin dashboard, so if you two decided to put it here, OK!]

frontends/api/src/mitxonline/hooks/organizations/queries.ts
ChristopherChudzicki: Mentioning: The contract code override type here won't be necessary if we merge https://github.com/mitodl/mitxonline/pull/3708

_Nothing to change here yet, since it would need a new client._

ChristopherChudzicki (changes requested):
@daniellefrappier18 this worked well. I left some small suggestions—I think they are all worth doing [with caveat on backend CSV generation...could be followup, and maybe you and Dan discussed it already.]

frontends/main/src/app-pages/ContractAdminPage/ContractAdminPage.tsx
dsubak: @ChristopherChudzicki @daniellefrappier18 

Danielle and I discussed it in person - I'm happy to give a bespoke export CSV projection if we'd like but I think it'd end up being pretty close to syntactic sugar over similar functionality to `/codes` w/ filtering and an arbitrarily large page_size.

1) I believe we are assuming that export would want to respect filters. It might be weird if you had it filtered to a specific status and search term and you hit export and got a huge superset of records in your export. Please correct me if this is not the case.
2) On the backend I deliberately _don't_ enforce a maximum page size in the pagination configuration to allow for pulling every row in one request, so my thinking was that we could specify a page size == max_learners (as returned by the summary stats in the `/contract/{id}` request we use elsewhere on the page).

With those in mind, we were basically like "It's essentially just /codes with a different serialization format" when you look at the inputs, outputs and behaviors needed. I do think we ought to strike the pagination from this function - as far as I understand it there shouldn't be the need for more than one request - but otherwise I don't think a bespoke endpoint for csv export is hugely valuable at this time.

That all having been said, if either 1 or 2 does not hold anymore, that'd push me towards a bespoke endpoint. Absolutely let me know if my assumptions are now invalid!

frontends/api/src/mitxonline/hooks/organizations/queries.ts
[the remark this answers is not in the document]
daniellefrappier18: done

frontends/main/src/app-pages/ContractAdminPage/ContractAdminPage.tsx
dsubak: (Also, ALL of this assumes an export fits nicely in memory in a single web request - if we needed to do the 10 ton solve of an async job to export stuff to s3 and allow ya'll to retrieve it by subsequent API request, whole thing would need to look different. I think this is understood by everyone at this point, but wanted to be explicit)

frontends/main/src/app-pages/ContractAdminPage/ContractAdminPage.tsx
daniellefrappier18: I implemented both suggestions.

frontends/api/src/mitxonline/hooks/organizations/queries.ts
daniellefrappier18: Once your PR merges and we cut a new client, I'll drop the union.

frontends/api/src/mitxonline/hooks/organizations/queries.ts
dsubak: Just as an FYI, that PR is now out in prod!

frontends/main/src/app-pages/ContractAdminPage/ContractAdminPage.tsx
```diff
@@ -461,14 +510,71 @@ const ContractAdminPageInternal: React.FC<ContractAdminPageInternalProps> = ({
...
+    if (!totalPurchased) return
+    setIsExporting(true)
+    try {
+      const allRows: ContractCode[] = []
+      let page = 1
+      let hasMore = true
+      while (hasMore) {
+        const data = await queryClient.fetchQuery(
+          managerOrganizationQueries.managerContractCodes({
```
dsubak, line 524: This is referenced in the above discussion, but didn't want it to get lost in the shuffle - do we want to respect any applied status or search term filters for the export?

frontends/main/public/sample-seat-assignments.csv
```diff
@@ -0,0 +1,6 @@
+id,name,email
```
dsubak: Should this have an ID column? IIRC, bulk_assign only needs an email and optionally a name per assignment.

Course, if the frontend needs it for something, feel free to ignore me!

dsubak (commented):
Looks pretty good to me - I was able to test everything listed and didn't find any issues.

I think we ought to sync up and make sure we're on the same page w/r/t CSV export (per Chris' comment thread), but once we've chased that to ground, I'd be happy to run this back through its paces and provide an approval!

frontends/main/src/app-pages/ContractAdminPage/ContractAdminPage.tsx
daniellefrappier18: Oh that's a good question. I personally would expect the export to just them export all

frontends/main/public/sample-seat-assignments.csv
daniellefrappier18: Good catch I updated the template to just email.

frontends/main/src/app-pages/ContractAdminPage/ContractAdminPage.tsx
dsubak: Discussed offline. Key takeaways are as follows!
- For now, the exported CSV will provide an unfiltered output. We might revisit this in the future, so any backend APIs we stand up should consider retaining filtering, but for now we're just gonna hand em everything.
- We're going to keep CSV export on the frontend for now. I'll file a ticket to stand up a tiny endpoint to move it to the backend; that endpoint should function really similarly to /codes (potentially including filtering), but we'll be able to drop the page_size arg and can use python csv builtins for actual payload construction. We can adopt that at any point in the future without introducing a breaking change.

frontends/main/src/app-pages/ContractAdminPage/ContractAdminPage.tsx
dsubak: We spoke w/ Peter/Linda/Steve and we're gonna go w/ unfiltered export for now!

frontends/main/src/app-pages/ContractAdminPage/ContractAdminPage.tsx
daniellefrappier18: @ChristopherChudzicki @dsubak As discussed for now we are going to keep the CSV export on the frontend with the plan to move the CSV generation to the backend in the future. I have swapped using the b2bApi client directly in favor of queryClient.fetchQuery

frontends/main/src/app-pages/ContractAdminPage/ContractAdminPage.tsx
daniellefrappier18: Just documenting the decision with @pdpinch @steven-hatch @skurchak that for now we are going to NOT filter the export.

frontends/api/package.json
```diff
@@ -16,6 +16,7 @@
     "./test-utils/mockAxios": "./src/test-utils/mockAxios.ts",
     "./test-utils": "./src/test-utils/index.ts",
     "./mitxonline-hooks/*": "./src/mitxonline/hooks/*/index.ts",
+    "./mitxonline-clients": "./src/mitxonline/clients.ts",
```
ChristopherChudzicki: **Request:** We don't need this export anymore, right? Can you remove it?

ChristopherChudzicki (approved):
Working well! Thanks for changing the query client 

One tiny request... https://github.com/mitodl/mit-learn/pull/3530/changes#r3499608315 IMO, no need for a re-review after client bump + that change.

frontends/api/package.json
daniellefrappier18: done

frontends/api/src/mitxonline/hooks/organizations/queries.ts
daniellefrappier18: Done

frontends/api/src/mitxonline/hooks/organizations/queries.ts
daniellefrappier18: done

PR #3539, opened 2026-06-29
Log missing/invalid edx_module_id in content-file API requests

mbertrand:
Closes mitodl/hq#11873

Logs an error  whenever an API request references an `edx_module_id` that has no backing content, so we get a
signal when AskTIM and other consumers are degraded by missing content.

For every requested id, existence is probed **directly against the source of
truth**, never inferred from empty search results. Two reasons are
distinguished:

- `not_in_db` — no `ContentFile` row (ETL/scrape gap)
- `not_in_index` — row exists but isn't embedded in Qdrant (embedding gap)

### How can this be tested?
- Pick a real `edx_module_id` from your local data, and also deindex a 

  ```python
  from vector_search.tasks import *
  
  real_cf = ContentFile.objects.exclude(edx_module_id='').first()
  unindexed_cf = ContentFile.objects.exclude(edx_module_id='').last()
  generate_embeddings([real_cf.run.id], "content_file", False)
  real_edx_module_id = print(real_cf.edx_module_id)
  print(real_edx_module_id )
  
  remove_embeddings([unindexed_cf.id], "content_file")
  unindexed_edx_module_id = unindexed_cf.edx_module_id
  print(unindexed_edx_module_id)  
  
  ```

- Log in as an admin user and try the following URLs.  The edx_module_id values will need to be url-encoded.

  **REST endpoint** — existing id is silent, bogus id logs `not_in_db`:

  `http://open.odl.local:8065/api/v1/contentfiles/?edx_module_id=<REAL_ID>`

   non-existent -> empty results AND an error log
  `http://open.odl.local:8065/api/v1/contentfiles/?edx_module_id=does-not-exist`

  **Vector endpoint** — only probes when there are no hits:

  non-existent id, no hits -> logs not_in_db
  `http://open.odl.local:8065/api/v0/vector_content_files_search/?edx_module_id=does-not-exist`

  real id that is in the DB but absent from Qdrant -> logs not_in_index
  `http://open.odl.local:8065/api/v0/vector_content_files_search/?edx_module_id=<REAL_UNINDEXED_ID>`

Watch the web container logs for lines like:

```
Missing ContentFile (not_in_db) for edx_module_id=does-not-exist [source=contentfiles_api]
```


shanbady (approved):
approving since it does what it needs to.

I did leave a small comment about performance when using exact=True - however since the edx_module_id is an indexed payload i think this will be minimal.

I also tried manually running this count with exact=True on prod and it was pretty fast

PR #3543, opened 2026-06-30
Mitxonline ETL: switch to consuming internal courses endpoint

shanbady:

Closes https://github.com/mitodl/hq/issues/11671

This PR allows us to be able to use the internal mitxonline courses api which includes variant and b2b courses

### How can this be tested?
1. checkout this branch
2. ssh into an rc mitxonline pod and get into a django shell
3. once in the shell execute the following to create an api key:
```python
from rest_framework_api_key.models import *
api_key, key = APIKey.objects.create_key(name="test-learn-etl-service")
print(key)
```
4. the key will only displayed/available once (its stored as a hash). note the api key down
5. in settings_course_etl.py set MITX_ONLINE_COURSES_API_URL = "https://api.rc.mitxonline.mit.edu/api/internal/courses/" and MITX_ONLINE_ETL_API_KEY to the api key you have just created.
6. restart celery and run the mitxonline backpopulate (optionally delete existing mitxionline courses) `./manage.py backpopulate_mitxonline_data` 
7. verify that mitxonline learning resources populate.
8. once done testing delete the test api key created on mitonline:
```python
from rest_framework_api_key.models import *
APIKey.objects.get(name="test-learn-etl-service").delete()
```

### Additional Context
1. This Pr also contains a small change to coerce the protocol used when fetching data to whatever the initial request was - this is because the mitonline api for some reason uses http for next urls.
2. when using mitxonline rc as the endpoint - you will end up with the same amount of resources as using the regular api since I dont think there are variant courses there.

merging this will be blocked on getting the actual api key secrets onto mitlearn rc/prod via https://github.com/mitodl/hq/issues/12088


learning_resources/etl/mitxonline.py
```diff
@@ -43,15 +43,21 @@
...
+    headers = {}
+    if settings.MITX_ONLINE_ETL_API_KEY:
+        headers["Authorization"] = f"Api-Key {settings.MITX_ONLINE_ETL_API_KEY}"
     while url:
+        # Preserve the scheme of the current request: the upstream API can
+        # return an http "next" URL even when we requested over https.
+        scheme = urlparse(url).scheme
         response = requests.get(
-            url, params=params, timeout=settings.REQUESTS_TIMEOUT
+            url, params=params, headers=headers, timeout=settings.REQUESTS_TIMEOUT
```
mbertrand: `_fetch_courses_by_ids` ([`mitxonline.py:583-597`](https://github.com/mitodl/mit-learn/blob/6b7cedbd9d223f71fa6ec1927daee5ea729aa988/learning_resources/etl/mitxonline.py#L583-L597)), called by the programs part of the ETL, passes `params={"id": ...}` which doesn't work with the new API.  So it ends up paginating the entire 219-course catalog and filters client-side — and it's called once per program in `transform_programs` ([`:616-622`](https://github.com/mitodl/mit-learn/blob/6b7cedbd9d223f71fa6ec1927daee5ea729aa988/learning_resources/etl/mitxonline.py#L616-L622)). That's roughly a 19× request blowup vs the v2 id-filtered path.

Fix options:
- Fetch the full catalog once and reuse it across programs (`extract_courses` already pulls the whole thing), or
- Use the detail route for each id: `GET /api/internal/courses/<id>/` works by integer PK (verified 200), which matches the `req_tree` course ids from `get_course_ids_from_req_tree`.

learning_resources/etl/mitxonline.py
```diff
@@ -43,15 +43,21 @@
 def _fetch_data(url, params=None):
     if not params:
         params = {}
+    headers = {}
+    if settings.MITX_ONLINE_ETL_API_KEY:
+        headers["Authorization"] = f"Api-Key {settings.MITX_ONLINE_ETL_API_KEY}"
```
mbertrand, line 46: The API returns **all published runs** per course — including B2B and variant runs that v2 filtered out. `_transform_course` ([`:352`](https://github.com/mitodl/mit-learn/blob/6b7cedbd9d223f71fa6ec1927daee5ea729aa988/learning_resources/etl/mitxonline.py#L352)) keeps all of them and publishes any run with `is_enrollable AND page.live`, so they surface in the catalog. (`include_in_learn_catalog` gates the course, not its runs.)

With RC data, published courses expose published B2B/variant published runs in API responses.   

Also, these runs can impact the assignment of `best_run` ([`models.py:581-634`](https://github.com/mitodl/mit-learn/blob/6b7cedbd9d223f71fa6ec1927daee5ea729aa988/learning_resources/models.py#L581-L634)), which picks purely by enrollment/start date with no B2B/variant awareness —  some of these runs now resolve to a B2B run as their `best_run` , which drives the displayed dates/price/URL and gates opensearch content-file indexing, so the wrong run could be shown and indexed in opensearch.

Example: http://api.open.odl.local:8065/api/v1/learning_resources/?readable_id=course-v1%3AMITxT%2BSTL.162x

Maybe we need a new field on `LearningResourceRun` to differentiate B2B/variant runs and exclude them from API results and calculation of best_run?

main/settings_course_etl.py
```diff
@@ -56,6 +56,9 @@
 MITX_ONLINE_BASE_URL = get_string("MITX_ONLINE_BASE_URL", None)
 MITX_ONLINE_PROGRAMS_API_URL = get_string("MITX_ONLINE_PROGRAMS_API_URL", None)
 MITX_ONLINE_COURSES_API_URL = get_string("MITX_ONLINE_COURSES_API_URL", None)
+# API key for the MITx Online internal-only secured endpoint. When set, it is
+# sent as an "Authorization: Api-Key <key>" header on MITx Online ETL requests.
+MITX_ONLINE_ETL_API_KEY = get_string("MITX_ONLINE_ETL_API_KEY", None)
```
mbertrand, line 61: [`README.md:300-301`](https://github.com/mitodl/mit-learn/blob/6b7cedbd9d223f71fa6ec1927daee5ea729aa988/README.md#L300-L301) and [`env/codespaces.env:29-30`](https://github.com/mitodl/mit-learn/blob/6b7cedbd9d223f71fa6ec1927daee5ea729aa988/env/codespaces.env#L29-L30) still point at `/api/v2/courses/` — maybe worth updating to the internal endpoint (and mentioning that `MITX_ONLINE_ETL_API_KEY` is now required for that ETL).

learning_resources/etl/mitxonline.py
```diff
@@ -43,15 +43,21 @@
...
     while url:
+        # Preserve the scheme of the current request: the upstream API can
+        # return an http "next" URL even when we requested over https.
+        scheme = urlparse(url).scheme
```
mbertrand: Nit: `scheme` is invariant across loop iterations — each rewritten `url` already carries the preserved scheme — so it could be captured once before the `while`. Not a bug, just slightly clearer.

mbertrand (changes requested):

learning_resources/etl/mitxonline.py
mbertrand: Not sure if we want these b2b/variant runs in the API results or not, but pretty sure they should not be assigned as the `best_run` for published courses.

mbertrand: Maybe can just set these runs as always unpublished, so they get their contentfiles indexed but that's it (assuming they don't have to show up anywhere else in mit-learn).

main/settings_course_etl.py
shanbady: I was intending this to be backwards compatable (local setups default to the public endpoint ~ less configuration for default environments)

learning_resources/etl/mitxonline.py
shanbady: that is a good point..let me see if i can filter those out some how

learning_resources/etl/mitxonline.py
shanbady: good catch. switched to just using the detail route `f"{settings.MITX_ONLINE_COURSES_API_URL}/{course_id}"` which is supported by both apis

learning_resources/etl/mitxonline.py
shanbady: I added a [is_b2b](https://github.com/mitodl/mit-learn/pull/3543/changes/d8ec5b35273962394c4f2fd653ae4aef8fb16362#diff-c143bb17b4eb35552c87ce587a727b382ef1921c758d4a5437c11d026ae53d16R317) attribute for runs which gets set based on b2b_contract from the api

learning_resources/etl/mitxonline.py
shanbady: fixed

learning_resources/etl/mitxonline.py
```diff
@@ -43,15 +43,21 @@
...
+            url, params=params, headers=headers, timeout=settings.REQUESTS_TIMEOUT
         ).json()
         results = response["results"]
         yield from results
         next_url = response.get("next")
         if next_url:
-            parsed = urlparse(next_url)
+            parsed = urlparse(next_url)._replace(scheme=scheme)
             url = f"{parsed.scheme}://{parsed.netloc}{parsed.path}"
             params = parse_qs(parsed.query)
```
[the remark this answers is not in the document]
shanbady: fixed + test

learning_resources/etl/mitxonline.py
```diff
@@ -578,17 +603,35 @@ def _fetch_courses_by_ids(course_ids):
...
-                params={
-                    "id": ",".join([str(courseid) for courseid in course_ids]),
-                    "live": True,
-                },
+        base_url = settings.MITX_ONLINE_COURSES_API_URL.rstrip("/")
+        headers = _mitxonline_api_headers()
+        courses = []
+        for course_id in course_ids:
+            response = requests.get(
+                f"{base_url}/{course_id}",
```
mbertrand: Maybe add a trailing slash to avoid a 301 redirect

mbertrand (approved):
LGTM, one question I have is if this will handle variant courses (other languages?) that are not B2B, or do those not exist?

Other than that, 1 minor nit about a trailing slash for individual course endpoint urls

learning_resources/etl/mitxonline.py
shanbady: added trailing slash as well as checks for non b2b translations/variants via https://github.com/mitodl/mit-learn/pull/3543/commits/f24b4260c92cb367b67a69742cb0b4b004b62cf0

PR #3546, opened 2026-06-30
fix(vector-search): align qdrant-client version with server

blarghmatey:
## Summary

- Production pods log repeated `DecodeError("Error parsing message with type 'qdrant.QueryResponse'")` on the Similar-Resources endpoint, surfaced as `Similar resources RPC error for resource NNN: Exception deserializing response!`
- Root cause: qdrant-client was pinned with `==1.18.0` (exact), which prevents automatic pickup of any future 1.18.x patch releases from the qdrant team
- Deployed Qdrant server: **v1.18.2** (set in `ol-infrastructure/src/bridge/lib/versions.py` as `QDRANT_VERSION`)
- qdrant-client **1.18.0** is the latest release on PyPI; there is no 1.18.1/1.18.2 client yet
- The proto schema for `QueryResponse` is identical between server 1.18.0 and 1.18.2 (diff shows only a comment change), so the error is likely triggered by the `cloud_inference=True + prefer_grpc=True` combination introduced via `QdrantCloudEncoder` in production

## Change

Replace `==1.18.0,<2` with `~=1.18.0` (compatible-release specifier), which:
- Keeps the resolved version at 1.18.0 today (only available release)
- Automatically picks up 1.18.x patch releases when the qdrant team publishes a compatible client
- Documents the server version and Query API minimum requirement in a comment

## What else may be needed

If the qdrant team does not publish a patch client before this is needed urgently, two interim mitigations exist:

1. **Downgrade the server** in `ol-infrastructure` `QDRANT_VERSION` from `v1.18.2` back to `v1.18.0` until a compatible client is released. (Note: v1.18.2 includes security fixes — see https://github.com/qdrant/qdrant/releases/tag/v1.18.2 — so this tradeoff should be weighed carefully.)
2. **Switch to HTTP/REST transport** in `vector_search/utils.py` by setting `prefer_grpc=False` on both `qdrant_client()` and `async_qdrant_client()`. This avoids the gRPC `DecodeError` entirely at the cost of marginally higher latency.

## Test plan

- [ ] Confirm CI passes (unit tests, ruff, pre-commit)
- [ ] Verify `uv lock` still resolves `qdrant-client==1.18.0` (no version change today)
- [ ] Monitor Similar-Resources endpoint error rate in production after any future qdrant-client patch release is picked up

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -88,7 +88,14 @@ dependencies = [
     "python-dateutil>=2.8.2,<3",
     "python-rapidjson>=1.8,<2",
     "pyyaml>=6.0.0,<7",
-    "qdrant-client[fastembed]==1.18.0,<2",
+    # Minimum server version for the Query API (QueryResponse gRPC type): qdrant v1.7.0.
+    # qdrant-client ~=1.18.0 targets qdrant server v1.18.x; deployed server is v1.18.2
+    # (ol-infrastructure/src/bridge/lib/versions.py QDRANT_VERSION).
+    # The prior ==1.18.0 pin blocked automatic pickup of future 1.18.x patch releases;
+    # switching to ~= lets uv upgrade within 1.18.x when the qdrant team publishes a
+    # client that resolves the gRPC QueryResponse DecodeError observed against server
+    # v1.18.2 with cloud_inference=True + prefer_grpc=True.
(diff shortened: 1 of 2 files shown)


shanbady (approved):
👍

PR #3547, opened 2026-06-30
fix(scim): per-user locking and async OpenSearch indexing

blarghmatey:
## Summary

- **Per-user SELECT FOR UPDATE lock** in `LearnUserAdapter.save()`: serializes concurrent SCIM PATCH requests for the same user across pods, eliminating the HTTP 409 conflicts seen during the 2026-06-30 alert storm where 5-8 pods PATCHed the same user IDs simultaneously.
- **Async OpenSearch indexing via Celery**: replaces any synchronous OpenSearch I/O during the SCIM request with a `reindex_user_learning_paths` Celery task dispatched via `transaction.on_commit()`, so the SCIM HTTP response is not blocked waiting for OpenSearch I/O.

### Fix 1 — Per-user row lock (`users/adapters.py`)

When `LearnUserAdapter.save()` is called for an existing user it now executes:

```python
with transaction.atomic():
    User.objects.select_for_update().get(pk=self.obj.pk)
    super().save()
```

`SELECT FOR UPDATE` acquires an exclusive row-level lock on the user record. Concurrent PATCH requests targeting the same user (on any pod) will block at this line until the first writer's outermost transaction commits, then proceed serially. New-user creation is unaffected (no lock needed for `INSERT`).

### Fix 2 — Async search indexing (`users/tasks.py`)

A new Celery task `reindex_user_learning_paths(user_id)` re-indexes all `LearningPath` resources authored by the updated user. The task is dispatched via `transaction.on_commit()` so it only runs after the database write is fully committed and the Celery worker always sees the up-to-date state.

```python
transaction.on_commit(lambda: reindex_user_learning_paths.delay(user_pk))
```

This prevents async executor thread exhaustion from synchronous OpenSearch I/O during the HTTP request cycle.

## Test plan

- [ ] `test_scim_patch_acquires_select_for_update_lock` — mocks `User.objects.select_for_update()` and asserts it is called with the correct PK during `LearnUserAdapter.save()` for an existing user.
- [ ] `test_scim_patch_dispatches_reindex_task_asynchronously` — patches `transaction.on_commit` to fire immediately and asserts `reindex_user_learning_paths.delay()` is called with the user's PK, not synchronously inline.
- [ ] Existing `test_save_related_creates_profile_and_favorites_if_missing` continues to pass (new-user path unchanged).
- [ ] Run `docker compose run --rm web uv run pytest users/adapters_test.py -v` to validate all three tests pass.

### DIFF
--- a/users/adapters.py
+++ b/users/adapters.py
@@ -60,21 +60,24 @@ class LearnUserAdapter(UserAdapter):
         with synchronous I/O.
         """
         newly_created = self.is_new_user
+        user_pk = self.obj.pk
         with transaction.atomic():
             if not newly_created:
                 # Lock the user row so concurrent PATCH requests for the
                 # same user block here rather than racing to the UPDATE.
                 User.objects.select_for_update().get(pk=self.obj.pk)
             super().save()
+            if not newly_created:
+                # Register the Celery dispatch inside the atomic block so
+                # on_commit fires after the transaction commits, not immediately
+                # (which would happen in autocommit mode if called outside).
+                transaction.on_commit(
+                    lambda: reindex_user_learning_paths.delay(user_pk)
+                )
         if newly_created:
             pm = get_plugin_manager()
             hook = pm.hook
             hook.user_created(user=self.obj, user_data={})
-        else:
-            # Dispatch search re-indexing only after the outermost transaction
-            # commits so the Celery worker always sees the committed state.
-            user_pk = self.obj.pk
-            transaction.on_commit(lambda: reindex_user_learning_paths.delay(user_pk))
 
     def _save_related(self):
         """
--- a/users/adapters.py
+++ b/users/adapters.py
@@ -8,6 +8,20 @@ from users.tasks import reindex_user_learning_paths
 
 User = get_user_model()
 
+# User model fields managed by SCIM (set by UserAdapter.from_dict()).
+# _save_user() restricts writes to these columns so that non-SCIM data
+# loaded before this request is not written back over a concurrent update.
+_SCIM_USER_FIELDS = (
+    "email",
+    "username",
+    "first_name",
+    "last_name",
+    "is_active",
+    "scim_username",
+    "scim_external_id",
+    "global_id",
+)
+
 
 class LearnUserAdapter(UserAdapter):
     """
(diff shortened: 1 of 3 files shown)


users/adapters.py
```diff
@@ -43,9 +48,32 @@ def from_dict(self, d):
...
+        with synchronous I/O.
         """
         newly_created = self.is_new_user
-        super().save()
+        user_pk = self.obj.pk
+        with transaction.atomic():
+            if not newly_created:
+                # Lock the user row so concurrent PATCH requests for the
+                # same user block here rather than racing to the UPDATE.
+                User.objects.select_for_update().get(pk=self.obj.pk)
```
rhysyngsun: There's already an atomic transaction [being created in the super class](https://github.com/mitodl/ol-django/blob/main/src/scim/mitol/scim/adapters.py#L146), so I think this lock should be moved there. I also think it should be setting `self.user` to ensure we have the latest view of the user, otherwise the `user.save()` call will potentially overwrite fields SCIM isn't touching with stale data. 

With that, this PR can be simplified down to moving the `transaction.on_commit` into an override for `_save_user()`, because that gets called within the transaction.

rhysyngsun (changes requested):

users/adapters.py
blarghmatey: Done in a6528b4. Moved the lock and `on_commit` into a `_save_user()` override so they run inside the parent's `transaction.atomic()`, and simplified `save()` to only handle the new-user plugin hook.

For the stale-field concern: I used `update_fields` rather than replacing `self.obj` with the locked row. Replacing `self.obj` would also discard the in-memory profile set by `from_dict()` on `self.obj.profile`, which `_save_related()` depends on — so we'd silently drop profile name/email-optin changes on every SCIM update. `update_fields` achieves the same outcome (non-SCIM columns are never written) without that issue.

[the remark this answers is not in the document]
blarghmatey: Not a real issue. The parent `UserAdapter._save_user()` is a one-liner — `self.obj.save()` — with no additional SCIM metadata logic, versioning, or ETag handling (verified at [mitol/scim/adapters.py#L137-L138](https://github.com/mitodl/ol-django/blob/main/src/scim/mitol/scim/adapters.py#L137-L138)).

Our override for existing users calls `self.obj.save(update_fields=_SCIM_USER_FIELDS)`, which is the same call with a field restriction — not a bypass of any parent logic. The asymmetry (`super()._save_user()` for new users only) is intentional: `update_fields` only applies to UPDATE statements, and a new user needs a full INSERT.

blarghmatey:
Superseded by https://github.com/mitodl/ol-django/pull/501

PR #3549, opened 2026-06-30
Add one click unsubscribe to non-transactional emails

rhysyngsun:

Closes https://github.com/mitodl/hq/issues/8424

This adds one-click unsubscribe. It manifests in 2 ways:
1. Some email clients, particularly Gmail will present and Unsubscribe button at the top of the email that allows the user to unsubscribe without leaving their mail client. This happens via an HTTP post by the client in the background.
2. There is also a traditional unsubscribe link in the email footer which navigates the user to a URL that performs that operation via HTTP GET and then redirects the user to the frontend. If there is an error, that is displayed, giving the user the option to login to unsubscribe manually.

I've also added a checkbox on the settings page that allows the user to opt out of emails without having to modify their subscriptions so they can put a non-destructive pause on these (and this is how the contactless unsubscribe works).

### How can this be tested?

- You can send yourself a test email with `./manage.py send_test_email digest --count 3 --email EMAIL`.
- It should have an unsubscribe link at the bottom, clicking it should update your user's `Profile.email_optin` to `False`
- Running the command again should not send the email.
- You should also be able to toggle the field value in the frontend on the settings page.


users/migrations/0011_user_unsubscribe_uuid.py
```diff
@@ -0,0 +1,19 @@
...
+class Migration(migrations.Migration):
+    dependencies = [
+        ("users", "0010_alter_user_global_id"),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name="user",
+            name="unsubscribe_uuid",
+            field=models.UUIDField(default=uuid.uuid4, null=True, unique=True),
```
shanbady, line 17: AddField with default=uuid.uuid4 and unique=True evaluates the callable default once, backfilling every existing user row with the same UUID, so the unique index creation fails and the migration crashes on any database with 2+ users.

I ran into this locally:
```
Traceback (most recent call last):
  File "/src/manage.py", line 29, in <module>
    execute_from_command_line(sys.argv)
  File "/opt/venv/lib/python3.12/site-packages/django/core/management/__init__.py", line 442, in execute_from_command_line
    utility.execute()
  File "/opt/venv/lib/python3.12/site-packages/django/core/management/__init__.py", line 436, in execute
    self.fetch_command(subcommand).run_from_argv(self.argv)
  File "/opt/venv/lib/python3.12/site-packages/django/core/management/base.py", line 412, in run_from_argv
    self.execute(*args, **cmd_options)
  File "/opt/venv/lib/python3.12/site-packages/django/core/management/base.py", line 458, in execute
    output = self.handle(*args, **options)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/venv/lib/python3.12/site-packages/django/core/management/base.py", line 106, in wrapper
    res = handle_func(*args, **kwargs)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/venv/lib/python3.12/site-packages/django/core/management/commands/migrate.py", line 356, in handle
    post_migrate_state = executor.migrate(
                         ^^^^^^^^^^^^^^^^^
  File "/opt/venv/lib/python3.12/site-packages/django/db/migrations/executor.py", line 135, in migrate
    state = self._migrate_all_forwards(
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/venv/lib/python3.12/site-packages/django/db/migrations/executor.py", line 167, in _migrate_all_forwards
    state = self.apply_migration(
            ^^^^^^^^^^^^^^^^^^^^^
  File "/opt/venv/lib/python3.12/site-packages/django/db/migrations/executor.py", line 252, in apply_migration
    state = migration.apply(state, schema_editor)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/venv/lib/python3.12/site-packages/django/db/migrations/migration.py", line 132, in apply
    operation.database_forwards(
  File "/opt/venv/lib/python3.12/site-packages/django/db/migrations/operations/fields.py", line 108, in database_forwards
    schema_editor.add_field(
  File "/opt/venv/lib/python3.12/site-packages/django/db/backends/base/schema.py", line 713, in add_field
    self.execute(sql, params)
  File "/opt/venv/lib/python3.12/site-packages/django/db/backends/postgresql/schema.py", line 48, in execute
    return super().execute(sql, None)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/venv/lib/python3.12/site-packages/django/db/backends/base/schema.py", line 201, in execute
    cursor.execute(sql, params)
  File "/opt/venv/lib/python3.12/site-packages/django/db/backends/utils.py", line 102, in execute
    return super().execute(sql, params)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/venv/lib/python3.12/site-packages/sentry_sdk/utils.py", line 1841, in runner
    return sentry_patched_function(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/venv/lib/python3.12/site-packages/sentry_sdk/integrations/django/__init__.py", line 645, in execute
    result = real_execute(self, sql, params)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/venv/lib/python3.12/site-packages/django/db/backends/utils.py", line 67, in execute
    return self._execute_with_wrappers(
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/venv/lib/python3.12/site-packages/django/db/backends/utils.py", line 80, in _execute_with_wrappers
    return executor(sql, params, many, context)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/venv/lib/python3.12/site-packages/django/db/backends/utils.py", line 84, in _execute
    with self.db.wrap_database_errors:
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/venv/lib/python3.12/site-packages/django/db/utils.py", line 91, in __exit__
    raise dj_exc_value.with_traceback(traceback) from exc_value
  File "/opt/venv/lib/python3.12/site-packages/django/db/backends/utils.py", line 87, in _execute
    return self.cursor.execute(sql)
           ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/venv/lib/python3.12/site-packages/opentelemetry/instrumentation/psycopg/__init__.py", line 380, in execute
    return _cursor_tracer.traced_execution(
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/venv/lib/python3.12/site-packages/opentelemetry/instrumentation/dbapi/__init__.py", line 761, in traced_execution
    return query_method(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/venv/lib/python3.12/site-packages/psycopg/cursor.py", line 117, in execute
    raise ex.with_traceback(None)
django.db.utils.IntegrityError: could not create unique index "users_user_unsubscribe_uuid_key"
DETAIL:  Key (unsubscribe_uuid)=(9740d88f-e8ce-4a76-906a-34c34ccc7dad) is duplicated.
```

users/views.py
```diff
@@ -0,0 +1,58 @@
...
+            with transaction.atomic():
+                sync_email_optin_to_keycloak(user, email_optin=False)
+                profile.email_optin = False
+                profile.save(update_fields=["email_optin"])
+        except KeycloakError:
+            log.exception("Failed to sync email_optin to Keycloak for user %s", user.id)
+            return False
+        return True
+
+    def get(self, request, token):  # noqa: ARG002
```
shanbady, line 49: this handler mutates state immediately with no confirmation step -   mail-scanner link prefetchers will silently unsubscribe users who never clicked. The footer link in email_base.html points at this endpoint.
RFC 8058 reserves one-click for the POST header path; the human-facing GET should render a confirmation page.

profiles/serializers.py
```diff
@@ -151,9 +155,32 @@ def update(self, instance, validated_data):
...
+                        instance.user, email_optin=instance.email_optin
+                    )
+                except KeycloakError as exc:
+                    log.exception(
+                        "Failed to sync email_optin to Keycloak for user %s",
+                        instance.user.id,
+                    )
+                    raise ValidationError(
+                        {
+                            "email_optin": (
```
shanbady, line 178: If keycloak fails to update - the save on the user object never happens/persists so they keep receiving marketing email they explicitly declined. we should consider saving their opt-out preference beforehand so atleast they dont get emails after explicitly opting out (may need to sort out keycloak reconciliation some other way )

shanbady (changes requested):
Left notes on some items - there are also some merge conflicts with main to resolve

profiles/serializers.py
rhysyngsun: I agree that would be ideal, but it's a lot more complicated to accomplish because keycloak pushes down updates too and could override the value before we manage to persist it. So for the time being that's why I'm purposely erroring here because it's better to error than to appear to succeed and then later have emails suddenly start sending again if keycloak pushes down stale state.

users/views.py
rhysyngsun: Good catch

rhysyngsun:
Closing this in favor of #3636 and #3637 as I had to split the work into 2 branches because the API changes now need to get merged and released first. I'll be incorporating @shanbady's feedback in those.

PR #3551, opened 2026-07-01
Migration to fix non-credit level for runs

shanbady:

Closes https://github.com/mitodl/hq/issues/10841

This PR adds a data migration to fix the state of a few runs on production that have the invalid level of "Non-Credit" (should be noncredit)

### How can this be tested?
1. grab some random runs and set their level to the invalid "Non-Credit":
```python

lrs = LearningResourceRun.objects.all()[:20]

for lr in lrs:
	lr.level = ['Non-Credit']
	lr.save()
```
2. checkout this branch
3. run the data migration via ` RUN_DATA_MIGRATIONS=True python manage.py migrate data_fixtures`
4. gio back into shell and see that the count of items with the invalid level is 0:
```python
LearningResourceRun.objects.filter(level__contains=['Non-Credit']).count()
```

### Additional Context
It seems like the same handful of runs have had this issue for some time now so it is likely just instances left in a bad state at some point.

### DIFF
--- a/data_fixtures/migrations/0025_replace_non_credit_level.py
+++ b/data_fixtures/migrations/0025_replace_non_credit_level.py
@@ -1,0 +1,34 @@
+# Generated manually
+"""Replace invalid non-credit level keys on learning resource runs."""
+
+from django.db import migrations
+
+OLD_LEVEL = "non-credit"
+NEW_LEVEL = "noncredit"
+
+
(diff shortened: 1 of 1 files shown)


shanbady:
spoke with @abeglova - these instances are just one-off orphans that we can simply delete instead of creating a migration/fixture for

PR #3552, opened 2026-07-01
Make the Profile.email_optin field writable

rhysyngsun:

Part of https://github.com/mitodl/hq/issues/8424

This makes the field writable so that I can implement UI to toggle this value in the UI in a future PR.

### How can this be tested?

Tests should pass

### DIFF
--- a/frontends/api/src/generated/v0/api.ts
+++ b/frontends/api/src/generated/v0/api.ts
@@ -4412,6 +4412,12 @@ export interface Profile {
   profile_image_medium: string
   /**
    *
+   * @type {boolean}
+   * @memberof Profile
+   */
+  email_optin?: boolean
+  /**
+   *
    * @type {string}
    * @memberof Profile
    */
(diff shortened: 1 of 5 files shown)


ChristopherChudzicki (commented):
I was confused because the field wasn't coming through, but seemed like a docker issue.

Small notes:

1. In PR description
    > This makes the field writable so that I can implement UI to toggle this value in the UI in a future PR.

    I think you meant readable.
2. Copilot's comments seem true, BUT:
    1. email optin isn't too sensitive.
    2. The `profile/{username}` endpoints are currently unusable for email usernames. Our router strips the stuff after the period. **Maybe we could just remove these? ** We don't use these endpoints, EXCEPT for `username = me` which is handled specially anyway.
3. Not really related, but I noticed that `/profile/me` 500s for unauthenticated users instead of 404

ChristopherChudzicki (approved):

PR #3489, opened 2026-06-17
Ab/password game

abeglova:
### What are the relevant tickets?
closes https://github.com/mitodl/hq/issues/11688

### Description (What does it do?)
This pr adds the password game/ hacksnack to learn. The pr for the library is here https://github.com/mitodl/hacksnack/pull/1

### Screenshots (if appropriate):
Screenshot 2026-06-17 at 7 42 06 PM
Screenshot 2026-06-17 at 7 54 29 PM

### How can this be tested?
got to 
http://open.odl.local:8062/games/hacksnack

you should get a 404

enable "hacksnack" as a feature flag in posthog. 

go to http://open.odl.local:8062/games/hacksnack
again. The game should load and work correctly

If you want the answers to see how the app responds to correct solutions you can get them in 
assets/puzzle_picker_list.json in This pr adds the password game/ hacksnack to learn. The pr for the library is here https://github.com/mitodl/hacksnack/pull/1

PR #3555, opened 2026-07-01
fix hacksnack postinstall

abeglova:
closes https://github.com/mitodl/hq/issues/11688

The hacksnack library has a postinstall step that copies image assets and puzzle data into ./public/games/hacksnack so that they are available to the game. I accidentally deleted it from package.json while rebasing https://github.com/mitodl/mit-learn/pull/3489

### How can this be tested?
Go to http://open.odl.local:8062/games/hacksnack

The game  should load


shanbady (approved):
LGTM

PR #3560, opened 2026-07-01
Add opt-in django-aqueduct typed settings (django-aqueduct>=0.9.0)

blarghmatey:
N/A

This branch adds an **opt-in, parallel** [django-aqueduct](https://github.com/mitodl/django-aqueduct)-based settings module for mit-learn. It has been **reworked onto django-aqueduct 0.9.0 (codegen v2)** — a ground-up rewrite in which the generator discovers settings by **static AST analysis** instead of importing the settings module, so generation is deterministic and secret-safe and recovers env-var aliases, required-ness, and conditional branches that the old live-import engine lost.

**Dependency:** `pyproject.toml` now requires `django-aqueduct[derivations,vault]>=0.9.0` (re-locked). The v1 live-import engine is removed upstream; v2 adds static discovery, managed-region regeneration, reusable `derivations`, a configurable Vault dev base, and the `check_aqueduct_settings` parity command. 0.7.1 made the generator emit `Annotated[..., NoDecode]` + a decoder for every list/dict field; 0.8.1 made its output `ruff format`-stable; 0.9.0 fixed the `[*]` wildcard decode, `get_list_of_str` discovery, opt-in URL typing, and subset dict-parity — letting this branch drop **all** of its earlier hand-written workarounds.

**New / reworked, opt-in settings modules (non-disruptive):**
- `main/aqueduct_settings.py` — a typed `AqueductSettings(pydantic_settings.BaseSettings)` model **regenerated from scratch** by `generate_aqueduct_settings` into managed `# >>> aqueduct:generated:*` regions, with all hand-written validators and field overrides in `# >>> aqueduct:preserved:*` regions that survive regeneration. Plus `DevAqueductSettings`, now built on `django_aqueduct.sources.dev.VaultDevBase`.
- `main/settings_aqueduct.py` / `main/settings_aqueduct_dev.py` — thin `DJANGO_SETTINGS_MODULE` shims; both now initialize Sentry from the validated model via the 0.7.0 `pre_configure` hook.
- `[tool.aqueduct]` in `pyproject.toml` records the generation + parity config so the file can be regenerated and drift-checked in CI.

**`main/envs.py` is untouched; `main/settings.py` changes by one line** — `django_aqueduct` is added to `INSTALLED_APPS` so the `generate_aqueduct_settings` / `check_aqueduct_settings` management commands are available. `main.settings` remains the default `DJANGO_SETTINGS_MODULE` everywhere it is already configured — this stays a fully non-disruptive, additive change.

**Regeneration fixed the review-flagged bugs**, because the machine no longer bakes its own runtime state into the file:

- **CRITICAL** — `AWS_ACCESS_KEY_ID` / `AWS_SECRET_ACCESS_KEY` were typed `bool` (default `False`, from `get_string(name, False)`); they are now `str | None`, so a real credential is never coerced to a boolean.
- **CRITICAL** — `APP_BASE_URL` now carries its `MITOL_APP_BASE_URL` alias and is enforced **required** (validator mirroring the legacy fail-fast) instead of a localhost default that masked missing config.
- **Recovered MITOL_* aliases** the v1 model dropped (it shipped with zero `validation_alias` entries): `SECURE_SSL_REDIRECT`←`MITOL_SECURE_SSL_REDIRECT`, `EMAIL_*`←`MITOL_EMAIL_*`, `DEFAULT_FROM_EMAIL`←`MITOL_FROM_EMAIL`, and ~300 others, recovered automatically by static discovery.
- `OIDC_LOGOUT_URL` is derived from `MITOL_API_BASE_URL`; `MIGRATION_MODULES` reproduces the `RUN_DATA_MIGRATIONS` branch (preserved validators).
- **Generation-machine snapshot defaults are gone**: `HOSTNAME` is dynamic again; `ANYMAIL`, `JWT_AUTH`, `OPENSEARCH_INDEX`, `MITOL_COOKIE_DOMAIN` no longer freeze the CI box's fake values.

**Preserved-region logic (celery-safe, derivation-backed):**
- `CELERY_BEAT_SCHEDULE` / `CELERY_BEAT_SCHEDULER` are rebuilt (their crontab schedule interval env vars are now typed fields) so the model is safe to run under celery worker/beat; `LOGGING` is attached from mitol-observability rather than left `None`.
- `DATABASES`, the Redis fallback chain, and `CACHES` use `django_aqueduct.derivations` — `database_config` is SQLite-safe and no longer applies `sslmode` to the sqlite fallback (a latent legacy bug).
- `FEATURES` reads through the model (`derivations.feature_flags`) so Vault/SSM-sourced flags are seen, not just `os.environ`.

**Vault dev base:** `DevAqueductSettings` now subclasses `VaultDevBase`, which configures Vault entirely from `VAULT_*` env vars and fails gracefully when unset — fixing the bare `KeyError('VAULT_ADDR')` and the hardcoded-placeholder mount/path/role wiring, and clarifying source precedence (Vault is prepended, so Vault values win).

**List env decoding (fully generator-handled in 0.9.0).** mit-learn's deployed `.env` files carry Python-literal and bare-wildcard list forms (`ALLOWED_HOSTS=[*]`, single-quoted regex lists) that pydantic-settings' default `json.loads` rejects — which crashed model construction under `manage.py` (litellm's import runs `load_dotenv()`, pulling a stray `.env` into `os.environ`). The generator now emits `Annotated[..., NoDecode]` + a decoder for every list/dict field, and 0.9.0 closed the two gaps this branch previously kept minimal overrides for — **both overrides are now removed**:
  - The generated decoder now strips one layer of surrounding brackets before its comma-split fallback, so the bare-wildcard `ALLOWED_HOSTS=[*]` parses to `["*"]` (verified under the stray-`.env` repro) without the one-field normalizer.
  - Static discovery now recognizes mitol's `get_list_of_str` as a `list[str]` reader, so `CORS_ALLOWED_ORIGINS`, `CORS_ALLOWED_ORIGIN_REGEXES`, `CSRF_TRUSTED_ORIGINS`, `OPEN_AI_EMBEDDING_MODELS`, and `ALLOWED_REDIRECT_HOSTS` get `NoDecode` + the container decoder automatically (they were previously discovered as `Any` and stored the raw env string).

**URL typing (`enrich_url_types`, 0.9.0).** 0.9.0 makes the `str`→`AnyUrl` promotion opt-in, default-validated, denylisted for Django's relative-URL settings, and paired with a `str` serializer for `model_dump()` — so the 49-field manual revert block this branch carried is **deleted**. Enabling `enrich_url_types` was evaluated but left **off**: it still promotes `APP_BASE_URL` (default `None`, promoted via the name-suffix fallback) to a `pydantic.Url` object, which this model's own `build_urls` validator then feeds to `urljoin()` (rejected as non-`str`). With it off the URL fields stay plain `str` — the drop-in contract — with no override block at all. `--enrich-usage`/`--enrich-runtime` are not used (usage found nothing previously; runtime is the one flag that imports the module).

**Security policy fixes** (kept from the original PR): `SECRET_KEY` and `WEBHOOK_SECRET` are **required** with no default; the insecure hardcoded fallbacks (`terribly_unsafe_default_secret_key`, `please-change-this`) are intentionally not reproduced.

N/A — no UI changes.

### How can this be tested?

- [ ] `uv lock` / `uv sync` resolve `django-aqueduct` to `0.9.0`.
- [ ] `DJANGO_SETTINGS_MODULE=main.settings_aqueduct uv run python manage.py check` passes with the required settings set in the environment.
- [ ] `DJANGO_SETTINGS_MODULE=main.settings_aqueduct_dev uv run python manage.py check` runs cleanly with `VAULT_ADDR` unset (runs without Vault) and reaches Vault auth when `VAULT_*` are set.
- [ ] `uv run python manage.py generate_aqueduct_settings --check` reports the model is up to date (no drift from a fresh render).
- [ ] `uv run python manage.py check_aqueduct_settings` reports parity with `main.settings` (deliberate divergences are documented in `[tool.aqueduct] parity_ignore`).
- [ ] `git diff --name-only -- '*.py' | xargs uv run ruff check --force-exclude` passes (the generated file is excluded from ruff so regeneration/`--check` is stable).

### Additional Context

- Deliberate parity divergences are enumerated with justifications in `[tool.aqueduct] parity_ignore` in `pyproject.toml` (**32 entries**): the `SECRET_KEY`/`WEBHOOK_SECRET`/AWS-cred policy fixes, conditional module locals, the typed input primitives the legacy read inline (including `HOST_IP`, which now backs `INTERNAL_IPS` through the source chain), and benign `get_float` int→float normalization. `DATABASES`/`DEFAULT_DATABASE_CONFIG`/`HEALTH_CHECK` were **dropped** — 0.9.0's parity check does a one-way subset compare for dict settings, so the keys Django's `ConnectionHandler` and django-health-check's `AppConfig.ready()` inject at runtime no longer read as divergences.
- The generated `main/aqueduct_settings.py` is excluded from ruff (`[tool.ruff] extend-exclude`). 0.8.1 made the generator output `ruff format`-stable for the ~9 rule groups it validated against, but mit-learn selects a much broader set — `ruff check` flags 68 issues in the generated regions and `ruff format` would rewrite them (causing `--check` drift) — so per the upstream note extend-exclude remains the documented fallback here.


blarghmatey:
django-aqueduct 0.6.0 is now published on PyPI. Removed the `[tool.uv.sources]` local-path override and TODO in `pyproject.toml`, re-ran `uv lock`/`uv sync` to resolve it from PyPI (sdist/wheel from files.pythonhosted.org), and re-verified `manage.py check` against both `main.settings_aqueduct` and `main.settings_aqueduct_dev` — same results as before, no longer blocked on that TODO.

blarghmatey:
Closing for now — we'll revisit and recreate this later.

PR #3561, opened 2026-07-02
Fix embeddings task oom thrashing

shanbady:

Closes https://github.com/mitodl/hq/issues/12115

This PR prevents/better controls OOM issues with celery tasks by controlling how many contentfiles are serialized at a time (via settings.QDRANT_CONTENT_FILE_SERIALIZATION_CHUNK_SIZE) before fanning out to celery. 

Currently what is happening is that each celery task serializes settings.QDRANT_CHUNK_SIZE contentfiles at once into memory for each task - if a task happens to have very large contentfile objects that are over the memory threshold, it results in an OOM kill - the probability of this goes up as we increase settings.QDRANT_CHUNK_SIZE.

### How can this be tested?
1. checkout main and restart celery
2. in another shell keep tabs on the memory usage of the celery container via ```docker stats mit-open-celery-1```
3. run the following script which generates a large contentfiles and attempts to embed them:
```python
from learning_resources.constants import LearningResourceType
from learning_resources.models import (
    ContentFile,
    LearningResource,
    LearningResourcePlatform,
    LearningResourceRun,
)

platform, _ = LearningResourcePlatform.objects.get_or_create(
    code="memory-test",
    defaults={"name": "Memory Test"},
)

resource,_ = LearningResource.objects.get_or_create(
    readable_id="memory-test-course",
    title="Memory Test Course",
    resource_type=LearningResourceType.course.name,
    platform=platform,
    published=True,
)

run,_ = LearningResourceRun.objects.get_or_create(
    learning_resource=resource,
    run_id="memory-test-run",
    title="Memory Test Run",
    published=True,
)

large_content = "Large embedding memory test content. " * 500_000

content_files = [
    ContentFile.objects.create(
        run=run,
        key=f"memory-test-{i}",
        title=f"Memory Test File {i}",
        content=large_content,
        checksum=f"memory-test-{i}",
        published=True,
    )
    for i in range(100)
]

cfids = [cf.id for cf in content_files]

from learning_resources_search.constants import CONTENT_FILE_TYPE
from vector_search.tasks import generate_embeddings



generate_embeddings.delay(
    cfids,
    CONTENT_FILE_TYPE,
    overwrite=True,
)
```
4. note the ballooning memory on the celery container
5. checkout this branch. 
6. restart celery and re-run the above script and note the steady usage.


vector_search/utils_test.py
```diff
@@ -513,6 +513,28 @@ def test_expected_document_chunks(mocker):
     assert len(chunked) == num_points_uploaded
 
 
+def test_embed_learning_resources_chunks_content_file_serialization(mocker, settings):
+    """
+    QDRANT_CHUNK_SIZE controls Celery task fanout; content files are serialized in
+    smaller batches so large task chunks do not hydrate every content blob at once.
+    """
+    settings.QDRANT_CHUNK_SIZE = 50
```
mbertrand: I think QDRANT_CHUNK_SIZE is actually irrelevant to this test?  Maybe remove the setting assignment and update the docstring?

mbertrand (approved):

vector_search/utils_test.py
shanbady: fixed

PR #3562, opened 2026-07-02
Bound program marketing-page embedding input to avoid token overflow

mbertrand:

Refs mitodl/hq#12117

Program marketing-page content files were ballooning to multi-MB (23MB for `UAI+B2C.1` in production) because they concatenated the LLM summaries of every content file in every child-course run. When embedded, a fully packed request overflowed OpenAI's 300k tokens/request limit (`max_tokens_per_request`), failing that chunk. The `generate_embeddings` chunk swallows the error, so it surfaced only as recurring `finalize_embeddings` `RuntimeError: 1 embedding chunk(s) failed` in leek.

Changes:

- **Headroom in the embedding request batch size** (`vector_search/utils.py`): request batches now target 90% of the 300k limit, since markdown header prefixes are prepended *after* the chunk-size split and inflate chunks past the nominal size. Guarded with `max(1, …)` so a misconfigured chunk-size override can't produce a zero-step batching loop.
- **Rework program children content** (`learning_resources/utils.py`): a program's children content is now assembled from the **marketing-page content files of its published child courses** (direct, and reached through child programs), each under a light `### {course title}` heading — instead of a capped subset of child-course run content-file summaries. Each course contributes only its single scraped marketing page, so the program page stays a sensible embedding input by construction. A 1M-char total cap remains as a backstop.

### How can this be tested?

1. Make sure you have  contentfiles for this program's courses:
   ```python
     from learning_resources.utils import build_program_children_content
     lr = LearningResource.objects.get(readable_id="program-v1:MITxT+18.01x")
     print(lr.id)
     print(lr.resources.values_list("id", flat=True))
   ```
   
   `./manage.py backpopulate_mitxonline_files --resource-ids <list of ids>`

2. **Children content is built from child-course marketing pages.**  In a shell, generate marketing pages for the programs and courses:
   ```python
   from learning_resources.tasks import marketing_page_for_resources
   from learning_resources.utils import build_program_children_content
   
   for lr in LearningResource.objects.filter(id__in=[program + course ids]):
       lr.save()   
   
   marketing_page_for_resources([program + course ids])
   lr =  LearningResource.objects.get(readable_id="program-v1:MITxT+18.01x")
   content = build_program_children_content(lr)
   ```
   Confirm the output contains a `### {course title}` section per published child course carrying that course's marketing-page text, that courses without a marketing page are absent, and `len(content) <= 1_000_000`.

4. **Embedding of a previously-failing program page succeeds.** After the re-scrape, embed it:
   ```python
   from vector_search.tasks import *
   cf = ContentFile.objects.get(learning_resource=lr, file_type="marketing_page")
   generate_embeddings.delay([cf.id], "content_file", overwrite=True)
   ```
   Confirm the task succeeds

5. **No more finalize failures.** Trigger a finalized chain covering that file (`embed_run_content_files.delay(lr.best)_run.id)`, or wait for `embed_new_content_files`) and verify it completes successfully

### DIFF
--- a/learning_resources/tasks_test.py
+++ b/learning_resources/tasks_test.py
@@ -646,6 +646,14 @@ def test_marketing_page_for_program_appends_children(mocker, settings):
         relation_type="PROGRAM_COURSES",
         position=0,
     )
+    models.ContentFile.objects.create(
+        learning_resource=child_course,
+        file_type=MARKETING_PAGE_FILE_TYPE,
+        file_extension=".md",
+        key="mktg-child-course",
+        content="Child Course marketing copy",
+        published=True,
+    )
 
     html_content = "<html><body><h1>Test Program</h1><p>Program info</p></body></html>"
     mocker.patch(
--- a/learning_resources/utils.py
+++ b/learning_resources/utils.py
@@ -748,19 +748,19 @@ def build_resource_summary_dict(resource):
 PROGRAM_CHILDREN_CONTENT_MAX_CHARS = 1_000_000
 
 
-def _reachable_course_ids(relationships, course_type):
-    """Course ids from a program's child relationships (direct + via child programs)."""
-    course_ids = set()
+def _course_ids_by_program(relationships, course_type):
+    """Map program id -> reachable published course ids (direct + via subprograms)."""
+    result = defaultdict(set)
     for rel in relationships:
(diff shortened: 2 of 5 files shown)


shanbady (approved):
LGTM 👍

PR #3566, opened 2026-07-02
feat(cohort-1): warehouse-pull Celery tasks for catalog sources

blarghmatey:
Part of https://github.com/mitodl/hq/issues/11510 (Cohort 1: DB-Backed Catalog Sources / Trino-pull)

Follow-up tracked in https://github.com/mitodl/mit-learn/issues/3746 (pre-cutover field gap on program runs).

Implements the Cohort 1 Trino-pull ETL pattern for MIT Learn's OL Data Platform migration: 6 new Celery tasks that pull pre-computed `integrations__learn__*` views from the warehouse and upsert them through the existing course/program loaders, replacing (once cut over) the equivalent API-based ETL pipelines for these sources.

**New tasks** (`learning_resources/tasks.py`): `SyncMITxOnlineCoursesTask`, `SyncMITxOnlineProgramsTask`, `SyncXProCoursesTask`, `SyncXProProgramsTask`, `SyncMITEdXCoursesTask`, `SyncOCWCoursesTask`. Program tasks `fetch_only` their child courses, so are scheduled 15 minutes after their courses task.

**New infrastructure** (`learning_resources/lib/warehouse.py`):
- `BaseWarehouseETLTask` — a Celery task base class that opens a warehouse connection, delegates to `fetch_and_upsert`, and handles connection lifecycle/error reporting (Sentry breadcrumbs).
- `iter_rows()` — backend-agnostic row iteration built on the plain DB-API 2.0 cursor surface (`execute`/`description`/`fetchmany`). Trino is the only backend wired up today (`settings.WAREHOUSE_BACKEND`); a `_connect_starrocks` stub documents the intended migration path — since Trino, StarRocks, and DuckDB all expose the same cursor surface, adding a backend is a connector fill-in, not an ETL-layer rewrite.
- **Full-refresh vs. incremental sync**: `run()` takes a `full_refresh: bool = True` kwarg. `full_refresh=True` (today's beat schedule default) pulls every row and prunes resources no longer present in the source — the self-healing baseline. `full_refresh=False` pulls only rows with `last_modified` greater than a persisted watermark (stored in Django's existing durable DB-backed cache, not a new model/migration) and skips pruning, since a partial pull must never be treated as the complete source state. Every `integrations__learn__*` view already exposes `last_modified` per the platform's marts contract, so this required no changes on the data platform side.

**New transforms** (`learning_resources/etl/catalog_sources.py`): pure `dict -> dict` functions mapping each view's flattened row contract (delimited strings for topics/instructors/runs, since the views intentionally avoid nested JSON) to the shape `loaders.load_courses`/`load_programs` expect.

**Per-source cutover switch** (`main/settings_celery.py`): `WAREHOUSE_ETL_CUTOVER_SOURCES`, a comma-separated list of `ETLSource` names. Listing a source removes its API-based beat entry (`mit_edx`, `mitxonline`, `xpro`). Deliberately *not* keyed on `TRINO_HOST`: during parallel validation `TRINO_HOST` is set and both pipelines must run, since the API-based ETL is the comparison baseline. The `import_all_*_files` tasks are never removed — the views carry catalog metadata only, not content files. An unrecognized source name raises `ImproperlyConfigured` rather than silently leaving a legacy task scheduled.

**Loader change**: `loaders.load_run`/`load_instructors`/`load_prices` now treat `None` as "not provided by this source, leave the existing value alone" — the same sentinel convention `load_topics` already uses for `topics_data`. The warehouse transforms pass `None` for prices (course and program runs) and instructors (program runs) so a sync doesn't wipe values the API-based ETL populated. An omitted key still defaults to `[]`, so this is additive for every existing caller.

**Bug fix**: `_SAFE_IDENTIFIER`'s regex used `$` instead of `\Z` to anchor the "safe view name" check. Python's `$` matches just before a trailing newline (not strictly end-of-string), so a `view_name` ending in `\n` slipped past the guard. Fixed to `\Z`.

**Not included: MicroMasters.** An earlier revision of this PR added `SyncMicromastersProgramsTask`. `main` has since removed the MicroMasters ETL entirely (https://github.com/mitodl/mit-learn/pull/3663), including migration `0117_unpublish_micromasters_resources`, which unpublishes and deindexes every `etl_source="micromasters"` resource. The warehouse task would have republished exactly what that migration retired, so it has been dropped.

### How can this be tested?
No live Trino endpoint is available yet in any environment (see https://github.com/mitodl/hq/issues/11509, "confirm MIT Learn deployment environment has network access to the Trino/Starburst endpoint" — still open), so there is no live-connectivity smoke test in this PR. All tests in the touched files pass against mocked warehouse connections:

```
docker compose run --rm web uv run pytest \
  learning_resources/tasks_test.py \
  learning_resources/lib/warehouse_test.py \
  learning_resources/etl/catalog_sources_test.py
# 122 passed
```

- `warehouse_test.py` covers connection dispatch (Trino/StarRocks/unknown backend), `iter_rows` batching/cursor-cleanup/SQL-injection-guard/`since`-filtering, and `BaseWarehouseETLTask`'s full-refresh vs. incremental branching (watermark read/write, prune flag, error paths) — all against a hand-built DB-API-2.0-shaped mock, no Django app/DB required.
- `tasks_test.py` covers each task's wiring (view name -> transform -> loader -> `clear_views_cache`) plus explicit full-refresh vs. incremental coverage, against a real Django test DB.
- `catalog_sources_test.py` covers the row-transform functions directly.
- `main/settings_test.py` covers `WAREHOUSE_ETL_CUTOVER_SOURCES` (default, per-source removal, unknown-name rejection).

Reviewers can also sanity-check the beat schedule wiring in `main/settings_celery.py` and confirm the `full_refresh=True` kwargs match the intended daily-parallel-validation cadence described in the cohort's implementation guide.

### Additional Context
This is Cohort 1 infrastructure landing ahead of the still-open network-access confirmation (https://github.com/mitodl/hq/issues/11509) and before parallel validation against the existing API-based ETL has started. Once Trino connectivity is confirmed from this deployment environment, the plan is to run both ETL paths in parallel for 2 weeks (≥99% record match) before cutover — tracked as a separate task in the epic, and now gated per source by `WAREHOUSE_ETL_CUTOVER_SOURCES`.

Where the data comes from: the `integrations__learn__*` views build on source application Postgres replicas (`stg__mitxonline__app__postgres__*`, `stg__mitxpro__app__postgres__*`, `stg__ocw__studio__postgres__*`), not the source APIs. `mit_edx` is the exception — it derives from `stg__edxorg__api__*`, since there is no DB replica of edx.org.


learning_resources/etl/catalog_sources.py
```diff
@@ -0,0 +1,345 @@
...
+    "WHU": PlatformType.whu.name,
+    "xPRO": PlatformType.xpro.name,
+}
+
+
+def _split(value: str | None, sep: str = ", ") -> list[str]:
+    """Split a delimited string column into stripped, non-empty parts."""
+    if not value:
+        return []
+    return [part.strip() for part in value.split(sep) if part.strip()]
```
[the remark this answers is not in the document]
blarghmatey: Fixed in 0422d0e9c: `_split`'s default separator changed from `", "` to `","` to match the plain-comma convention used elsewhere in learning_resources/etl (ocw.py, podcast.py, sloan.py). Verified against the actual ol-data-platform dbt views (they `array_join` on `", "`) that this is a no-op for real data, since each part is `.strip()`'d either way — but the plain comma is also correct if a view ever emits unspaced commas.

learning_resources/lib/warehouse.py
```diff
@@ -0,0 +1,254 @@
...
+    from trino.auth import BasicAuthentication
+    from trino.dbapi import connect
+
+    return connect(
+        host=settings.TRINO_HOST,
+        port=settings.TRINO_PORT,
+        user=settings.TRINO_USER,
+        auth=BasicAuthentication(settings.TRINO_USER, settings.TRINO_PASSWORD),
+        catalog=settings.TRINO_CATALOG,
+    )
```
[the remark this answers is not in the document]
blarghmatey: Fixed in 0422d0e9c: `_connect_trino` now raises `ImproperlyConfigured` with a clear message when `TRINO_HOST`/`TRINO_USER` are unset, instead of building `BasicAuthentication(None, None)` and letting `connect()` fail confusingly. Auth is now `None` (not a garbage-credential object) when user/password aren't both set.

learning_resources/tasks.py
```diff
@@ -810,3 +815,188 @@ def cleanup_deleted_content_files():
...
+# Each task pulls its integrations__learn__* view from the OL Data
+# Platform warehouse (Trino today; see learning_resources.lib.warehouse for
+# the backend abstraction), transforms rows with
+# learning_resources.etl.catalog_sources, and upserts through the same
+# loaders.load_courses/load_programs used by the API-based ETL pipelines
+# above. Program tasks fetch_only their child courses, so a program task
+# should be scheduled after its courses task has run at least once.
+# ---------------------------------------------------------------------------
+
+_INTEGRATIONS_CATALOG = "ol_warehouse_production.integrations"
```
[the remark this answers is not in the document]
blarghmatey: Fixed in 0422d0e9c: verified against the installed `trino` client that `connect(catalog=...)` is threaded through `ClientSession`, so 2-part `schema.table` references resolve against the connection's default catalog. `view_name`s are now schema-only (`integrations.integrations__learn__*`), renamed `_INTEGRATIONS_CATALOG` -> `_INTEGRATIONS_SCHEMA` — `settings.TRINO_CATALOG` alone now controls which catalog gets queried.

main/settings_celery.py
```diff
@@ -202,6 +202,53 @@
...
+        # baseline). Each task also supports full_refresh=False for a
+        # cheaper, watermark-based incremental pull — see
+        # learning_resources.lib.warehouse.BaseWarehouseETLTask — for a
+        # future higher-frequency schedule tier once Cohort 1 is validated;
+        # not wired up here yet.
+        "warehouse-sync-mitxonline-courses-every-1-days": {
+            "task": "learning_resources.tasks.SyncMITxOnlineCoursesTask",
+            "schedule": crontab(minute=0, hour=10),  # 6:00am EST
+            "kwargs": {"full_refresh": True},
+        },
```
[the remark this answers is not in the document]
blarghmatey: Fixed in 0422d0e9c: corrected to "6:00am EDT / 5:00am EST".

main/settings_celery.py
```diff
@@ -202,6 +202,53 @@
                 minute=0, hour=4
             ),  # 04:00 UTC (midnight ET during DST, 11pm ET during standard time)
         },
+        # Cohort 1 warehouse-pull catalog sources (learning_resources.tasks.Sync*Task).
+        # Scheduled to run in parallel with the existing API-based ETL tasks
+        # above during validation; see implementation_guide_01_db_catalog.md.
+        # Programs run 15 minutes after their courses task so fetch_only child
+        # course lookups find freshly-synced course resources.
```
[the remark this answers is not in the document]
blarghmatey: Fixed in 0422d0e9c: the 7 warehouse-sync beat entries are now gated behind `TRINO_HOST` being configured (read directly in settings_celery.py, since it loads before settings_course_etl per main/settings.py's import order — so it can't just reference the settings_course_etl-defined TRINO_HOST constant). Verified empirically via a direct settings import: 0 entries registered without TRINO_HOST set, 7 with it set.

learning_resources/etl/catalog_sources.py
```diff
@@ -0,0 +1,345 @@
...
+    return str(value).strip().lower() == "true"
+
+
+def _parse_datetime(value):
+    if not value:
+        return None
+    try:
+        return parse(value)
+    except (ValueError, TypeError, OverflowError):
+        return None
```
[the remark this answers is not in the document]
blarghmatey: Fixed in 0422d0e9c: `_parse_datetime` now normalizes to UTC via `.replace(tzinfo=UTC)`, matching the identical existing convention in learning_resources/etl/mitxonline.py and xpro.py.

learning_resources/etl/catalog_sources.py
```diff
@@ -0,0 +1,356 @@
...
+        "courses": [
+            _course_stub(course_id, PlatformType.edx.name)
+            for course_id in _split(row.get("courses"))
+        ],
+        "runs": [_program_run(row, title)],
+        "published": _parse_bool(row.get("published")),
+        "image": _image(row.get("image_url")),
+        "url": row.get("url"),
+        "description": clean_data(row.get("description")),
+    }
```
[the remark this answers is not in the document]
blarghmatey: Confirmed and fixed in 131b156fd. Traced it through loaders.py: `load_program` does `topics_data = program_data.pop("topics", [])` then unconditionally calls `load_topics`, which does `resource.topics.set(topics)` whenever `topics_data is not None" — so a missing key defaults to `[]`, which clears topics every run, exactly as flagged. integrations__learn__micromasters_programs has no topics column at all (unlike every other Cohort 1 view), so there's nothing to populate it with yet — the fix sets `"topics": None` explicitly, which is loaders.py's own existing sentinel for "not provided, leave alone" (the same None-vs-[] distinction already governs `departments_data = program_data.pop("departments", None)` a few lines below). Added a regression test asserting `topics is None`, not omitted or `[]`.

learning_resources/etl/catalog_sources.py
```diff
@@ -0,0 +1,365 @@
...
+# XPRO_PLATFORM_TRANSFORM in learning_resources.etl.xpro); course/program
+# rows for the same xPRO program share one platform.
+XPRO_PLATFORM_TRANSFORM = {
+    "Emeritus": PlatformType.emeritus.name,
+    "Global Alumni": PlatformType.globalalumni.name,
+    "Simplilearn": PlatformType.simplilearn.name,
+    "Susskind": PlatformType.susskind.name,
+    "WHU": PlatformType.whu.name,
+    "xPRO": PlatformType.xpro.name,
+}
```
mbertrand: Move to `constants.py` alongside other dicts?

learning_resources/etl/catalog_sources.py
```diff
@@ -0,0 +1,365 @@
...
+            continue
+        runs.append(
+            {
+                "run_id": run_id,
+                "title": title,
+                "published": _parse_bool(is_live),
+                "start_date": _parse_datetime(start_on),
+                "end_date": _parse_datetime(end_on),
+                "instructors": instructors,
+                "prices": [],
```
mbertrand: `load_run` unconditionally overwrites `run.prices` (loaders.py:327-328) and `load_prices(run, [])` clears `resource_prices` — omitting the key doesn't help. Since these tasks upsert the same rows as the API ETL (same etl_source + readable_id), every warehouse sync will wipe displayed prices for mitxonline/xpro/mit_edx runs until the next API run restores them — a daily, user-visible flap once the beat schedule goes live. Consider making parallel validation read-only against prod rows, or adding prices to the views first.

learning_resources/etl/catalog_sources.py
```diff
@@ -0,0 +1,365 @@
...
+def transform_micromasters_program(row: dict) -> dict:
+    """Transform an integrations__learn__micromasters_programs row.
+
+    MicroMasters courses are hosted on edX; child course readable_ids are
+    the edX course keys, so program->course linking matches on
+    ``PlatformType.edx``, same as the API-based MicroMasters ETL.
+    """
+    title = row["title"]
+    return {
+        "readable_id": row["readable_id"],
```
mbertrand: The view emits `cast(program_id as varchar)` (`"3"`, `"4"`), but the API ETL synthesizes `micromasters-program-{id}` (micromasters.py:126) — prod currently has `micromasters-program-3/4` (verified via the public API). Passing the bare id through creates duplicate programs, and the prune step then unpublishes the real ones; the next API run reverses it. Prefix here (and use the same prefixed value for `run_id`, which the API ETL also prefixes — micromasters.py:137), or fix the view in ol-data-platform.

learning_resources/lib/warehouse.py
```diff
@@ -0,0 +1,278 @@
...
+                message=f"{self.name} failed",
+                data={"view_name": self.view_name, "full_refresh": full_refresh},
+                level="error",
+            )
+            raise
+        finally:
+            conn.close()
+
+        if not full_refresh:
+            self._set_watermark(datetime.now(tz=UTC))
```
mbertrand: The watermark is stamped *after* the fetch completes, so rows modified between query execution and completion fall outside both this pull and the next incremental window (skipped until a full refresh heals them). Capture the timestamp before executing the query, or use the max `last_modified` seen. Not urgent while incremental is unscheduled, but worth fixing before that tier goes live.

learning_resources/etl/catalog_sources.py
```diff
@@ -0,0 +1,365 @@
...
+    per-part ``.strip()`` makes the two equivalent for well-formed input,
+    and the plain comma is also correct if a view ever emits "a,b" without
+    the space.
+    """
+    if not value:
+        return []
+    return [part.strip() for part in value.split(sep) if part.strip()]
+
+
+def _parse_bool(value) -> bool:
```
mbertrand, line 51: `_parse_bool(1)` returns `False`. Trino returns real bools, but StarRocks (MySQL protocol) returns `1`/`0` — the migration this module explicitly plans for — which would silently unpublish everything. Suggest: `if isinstance(value, (bool, int)): return bool(value)`.

learning_resources/etl/catalog_sources.py
```diff
@@ -0,0 +1,365 @@
...
+            ),
+        },
+        "runs": [
+            {
+                "run_id": readable_id,
+                "title": title,
+                "published": published,
+                "url": row.get("url"),
+                "semester": row.get("term"),
+                "year": row.get("year"),
```
mbertrand: `year` lands in an `IntegerField` — `"2025"` coerces fine, but an empty string would raise on save. Cheap to guard here, e.g. `int(row["year"]) if row.get("year") else None`.

learning_resources/lib/warehouse.py
```diff
@@ -0,0 +1,278 @@
...
+            batch = cur.fetchmany(batch_size)
+            if not batch:
+                break
+            for row in batch:
+                yield dict(zip(columns, row))
+    finally:
+        cur.close()
+
+
+class BaseWarehouseETLTask(Task):
```
mbertrand, line 181: `BaseWarehouseETLTask` doesn't set `acks_late=True`. The other long-running ETL tasks in this module do (e.g. `get_ocw_data`, `ingest_edx_run_archive`, `import_*_files`). A worker lost mid-pull loses the work and the message is already acked. Low-impact today (daily `full_refresh=True` is idempotent/self-healing, and a failed incremental leaves the watermark untouched so the next run re-covers the gap), but `acks_late=True` on the base would match the rest of the ETL task fleet. Optional.

learning_resources/etl/catalog_sources.py
```diff
@@ -0,0 +1,365 @@
...
+        "description": clean_data(row.get("description")),
+    }
+
+
+# ---------------------------------------------------------------------------
+# MicroMasters
+# ---------------------------------------------------------------------------
+
+
+def transform_micromasters_program(row: dict) -> dict:
```
mbertrand: Distinct from the prices-wipe and `readable_id` issues already flagged: this transform also omits `certification`, `certification_type`, `start_date`/`end_date`, `availability`, and `pace`, all of which the API ETL sets (micromasters.py:133-149). Those are non-destructive — loaders only write keys present in the dict — but post-cutover they'd freeze at their last API-written values. Worth recording in the cutover plan: extend the view with these columns, or explicitly keep the API path as owner of them.

`instructors` is the exception, and it's destructive: `_program_run` provides no `instructors` key, so `load_run` pops the default `[]` and `load_instructors(run, [])` deletes every `RunInstructorRelationship` for the run (loaders.py:224). Program-run instructors will be wiped on every warehouse sync — same class of issue as prices, and this applies to the mitxonline/xpro program runs too.

mbertrand (changes requested):

learning_resources/etl/catalog_sources.py
blarghmatey: https://github.com/mitodl/mit-learn/pull/3565 is aimed at handling this shift in responsibilities as we cut over different data flows.

learning_resources/etl/catalog_sources.py
blarghmatey: Fixed in e005b7d5d: moved `XPRO_PLATFORM_TRANSFORM` into `learning_resources/etl/constants.py` and import it from both `xpro.py` and `catalog_sources.py` instead of maintaining two copies of the same dict.

learning_resources/etl/catalog_sources.py
blarghmatey: Fixed in e005b7d5d: `transform_micromasters_program` now prefixes `readable_id`/`run_id` with `learning_resources.etl.micromasters.READABLE_ID_PREFIX` ("micromasters-program-"), matching the API-based ETL. Added a regression test asserting the prefix is applied.

learning_resources/lib/warehouse.py
blarghmatey: Fixed in e005b7d5d: the watermark is now captured immediately before `fetch_and_upsert` is called, not after it returns, so a row modified mid-fetch is still covered by the next incremental window. Added a freezegun-based regression test that advances the clock inside `fetch_and_upsert` and asserts the stored watermark is the pre-fetch time.

learning_resources/etl/catalog_sources.py
blarghmatey: Fixed in e005b7d5d: `_parse_bool` now treats `bool | int` as native booleans (`bool(value)`), so StarRocks-style 1/0 BOOLEAN columns parse correctly instead of silently becoming `False`. Added `(1, True)`/`(0, False)` cases to the parametrized test.

learning_resources/etl/catalog_sources.py
blarghmatey: Fixed in e005b7d5d: `year` now coerces via `int(row["year"]) if row.get("year") else None`, so an empty string becomes `None` instead of raising on save. Added a regression test.

learning_resources/lib/warehouse.py
blarghmatey: Fixed in e005b7d5d: added `acks_late = True` to `BaseWarehouseETLTask`, matching the rest of the ETL task fleet.

learning_resources/etl/catalog_sources.py
blarghmatey: Fixed directly in e005b7d5d, independent of #3565 (that PR's pull-vs-push ownership guard can't actually arbitrate this: both the API-based ETL and this warehouse-pull task pass the identical `etl_source` value and are both "pull" from #3565's guard's point of view, so it can't tell them apart or protect one from the other — see the PR-level note on merge order).

The actual fix: `_parse_runs` now sets `"prices": None` instead of `[]`, and `loaders.load_run`/`load_prices` (and `load_instructors`) now treat `None` as "not provided by this source, leave the existing value alone" — the same sentinel convention `load_topics` already uses for `topics_data`. An omitted key still defaults to `[]` (unchanged behavior for every existing caller, all of which always pass real data), so this is additive, not a behavior change for the current API-based pipelines.

learning_resources/etl/catalog_sources.py
blarghmatey: The instructors-wipe half is fixed in e005b7d5d: `_program_run` now explicitly sets `"instructors": None` (and `"prices": None`), which `loaders.load_run`'s new None-sentinel treats as "leave alone" rather than clearing — this now applies to mitxonline/xpro/micromasters program runs alike, as you flagged.

The certification/certification_type/dates/availability/pace gap is left as-is per your note — those keys are simply omitted from the transform dicts (not set to `[]`/falsy), and `upsert_course_or_program`/`load_program` only write keys present in the dict, so they freeze at whatever the API-based ETL last set rather than getting clobbered. Non-destructive, as you said; recording it here as the known gap to close (extend the views, or keep the API path authoritative for these fields) before cutover.

learning_resources/lib/warehouse.py
```diff
@@ -0,0 +1,278 @@
...
+        # input, so a formatted literal is safe here (no parameterization
+        # needed) — and `TIMESTAMP '...'` literal syntax is understood by
+        # Trino, StarRocks/MySQL, and DuckDB alike.
+        watermark = since.strftime(_WATERMARK_SQL_FORMAT)[:-3]
+        query += f" WHERE last_modified > TIMESTAMP '{watermark}'"
+
+    cur = conn.cursor()
+    try:
+        cur.execute(query)
+        columns = [d[0] for d in cur.description]
```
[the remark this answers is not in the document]
blarghmatey: Fixed in e005b7d5d: `iter_rows` now reads `cur.description` from the first fetched batch rather than immediately after `execute()`, since PEP 249 leaves the timing of when a driver populates `description` up to the driver. Added a regression test with a mock cursor where `description` is `None` until after the first `fetchmany` call. (Traced through the installed `trino` client's `dbapi.py`: Trino itself populates it synchronously inside `execute()`, so this wasn't an active bug for the Trino backend today, but the fix costs nothing and removes the assumption for the planned StarRocks backend.)

blarghmatey:
Pushed `e005b7d5d` addressing the outstanding review feedback (individual replies on each thread above):

- **Prices/instructors wipe during parallel validation** (@mbertrand): fixed by making `loaders.load_run`/`load_instructors`/`load_prices` treat `None` as "not provided, leave existing value alone" — the same sentinel `load_topics` already uses for `topics_data`. The warehouse-pull transforms now pass `None` for prices (course & program runs) and instructors (program runs), instead of `[]`.
- **MicroMasters `readable_id`/`run_id` mismatch** (duplicate programs + unpublish flapping): fixed — now prefixed with `micromasters-program-`, matching the API-based ETL.
- `XPRO_PLATFORM_TRANSFORM` de-duplicated into `etl/constants.py`.
- `_parse_bool` int handling, OCW `year` empty-string guard, `acks_late=True`, watermark-before-fetch, and the `cur.description` timing all fixed — see individual thread replies for detail.
- All touched suites pass locally (`catalog_sources`, `warehouse`, `loaders`, `xpro`, `tasks`, `pipelines`): 488 passed.

**On merging #3565 into this PR**: I don't think it should be merged in, and I don't think it actually covers the parallel-validation risk that prompted the "aimed at handling this shift in responsibilities" comment on the prices thread.

#3565's `ETLSourceOwnership` guard is a binary pull/push switch keyed on `(etl_source, resource_type)`. It's built for the case where exactly one pipeline is authoritative at a time and you flip a flag at cutover (e.g. SEE/Sloan moving from a Celery pull to a Dagster push asset). It can't arbitrate between *two simultaneous pull writers* of the same `etl_source` — which is exactly this PR's situation during the 2-week parallel-validation window: the existing API-based ETL and these new warehouse-pull tasks both pass e.g. `ETLSource.mitxonline.name` into `load_courses`/`load_programs`, and #3565's guard sees them as indistinguishable ("pull" either way). Setting that pair to `push` would block *both* callers, not just one.

So the two are complementary but solve different problems, and neither blocks the other at the code level (verified: no line-level overlap — #3565 touches the top of `load_courses`/`load_programs` and `sync_canvas_courses`; this PR's fixes touch `load_run`/`load_instructors`/`load_prices`). #3565 looks close to mergeable on its own merits (the CHANGES_REQUESTED-adjacent bot findings there are already resolved) and is worth landing independently for the sources it's actually meant to protect. It just isn't the fix for this PR's validation-window risk — the prices/instructors/readable_id fixes above are.

main/settings_celery.py
```diff
@@ -205,6 +205,66 @@
...
+                "kwargs": {"full_refresh": True},
+            },
+            "warehouse-sync-micromasters-programs-every-1-days": {
+                "task": "learning_resources.tasks.SyncMicromastersProgramsTask",
+                "schedule": crontab(minute=30, hour=11),
+                "kwargs": {"full_refresh": True},
+            },
+        }
+    )
+
```
mbertrand, line 268: If `TRINO_HOST` is set, maybe these old ETL scheduled tasks should not be enabled?

```
        "update_edx-courses-every-1-days": {
            "task": "learning_resources.tasks.get_mit_edx_data",
            "schedule": crontab(minute=0, hour=5),
        },
        "update-micromasters-programs-every-1-days": {
            "task": "learning_resources.tasks.get_micromasters_data",
            "schedule": crontab(minute=0, hour=5),  # 1:00am EST
        },
        "update-mitxonline-courses-every-6-hours": {
            "task": "learning_resources.tasks.get_mitxonline_data",
            "schedule": crontab(minute=0, hour="*/6"),
        },
        "update-xpro-courses-every-1-days": {
            "task": "learning_resources.tasks.get_xpro_data",
            "schedule": crontab(minute=0, hour=5),  # 1:00am EST
        },
```

main/settings_celery.py
```diff
@@ -205,6 +205,66 @@
...
+            "warehouse-sync-mit-edx-courses-every-1-days": {
+                "task": "learning_resources.tasks.SyncMITEdXCoursesTask",
+                "schedule": crontab(minute=0, hour=11),
+                "kwargs": {"full_refresh": True},
+            },
+            "warehouse-sync-ocw-courses-every-1-days": {
+                "task": "learning_resources.tasks.SyncOCWCoursesTask",
+                "schedule": crontab(minute=15, hour=11),
+                "kwargs": {"full_refresh": True},
+            },
```
mbertrand, line 265: This is currently handled via webhooks - is this meant to replace those webhooks or complement them (ie fill any missing gaps caused by webhooks that failed to be delivered or processed for whatever reason)?

mbertrand (commented):
Looks great, just a couple more questions.

main/settings_celery.py
mbertrand: Another question - @pdpinch was wondering if it might be possible to send webhooks from data-platform instead whenever there is a new or updated course/program that should be processed from these sources.  And is data-platform getting this data from the source API's (like mitlearn currently does) or from the source databases? 

Also, the manner in which mitxonline courses are ingested is changing right now so that B2B/invariant course runs are ingested (currently they are not) - https://github.com/mitodl/mit-learn/pull/3543

main/settings_celery.py
blarghmatey: The intent is to replace the webhooks, though it does create a tradeoff of frequency/timeliness. This is largely an opportunistic body of work so there may be some additional design work necessary before it all lands. If timeliness is a core requirement for OCW data then we can remove this source from the current PR

learning_resources/etl/catalog_sources.py
blarghmatey: Verified on the branch head: `XPRO_PLATFORM_TRANSFORM` now lives in `learning_resources/etl/constants.py:27` and is imported by both `xpro.py:19` and `catalog_sources.py:27`. Resolving.

learning_resources/etl/catalog_sources.py
blarghmatey: Verified on the branch head: `catalog_sources.py:108` passes `"prices": None` and `loaders.py:246`/`loaders.py:335` short-circuit on the `None` sentinel, so a warehouse sync no longer clears `resource_prices`. Resolving.

learning_resources/etl/catalog_sources.py
blarghmatey: Verified on the branch head: `catalog_sources.py:353` builds `readable_id` as `f"{MICROMASTERS_READABLE_ID_PREFIX}{row[readable_id]}"`, matching the API ETL, with the same prefixed value used for `run_id`. Resolving.

learning_resources/lib/warehouse.py
blarghmatey: Verified on the branch head: `warehouse.py` captures `fetch_started_at` before calling `fetch_and_upsert` (line 234 comment explains why) and stamps that value at line 256 on success, so rows modified mid-fetch stay inside the next incremental window. Resolving.

learning_resources/etl/catalog_sources.py
blarghmatey: Verified on the branch head: `_parse_bool` now short-circuits on `isinstance(value, bool | int)` and returns `bool(value)`, so StarRocks 1/0 BOOLEAN columns parse correctly. Resolving.

learning_resources/etl/catalog_sources.py
blarghmatey: Verified on the branch head: `catalog_sources.py:324` reads `int(row["year"]) if row.get("year") else None`. Resolving.

learning_resources/lib/warehouse.py
blarghmatey: Verified on the branch head: `acks_late = True` is set on `BaseWarehouseETLTask` (`warehouse.py:222`). Resolving.

main/settings_celery.py
blarghmatey: Added in c087ba981 — but keyed on its own setting rather than `TRINO_HOST`, because gating on `TRINO_HOST` would break the thing it's meant to protect: during the 2-week parallel-validation window `TRINO_HOST` *is* set and both pipelines have to run, since the API-based ETL is the comparison baseline. Cutover is also staged per source rather than all-at-once, which a single host check can't express.

So: `WAREHOUSE_ETL_CUTOVER_SOURCES`, a comma-separated list of `ETLSource` names. Listing a source pops its API-based beat entry:

| source | entry removed |
|---|---|
| `micromasters` | `update-micromasters-programs-every-1-days` |
| `mit_edx` | `update_edx-courses-every-1-days` |
| `mitxonline` | `update-mitxonline-courses-every-6-hours` |
| `xpro` | `update-xpro-courses-every-1-days` |

Two deliberate exclusions: the `import_all_*_files` tasks stay scheduled at every setting, since the `integrations__learn__*` views carry catalog metadata only and not content files; and an unrecognized source name raises `ImproperlyConfigured` at settings load rather than being ignored, because the silent-failure direction here is two pipelines both writing the same rows after a believed cutover. Tests in `main/settings_test.py`.

---

**On webhooks from data-platform** (@pdpinch's question): worth pursuing, and a better fit than polling for the sources where timeliness matters — but it's a data-platform-side change (an emit-on-materialize hook per `integrations__learn__*` model) rather than something this PR can do, so I'd rather land the pull tasks as the self-healing baseline and treat push as the follow-on. Happy to write that up as a separate issue against ol-data-platform.

**Where data-platform gets this data:** source databases, not the source APIs — with one exception. Tracing the views' lineage:

- `integrations__learn__mitxonline_courses` → `int__mitxonline__courses` → `stg__mitxonline__app__postgres__courses_course` / `cms_coursepage`
- `integrations__learn__xpro_courses` → `int__mitxpro__courses` → `stg__mitxpro__app__postgres__courses_course` / `cms_coursepage`
- `integrations__learn__ocw_courses` → `int__ocw__courses` → `stg__ocw__studio__postgres__websites_website` / `websitecontent`
- `integrations__learn__mit_edx_courses` → `stg__edxorg__api__course` / `courserun` — **API-sourced**, since there's no DB replica of edx.org

That's a meaningful difference from the current MIT Learn ETL, which reads the public APIs for all of them: for mitxonline/xpro/OCW the warehouse path sees the application's own tables, so it isn't bounded by what the API chooses to serialize.

**On #3543 (B2B/invariant runs):** the warehouse path already carries them. `int__mitxonline__course_runs` applies no `b2b_contract_id` filter, and the integrations view aggregates every run row for a course, so B2B runs are included without a view change. `stg__mitxonline__app__postgres__courses_courserun` also exposes `b2b_contract_id` directly, so if MIT Learn ends up needing to *distinguish* B2B runs rather than just ingest them, that's an available column rather than new plumbing. Since #3543 is merged, that's something parallel validation should explicitly check for mitxonline.

Leaving this thread open for the webhook question rather than resolving it.

main/settings_celery.py
blarghmatey: Following up with the concrete freshness tradeoff, since my earlier answer was hand-wavy about it.

`integrations__learn__ocw_courses` is built from ocw-studio's **Postgres replica** (`int__ocw__courses` → `stg__ocw__studio__postgres__websites_website` / `websitecontent` / `websitestarter`), not from any OCW API. So the warehouse path sees the same data ocw-studio itself writes — the gap isn't fidelity, it's latency: a publish becomes visible to MIT Learn only after the raw ingest lands and the dbt models materialize, then after the next daily beat run, versus a webhook that fires on publish.

That makes this the one Cohort 1 source where replacing the existing mechanism is a strict downgrade in timeliness, since OCW is the only one of the seven with no API-based beat entry at all today — it's webhook-only. For the other six the warehouse task replaces a daily/6-hourly poll, so the latency change is marginal or favorable; here it's poll-instead-of-push.

Two options, and I don't think this is mine to pick:

1. **Drop `SyncOCWCoursesTask` from this PR** — webhooks stay authoritative, and OCW comes back later as a reconciliation pass (a low-frequency `full_refresh` that heals dropped webhook deliveries without owning the fast path). This is the conservative one.
2. **Keep it as-is** for parallel validation only, and treat webhook replacement as an explicitly separate decision at cutover time — which is now enforceable, since after c087ba981 cutover is a per-source setting rather than an implicit consequence of merging. Note OCW has no entry in that map precisely because it has no beat task to remove; cutting OCW over means disabling the webhook, which is a deliberate separate act.

I lean toward (1) if you think publish-to-visible latency is user-visible for OCW, and (2) if the daily cadence is acceptable during validation. Your call — happy to strip the task out of this PR if you prefer the first.

learning_resources/etl/catalog_sources.py
blarghmatey: Closing this out — the code it points at no longer exists.

`main` removed the MicroMasters ETL entirely in 5cd79dfb1 (#3663) after this review: the module, the `ETLSource.micromasters` enum member, the beat entry, and migration `0117_unpublish_micromasters_resources`, which unpublishes and deindexes every `etl_source="micromasters"` resource. `SyncMicromastersProgramsTask` would have republished exactly what that migration retired, so I've dropped the task, `transform_micromasters_program`, and its beat entry from this PR (7769ba2e1), and rebased onto main.

That resolves the destructive half here by deletion rather than by fix.

The non-destructive half you flagged still applies to the **mitxonline and xpro program runs**, which remain in this PR. `_program_run` sets only `run_id`/`title`/`published`/`url` (plus the `None` sentinels for instructors/prices), so `certification`, `certification_type`, `start_date`/`end_date`, `availability` and `pace` are omitted.

One correction to my earlier "they freeze at their last API-written values" framing — that's right for rows that already exist, since `load_run` does `update_or_create(defaults={**run_data})` and only writes keys present in the dict. But for a resource the warehouse path creates *first*, those fields get Django model defaults rather than API values: `certification` is `models.BooleanField(default=False)`, `availability` is null, `pace` empty, dates null. During parallel validation the API ETL backfills within hours so it's invisible; post-cutover nothing backfills it. There's also a knock-on — `_resolve_run_prices` forces `prices = []` when `learning_resource.certification is False`, so a warehouse-created resource would suppress prices even once a source supplies them.

Recording it here as the pre-cutover gap for mitxonline/xpro: either extend the views with these columns, or keep the API path authoritative for them. The per-source `WAREHOUSE_ETL_CUTOVER_SOURCES` switch added in c59ee02e7 means that's now an explicit decision per source rather than an implicit consequence of merging.

blarghmatey: Tracked as https://github.com/mitodl/mit-learn/issues/3746 (scoped to mitxonline/xpro program runs, gated as a pre-cutover blocker on `WAREHOUSE_ETL_CUTOVER_SOURCES`). Resolving here.

blarghmatey:
Rebased onto `main` (was 163 commits behind) and made two substantive changes. All checks green.

**MicroMasters warehouse task removed.** `main` retired the MicroMasters ETL entirely in #3663 after this review — the module, the `ETLSource.micromasters` enum member, the beat entry, and migration `0117_unpublish_micromasters_resources`, which unpublishes and deindexes every `etl_source="micromasters"` resource. `SyncMicromastersProgramsTask` would have republished exactly what that migration retired, so the task, `transform_micromasters_program`, its tests, and its beat entry are gone. This PR now covers 6 tasks across 4 sources, not 7 across 5.

**Per-source cutover switch added** (@mbertrand's question on the legacy beat entries): `WAREHOUSE_ETL_CUTOVER_SOURCES`, a comma-separated list of `ETLSource` names, where listing a source removes its API-based beat entry. Not keyed on `TRINO_HOST` as suggested, because during parallel validation `TRINO_HOST` is set and both pipelines must run — the API path is the comparison baseline, so that gate would disable the thing validation depends on. Details and the source→entry map are in the thread.

**Thread status:** 8 resolved, 2 left open for you —

- **Legacy beat entries / webhooks**: the gate is implemented, but @pdpinch's webhook question is still open. Short version: the views build on source Postgres replicas (`stg__mitxonline__app__postgres__*`, `stg__mitxpro__app__postgres__*`, `stg__ocw__studio__postgres__*`), not the source APIs — `mit_edx` is the exception, deriving from `stg__edxorg__api__*`. Emit-on-materialize webhooks from the platform are worth doing but are a data-platform-side change; happy to open an issue against ol-data-platform.
- **OCW**: still your call whether to drop `SyncOCWCoursesTask` from this PR. OCW is the only Cohort 1 source that's webhook-only today, so it's the one case where this is a strict timeliness downgrade rather than a swap for an existing poll.

**Follow-up filed:** https://github.com/mitodl/mit-learn/issues/3746 — the program-run field gap you flagged (`certification`, `certification_type`, dates, `availability`, `pace`), now scoped to mitxonline/xpro since the MicroMasters instance is moot. One correction to what I said earlier: it's non-destructive for *existing* rows, but a resource the warehouse path creates first gets model defaults rather than API values (`certification` defaults to `False`), and `_resolve_run_prices` suppresses prices when `certification is False`. Gated as a pre-cutover blocker on `WAREHOUSE_ETL_CUTOVER_SOURCES`.

Also relevant to validation: `int__mitxonline__course_runs` applies no `b2b_contract_id` filter, so the warehouse path already carries the B2B/invariant runs that #3543 added on the API side.

blarghmatey:
Closing in favor of a StarRocks-targeted stacked PR series — the org has decided to route new cross-VPC warehouse access through StarRocks rather than Trino/Starburst (Trino usage is being drawn down), and Trino network access from MIT Learn was never actually wired up (blocked on `hq#11509`/`hq#4027` the whole time this PR was open).

Replacement stack:
- #3807 — StarRocks warehouse-pull machinery (carries forward this PR's `BaseWarehouseETLTask`/`iter_rows` design and the `loaders.py` None-sentinel fix, retargeted to StarRocks)
- MicroMasters certificate ingest (resolves `hq#12954`) — stacked on #3807, not yet opened
- mitxonline, xpro, mit_edx, ocw courses/programs — stacked follow-ons reusing this PR's per-source transform logic, not yet opened

Tracked in witan: `tk-cohort-1-db-backed-catalog-sources-trino-pull-e15c82` and children.

PR #3576, opened 2026-07-03
feat: Podcasts Listing Page

ahtesham-quraish:
https://github.com/mitodl/hq/issues/12147

As a user, I want a central place to view podcasts from MIT, including new podcasts, recommended podcasts, and be able to view all podcasts.

### Desktop

image

### Mobile

image

### How can this be tested?
For testing you need to visit the following URL `http://open.odl.local:8062/podcasts` it will load the page, for featured podcasts you need to create the learningPath list from the following urls `http://open.odl.local:8062/learningpaths` and add few podcasts from search page and then set the id of this newly created list in frontend env `NEXT_PUBLIC_PODCASTS_FEATURED_LIST_LEARNINGPATH_ID=<learningPath_id>` it will bring the featured Podcasts from this list.


frontends/main/src/app/(site)/podcasts/page.tsx
```diff
@@ -0,0 +1,14 @@
+import React from "react"
+import { Metadata } from "next"
+import { standardizeMetadata } from "@/common/metadata"
+import { PodcastsListingPage } from "@/app-pages/PodcastPage/PodcastsListingPage"
+
+export const metadata: Metadata = standardizeMetadata({
+  title: "MIT Learn | Podcasts",
```
ChristopherChudzicki: Write now the title is `MIT Learn | Podcasts | MIT Learn`... Should just be `"Podcasts"` here.  The site name gets appended automatically.

frontends/main/src/app-pages/PodcastPage/PodcastsListingPage/NowPlayingSection.tsx
```diff
@@ -0,0 +1,232 @@
...
+    lineHeight: "18px",
+  },
+}))
+
+const NowPlayingCard = styled.div(({ theme }) => ({
+  display: "flex",
+  gap: "48px",
+  padding: "40px",
+  alignItems: "flex-start",
+  backgroundColor: theme.custom.colors.lightGray1,
```
ChristopherChudzicki: This should be `lightGray0`. Smoot-design actually has it, but in in v6.28; learn is on 6.27

frontends/main/src/app-pages/PodcastPage/PodcastsListingPage/PodcastsListingPage.tsx
```diff
@@ -0,0 +1,187 @@
...
+        if (!audioUrl) return null
+        return {
+          audioUrl,
+          title: playingEpisode.title || "Untitled Episode",
+          podcastName: playingEpisode.offered_by?.name || "Podcast",
+        }
+      })()
+    : null
+
+  useEffect(() => {
```
ChristopherChudzicki: There's a lot of logic near here that is duplicated exactly in three files:
- frontends/main/src/app-pages/PodcastPage/PodcastDetailPage.tsx
- frontends/main/src/app-pages/PodcastPage/PodcastListingPage.tsx
- frontends/main/src/app-pages/PodcastPage/PodcastEpisodeDetailPage.tsx

All three files have:
- `useEffect` setting/unsetting `--mit-player-height`
- an IIFE calculating `currentTrack`
- `handlePlayClick`, `handlePauseClick`


Let's move all three to a shared hook:

```ts
const usePodcastPlayer = (playerRef, isMobile) => {
  useEffect(/* CSS effect for --mit-player-height */)
  const toggle = (epside, podcastName) => {
    /* single callback for play/pause toggling; simplifies consumers */
  } 
  const currentTrack = etc

  return {  toggle, currentTrack }
}
```

NOTE:
- some places use play/pause/toggle, but I think they could all just use toggle
- currentTrack is a little tricky because of the parent podcast title issue... I think we should resolve that first (possibly outside this pr)

frontends/main/src/app-pages/PodcastPage/PodcastsListingPage/PodcastsListingPage.test.tsx
```diff
@@ -0,0 +1,265 @@
...
+import {
+  EPISODES_PAGE_SIZE,
+  SERIES_FEATURED_COUNT,
+  SERIES_MORE_COUNT,
+} from "./constants"
+import { PodcastsListingPage } from "./PodcastsListingPage"
+
+const ENV_KEY = "NEXT_PUBLIC_PODCASTS_FEATURED_LIST_LEARNINGPATH_ID"
+
+jest.mock("../PodcastPlayer", () => {
```
ChristopherChudzicki: This is a pretty heavy / complex mock; we are using it (we check play / pause in this testfile). Howveer, both `PodcastDetailPage.test.tsx` and `PodcastEpisodeDetailPage.test.tsx` mock `./PodcastPlayer` similarly. Let's have a shared `mockPodcastPlayer` to (A) reuse and (B) isolate the complexity of this mock. It could live aside `PodcastPlayer` as `PodcastPlayer.test-utils.tsx ` perhaps.

frontends/main/src/app-pages/PodcastPage/PodcastsListingPage/EpisodeItem.tsx
ChristopherChudzicki, line 1: - PodcastDetailPage defines `EpisodeItem`, and then it's used again by `PodcastEpisodeDetailPage.tsx`
- Then `PodcastsListingPage.tsx` defines this file

I think we should unify them into a single component:
- They are the same component in figma, though there are some overrides that make styles different
    - In particular, the podcast listing version has 24px vertical padding vs 28px on the other two. This seems like a mistake. We might want to check with @mbilalmughal ... Seems like they should be the same style
    - There is one clearly intentional difference in Figma... whether the podcast title is shown. The podcast listing page is supposed to show the title, podcast detail + episode detail do not, since those are scoped to a single podcast.
        - _note: we don't actually show the podcast title on the new podcast listing page either; Figma shows it, but it's not implemented here.... Same underlying issue as https://github.com/mitodl/mit-learn/pull/3576/changes#r3540458476_

**Additionally:**  Both implementations have some accessibility issues:
-  the EpisodeItem is an anchor tag, but expose to screenreaders as a list item via `role="listitem"` override. Screenreader users won't know it's a link.
- the anchor tag + play button are nested ... Not supposed to nest interactive elements

I don't think we need to fix those a11y issues in _this pr_ (we should fix them, though). But we should not duplicate the EpisodeItem implementations—then we'd have to fix it in two places.

**In summary:** Unify now, address a11y issue later in a single place.

So, let's reuse the existing EpisodeItem!

frontends/main/src/app-pages/PodcastPage/PodcastsListingPage/HeroSection.tsx
```diff
@@ -0,0 +1,75 @@
...
+}
+
+const HeroSection: React.FC<HeroSectionProps> = ({
+  totalSeries,
+  totalEpisodes,
+}) => {
+  return (
+    <HeroSectionWrap>
+      <PodcastContainer>
+        <HeroHeading variant="h1">Podcasts from across MIT</HeroHeading>
```
ChristopherChudzicki: **Accessibility Issue:** The podcast listing page currently has no heading elements!

This should be `component="h1"`

Additionally, I'd recommend:

```
A... h1 = Podcasts from across MIT
B...h2 = Now Playing: <Big Featured Podcast's Title> 
C... h2 = Latest Episodes
D... h2 = Podcasts Across MIT
E... h2 = More Podcasts
```
_Note: ☝️ h1/h2 are about the element, not the visual variant property._

What do you think of the suggested headings above?

frontends/main/src/app-pages/PodcastPage/PodcastsListingPage/EpisodeItem.test.tsx
```diff
@@ -0,0 +1,189 @@
...
+    last_modified: "2024-05-03T00:00:00Z",
+    podcast_episode: {
+      id: 1,
+      podcasts: [1],
+      duration: "PT2H",
+      audio_url: "https://example.com/audio.mp3",
+      episode_link: "https://example.com/link",
+    },
+    ...overrides,
+  }) as unknown as LearningResource
```
ChristopherChudzicki: Not a huge deal, but I would get rid of this, or change it simply to
```ts
const makeEpisode = factories.learningResources.podcastEpisode
```
if you want a shorter name.

- Most of the overrides are unnecessary
- Our factories merge overrides deeply, so if you do need to override an object, you can override just the piece you care about
- In a few places you assert on `"Episode Title"`; Personally, I prefer just to assert on `episode.title`.
- If you need a specific override, you can declare it in the test **that cares about that specific override**
     - This is the main reason to just use `factories.learningResources.podcastEpisode` directly. Unless you're actually sharing / re-using a complicate test case, it's clearer to keep the overrides co-located with the test that actually cares about them.

frontends/main/src/app-pages/PodcastPage/PodcastsListingPage/PodcastsListingPage.tsx
```diff
@@ -0,0 +1,187 @@
...
+  const handlePauseClick = () => playerRef.current?.pause()
+
+  const currentTrack: PodcastTrack | null = playingEpisode
+    ? (() => {
+        const audioUrl = getEpisodeAudioUrl(playingEpisode)
+        if (!audioUrl) return null
+        return {
+          audioUrl,
+          title: playingEpisode.title || "Untitled Episode",
+          podcastName: playingEpisode.offered_by?.name || "Podcast",
```
ChristopherChudzicki: We want this to be the podcast title, not `offered_by`. 
- We shouldn't use `offered_by`. It's not populated, anyway. (You and I both noticed it was `null` on episode responses).
- We can't currently use the podcast title... It's not exposed on the episode API results. 

IMO, we should open a separate PR to get the podcast title exposed on the `resource.podcast_episode` object.

⚠️ The `offered_by` fallback is used elsewhere (e.g., the EpisodeItem card overline value). We should not use that as the fallback value anywhere, imo.

frontends/main/src/app-pages/PodcastPage/PodcastsListingPage/LatestEpisodesSection.tsx
```diff
@@ -0,0 +1,107 @@
...
+  hasMoreEpisodes,
+  isPlayable,
+}) => {
+  return (
+    <Section>
+      <SectionHeader>
+        <SectionTitle variant="subtitle1">Latest Episodes</SectionTitle>
+        <Link color="red" href={SEARCH_PODCAST_EPISODES}>
+          <SectionLink>All episodes</SectionLink>
+        </Link>
```
ChristopherChudzicki: **Request:** The "All Episodes" link should directly use our link from `ol-components`, probably with `variant="red"`. For some reason, Figma just has this as a plain text node:

Image

Plain text is bad, no hover state, plus drift from other links. Confirm with @mbilalmughal, but yeah, this should be a Link component.

**Suggestion:** Review figma first for things that should probably be a component. It's easy to miss things... helps to ask your agent to review also, and scan for places where Figma should be using design tokens but isn't.

ChristopherChudzicki (changes requested):
Thanks for putting this together. The page is close. I've left some comments to help organize the code plus fix some issues. 

My inline comments fall in a few groups:
1. **Unify the duplicated/triplicated player pieces:** a shared [player hook](https://github.com/mitodl/mit-learn/pull/3576/changes#r3539115585), a single [`EpisodeItem`](https://github.com/mitodl/mit-learn/pull/3576/changes#r3539562074), and a shared [`PodcastPlayer` mock](https://github.com/mitodl/mit-learn/pull/3576/changes#r3539124165). (The hook can be extracted now with the current fallback; the podcast-name line swaps to the real title when that lands.)
2. **Accessibility:** [no real headings on the page](https://github.com/mitodl/mit-learn/pull/3576/changes#r3539664716) (fix here). Additionally, there's some a11y issues in `EpisodeItem`, but I think we can fix those in a followup after it's de-duplicated.
3. **Smaller comments:** 
    - [factory defaults over hardcoded overrides](https://github.com/mitodl/mit-learn/pull/3576/changes#r3540421679)
    - [double-appended page title](https://github.com/mitodl/mit-learn/pull/3576/changes#r3537957822)
    - ["All episodes" should use our Link](https://github.com/mitodl/mit-learn/pull/3576/changes#r3540531810)
    - [`lightGray0`](https://github.com/mitodl/mit-learn/pull/3576/changes#r3538447275) (blocked on a smoot bump... Would be nice to do that as a separate PR).
4. **Podcast Title**: [My comment at :109](https://github.com/mitodl/mit-learn/pull/3576/changes#r3540458476). Figma shows each card with an overline naming the **parent podcast** (e.g., Chalk Radio) above the episode title. This PR doesn't render it. More on that below.

**Address Before Merging:** Groups 1, 2, 3 above.

**Unsure:** I'm unsure if we *need* to address 4 before merging. That's a product question for Ferdi/Peter. On the one hand, we could improve it later. On the other hand, **the whole point of "Recent Episodes" is to aggregate episodes across podcasts, so showing the podcast title is important.**

## Podcast title on episode cards (the trickiest)

[My comment at :109](https://github.com/mitodl/mit-learn/pull/3576/changes#r3540458476). Figma shows each card with an overline naming the **parent podcast** (e.g., Chalk Radio) above the episode title. We don't render it today (the overline is wired to `offered_by`, which is the wrong field and null anyway).

Either way I'd fix it by exposing `resource.podcast_episode.podcast_title`. Wrinkle: the model allows multiple podcast parents, but we've agreed we don't want that and afaik other platforms don't either, so expose the **primary** parent's title (matching the `podcasts[0]` we already use for the episode link).

Happy to discuss the title issue more in morning.

frontends/main/src/app-pages/PodcastPage/PodcastsListingPage/HeroSection.tsx
ahtesham-quraish: I have made two headings as h1 and h2 respectively

ChristopherChudzicki:
@ahtesham-quraish One issue I forgot to mention in yesterday's review: Series vs Podcast naming.

Recall that the Figma designs originally used "Series" language, I guess because podcasts are a series of episodes, in analogy to the "Video Series" we have. We abandoned that language because it is highly non-standard: Apple Podcasts, Substack, Spotify, etc, all use "Podcast" language.

The code still uses "Series" through: `featuredSeries`, `PodcastSeriesSection`, `SeriesDescription`... Many dozens of variables / functions / files using Series.

**Request:** Fully remove the "Series" language. The UI uses "Podcasts", the URL uses "podcast", and the internal data models use "podcast". That's what we should use here.

ahtesham-quraish:
> @ahtesham-quraish One issue I forgot to mention in yesterday's review: Series vs Podcast naming.
> 
> Recall that the Figma designs originally used "Series" language, I guess because podcasts are a series of episodes, in analogy to the "Video Series" we have. We abandoned that language because it is highly non-standard: Apple Podcasts, Substack, Spotify, etc, all use "Podcast" language.
> 
> The code still uses "Series" through: `featuredSeries`, `PodcastSeriesSection`, `SeriesDescription`... Many dozens of variables / functions / files using Series.
> 
> **Request:** Fully remove the "Series" language. The UI uses "Podcasts", the URL uses "podcast", and the internal data models use "podcast". That's what we should use here.

I have refactored the code can you please review again ?

frontends/main/src/app-pages/PodcastPage/PodcastsListingPage/NowPlayingSection.test.tsx
```diff
@@ -0,0 +1,133 @@
...
+import { factories } from "api/test-utils"
+import { ResourceTypeEnum } from "api/v1"
+import type { LearningResource } from "api/v1"
+import { renderWithProviders, screen, user } from "@/test-utils"
+import NowPlayingSection from "./NowPlayingSection"
+
+const makeEpisode = (overrides = {}): LearningResource =>
+  factories.learningResources.podcastEpisode({
+    title: "Now Playing Episode",
+    description: "<p>Featured episode description</p>",
```
pdpinch: I don' t think this matters, but it would be more correct: 


```suggestion
    description: "<p>Now Playing Episode description</p>",
```

frontends/main/src/app-pages/PodcastPage/PodcastsListingPage/NowPlayingSection.tsx
```diff
@@ -0,0 +1,233 @@
...
+            src={nowPlaying.image.url}
+            alt={
+              nowPlaying.image.alt ??
+              nowPlaying.title ??
+              "Podcast episode cover"
+            }
+          />
+        )}
+        <NowPlayingBody>
+          <FeaturedBadge>FEATURED</FeaturedBadge>
```
pdpinch, line 186: I don't know why we need a featured badge in addition to the "Now Playing" header. I presume that the episode is chosen by publish date, so it's not technically "featured" in the sense that it wasn't chosen by a human. 

Contrast this with the featured podcasts lower on the page. These come from a list that is hand-picked by a human.

pdpinch (commented):
A couple of small comments. I don't want to hold this up, so if you are uncertain or you are getting contradictory comments, please just skip them for now and we can address them in a follow-up.

ChristopherChudzicki (changes requested):
Changes look good:
- unified `usePodcastPlayer.ts`
- single reusable mock for PodcastPlayer
- single EpisodeItem
- "All Episodes" uses a link
- other small cleanups

I still see some issues:

1. **Headings:** You added an h1 and a single h2. I'd suggest adding all the h2's mentioned in https://github.com/mitodl/mit-learn/pull/3576#discussion_r3539664716 unless there's a good reason not to. They help screenreaders navigate.
    - **SUGGESTION:** I'd also ad an `assertHeadings` test. They can be a bit fragile (break when we add more) but the fragility doesn't bother me. They are very easy to fix, an they have value... Assures meaningful headings are nested and we aren't skipping levels.
2. **Offered by as fallback title still in use:** E.g., here: https://github.com/mitodl/mit-learn/pull/3576/changes#diff-5c34dde8b81e39e81546eeb7d8c6aecf2a4d49eba1c38d6960af8a549358c58eR86    
3. Very minor, but https://github.com/mitodl/mit-learn/pull/3576#discussion_r3538447275 would be good to do
    - I opened https://github.com/mitodl/mit-learn/pull/3600... if it looks good to you, merge it, and use the new color.

**Suggestion:** Let's merge #3592 first, then you can use it here to resolve (2) easily.

frontends/main/src/app-pages/PodcastPage/PodcastsListingPage/LatestEpisodesSection.tsx
```diff
@@ -0,0 +1,119 @@
...
+        <EpisodeList role="list">
+          {episodes.map((episode) => {
+            const parentPodcastId = getEpisodeParentPodcastId(episode)
+            return (
+              <EpisodeItem
+                role="listitem"
+                key={episode.id}
+                isMobile={isMobile}
+                episode={episode}
+                overline={episode.offered_by?.name}
```
ahtesham-quraish: @ChristopherChudzicki this is as per design I think we are displaying the podcast Name there, do you want me to remove it?

Image

ahtesham-quraish:
>Suggestion: Let's merge https://github.com/mitodl/mit-learn/pull/3592 first, then you can use it here to resolve (2) >easily.

I have checked out this `ahtesham/podcast-episode` branch from this branch `ahtesham/podcast-listing-page-branch` So I think this fine to merge this PR first

frontends/main/src/app-pages/PodcastPage/PodcastsListingPage/PodcastsListingPage.test.tsx
ahtesham-quraish: I have done this already can you check please? I am confused what is left here

frontends/main/src/app-pages/PodcastPage/PodcastsListingPage/NowPlayingSection.tsx
ahtesham-quraish: @mbilalmughal What do you think should we remove the featured badge or something?

ahtesham-quraish:
>Very minor, but https://github.com/mitodl/mit-learn/pull/3576#discussion_r3538447275 would be good to do
I opened https://github.com/mitodl/mit-learn/pull/3600... if it looks good to you, merge it, and use the new color.

when #3600 will be merged i will rebase then will resolve the mentioned comment

frontends/main/src/app-pages/PodcastPage/PodcastsListingPage/LatestEpisodesSection.tsx
ChristopherChudzicki: @ahtesham-quraish You are right! Keep it!

I do think it would be good to add the podcast name before it to match the designs, now that https://github.com/mitodl/mit-learn/pull/3592 is merged

Screenshot 2026-07-10 at 10 35 26 AM

The offered_by will have to wait for the podcast offered_by ETL fix to actually work, but you can keep the code for it here. (Did you make an issue for that?)

ahtesham-quraish: image


I have made this change

ChristopherChudzicki (approved):
👍

PR #3577, opened 2026-07-06
fix: let keyboard users skip resource carousels via a reusable skip link

daniellefrappier18:

Fixes https://github.com/mitodl/hq/issues/11575

Keyboard users had to Tab through every focusable element in every card of a resource carousel (title link + bookmark button ≈ 2 stops × ~12 cards) before reaching anything after it. The originating ticket is about the home page's Featured Courses carousel, but they're all the same `ResourceCarousel` component.

- **New `SkipLink` component** (`ol-components`) — a small compound component (`Container` / `Trigger` / `Target`). The trigger is visually hidden until focused. It covers two shapes:
  - *Skip past a self-contained block* — `Container` generates its own target and renders a sentinel after the block (used by the carousel).
  - *Skip to an existing landmark* — a standalone `Trigger targetId="…"` pointing at an element already in the page (used by the video pages).
- **Featured Courses carousel**  — cards are plain tabbable content, wrapped in a `SkipLink` so keyboard users can jump the whole carousel via a "Skip {title}" link. Also reverted the `activeIndex` prop/effect added to `CarouselV2`.
- **Video pages** — migrated the existing hand-rolled skip links in `VideoDetailPage` and `VideoSeriesDetailPage` onto the shared component, and fixed a pre-existing bug where the revealed skip link was hidden behind the fixed site header — it now reveals just below the header, above the app-bar z-index, on a single line.

- [ ] Desktop screenshots
Screenshot 2026-07-09 at 1 48 03 PM

Screenshot 2026-07-09 at 1 48 22 PM

Screenshot 2026-07-09 at 1 47 49 PM

Screenshot 2026-07-09 at 1 47 41 PM

Using NVDA

https://github.com/user-attachments/assets/50f473e0-24c4-495f-ab45-fa0f122ceabd

### How can this be tested?

**Featured Courses carousel**
1. Go to the home page. Tab down the page until focus is about to enter the Featured Courses section.
2. The first focusable element in the section is a **"Skip Featured Courses"** link  just above the section heading.
3. Press **Enter** — focus jumps past all the cards to the content after the carousel.

**Video pages** (e.g. any `/video/<id>/<slug>` page)
1. Load the page and press **Tab** — the first stop is a skip link, below the fixed header.
2. Tab through **"Skip to main content"**, **"Skip to video player"**
3. Open a video that belongs to a playlist (for example: http://open.odl.local:8062/video/137323/space-chemistry-with-brett-mcguire?playlist=136851) and confirm a third link, **"Skip to more videos"**, appears and jumps to the "more from this playlist" section further down the same page. Note: It only renders when the video belongs to a playlist and the router picks VideoDetailPage (non-OCW).

**Screen reader (optional)**
- With NVDA/VoiceOver, confirm each skip link is announced as a link and that activating it lands focus in the expected place.


ChristopherChudzicki (changes requested):
Thank you for the UTM setup + JAWS guide. I got through Step 8 and have copy/paste working. Going to wait till I get home to finish the NVDA install rather than do that on the bus wifi. So, unfortunately, I haven't quite tested this on a screenreader yet.

I did test the keyboard navigation. It largely works as advertised. I found one bug (mentioned below) but more than that, I have some a11y/UX concerns about the behavior.

First, though, I want to mention: As far as I can tell, this is **not** one of the normal "carousel" patterns on https://www.w3.org/WAI/ARIA/apg/patterns/carousel/. I don't think that document mentions roving tabindex. It's focused on more traditional single-slide carousels (like big featured product auto-playing carousels on ecommerce sites). The behavior here is similar to grid+gridcell (https://www.w3.org/WAI/ARIA/apg/patterns/grid/, "A grid widget is a container that enables users to navigate the information or interactive elements it contains using directional navigation keys, such as arrow keys...")

The concerns I have are regarding discoverability:

1. **Discoverability, Sighted Users:** Keyboard-navigating sighted users don't have any way to discover these new keyboard controls, right? Consequence: They can't focus any elements in the 2nd+ card, unless they guess that arrow keys work.
    - I don't want to overstate this. Sighted keyboard users would be used to arrow keys from radiobuttons (definitely) and hopefully menus and tablist, where roving tabindex really is the sanctioned wcag pattern. But radio buttons, menus, tablist are all common operating level interfaces where the keyboard controls mimic the OS controls. This carousel doesn't mimic OS behavior.
2. **Discoverability, Screenreader Users:** Keyboard users navigating with a screenreader could potentially discover the keyboard arrow controls if they were associated with a role that usually has a roving tabindex (listbox, grid, menu, tablist, radiogroup; unsure if there are others). But `group` does not usually have a roving tabindex, right?

### An Alternative
My reading of the original issue is that the real keyboard-navigation ux issue is "User needs to tab through list of ~12x2 = 24 items per carousel. That's a lot of tabbing".

Am I correct that that is the main issue?

If so, one alternative might be a "skip link" ([WCAG G123 Adding a link at the beginning of a block of repeated content to go to the end of the block](https://www.w3.org/WAI/WCAG22/Techniques/general/G123)). Something like this video:

https://github.com/user-attachments/assets/8381f2d8-227f-48c4-ac40-9604b8c10650


What do you think? 
- Everything is tabbable, very standard behavior
- Carousels are skippable for keyboard users, and the mechanism is easily discovered

Could be implemented as a generic trio of components usable anywhere:
```tsx
<SkipLink.Container>
    maybe very minimal content here but probably not...
    <SkipLink.Trigger />
    content here
    <SkipLink.Target />
</SkipLink.Container>
```

### Roving Index Pagination Bug
See video below... The roving index works, until I try to transition from the 4th to the 5th slide. Then the carousel jumps all the way to the end.

https://github.com/user-attachments/assets/9ca7eff4-9c55-44b9-b11b-de0e32597f88

I believe the issue is here: https://github.com/mitodl/mit-learn/pull/3577/changes#diff-10bdf34da766cb0cf478aab89b0f935d849dec810e9c1c28307bdaa42b4a4da4R214. I think scrollTo expects a page, but I haven't worked with embla too much.

### My Suggestion:
The skipping approach seems simpler, so I lean toward that. If you think there's a good reason to keep the roving tabindex, I think we should use the appropriate roles (grid/gridcell) and add some visual affordance for keyboard-navigating sighted users. And fix the bug above.

frontends/ol-components/src/components/SkipLink/SkipLink.tsx
```diff
@@ -0,0 +1,259 @@
...
+const SkipLinkContext = React.createContext<SkipLinkContextValue | null>(null)
+
+const Root = styled.div({
+  position: "relative",
+})
+
+// Shared visual chrome for every trigger — hidden but focusable by default,
+// styled to match the app's existing skip-link treatment. Only the `:focus`
+// reveal placement differs between the block and page variants below.
+const skipLinkChrome = (theme: Theme) => ({
```
ChristopherChudzicki: Here and below in `revealBase`:

```suggestion
const skipLinkChrome = (theme: Theme): CSSObject => ({
```

This way:
- no need for `as const`
- autocomplete works
    - though...that matters less with LLMs doing a lot of the editing nowadays

frontends/ol-components/src/components/SkipLink/SkipLink.tsx
```diff
@@ -0,0 +1,259 @@
...
+  )
+  return (
+    candidates.find(
+      (el) =>
+        !!(
+          marker.compareDocumentPosition(el) & Node.DOCUMENT_POSITION_FOLLOWING
+        ),
+    ) ?? null
+  )
+}
```
ChristopherChudzicki: Two requests here:
1. expand the FOCUSABLE_SELECTOR list.
2. Do the "is it really focusable" check mentioned below.

### FOCUSABLE_SELECTOR

video and iframe are realistic for us; contenteditable is possible, too, on some staff pages.

```ts
const FOCUSABLE_SELECTOR = [
  "a[href]",
  "area[href]",
  "button:not([disabled])",
  'input:not([disabled]):not([type="hidden"])',
  "select:not([disabled])",
  "textarea:not([disabled])",
  "iframe",
  "object",
  "embed",
  "audio[controls]",
  "video[controls]",
  "details > summary:first-of-type",
  '[contenteditable]:not([contenteditable="false"])',
  '[tabindex]:not([tabindex="-1"])',
].join(",")
```

### Check if it's really focusable
An element whose tag name can receive focus may or may not actually be focusable depending on ancestors. **For example:** The querySelector could match a button or an anchor whose ancestor has `display: none`. Such elements aren't actually focusable.

I'm not aware of any cases where this happens today, but it's fairly easy to avoid: *try to focus* then move on if you couldn't.

```ts
const focusFirstAfter = (marker: HTMLElement): boolean => {
  const doc = marker.ownerDocument
  const candidates = doc.querySelectorAll<HTMLElement>(FOCUSABLE_SELECTOR)
  for (const el of candidates) {
    if (marker.compareDocumentPosition(el) & Node.DOCUMENT_POSITION_FOLLOWING) {
      el.focus()
      if (doc.activeElement === el) return true // actually took focus
    }
  }
  return false
}
// then the caller does:      if (!focusFirstAfter(target)) target.focus()
```

**Edit:** I guess tabindex makes "next focusable" even trickier

ChristopherChudzicki (changes requested):
This is working well now. One potential issue worry about in the future is that "next focusable element" is actually pretty tricky to do. See https://github.com/mitodl/mit-learn/pull/3577/changes#r3560658286

For that reason, I think the alternative below is worthwhile.

If you need to move on to other stuff OR think the extra tab stop (for users who activate the skip) is not worth it, im ok with the current approach, but we should make the changes mentioned above.

**Request:** Do alternative OR the changes in https://github.com/mitodl/mit-learn/pull/3577/changes#r3560658286

---

### Current approach + alternative

When i suggested this, I was expecting that we'd always focus the target. But you're absolutely right that the next element should be a real focusable thing with a relevant role/label.

My understanding is that most sites do this by having an explicit target, e.g., a `<main tabIndex="-1" />`. We do NOT want to require an explicit target, because this is used in some cases (like the resource carousel) all over the site where we don't want to require tying it explicitly to what is next. That would be fragile.

The "next focusable thing" has downsides though:
- **it's location is unpredictable.** It could move the user way paste where they were. In practice, I don't think that's an issue anywhere on our site.
- **"Next focusable" is tricky:** There's no clean way to query what the next focusable element is. 

If that becomes a problem...An alternative that avoids this issue altogether would be:

-  to make your sentinel target a `<a tabindex="-1" id="target">Return to LABEL </a>` that goes back to the trigger
- and ALWAYS use that, rather than finding next focusable element.

Then it's a real meaningful element with a relevant label. With `tabindex="-1"`, so that it doesnt add a (second) new tab stop.

frontends/ol-components/src/components/SkipLink/SkipLink.tsx
daniellefrappier18: I went with the alternative. It seemed like the safest of the two paths. The Container-owned `Target` is now a real `<a tabindex="-1">Return to{label}</a>` at the end of the block, and the trigger always lands focus there. That deletes `firstFocusableAfter` / `FOCUSABLE_SELECTOR` entirely. Seems to work well with NVDA. 

https://github.com/user-attachments/assets/b711891c-4ed9-4422-a7c6-9d7eebadffcd

frontends/main/src/page-components/ResourceCarousel/ResourceCarousel.tsx
```diff
@@ -261,92 +267,99 @@ const ResourceCarousel: React.FC<ResourceCarouselProps> = ({
...
-                arrowsContainer={ref}
-                mobileBleed="symmetric"
-                mobileGutter={16}
-              >
-                {isLoading || childrenLoading
-                  ? Array.from({ length: 6 }).map((_, index) => (
-                      <ResourceCard
-                        isLoading
+      <SkipLink.Container label={title}>
+        <SkipLink.Trigger />
```
ChristopherChudzicki: Potential small improvement would be to only offer the skip if there's more than say 3 cards

Image

ChristopherChudzicki (approved):
👍 Looks good. I think you've settled on the clean/robust solution.


Left a small suggestion.

PR #3578, opened 2026-07-06
desktop dashboard design updates

gumaerc:
Closes https://github.com/mitodl/hq/issues/12216

This PR mainly addresses design updates to the dashboard cards in desktop view. As part of this, the "Module" card type indicator was restored on B2B contract pages as described in the above issue. However, a large amount of other design changes were implemented from Figma. There are no issues for these, but I will try and do by best to summarize:

#### Figma links
- https://www.figma.com/design/mSDdN8giMcncBwNgwcVJL9/UX-Wireframes?node-id=19084-489146&m=dev
- https://www.figma.com/design/mSDdN8giMcncBwNgwcVJL9/UX-Wireframes?node-id=19084-487069&m=dev

##### Card type & progress indication
- Replaced the plain "Course"/"Module" text label with a ProgressBadge + Separator + type-label combo on every card (Enrolled, Unenrolled, Program-as-Course, Program Enrollment)
- ProgressBadge shows "Not Started" / "In Progress" / "Completed" (with a checkmark icon for Completed), color-coded by status
- Renamed EnrollmentStatusIndicator to EnrollmentStatus/EnrollmentStatusIcon; simplified it by dropping the showNotComplete prop and swapping custom image icons for Remix icon components
- The status icon is now scoped to only compact module rows inside ProgramAsCourseCard, and is hidden there for B2B contract enrollments; it no longer appears on default single-run cards or in the sibling-runs accordion header
- CoursewareCard, EnrolledCourseCard, and UnenrolledCourseCard gained an isModule prop to drive "Module" vs "Course" labeling; B2B contract enrollments are automatically labeled "Module"

##### Certificate & upgrade messaging
- "Certificate track" (UpgradedBanner) now shows for any non-B2B verified enrollment regardless of whether a certificate has already been issued, coexisting with the "View Certificate" link instead of disappearing once earned
- On ProgramEnrollmentCard, the certificate link moved out of the title section and into the button row next to "View", matching EnrolledCourseCard and UnenrolledCourseCard
- Fixed alignment of the "View Certificate" button on the program-as-course card

##### End date & date summary placement
- End date text now renders consistently below the title on desktop for both compact and default layouts, replacing the old "compact-meta-row" pattern that grouped the end date and CTA button together
- On mobile, compact-layout cards omit the end date entirely; default layout still shows it
- Popover trigger styling switched from body2 to subtitle3 typography

##### Multiple enrollment runs
- Added a sibling-runs accordion on EnrolledCourseCard desktop that appears when a course has more than one active enrollment, showing a per-run status icon (or upcoming/expired icon) and a "Course runs (N)" count

##### Typography & spacing
- CardTypeText, SubtitleLink/SubtitleLinkRoot, and UpgradedBannerRoot switched from subtitle3/subtitle4 to body3 typography
- TitleHeading and TitleLink now explicitly set subtitle2 typography
- SubtitleLink color changed from conditional red/gray by layout to a flat red, with a tighter icon gap
- Removed the negative left margin on the "More options" menu button
- Added a small Ellipse separator dot for stacked meta segments, alongside the existing Separator bar

##### Other
- Fixed a spacing/gap bug in the card type text on ProgramEnrollmentCard
- General visual cleanup pass on UnenrolledCourseCard styling

image
image
image
image
image
image

### How can this be tested?
- Make sure you have an instance of MITx Online started and connected to Learn as described in the readme
- Your test user should have at least one B2B contract assigned to them as well as a B2C enrollment for each type of content (Course, Program, "Program As Course") with different enrollments in different states (not started, started, completed) for each type
- View the B2B contract in the dashboard. Ensure that the design changes noted above and in the Figma look correct and that the card type indicator says "Module"
- View the B2C enrollments on your dashboard home page / program dashboard and ensure that the UI changes are consistent

### DIFF
--- a/frontends/main/src/app-pages/DashboardPage/CoursewareDisplay/EnrolledCourseCard.tsx
+++ b/frontends/main/src/app-pages/DashboardPage/CoursewareDisplay/EnrolledCourseCard.tsx
@@ -8,6 +8,7 @@ import {
   CoursewareButtonLink,
   MenuButton,
   Separator,
+  Ellipse,
   SubtitleLink,
   SubtitleLinkRoot,
   TitleHeading,
@@ -42,12 +43,27 @@ const formatUpgradeTime = (daysFloat: number) => {
   if (daysFloat < 0) return ""
   const days = Math.floor(daysFloat)
   if (days > 1) {
-    return ` · ${days} days remaining`
+    return `${days} days remaining`
   } else if (days === 1) {
-    return ` · ${days} day remaining`
+    return `${days} day remaining`
   }
-  return " · Less than a day remaining"
+  return "Less than a day remaining"
 }
+
+const RedEllipse = styled(Ellipse)(({ theme }) => ({
+  marginLeft: "6px",
+  marginRight: "6px",
+  backgroundColor: theme.custom.colors.red,
+}))
+
+const EnrolledTitleLink = styled(TitleLink)<{
+  enrollmentstatus: EnrollmentStatus
+}>(({ enrollmentstatus, theme }) => ({
+  color:
+    enrollmentstatus === EnrollmentStatus.Completed
+      ? theme.custom.colors.silverGrayDark
+      : theme.custom.colors.black,
+}))
 
 const UpgradeBanner: React.FC<
   {
@@ -56,7 +72,6 @@ const UpgradeBanner: React.FC<
     certificateUpgradePrice?: string | null
     productId?: number | null
     onError?: (error: Error) => void
-    layout?: "default" | "compact"
   } & React.HTMLAttributes<HTMLDivElement>
 > = ({
   canUpgrade,
@@ -64,7 +79,6 @@ const UpgradeBanner: React.FC<
   certificateUpgradePrice,
   productId,
   onError,
-  layout = "default",
   ...others
 }) => {
   const replaceBasketItem = useReplaceBasketItem()
@@ -95,16 +109,19 @@ const UpgradeBanner: React.FC<
     : null
 
   return (
-    <SubtitleLinkRoot layout={layout} {...others}>
-      <SubtitleLink layout={layout} href="#" onClick={handleUpgradeClick}>
+    <SubtitleLinkRoot {...others}>
+      <SubtitleLink href="#" onClick={handleUpgradeClick}>
         <RiArrowUpCircleLine size="16px" />
         {`Upgrade for certificate - ${formattedPrice}`}
       </SubtitleLink>
       {calendarDays !== null && (
-        <NoSSR>
-          {/* This uses local time. */}
-          {formatUpgradeTime(calendarDays)}
-        </NoSSR>
+        <>
+          <RedEllipse />
+          <NoSSR>
+            {/* This uses local time. */}
+            {formatUpgradeTime(calendarDays)}
+          </NoSSR>
+        </>
       )}
     </SubtitleLinkRoot>
   )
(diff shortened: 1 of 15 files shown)


daniellefrappier18 (commented):
This spacing seems strange to me but this is following the figma spec. So this is more of a question for @steven-hatch 

Image

But if you give the button top 10px it looks a lot better. 

Image

frontends/main/src/app-pages/DashboardPage/CoursewareDisplay/EnrolledCourseCard.tsx
```diff
@@ -162,15 +181,22 @@ export const EnrolledCourseCard = ({
...
   Component,
   className,
 }: EnrolledCourseCardProps) => {
   const course = enrollment.run.course
   const run = enrollment.run
+  const isCompact = layout === "compact"
   const isContractPageResource = Boolean(enrollment.b2b_contract_id)
+  const cardTypeLabelText =
+    isModule || isContractPageResource ? "Module" : "Course"
+  const cardTypeLabel =
```
daniellefrappier18, line 194: on UnenrolledCourseCard the cardTypeLabel is `cardTypeLabel  = isModule && layout === "compact" ? null : <CardTypeText/>` but here it's whenever isModule OR isCompact. Is that intentional?

gumaerc: Hmm, maybe I need to rethink the labeling here. The issue is that "module" is an overloaded term. In `Program` objects with `display_mode` set to "course" aka "program as course" in `ProgramAsCourseCard`, individual course requirements from the `Program` are displayed and referred to as "modules" in the code but the card type label is not displayed because of the compact layout. However, in a B2B contract displayed via `ContractContent`, the programs there do not have `display_mode` set to "course" but they are still referred to as "Module" in the UI. So, I'm using `isModule && isCompact` to mean "we're truly rendering a card for a module row within `ProgramAsCourse`", but maybe I need to be more explicit about that? So, the reason this is so confusing is because "module" as a canonical term means a part of a course, which we represent using the `Program` structure from MITx Online, designating `display_mode` as "course" which changes the terminology of the parent / child relationship. However, we break this rule by also referring to it as a module when this flag is *not* true. Hopefully that makes sense?

daniellefrappier18 (approved):
This looks good. I only found that one weird spacing issue and it's certainly not a blocker.

frontends/main/src/app-pages/DashboardPage/CoursewareDisplay/EnrolledCourseCard.tsx
daniellefrappier18: ah I see, I think you even mentioned this and I forgot.

gumaerc:
@daniellefrappier18 Thanks for the review. With the spacing issue there, I did talk with Steve about this. In Figma, you'll notice that the first dot separator is actually an SVG, and the second one in the text is a unicode character directly in the text. In the code, I brought some sanity to this by making them both simple styled divs. The spacing may not match exactly to Figma because of the difference in implementation, but this ensures that the separators look the same and aren't being rendered two different ways. We wanted the red text closer together though, hence why the left gray separator has a 12 pixel margin on either side and the red has 6 pixels on either side. It's supposed to distinguish that the start / end date popover trigger and the upgrade banner are two different things, and the time left to upgrade belongs with the upgrade banner. @steven-hatch that makes sense right?

daniellefrappier18:
> @daniellefrappier18 Thanks for the review. With the spacing issue there, I did talk with Steve about this. In Figma, you'll notice that the first dot separator is actually an SVG, and the second one in the text is a unicode character directly in the text. In the code, I brought some sanity to this by making them both simple styled divs. The spacing may not match exactly to Figma because of the difference in implementation, but this ensures that the separators look the same and aren't being rendered two different ways. We wanted the red text closer together though, hence why the left gray separator has a 12 pixel margin on either side and the red has 6 pixels on either side. It's supposed to distinguish that the start / end date popover trigger and the upgrade banner are two different things, and the time left to upgrade belongs with the upgrade banner. @steven-hatch that makes sense right?

I should have been more clear on what spacing I thought was off. I was referring to kebab menu not aligning with the top of the course name.

gumaerc:
@daniellefrappier18 Okay, thanks. I discussed this with Steve on Slack and he wants to get these changes out, then ask Bilal what he thinks when he's back and we can make an adjustment PR later. I think this is good enough now and we've made enough changes that this should go out and we should get some feedback.

daniellefrappier18:
> @daniellefrappier18 Okay, thanks. I discussed this with Steve on Slack and he wants to get these changes out, then ask Bilal what he thinks when he's back and we can make an adjustment PR later. I think this is good enough now and we've made enough changes that this should go out and we should get some feedback.

Definitely! I would not hold this up for a spacing issue.

PR #3581, opened 2026-07-07
Fix: header stat counts not updating after bulk-assign or revoke on Contract Admin page

daniellefrappier18:

The Contract Admin page header shows stat counts (Unassigned, Pending claim, etc.) derived from the contract detail query. Bulk-assigning seats or revoking a code mutates those counts server-side, but the `contractDetail` query wasn't being invalidated on mutation, so the header stayed stale until a full page refresh.

- [ ] Desktop screenshots

https://github.com/user-attachments/assets/23a10011-5497-4d91-9631-16eeae4af253

### How can this be tested?

1. Navigate to a Contract Admin page for an org with an active contract
2. Note the Unassigned and Pending claim counts in the header.
3. Bulk-assign one or more seats via email (paste an email, click Assign Seats, confirm the send)
4. Confirm the header's Unassigned count decreases and Pending claim increases immediately — no manual page refresh needed.
5. From the codes table, open More actions on an assigned-but-unredeemed code and click Release seat (revoke), then confirm.
6. Confirm the header's Unassigned count increases and Pending claim decreases immediately.

### DIFF
--- a/frontends/api/src/mitxonline/hooks/organizations/index.ts
+++ b/frontends/api/src/mitxonline/hooks/organizations/index.ts
@@ -45,6 +45,14 @@ const useBulkAssignSeats = () => {
           vars.parent_lookup_organization,
         ),
       })
+      // Assigning seats moves them from unassigned to assigned, which changes
+      // the header stat counts on the contract admin page.
+      queryClient.invalidateQueries({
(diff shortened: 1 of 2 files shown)


ahtesham-quraish (approved):
LGTM!

PR #3585, opened 2026-07-08
Fix embedding summarizer task when overwrite=True

shanbady:

Closes https://github.com/mitodl/hq/issues/12244

This PR ensures that we don't continuously regenerate summaries when running the mitxonline etls

### How can this be tested?
1. checkout main
2. go to django admin and configure a ContentSummarizerConfig object (use [prod as a reference](https://api.learn.mit.edu/admin/learning_resources/contentsummarizerconfiguration/1/change/) for mitxonline
3. find some contentfile ids for .srt files (that are also part of a course in a platform that summarizer config is set to - in this case mitxonline) and run the generate embeddings task with overwrite=True via:
```python
generate_embeddings.apply([[17964642, 17964479, 17964266, 17964348, 17964067, 17964323, 17964168,], 'content_file', True])
```
4. see in the output that the model configured for the summarizer is shown and that the contentfiles have summaries
5. re-run the exact same snippet and note that it keeps regenerating summaries
6. checkout this branch and repeat - note that summaries are only generated once

### DIFF
--- a/vector_search/utils.py
+++ b/vector_search/utils.py
@@ -616,6 +616,9 @@ def _content_file_stored_checksum_changed(serialized_document: dict) -> bool:
     if not point:
         return False
     stored_checksum = (point.payload or {}).get("checksum")
+    # Missing checksums should not force an expensive summary rewrite by themselves.
+    # should_generate_content_embeddings still treats them as changed so embeddings
+    # can repair older Qdrant points without overwriting existing summaries.
     return stored_checksum is not None and stored_checksum != serialized_document.get(
         "checksum"
     )
--- a/vector_search/utils.py
+++ b/vector_search/utils.py
@@ -591,12 +591,9 @@ def should_generate_resource_embeddings(serialized_document):
     return True
 
 
-def should_generate_content_embeddings(
+def _retrieve_content_file_point(
     serialized_document: dict, point_id: str | None = None
-) -> bool:
-    """
-    Determine if we should generate embeddings for a content file
-    """
+):
     client = qdrant_client()
     if not point_id:
         # we just need metadata from the first chunk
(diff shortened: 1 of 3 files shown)


vector_search/utils.py
```diff
@@ -889,7 +908,10 @@ def process_batch(docs_batch, summaries_list):
                     .filter(run__id=resource.get("run_id"))
                     .exists()
                 ):
-                    summaries_list.append(resource["id"])
+                    if _content_file_stored_checksum_changed(resource):
```
mbertrand: When `overwrite=False`, `docs_batch` has already been filtered to files with no existing chunk-0 point, so this always returns `False` — but still costs one Qdrant retrieve per file. `if overwrite and _content_file_stored_checksum_changed(resource):` preserves behavior and skips the wasted round-trips.

vector_search/utils.py
```diff
@@ -935,20 +957,32 @@ def process_batch(docs_batch, summaries_list):
...
+            process_batch(
+                current_batch_docs,
+                fill_summary_content_ids,
+                changed_summary_content_ids,
+            )
             current_batch_docs = []
             gc.collect()
 
-        if summary_content_ids:
+        if fill_summary_content_ids:
```
mbertrand: Summarization runs after all batches have already upserted points with the new checksum. If the run crashes or the summarizer fails for some files, the next run sees matching checksums and so I think the stale summary would never be regenerated. Is `generate_summary_flashcards` the intended recovery path? If so, worth a note here.

vector_search/utils.py
```diff
@@ -610,10 +607,31 @@ def should_generate_content_embeddings(
...
+        return response[0]
+    return None
+
+
+def _content_file_stored_checksum_changed(serialized_document: dict) -> bool:
+    point = _retrieve_content_file_point(serialized_document)
+    if not point:
+        return False
+    stored_checksum = (point.payload or {}).get("checksum")
+    return stored_checksum is not None and stored_checksum != serialized_document.get(
```
mbertrand, line 622: A missing stored checksum is treated as "unchanged" here, but as "changed" in `should_generate_content_embeddings`. If that's intentional (conservative about expensive summary calls), a short comment might keep someone from "fixing" the asymmetry later.

vector_search/utils.py
```diff
@@ -935,20 +957,32 @@ def process_batch(docs_batch, summaries_list):
...
+            )
             current_batch_docs = []
             gc.collect()
 
-        if summary_content_ids:
+        if fill_summary_content_ids:
+            ContentSummarizer().summarize_content_files_by_ids(
+                fill_summary_content_ids, overwrite=False
+            )
+        if changed_summary_content_ids:
```
mbertrand: With this change, `overwrite=True` can no longer force summary regeneration for unchanged content (e.g. after switching the summarizer model/prompt) — `generate_summary_flashcards` would be the only path. Just confirming that's deliberate.

shanbady: I think that is fine. I imagine that would likely want to force re-generate via management command after making a model switch

vector_search/utils.py
shanbady: good catch. moved things around so that the checksum is only updated after successful summarization (it  will get retried on the next etl run) - added tests

vector_search/utils.py
shanbady: added comment

vector_search/utils.py
shanbady: fixed

mbertrand (approved):
👍 assuming tests pass

PR #3592, opened 2026-07-09
feat: Add parent podcast information while fetching the episode info  

ahtesham-quraish:
https://github.com/mitodl/hq/issues/12264

As a user viewing a podcast episode page, I want the episode API to include summary information about the podcast it belongs to, so the frontend can render the parent podcast's title and image without making a second API request.

### How can this be tested?
For testing you just need to visit the following url `http://open.odl.local:8062/podcast/2115/podcast_episode/2116/the-electrical-grid-and-beyond-with-david-hsu` please replace the ids and slug from the url according to your local data and see the podcast parent array inside the episode object is populated by the backend.

### DIFF
--- a/frontends/main/src/app-pages/PodcastPage/PodcastEpisodeDetailPage.test.tsx
+++ b/frontends/main/src/app-pages/PodcastPage/PodcastEpisodeDetailPage.test.tsx
@@ -63,8 +63,7 @@ const setupApis = ({
         {
           id: podcast.id,
           title: podcast.title!,
-          readable_id: podcast.readable_id!,
-          image: podcast.image ?? null,
+          readable_id: podcast.readable_id,
         },
       ],
       ...episodeOverridesEpisode,
@@ -211,8 +210,7 @@ describe("PodcastEpisodeDetailPage", () => {
       {
         id: podcast.id,
         title: podcast.title!,
-        readable_id: podcast.readable_id!,
-        image: podcast.image ?? null,
+        readable_id: podcast.readable_id,
       },
     ]
 
--- a/frontends/main/src/app-pages/PodcastPage/PodcastEpisodeDetailPage.test.tsx
+++ b/frontends/main/src/app-pages/PodcastPage/PodcastEpisodeDetailPage.test.tsx
@@ -52,7 +52,24 @@ const setupApis = ({
   moreEpisodes,
 }: SetupOptions = {}) => {
   const podcast = makePodcast(podcastOverrides)
-  const episode = makePodcastEpisode(episodeOverrides)
+  const episodeOverridesEpisode = (
+    episodeOverrides as Partial<PodcastEpisodeResource>
+  ).podcast_episode
+  const episode = makePodcastEpisode({
+    ...episodeOverrides,
+    podcast_episode: {
+      podcasts: [podcast.id],
+      parent_podcasts: [
+        {
+          id: podcast.id,
+          title: podcast.title!,
+          readable_id: podcast.readable_id!,
+          image: podcast.image ?? null,
+        },
+      ],
+      ...episodeOverridesEpisode,
+    },
+  } as Partial<LearningResource>)
 
   setMockResponse.get(
     urls.learningResources.details({ id: episode.id }),
@@ -189,6 +206,15 @@ describe("PodcastEpisodeDetailPage", () => {
     const episode = makePodcastEpisode()
     episode.podcast_episode.audio_url = "https://example.com/ep.mp3"
     const podcast = makePodcast()
+    episode.podcast_episode.podcasts = [podcast.id]
+    episode.podcast_episode.parent_podcasts = [
+      {
+        id: podcast.id,
+        title: podcast.title!,
+        readable_id: podcast.readable_id!,
+        image: podcast.image ?? null,
+      },
+    ]
 
     setMockResponse.get(
       urls.learningResources.details({ id: episode.id }),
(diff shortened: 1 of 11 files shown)


pdpinch:
I understand this will make the front-end more performant by reducing the number of API calls. But does it increase the complexity of the backend query? Will that still be performant?

frontends/main/src/app-pages/PodcastPage/PodcastEpisodeDetailPage.test.tsx
```diff
@@ -189,6 +206,15 @@ describe("PodcastEpisodeDetailPage", () => {
     const episode = makePodcastEpisode()
     episode.podcast_episode.audio_url = "https://example.com/ep.mp3"
     const podcast = makePodcast()
+    episode.podcast_episode.podcasts = [podcast.id]
+    episode.podcast_episode.parent_podcasts = [
+      {
+        id: podcast.id,
+        title: podcast.title!,
+        readable_id: podcast.readable_id!,
```
ChristopherChudzicki: Already non-nullable, no reason for the `!` assertion.

learning_resources/serializers.py
```diff
@@ -441,17 +441,34 @@ class Meta:
         exclude = ("learning_resource", "author", *COMMON_IGNORED_FIELDS)
 
 
+class PodcastEpisodeParentSerializer(serializers.Serializer):
+    """Minimal parent-podcast summary embedded in an episode."""
+
+    id = serializers.IntegerField(source="parent_id")
+    title = serializers.CharField(source="parent.title")
+    readable_id = serializers.CharField(source="parent.readable_id")
+    image = LearningResourceImageSerializer(source="parent.image", allow_null=True)
```
ChristopherChudzicki: Correct me if I'm wrong, but nothing on the current (`main`) frontend or #3576 needs the `image` property, here, right? Just `{ id, title }`?

**Request:** Unless there's a good reason, let's hold off on adding `image`. You DID add the prefetch appropriately, so there's no N+1, but it's an extra join on the DB and we aren't using it. So, let's not expand the api surface / let's not do the join.

learning_resources/models.py
```diff
@@ -374,7 +374,7 @@ def for_serialization(self):
                 "parents",
                 queryset=LearningResourceRelationship.objects.filter(
                     relation_type=LearningResourceRelationTypes.PODCAST_EPISODES.value,
-                ),
+                ).select_related("parent__image"),
```
ChristopherChudzicki: If you agree with https://github.com/mitodl/mit-learn/pull/3592/changes#r3555565456, then make sure to remove this piece.

ChristopherChudzicki (changes requested):
This works well / looks good, with the one request mentioned below (remove `image` field unless we need it).

Thanks for updating `frontends/main/src/app/(site)/podcast/[podcastId]/podcast_episode/[episodeId]/[slug]/page.tsx` to avoid the extra query. That's a nice side-benefit. The real win comes in #3576

ahtesham-quraish:
> I understand this will make the front-end more performant by reducing the number of API calls. But does it increase the complexity of the backend query? Will that still be performant?

## Does embedding `parent_podcasts` hurt backend performance?

**Short answer: No meaningful increase in cost, and no N+1.**

### What actually changed in the query

The parent-podcast relationship was **already being prefetched** — the `_podcasts` Prefetch exists in `for_serialization()` because the serializer's existing `podcasts` field needs the parent IDs. The **only** change is appending `.select_related("parent")` to that already-existing prefetch:

```python
Prefetch(
    "parents",
    queryset=LearningResourceRelationship.objects.filter(
        relation_type=LearningResourceRelationTypes.PODCAST_EPISODES.value,
    ).select_related("parent"),   # ← the only addition
    to_attr="_podcasts",
),
```

So:

- **No new database round-trip is added.** Django already fired one prefetch query for `parents`. We did not add a second query.
- `select_related("parent")` widens that *same* query with a single `JOIN` (the parent `LearningResource` row) instead of issuing lazy per-row lookups.

> Note: we intentionally dropped `image` from the payload. It wasn't being used by the frontend, so there's no reason to expand the API surface or pay for the extra `LearningResourceImage` join. Only `id`, `title`, and `readable_id` are returned.

### Why there's no N+1

Without `select_related`, reading `relationship.parent.title` and `relationship.parent.readable_id` in the serializer would trigger an extra lazy query **per parent, per episode** — the classic N+1 trap. The `select_related("parent")` pre-joins that row into the single prefetch query, eliminating it entirely.

### Why the JOIN cost is negligible

- A podcast episode belongs to **very few** podcasts (typically exactly 1), so `_podcasts` is a tiny set — usually one row.
- The added JOIN is on an indexed foreign key (`LearningResourceRelationship.parent_id → LearningResource.id`).
- On the cached `list` endpoint, non-episode resources have an **empty** `_podcasts` set, so they contribute no extra joined rows.

### Net effect

| | Before | After |
|---|---|---|
| Queries for parent data | 1 prefetch (IDs only) | 1 prefetch (IDs + title/readable_id via JOIN) |
| Risk of N+1 | Yes, if title accessed | No |
| Extra round-trips | — | 0 |

We trade **one indexed JOIN on a single existing query** for **one fewer HTTP request on the frontend**. Clear net win — the backend stays performant, and the `list` endpoint remains cached on top of that.

ChristopherChudzicki:
@ahtesham-quraish ugh I merged this by accident. I meant to approve.

PR #3595, opened 2026-07-09
Retry generate_embeddings on gRPC UNAVAILABLE with minutes-scale backoff

mbertrand:
None directly. Tangentially related to earlier embedding-pipeline reliability work: [mitodl/hq#12115](https://github.com/mitodl/hq/issues/12115) (embeddings task OOM thrashing) and [mitodl/hq#12018](https://github.com/mitodl/hq/issues/12018) (skip waiting on Qdrant payload operations).

Prompted by a burst of `generate_embeddings` failures in Sentry overnight July 8–9 (gRPC `UNAVAILABLE`: *"Received http2 header with status: 502"*).

Our nightly embedding runs periodically saturate the Qdrant Cloud cluster's CPU (2 vCPUs per node). During those short throttle episodes, Qdrant's edge proxy returns 502s, which reach the task as gRPC `UNAVAILABLE` errors. `generate_embeddings` currently treats `UNAVAILABLE` as fatal, so each affected chunk fails immediately and its resources stay unembedded until the next overwrite run.

Two changes, both confined to `generate_embeddings`:

- Treat `UNAVAILABLE` as transient and retryable, same as `DEADLINE_EXCEEDED`.
- Scale the retry backoff from seconds (max 1s/2s/4s) to jittered minutes (max 2m/4m/8m), so retries land after a throttle episode has (hopefully) passed instead of hitting the cluster while it's still saturated.

Retries re-enter the queue through the task's existing `200/m` rate limit, and the full jitter spreads them out. If Qdrant is still unavailable after all retries (~14 min worst case), the task fails terminally exactly as it does today.

### How can this be tested?
Stop the local Qdrant container (`docker compose stop qdrant`), trigger an embedding task, and watch the celery logs — the task should schedule a retry with a minutes-scale countdown instead of failing. Start Qdrant again and the retry succeeds.

### DIFF
--- a/vector_search/tasks.py
+++ b/vector_search/tasks.py
@@ -131,9 +131,12 @@ def _queue_program_content_file_embedding_tasks(index_tasks, program_ids, overwr
 
 
 def _retry_countdown(retries: int) -> int:
-    """Full-jitter exponential backoff (mirrors retry_backoff=True), capped at 10m."""
+    """
+    Full-jitter exponential backoff capped at 10m. Minutes-scale so retries land
+    after a qdrant CPU-throttle episode instead of hammering it while saturated.
+    """
     return get_exponential_backoff_interval(
-        factor=1, retries=retries, maximum=600, full_jitter=True
+        factor=120, retries=retries, maximum=600, full_jitter=True
     )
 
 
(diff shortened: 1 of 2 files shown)


shanbady (approved):
👍

PR #3411, opened 2026-06-03
Ignore both SCM and DEDP micromasters programs

mbertrand:
### What are the relevant tickets?
N/A

### Description (What does it do?)

Tweaks the micromasters ETL to ignore both `/dedp/` and `/scm/` programs.

### How can this be tested?
- Set `MICROMASTERS_CATALOG_API_URL=https://micromasters.mit.edu/api/v0/catalog/` in your `backend.local.env` file
- On the main branch, run `./manage.py backpopulate_micromasters_data`.  Then go to django admin url http://open.odl.local:8065/admin/learning_resources/learningresource/?etl_source=micromasters&resource_type__exact=program.  You should see a "Supply Chain Management" program there, and it should be published.
- Switch to this branch, restart containers, run the command again.
- Go back to django admin. The "Supply Chain Management" program should now be unpublished.

### Additional Context

Using a env variable / django setting for IGNORE_URLS was considered but ultimatedly decided to leave as a constant and update as necessary.

PR #3597, opened 2026-07-09
Ignore POM micromasters programs

mbertrand:
N/A — follow-up to #3411

Adds `/pom/` to the micromasters ETL `IGNORE_URLS` pattern so Principles of Manufacturing programs are excluded, same as the existing `/dedp/`, `/scm/`, and `/fin/` exclusions. Updates the exclusion unit test with `/pom/` (excluded) and `/pomelo/` (boundary, not excluded) cases.

### How can this be tested?
- set `MICROMASTERS_CATALOG_API_URL=https://micromasters.mit.edu/api/v0/catalog/` in `backend.local.env`, run `./manage.py backpopulate_micromasters_data`, and confirm no POM program appears at http://open.odl.local:8065/admin/learning_resources/learningresource/?etl_source=micromasters&resource_type__exact=program

### DIFF
--- a/learning_resources/etl/micromasters.py
+++ b/learning_resources/etl/micromasters.py
@@ -21,7 +21,7 @@ from learning_resources.models import LearningResource, default_pace
 
 OFFERED_BY = {"code": OfferedBy.mitx.name}
 READABLE_ID_PREFIX = "micromasters-program-"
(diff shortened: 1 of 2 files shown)


shanbady (approved):
👍

PR #3601, opened 2026-07-10
fix: Reconcile inconsistent parent podcast title resolution used by the podcast audio player

ahtesham-quraish:
https://github.com/mitodl/hq/issues/12245

As a developer maintaining the podcast pages, I want a single canonical way to resolve the "podcast name" shown in the audio player's now-playing track, so that the three podcast pages don't each compute it differently and inconsistently.

image

### How can this be tested?
For testing you need to check the podcast listing page and episode page and check the if the player has proper information of podcast title, check the attached screenshot.


frontends/main/src/app-pages/PodcastPage/PodcastsListingPage/NowPlayingSection.tsx
daniellefrappier18: MUI's Typography maps variant="subtitle2" to `<h6>` by default. So non-heading strings become spurious headings, polluting the screen-reader heading outline. This also happens in the offered_by • platform meta line → `<h6>` and in:

[PodcastSection.tsx:189] — the All podcasts link text → `<h6>` inside an <a> (heading nested in a link)
[PodcastSection.tsx:200]— FEATURED eyebrow → `<h6>`

Separately, real card titles jump levels: SectionTitle is `<h2>` but FeaturedPodcastTitle is variant="h4" and MorePodcastTitle is variant="h5" (no `<h3>` between them).

Suggested Fix: pass component="span" (or "p") to the label/meta/link Typographies so they keep their style but shed heading semantics, and pick heading variants that don't skip a level for the card titles.

frontends/main/src/app-pages/PodcastPage/PodcastsListingPage/PodcastsListingPage.tsx
daniellefrappier18, line 84: Since PodcastsListingPage fetches entirely client-side when episodesData/podcastData are undefined, totalEpisodes/totalPodcasts fall back to 0 so on first paint the hero will render: 0 podcasts • 0 episodes • Updated daily, and the "Latest Episodes" / "Podcasts across MIT" sections render their headers over empty bodies.

Suggested Fix: gate the sections on loading (skeleton) and render empty/error states.

frontends/main/src/app-pages/PodcastPage/PodcastsListingPage/PodcastsListingPage.tsx
```diff
@@ -94,8 +94,7 @@ export const PodcastsListingPage: React.FC = () => {
   const morePodcasts = podcasts
```
daniellefrappier18: `featuredPodcasts` come from the curated learning path, while morePodcasts come from `useLearningResourcesList({ resource_type: [Podcast], sortby: New, limit: 5 })`. Nothing excludes the featured ids from the "More Podcasts" list, so a featured podcast that's also among the newest will render in both sections.

Suggested Fix: filter the featured ids out of morePodcasts before rendering.

frontends/main/src/app-pages/PodcastPage/PodcastEpisodeDetailPage.tsx
daniellefrappier18, line 235: For an episode belonging to multiple podcasts, the header/breadcrumb and the player bar disagree on which one to name. The header/breadcrumb picks the parent matching the URL's podcastId, but in helpers.ts the player bar always takes index [0]: `return episode.podcast_episode?.parent_podcasts?.[0]?.title ?? null`
So on the same page the two resolve differently: for an episode with parent_podcasts = [A, B] where the user is on B's URL, the header says B and the player bar says A. (Single-parent episodes are unaffected — find and [0] return the same podcast.)

daniellefrappier18 (changes requested):
Looks like you have some merge conflicts to resolve as well

daniellefrappier18 (approved):
LGTM

PR #3600, opened 2026-07-10
bump smoot-design

ChristopherChudzicki:
### What are the relevant tickets?
Related to #3576 

### Description (What does it do?)
Bumps smoot-design; per https://github.com/mitodl/smoot-design/releases, the only change that's really relevant here is new color `lightGray0`.

### How can this be tested?
1. Look at the frontend. It should look same.

### Additional Context
Bumping for use in  #3576

PR #3604, opened 2026-07-10
Sync pyproject.toml version with settings.py VERSION (0.73.4)

ChristopherChudzicki:
N/A

`pyproject.toml`'s `[project].version` was reset to `0.70.10` in #3220 while `main/settings.py` `VERSION` is `0.73.4`. The `test_bump_my_version_format` test (also added in #3220) asserts the two are equal, so it fails on `main` and on every branch cut from it — the failure on unrelated PRs (e.g. #3600) is this, not flakiness. This bumps `pyproject.toml` to `0.73.4` to match.

**Heads up — this will re-break on the next release.** The release-script bot (Doof) only bumps `settings.py`, not `pyproject.toml`, so they'll desync again and this test will fail after the next release. The real fix is to wire the release process to keep both in sync (or run bump-my-version) as part of the release-resource migration; the equality assertion was effectively merged ahead of that. Raising in standup.

### How can this be tested?
CI passes.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "mit-learn"
-version = "0.70.10"
+version = "0.73.4"
 description = "Search index for use with other MIT applications."
 authors = [{ name = "MIT ODL" }]
 requires-python = ">=3.12,<3.13"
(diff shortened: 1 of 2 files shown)


mbertrand (approved):
👍

PR #3607, opened 2026-07-13
Add learning material content to embeddings context

shanbady:

Closes https://github.com/mitodl/hq/issues/12073

This PR adds the content of learning materials (document type resources) to the embedding context - this is similar to what we do in opensearch where the contents are searched [via a subquery](https://github.com/mitodl/mit-learn/blob/47bdde3e0e29ab95b9ce5d0508703bab8cc21d73/learning_resources_search/api.py#L236)

### How can this be tested?
1. checkout this branch
2. Create a test learning material with contents we can search for later:
```python
from learning_resources.factories import LearningResourceFactory

CONTENT = (
    "This material covers zymurgy techniques for fermenting kelp "
    "inside hyperbolic paraboloid greenhouses on Mars."
)

# Test resource: generic title/description, sentinel ONLY in the content file
res = LearningResourceFactory.create(
    resource_type="document",
    title="Untitled material A",
    description="A learning material.",
    full_description="A learning material.",
    published=True,
)
ContentFile.objects.create(
    direct_learning_resource=res,   # must be the direct FK, not run/learning_resource
    key="test-sentinel-content",
    title="Test content",
    content=CONTENT,
    published=True,
)

# Control resource: identical metadata, no content file
control = LearningResourceFactory.create(
    resource_type="document",
    title="Untitled material B",
    description="A learning material.",
    full_description="A learning material.",
    published=True,
)
print(res.id, control.id)
```
3. generate embeddings for the materials:
```python
from vector_search.utils import embed_learning_resources
embed_learning_resources([res.id, control.id], "document", overwrite=True)
```
4. go to the search page with the [vector_search=true](http://open.odl.local:8062/search?q=fermenting+kelp+in+greenhouses+on+Mars+zymurgy&vector_search=true) param and note that the learning material we just created is surfaced.
 

### Additional Context
The embedding context adds a markdown header of "# content" to `steer` the results and help distinguish the title + description from contents. This is similar to what we do with markdown course info documents and I think will have a positive impact for this case.


vector_search/utils.py
```diff
@@ -463,11 +465,28 @@ def _chunk_markdown_documents(text, metadata):
...
+    content = "\n\n".join(
+        content_file["content"]
+        for content_file in document.get("content_files") or []
+        if content_file.get("content")
+    )
+    if content:
+        max_chars = settings.LEARNING_RESOURCE_EMBEDDING_CONTENT_MAX_CHARS
+        if max_chars and len(content) > max_chars:
+            content = content[:max_chars]
+        context = f"{context}\n\n# Content\n{content}"
```
mbertrand: **Already-embedded documents/videos never get content added to their vectors — even with `overwrite=True`.**

This function is also used by `should_generate_resource_embeddings`, which re-derives the embedding text from the stored Qdrant payload and compares it to the fresh doc's. Stored payloads already contain `content_files` (the serializer field predates this PR), so after deploy both sides fold in content and match — even though the stored vector was built from title + description only. `overwrite` never reaches `_process_resource_embeddings`, so the skip always applies; only `--recreate-collections` forces re-embedding.

Reproduced locally by simulating a resource embedded before this change:

```python
from learning_resources.factories import LearningResourceFactory
from learning_resources_search.serializers import serialize_bulk_learning_resources
import vector_search.utils as vsu
from vector_search.constants import RESOURCES_COLLECTION_NAME

res = LearningResourceFactory.create(resource_type="document", published=True)
cf = ContentFile.objects.create(
    direct_learning_resource=res, key="s", content="zymurgy kelp " * 50, published=True
)

# embed the way the OLD code did (metadata only)
orig = vsu._learning_resource_embedding_context
vsu._learning_resource_embedding_context = lambda d: (
    f"{d.get('title')} {d.get('description')} {d.get('full_description')}"
)
vsu.embed_learning_resources([res.id], "document", overwrite=True)
vsu._learning_resource_embedding_context = orig

doc = next(iter(serialize_bulk_learning_resources([res.id])))
pid = vsu.vector_point_id(vsu.vector_point_key(doc))
client = vsu.qdrant_client()
before = client.retrieve(RESOURCES_COLLECTION_NAME, ids=[pid], with_vectors=True)[0]

print(vsu.should_generate_resource_embeddings(doc))  # False — skipped
vsu.embed_learning_resources([res.id], "document", overwrite=True)
after = client.retrieve(RESOURCES_COLLECTION_NAME, ids=[pid], with_vectors=True)[0]
print(before.vector == after.vector)  # True — never re-embedded
# cleanup: delete cf before res (res.delete() alone hits a pre-existing constraint bug — see summary)
```

Fix: pass `overwrite` into `_process_resource_embeddings` to bypass the check, or add an embedding-version marker to the comparison.

main/settings.py
```diff
@@ -844,6 +844,10 @@ def get_all_config_keys():
     "CONTENT_SUMMARIZER_FLASHCARD_QUANTITY", 10
 )
 
+LEARNING_RESOURCE_EMBEDDING_CONTENT_MAX_CHARS = get_int(
+    name="LEARNING_RESOURCE_EMBEDDING_CONTENT_MAX_CHARS",
+    default=32000,
+)
```
mbertrand: **32,000 chars can exceed the 8,192-token input cap of `text-embedding-3-*`, and one over-limit doc fails its whole batch — permanently.**

Real course files run denser than the ~4 chars/token of clean prose: three edX files in my local DB measure 8,439–11,281 tokens in their first 32k chars. Reproduced through the actual code path with one of them (a UAI course policies file — any OCR'd, LaTeX-heavy, or non-English file gets similarly dense):

```python
import tiktoken
from vector_search.encoders.litellm import LiteLLMEncoder

cf = ContentFile.objects.get(
    edx_module_id="block-v1:UAI_SOURCE+UAI.ENT.1.AR+1T2026+type@policies+block@assets"
)
content = cf.content[: settings.LEARNING_RESOURCE_EMBEDDING_CONTENT_MAX_CHARS]
print(len(tiktoken.get_encoding("cl100k_base").encode(content)))  # 11,281 tokens

doc = f"Title Description Full description\n\n# Content\n{content}"
LiteLLMEncoder("text-embedding-3-small").embed_documents([doc, "hello world"])
# litellm.BadRequestError (400): "Invalid 'input[0]': maximum input length is 8192 tokens."
# whole batch fails; "hello world" alone embeds fine
```

Batches are `QDRANT_CHUNK_SIZE` resources, and `generate_embeddings` treats a 400 as non-transient, so every resource in the batch permanently fails to embed.

Fix: truncate by tokens (tiktoken is already a dep; `LiteLLMEncoder.token_encoding_name` resolves the encoding), or lower the default to ~20k chars for real headroom.

vector_search/utils.py
```diff
@@ -463,11 +465,28 @@ def _chunk_markdown_documents(text, metadata):
...
         f"{document.get('description')} {document.get('full_description')}"
     )
+    content = "\n\n".join(
+        content_file["content"]
+        for content_file in document.get("content_files") or []
+        if content_file.get("content")
+    )
+    if content:
+        max_chars = settings.LEARNING_RESOURCE_EMBEDDING_CONTENT_MAX_CHARS
+        if max_chars and len(content) > max_chars:
```
mbertrand: Minor: setting this to 0 disables *truncation*, not content — unbounded content instead of none. Document that, or make 0 mean "skip content".

vector_search/utils_test.py
```diff
@@ -1024,6 +1027,73 @@ def test_should_generate_for_changed_resource(mocker):
     assert result is True
 
 
+@pytest.mark.parametrize(
+    "resource_type_group",
+    [LEARNING_MATERIAL_RESOURCE_TYPE_GROUP, "course", "program"],
+)
+def test_embedding_context_includes_content_files(resource_type_group):
```
mbertrand: These tests hand-craft the `content_files` dicts, so if the serializer ever stopped emitting that field the feature would silently no-op with green tests. One end-to-end test closes that (`serialize_bulk_learning_resources` comes from `learning_resources_search.serializers`):

```python
def test_embedding_context_includes_serialized_content_files():
    """Content from a real serialized document resource ends up in the context"""
    resource = LearningResourceFactory.create(resource_type="document", published=True)
    ContentFileFactory.create(
        direct_learning_resource=resource, content="sentinel text", published=True
    )
    serialized = next(iter(serialize_bulk_learning_resources([resource.id])))
    assert "sentinel text" in vs_utils._learning_resource_embedding_context(serialized)
```

Also: the `resource_type_group` parametrization is decorative — the function never reads it, and only the document/video serializers emit `content_files` (the docstring's "regardless of resource type" overstates for the same reason). A test that a content-only change flips `should_generate_resource_embeddings` to `True` would cover the staleness subtlety from my other comment.

mbertrand (changes requested):

main/settings.py
shanbady: I ended up creating a utility method that infers the model+token limits via litellm and uses that or falls back to a 20k character default so gives us best of both (i removed the setting and just made the fallback a constant

vector_search/utils.py
shanbady: removed this code altogether (see above)

vector_search/encoders/utils.py
```diff
@@ -24,3 +31,49 @@ def sparse_encoder():
...
+
+            return encoding.decode(tokens[:max_input_tokens])
+        except Exception:
+            logger.exception(
+                "Failed to truncate using tiktoken encoding '%s'",
+                token_encoding_name,
+            )
+
+    # Fall back to a simple character limit.
+    return text[:DEFAULT_TRUNCATE_CHARS]
```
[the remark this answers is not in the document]
shanbady: added cached logging mechanism

vector_search/utils_test.py
shanbady: fixed

vector_search/encoders/utils.py
```diff
@@ -24,3 +31,80 @@ def sparse_encoder():
...
+    Returns:
+        Truncated text.
+    """
+    max_input_tokens = None
+
+    try:
+        info = get_model_info(model)
+        max_input_tokens = info.get("max_input_tokens")
+    except Exception:
+        logger.exception("Failed to determine max_input_tokens for model %s", model)
```
mbertrand: This got triggered locally, could get very noisy in sentry maybe, depending on whether or not the model selected there supports this or not.  Would `info` or even `debug` be adequate?

mbertrand (approved):
👍 , one suggestion on logging failure to retrieve max_input_tokens

vector_search/encoders/utils.py
shanbady: makes sense. switched to debug

vector_search/utils.py
```diff
@@ -463,11 +469,29 @@ def _chunk_markdown_documents(text, metadata):
 def _learning_resource_embedding_context(document):
     """
```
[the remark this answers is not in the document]
shanbady: this is not actually an issue.

PR #3610, opened 2026-07-14
improve vector search endpoint performance 

shanbady:

Closes https://github.com/mitodl/hq/issues/12255

This PR resolves some issues with the performance on the vector search endpoint which makes sync_to_async calls using thread_sensitive=True which ends up queuing calls on a single shared thread.

### How can this be tested?
Testing the difference locally is a bit difficult (requires actual load and same settings as prod) but you can see the effect of thread_sensitive=True vs thread_sensitive=False by running the following script which simulates the situation:
```python
import asyncio, time
from asgiref.sync import sync_to_async

def blocking_sync_work(n):
    time.sleep(0.25)   # stand-in for PG hydration + DRF serialization
    return n

async def run(thread_sensitive, concurrency=8):
    start = time.monotonic()
    await asyncio.gather(*[
        sync_to_async(blocking_sync_work, thread_sensitive=thread_sensitive)(i)
        for i in range(concurrency)
    ])
    return time.monotonic() - start

async def main():
    before = await run(thread_sensitive=True)    # current prod behavior
    after  = await run(thread_sensitive=False)   # the fix
    print(f"thread_sensitive=True  (shared 1 thread): {before:.2f}s")
    print(f"thread_sensitive=False (thread pool):     {after:.2f}s")

asyncio.run(main())
```


vector_search/views_test.py
```diff
@@ -940,7 +940,7 @@ def test_content_file_search_restricts_resource_query_to_best_run(
     assert not any(getattr(c, "key", None) == "resource_readable_id" for c in must)
 
 
-@pytest.mark.django_db
+@pytest.mark.django_db(transaction=True)
```
mbertrand, line 954: Not this line itself, but its sibling just below: `test_content_file_search_no_best_run_metadata_only` (line 1034) still has plain `@pytest.mark.django_db`. With `best_run_ids_for_resources` now running on a pool thread (its own DB connection), it can no longer see the uncommitted course + unpublished-run fixtures — so it returns `[]` because the *resource isn't found*, not because the run is unpublished. The assertion (`{course.readable_id}`) passes either way, which means the "unpublished runs are excluded" behavior is no longer actually exercised.

Same one-line fix as the two tests you already updated:

```python
@pytest.mark.django_db(transaction=True)
def test_content_file_search_no_best_run_metadata_only(
```

vector_search/views.py
```diff
@@ -391,9 +393,13 @@ async def _async_vector_hits(  # noqa: PLR0913
...
         if search_collection == RESOURCES_COLLECTION_NAME:
-            return await sync_to_async(_resource_vector_hits)(search_result)
+            return await sync_to_async(_resource_vector_hits, thread_sensitive=False)(
+                search_result
+            )
         else:
-            return await sync_to_async(_content_file_vector_hits)(search_result)
+            return await sync_to_async(
+                _content_file_vector_hits, thread_sensitive=False
+            )(search_result)
```
mbertrand: +1 to Copilot's comments here and at line 711 — with one caveat: Sentry's suggested fix (moving the ORM work back out of the thread pool) would reintroduce the exact single-shared-thread bottleneck this PR removes. The `close_old_connections()` wrapper is the right shape.

Two specifics the bot comments don't spell out:

- `CONN_MAX_AGE=0` (main/settings.py:283) is silently unenforced on pool threads: Django only runs `close_old_connections` via the `request_started`/`request_finished` signals, which fire on the handler thread, and `connections` is thread-local. So each of up to `min(32, cpu + 4)` executor threads per worker keeps its lazily-opened Postgres connection until process restart.
- The sharper failure mode isn't connection *count*, it's staleness: if anything reaps an idle connection (pgbouncer/RDS-proxy idle timeout, NAT, DB failover), the next request landing on that thread raises `OperationalError` and Django never recovers it — `close_if_unusable_or_obsolete` never runs on these threads, and `CONN_HEALTH_CHECKS` can't help for the same reason. That thread then throws a 500 on every request it picks up until the process restarts.

A small helper covers all three ORM offload sites (the two hits calls here and `best_run_ids_for_resources` at line 709):

```python
from django.db import close_old_connections


def db_sync_to_async(func):
    """Offload sync DB work to the thread pool, with per-call connection cleanup."""

    def wrapper(*args, **kwargs):
        close_old_connections()
        try:
            return func(*args, **kwargs)
        finally:
            close_old_connections()

    return sync_to_async(wrapper, thread_sensitive=False)
```

then e.g. `return await db_sync_to_async(_resource_vector_hits)(...)`. The two `serialize()` closures don't hit the DB today, so they can stay plain `sync_to_async` (or use the helper too, for uniformity — it's harmless).

vector_search/views.py
```diff
@@ -631,7 +637,7 @@ def serialize():
                         response, context={"request": request}
                     ).data
 
-                response_data = await sync_to_async(serialize)()
+                response_data = await sync_to_async(serialize, thread_sensitive=False)()
```
mbertrand: Follow-up to my comment at line 402: I'd route this call and its twin at line 743 through `db_sync_to_async` as well, rather than leaving them plain.

To be clear about the risk: these closures don't touch the DB today — by the time they run, `hits` are already plain dicts (model serialization happens upstream in `_resource_vector_hits` / `_content_file_vector_hits`, which the helper covers), and both response serializers' `get_results` / `get_count` / `get_metadata` only shape those dicts. The latent risk is a future change adding DB access to the response serializers (e.g. a `SerializerMethodField` reading `context["request"].user`) — that would silently start opening uncleaned pool-thread connections, and nothing would catch it. The helper costs nothing here, and it makes the rule uniform and easy to enforce in review: anything offloaded to the pool goes through `db_sync_to_async`.

mbertrand (changes requested):

vector_search/views_test.py
shanbady: fixed

vector_search/utils.py
```diff
@@ -1070,6 +1096,15 @@ async def check_missing_content_file_ids(edx_module_ids, collection_name):
...
+    # Trim edge whitespace so the exact-match DB/Qdrant probes below can't
+    # produce false misses for ids that differ only by transport junk.
+    edx_module_ids = [
+        eid
+        for eid in ((eid or "").strip() for eid in edx_module_ids)
+        if is_loggable_missing_content_id(eid)
+    ]
+    if not edx_module_ids:
+        return
     present = await sync_to_async(present_edx_module_ids)(edx_module_ids)
```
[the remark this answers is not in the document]
mbertrand: Worth doing this

mbertrand (approved):
LGTM, but worth addressing the final sentry comment before merging

PR #3621, opened 2026-07-15
Ab/program content files

abeglova:
closes https://github.com/mitodl/hq/issues/12338
closes https://github.com/mitodl/hq/issues/12382
closes https://github.com/mitodl/hq/issues/11833

This pr 
1) Updates the programs search index to index content files
2) Updates the courses index to index content files linked to a learning resource instead of a learning resource run

This adds marketing pages to search for both courses and programs

### How can this be tested?
run docker-compose run web ./manage.py  recreate_index --all

run docker-compose run web ./manage.py  backpopulate_mitxonline_data

From the shell run
```
from learning_resources.tasks import *

for lr in LearningResource.objects.filter(etl_source='mitxonline'):
   if lr.url:
         lr.save()

scrape_marketing_pages()
```

Go to
http://open.odl.local:8062/search?min_score=0&q=DEDP&resource_type_group=program

you should see two programs

https://learn.mit.edu/search?q=DEDP&resource_type_group=program&min_score=0

has no results.

I did have to set min_score=0 to see the programs. I'm hoping that with more resources in the index they would be visible without that. If that is not the case on rc i will do a followup pr


learning_resources_search/indexing_api.py
```diff
@@ -430,43 +431,35 @@ def index_percolators(ids, index_types):
     )
 
 
-def index_course_content_files(learning_resource_ids, index_types):
```
abeglova, line 433: This was already orphaned and not called anywhere prior to this pr. Removing it

pdpinch:
Do you think this will actually close https://github.com/mitodl/hq/issues/11833 or just mitigate it?

learning_resources_search/tasks.py
```diff
@@ -784,11 +844,101 @@ def get_update_resource_files_tasks(blocklisted_ids, etl_source):
...
+    Get list of tasks to update program content files.
+
+    Args:
+        etl_source(str): ETL source filter for the task
+    """
+    program_update_query = LearningResource.objects.filter(
+        published=True, resource_type=PROGRAM_TYPE
+    ).order_by("id")
+
+    if etl_source:
```
mbertrand, line 931: This program path is missing the `RESOURCE_FILE_ETL_SOURCES` guard that the course equivalent applies. `get_update_resource_files_tasks` short-circuits with `if etl_source is None or etl_source in RESOURCE_FILE_ETL_SOURCES:` and, when `etl_source` is `None`, further restricts the query to `etl_source__in=RESOURCE_FILE_ETL_SOURCES`. Here, when `etl_source is None`, we iterate over every published program of every source. It's harmless today (non-file sources simply match no content files), but it diverges from the course path and may run some needless queries.

learning_resources_search/tasks.py
```diff
@@ -784,11 +844,101 @@ def get_update_resource_files_tasks(blocklisted_ids, etl_source):
...
+                    )
+                ]
+            )
+
+        return index_tasks
+    else:
+        return []
+
+
+def get_update_program_files_tasks(etl_source):
```
mbertrand, line 917: This function is very similar to `get_update_resource_files_tasks` (run-level index + resource-level index + run-level deindex + resource-level deindex), differing only by `resource_type=PROGRAM_TYPE` and the dropped blocklist/guard. The same four-block pattern is also duplicated in the recreate path (`start_recreate_index`, the `if PROGRAM_TYPE in indexes:` block mirrors the course block), which is what pushed `start_recreate_index` over the complexity limit and required the new `# noqa: C901`.

Might be worth extracting a helper, e.g. `content_file_index_tasks(resource_id, index_type, resource_type)` returning the run+resource index/deindex signatures, and calling it from all four sites. This is de-duplicating already-duplicated code rather than premature abstraction, and it would let the `C901` suppression be removed.  Could be done in a followup PR though.

learning_resources_search/tasks.py
```diff
@@ -694,6 +708,49 @@ def start_recreate_index(self, indexes, remove_existing_reindexing_tags):
                     )
                 ]
 
+        if PROGRAM_TYPE in indexes:
```
mbertrand, line 711: Test-coverage gap on the core new behavior. In `test_start_recreate_index`, the programs are created with `ProgramFactory.create_batch(4, courses=[])` and no `ContentFileFactory`, so this block emits zero `index_content_files.si` calls and the program branch is never asserted against `index_files_mock`. Same in `test_start_update_index` (programs have no content files), so `get_update_program_files_tasks` is never exercised with data; and the resource-level (marketing-page) branch added to `get_update_resource_files_tasks` is untested too since test courses only get run-level `ContentFile`s.

Since indexing program content files (#12338) and resource-level content files (#12382) are the point of the PR, a regression in these paths would pass CI. Suggest extending the existing tests: attach both a run-level and a direct (`learning_resource=`) `ContentFile` to a program, and assert `index_content_files.si` is called with `resource_type=PROGRAM_TYPE`; likewise add a direct `ContentFile` to a course and assert the default-type call.

mbertrand (approved):
LGTM overall, found a few nitpicks but I don't think any of them are blockers, so feel free to fix, add to a followup PR, or ignore.

learning_resources_search/tasks.py
abeglova: fixed

learning_resources_search/tasks.py
abeglova: fixed

learning_resources_search/tasks.py
abeglova: I'll handle this in a followup

mbertrand (commented):
👍

mbertrand (approved):

PR #3628, opened 2026-07-16
refactor: Extract a PodcastPageShell for shared player/breadcrumb wiring  and Extract a shared audio engine from the two podcast players

ahtesham-quraish:
https://github.com/mitodl/hq/issues/12384
https://github.com/mitodl/hq/issues/12387
https://github.com/mitodl/hq/issues/12335

- As a developer, I want the podcast player wiring and breadcrumb bar in one shell, so the three podcast pages stop copy-pasting the same setup.
- As a podcast listener, I want a clear message when a page is loading or fails, so I don't stare at a blank layout or silently-empty text.
- As a developer maintaining the podcast players, I want the audio playback logic to live in one place, so I can fix bugs and add features once instead of keeping two ~500-line files in sync.

- [ ] Desktop screenshots

### How can this be tested?
For testing we need to test all the podcasts pages like `/podcasts` page and podcast detail page and episode detail page as well, we should also check if the player plays the audio fine. We should check the embed page as well whose url is given here `http://open.odl.local:8062/podcast/embed/137277`


frontends/main/src/app-pages/PodcastPage/AudioPlayer.styled.ts
```diff
@@ -0,0 +1,150 @@
...
+export const PlayPauseButton = styled("button", {
+  shouldForwardProp: (prop) =>
+    prop !== "buttonSize" && prop !== "mobileButtonSize",
+})<{ buttonSize?: number; mobileButtonSize?: number }>(
+  ({ theme, buttonSize = 48, mobileButtonSize }) => ({
+    background: "none",
+    border: "none",
+    cursor: "pointer",
+    padding: 0,
+    // Fixed size + overflow:hidden keeps the spinner clipped inside the button.
```
gumaerc: It seems like this comment should be moved down to where the fixed size and overflow are actually set?

frontends/main/src/app-pages/PodcastPage/AudioPlayer.styled.ts
```diff
@@ -0,0 +1,150 @@
...
+    WebkitAppearance: "none",
+    flex: 1,
+    height: `${trackHeight}px`,
+    borderRadius: `${trackHeight / 2}px`,
+    cursor: "pointer",
+    outline: "none",
+    border: "none",
+    padding: 0,
+    background: `linear-gradient(to right, ${theme.custom.colors.red} ${percent}%, ${theme.custom.colors.lightGray2} ${percent}%)`,
+    "&::-webkit-slider-thumb": {
```
gumaerc, line 121: There seems to be some kind of issue with transparency here. The circular thumb at the end of the slider obscures the bar behind it, leaving two little notches of grey background on the left hand side of the drag handle.

Image

gumaerc (approved):
Everything seems to work well for me. I just noted a couple nitpick problems. The comment is whatever, but we should probably fix the issue with the track thumb.

frontends/main/src/app-pages/PodcastPage/AudioPlayer.styled.ts
ahtesham-quraish: > There seems to be some kind of issue with transparency here. The circular thumb at the end of the slider obscures the bar behind it, leaving two little notches of grey background on the left hand side of the drag handle.
> 
> Image

I am looking into this before merge.

ahtesham-quraish: fixed the issue

PR #3629, opened 2026-07-16
Sanitize LLM-generated summaries/flashcards before saving ContentFile

mbertrand:

Fixes mitodl/hq#12369

The summarizer LLM occasionally emits characters that a valid Python string can hold but Postgres cannot store: **NUL (`\x00`)** and **lone UTF-16 surrogates (`\ud800`–`\udfff`)**. When either appears in a generated summary or flashcard, `content_file.save()` in `ContentSummarizer.summarize_single_content_file` fails (`DataError: unsupported Unicode escape sequence` for NUL; client-side `UnicodeEncodeError` for lone surrogates). The save happens inside a `transaction.atomic()` block, so nothing persists — the file is then flagged as missing a summary by `embeddings_healthcheck` and retried indefinitely, failing identically each time. (Sentry OPEN-NEXT-1YW: 368 occurrences since Aug 2025; French-language `.srt` content is most prone.)

This adds a small `sanitize_llm_text` utility and applies it at the two points where LLM output enters the summarizer:

- **`learning_resources/utils.py`** — new `sanitize_llm_text(text)`: strips NUL and replaces lone surrogates via `encode("utf-8", errors="replace").decode("utf-8")`. Clean text (including accented non-ASCII) passes through unchanged; the result is always strictly UTF-8-encodable.
- **`learning_resources/content_summarizer.py`** — `_generate_summary` returns sanitized text; `_generate_flashcards` sanitizes the string values of each parsed flashcard dict.

NUL and lone surrogates are the only two things a valid Python `str` can contain that Postgres `text`/`jsonb` reject, so this closes both failure modes.

**Scope notes:** fixes the summarizer path only (source of all 368 Sentry events). A `ContentFile.save()` model-level guard was considered and rejected — bulk `queryset.update()` writes bypass `save()`, and it would scan the large `content` field on every ETL write. No migration or backfill: the affected files self-heal on the next `embeddings_healthcheck` retry once this deploys.

### How can this be tested?

Make sure you have an enabled `ContentSummarizer` for mitxonline (check in django admin, create if not).

The following will error on main, should work on this branch:

```python
from unittest.mock import MagicMock, patch
from learning_resources.content_summarizer import ContentSummarizer

# Pick a real mitxonline content file that will actually summarize:
# non-empty content + platform's active summarizer config allows its type/extension.
cf = next(
    (
        c
        for c in ContentFile.objects.filter(
            run__learning_resource__etl_source="mitxonline"
        )
        .exclude(content="")
        .exclude(content__isnull=True)
        .iterator()
        if summarizer._can_process_content_file(c)
    ),
    None,
)
assert cf is not None, "No processable mitxonline content file found — check the summarizer config is active and its allowed extensions/content types match ingested files"
print("using content file", cf.id, cf.file_extension, cf.content_type)

# Force the LLM to emit the two Postgres-unstorable characters.
with patch("learning_resources.content_summarizer.ChatLiteLLM") as MockLLM:
    inst = MockLLM.return_value
    resp = MagicMock(); resp.content = "summary with nul \x00 and surrogate \ud800"
    inst.invoke.return_value = resp
    inst.with_structured_output.return_value.invoke.return_value = {
        "flashcards": [{"question": "q\x00", "answer": "a\ud800"}]
    }
    print(summarizer.summarize_single_content_file(cf.id, overwrite=True))

cf.refresh_from_db()
print(repr(cf.summary))
print(repr(cf.flashcards))
```

Before this change the `generate_embeddings` step fails on `cf.save()`; after it, the summary is populated with sanitized text.


learning_resources/utils.py
```diff
@@ -798,6 +798,11 @@ def truncate_to_tokens(text: str, max_tokens: int, model: str = "gpt-4o") -> str
     return encoding.decode(tokens[:max_tokens])
 
 
+def sanitize_llm_text(text: str) -> str:
+    """Strip characters Postgres can't store: NUL and lone surrogates."""
```
abeglova: Can you change strip to replace in the docstring since that is what the function is doing?

abeglova (approved):
LGTM

PR #3634, opened 2026-07-17
fix: raise SCAN COUNT in clear_views_cache to cut Redis scan time + clear cache less often during website_content requests

ChristopherChudzicki:
- Closes https://github.com/mitodl/hq/issues/12443

This PR:
1. sets an itersize on `cache.delete_pattern` to increase its performance; with 300k keys, it took about 30 seconds with default size (10) and 0.45 seconds with size 1000.
2. Changes website_content views slightly:
    - only clear cache after database changes are committed

### How can this be tested?
1. View `/news` on the frontend ... This ensures something at least is in your views cache. (FeedSourceViewSet caches its listing)
2. In a django shell, `./manage.py shell` and `from main.utils import clear_views_cache`, `clear_views_cache()`. It should succeed. It should return a nonzero number—number of keys deleted—so you know it still works.
3. Try creating a new news page via https://learn.mit.dev/website_content/news/new ... save it. 
    - if you save it without publishing, it should NOT show up on `/news` frontend page
    - when you publish it , it SHOULD show up
    - if you change the article's title and publish it again, it should update on `/news` frontend listing page.

### DIFF
--- a/main/utils.py
+++ b/main/utils.py
@@ -589,9 +589,8 @@ def clear_views_cache(key_prefix: str | None = None) -> int:
         # Redis with the Celery broker + result backend) needs ~keyspace/10
         # round-trips. Raising it ~100x cuts a multi-second scan to a fraction.
         return cache.delete_pattern(pattern, itersize=1000)
-    if hasattr(cache, "keys"):
-        return cache.delete_many(cache.keys(pattern)) or 0
-    # Backends without pattern/key enumeration (e.g. LocMemCache) can only flush.
+    # Backends without SCAN-based pattern deletion (DummyCache in tests,
+    # LocMemCache) can't clear selectively -- flush everything instead.
     cache.clear()
     return 0
 
--- a/main/utils.py
+++ b/main/utils.py
@@ -584,7 +584,11 @@ def clear_views_cache(key_prefix: str | None = None) -> int:
     )
 
     if hasattr(cache, "delete_pattern"):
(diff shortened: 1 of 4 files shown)


main/utils.py
```diff
@@ -584,10 +584,13 @@ def clear_views_cache(key_prefix: str | None = None) -> int:
     )
 
     if hasattr(cache, "delete_pattern"):
-        return cache.delete_pattern(pattern)
-    if hasattr(cache, "keys"):
-        return cache.delete_many(cache.keys(pattern)) or 0
```
ChristopherChudzicki, line 589: Claude wanted to delete this branch of this func, and I agree. Its rationale:

- `RedisCache` has both. `delete_pattern` and `keys`
    - `delete_pattern` → https://github.com/jazzband/django-redis/blob/6.0.0/django_redis/cache.py#L107
    - `keys` → https://github.com/jazzband/django-redis/blob/6.0.0/django_redis/cache.py#L140
- `DummyCache` which we use in tests has neither method
    - https://github.com/django/django/blob/4.2.30/django/core/cache/backends/dummy.py
- `LocMemCache` also has neither , as the comment indicates

mbertrand (approved):
It actually took a few minutes initially when I published some new articles before they showed up on the news page, but I think that's because the tasks to convert articles to news items got backed up behind a bunch of contentfile-related tasks that were running.  Once I cleared those out, they showed up pretty much instantly.

PR #3636, opened 2026-07-17
Add unsubscribe API

rhysyngsun:

Part of https://github.com/mitodl/hq/issues/8424

This PR adds an API to unsubscribe a given user based on a passed access token. This permits an unsubscribe to occur without login. This was pulled out of https://github.com/mitodl/mit-learn/pull/3549 because that has frontend changes and based on feedback we need to have a confirmation in front of unsubscribe links in the footer so this requires the API to be put into the OpenAPI schema separately from that PR.

### How can this be tested?

Tests should pass

---

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>


profiles/serializers.py
```diff
@@ -151,9 +155,32 @@ def update(self, instance, validated_data):
...
+                        {
+                            "email_optin": (
+                                "Unable to update email preferences at this time."
+                            )
+                        }
+                    ) from exc
+
             update_image = "image_file" in validated_data
             instance.save(update_image=update_image)
             return instance
```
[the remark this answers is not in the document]
rhysyngsun: This is intentional, we don't want keycloak to override the local state and if the operation fails we need to give that signal back to the user

users/migrations/0011_user_unsubscribe_uuid.py
```diff
@@ -0,0 +1,19 @@
...
+
+class Migration(migrations.Migration):
+    dependencies = [
+        ("users", "0010_alter_user_global_id"),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name="user",
+            name="unsubscribe_uuid",
```
mbertrand, line 14: This migration will fail if more than 1 user exists.

`AddField` evaluates the `uuid.uuid4` default **once**, so every existing row is backfilled with the *same* UUID and the `unique=True` index can't be built. Tests pass only because the test DB creates `users_user` empty (0 rows → no collision); RC/prod has real users, so `migrate` will abort.

Reproduced against my local 3-users table: 
```bash
psycopg.errors.UniqueViolation: could not create unique index "users_user_unsubscribe_uuid_key"
DETAIL:  Key (unsubscribe_uuid)=(df797a77-3bfc-441b-bde7-868d87b90b2c) is duplicated.
```

mbertrand (changes requested):

users/migrations/0011_user_unsubscribe_uuid.py
rhysyngsun: I had that addressed at some point, Claude must've stomped over it. Fixed now by setting the field to default to `None`.

mbertrand (approved):
👍

Issue #3624, opened 2026-07-16
mitlearn-redis-production memory saturation from unbounded Celery result retention


feoh:
### Description/Context

On-call was paged with 4 CloudWatch `DatabaseMemoryUsagePercentage` alerts for `mitlearn-redis-production` on 2026-07-15, all crossing the 90% threshold at ~99.98%. This is a recurring page. This issue documents the root-cause investigation and tracks the immediate mitigation; deeper architectural fixes are tracked in #3622 and #3623.

**Sample alert:** `DatabaseMemoryUsagePercentage - Threshold Crossed: [99.98806233455291, 90.96114310065028] > 90.0` on `CacheClusterId: mitlearn-redis-production-003` (also fired for `-001`), account `610119931565`, region `us-east-1`.

#### Topology

`mitlearn-redis-production` is a **single-shard** Valkey 7.2.6 replication group — 1 primary + 2 replicas (`-001` / `-002` / `-003`), each `cache.r7g.xlarge` (~26 GB). The alerts fire together because all three nodes hold the **same** dataset. This one instance simultaneously serves as:

- Celery **broker** (`CELERY_BROKER_URL`)
- Celery **result backend** (`CELERY_RESULT_BACKEND`)
- **RedBeat** scheduler store (`redbeat_redis_url`)
- Django **`redis` response cache** (`CACHES["redis"]`, LOCATION = `CELERY_BROKER_URL`)

Config: https://github.com/mitodl/mit-learn/blob/2c6caa68179725c90e17237acb50bc1125ed4573/main/settings_celery.py#L20-L26 and https://github.com/mitodl/mit-learn/blob/2c6caa68179725c90e17237acb50bc1125ed4573/main/settings.py#L537-L540

Parameter group `mitlearn-redis-production-72-parameter-group`: `maxmemory-policy = allkeys-lru`, `maxmemory-samples = 3`, `reserved-memory-percent = 25`.

#### What the metrics show (CloudWatch, primary node `-001`)

Memory was stable for 36+ hours, then spiked to 99.98% in a single hour while the **item count barely moved** and **evictions stayed at 0** — ruling out a view-cache key explosion:

| Time (UTC bucket) | DBMemoryUsage % | BytesUsedForCache | CurrItems | Evictions |
|---|---|---|---|---|
| 2026-07-15 18:11 | 59.25 | 12.56 GB | 439,306 | 0 |
[...]

Issue #3622, opened 2026-07-16
Move Celery result backend off the shared mitlearn broker/cache Redis


feoh:
### Description/Context

`mitlearn-redis-production` is a single-shard Valkey 7.2.6 replication group (1 primary + 2 replicas, `cache.r7g.xlarge`, ~26 GB) that simultaneously serves as the **Celery broker**, the **Celery result backend**, the **RedBeat scheduler store**, and the Django **`redis` response cache**. All of these compete for the same memory, so a burst in any one of them (notably Celery task results from a catalog reindex fan-out) can saturate the instance and page on-call with `DatabaseMemoryUsagePercentage` alerts.

Relevant config:
- https://github.com/mitodl/mit-learn/blob/2c6caa68179725c90e17237acb50bc1125ed4573/main/settings_celery.py#L20-L26 (`CELERY_BROKER_URL` and `CELERY_RESULT_BACKEND` both default to `REDIS_URL`)
- https://github.com/mitodl/mit-learn/blob/2c6caa68179725c90e17237acb50bc1125ed4573/main/settings.py#L537-L540 (Django `redis` cache also points at `CELERY_BROKER_URL`)

Observed impact (CloudWatch, prod, 2026-07-15): memory jumped from a stable ~13.5 GB (~60%) to 21.2 GB (99.98%) in a single hour, driven by a ~140x spike in Redis list commands (Celery chord result collection) — see the root-cause issue for full data.

### Plan/Design

Give Celery task results their own memory budget so index fan-outs cannot starve the broker and the Django response cache. Options, roughly in order of preference:

1. Point `CELERY_RESULT_BACKEND` at a **separate ElastiCache instance** (or at minimum a separate logical Redis DB) sized for result churn.
2. Alternatively, use the **database result backend** (`django-celery-results`) for the high-volume fan-out tasks so results don't live in Redis at all.
3. If kept on Redis, ensure the result backend's eviction/expiry cannot evict live broker queue entries.

Coordinate the infra change (Pulumi / ElastiCache) with the settings change. Verify `redbeat_redis_url` and the Django `redis` cache remain on the broker instance (or are likewise separated) as appropriate.

PR #3639, opened 2026-07-17
Batch catalog reindex into sequential chords to bound Redis broker pressure

### LINKED ISSUE #3623: Batch catalog reindex to avoid a single catalog-wide Celery chord
### Description/Context

`start_recreate_index` and `start_update_index` build **one enormous `celery.group`** — a subtask per 100-item chunk (`OPENSEARCH_INDEXING_CHUNK_SIZE`, default 100) of *every* content file and learning resource in the catalog — then wrap it in a chain, which Celery executes as a single catalog-wide **chord**:

- https://github.com/mitodl/mit-learn/blob/2c6caa68179725c90e17237acb50bc1125ed4573/learning_resources_search/tasks.py#L692-L700 (`start_recreate_index`: `celery.group(index_tasks)` then `self.replace(celery.chain(index_tasks, finish_recreate_index.s(...)))`)
- https://github.com/mitodl/mit-learn/blob/2c6caa68179725c90e17237acb50bc1125ed4573/learning_resources_search/tasks.py#L746-L757 (`start_update_index`: same pattern)
- https://github.com/mitodl/mit-learn/blob/2c6caa68179725c90e17237acb50bc1125ed4573/main/settings.py#L584-L588 (chunk size default 100)

Celery's Redis result backend synchronizes a chord by having **every** subtask `RPUSH` its result into a per-group list and poll `LLEN` against the expected count. With thousands of subtasks over the full catalog, this produced a ~140x spike in Redis list commands (80k/hr → 11.4M/hr) and drove `mitlearn-redis-production` memory from ~60% to 99.98% in one hour on 2026-07-15. These reindexes are launched from the `recreate_index` / `update_index` management commands (manual / deploy-time), so the saturation recurs and pages on-call each time. Full data in the root-cause issue.

Note: `finish_recreate_index(results, backing_indices)` genuinely consumes the subtask results (`merge_strings(results)`), so the results cannot simply be ignored — https://github.com/mitodl/mit-learn/blob/2c6caa68179725c90e17237acb50bc1125ed4573/learning_resources_search/tasks.py#L924-L953

### Plan/Design

Bound the in-flight footprint of a reindex so a single run cannot saturate the shared Redis:

- Split the catalog-wide chord into **sequential batches** (e.g. chord-per-index or chord-per-N-chunks), so only one batch's results occupy Redis at a time.
- Consider driving the fan-out from a producer that throttles how many subtasks are in flight, rather than materializing the entire group up front.
- Combine with moving the result backend off the broker Redis (see companion issue) for defense in depth.

Preserve current behavior: error aggregation via `merge_strings(results)` and the final alias swap in `finish_recreate_index` must still run once all batches complete.

feoh:
Closes #3623

`start_recreate_index` and `start_update_index` previously materialized a subtask for every ~100-item chunk of the entire catalog and dispatched all of them as **one giant celery chord**. With the Redis result backend, every subtask RPUSHes its result into a single per-group list, so a full reindex produced a massive burst of Redis list/set commands and result keys on the shared `mitlearn-redis` broker/cache instance — the failure mode behind the 2026-07-15 memory saturation (#3624) and the ongoing paging.

This change bounds the in-flight footprint:

- **`replace_with_batched_reindex`** splits the index subtask list into batches of at most `OPENSEARCH_REINDEX_BATCH_SIZE` (new setting, default 100) and dispatches only the first batch as a chord.
- **`continue_reindex_batches`** (each batch chord's callback) accumulates that batch's error strings, then dispatches the next batch — so only one batch's chord bookkeeping lives on Redis at a time — and finally invokes the original finish signature with all accumulated results.
- Batch state (remaining subtask signatures, interim errors, the finish signature) is stored in the **durable database cache** keyed by the originating task's run id, keeping the multi-MB batch snapshot off Redis entirely. Keys are deleted when the run completes, with a 7-day TTL as a leak backstop.
- Existing semantics are preserved: `finish_recreate_index` still receives the merged error strings and performs the alias swap exactly once after all batches complete (`merge_strings` aggregation unchanged), and `finish_update_index` still clears the views cache. Callers waiting on the original task still wait for the full chain via `Task.replace`.

If there are zero index subtasks, the task skips straight to the finish signature (previously an empty chord was dispatched).

### How can this be tested?
- `pytest learning_resources_search/tasks_test.py` — includes new tests covering batch splitting/storage (`test_replace_with_batched_reindex`), the no-op path, sequential dispatch + error accumulation (`test_continue_reindex_batches_*`), cache cleanup on completion, and loud failure when batch state is missing/expired.
- Full affected suite run locally: `learning_resources_search/tasks_test.py`, `learning_resources/tasks_test.py`, `vector_search/tasks_test.py`, `learning_resources_search/management` — 163 passed.
- Manual validation: run `./manage.py recreate_index --all` (or `update_index`) against a populated environment and observe on the Redis side that chord result lists stay bounded (≤ `OPENSEARCH_REINDEX_BATCH_SIZE` entries) instead of growing to catalog size, while the final alias swap still occurs once at the end.

### Additional Context
- Companion mitigations already merged: #3625 (`CELERY_RESULT_EXPIRES` bound). #3622 (separate result backend) remains available as defense in depth.
- `OPENSEARCH_REINDEX_BATCH_SIZE` is env-configurable; 100 subtasks/batch ≈ 10k documents in flight per batch at the default chunk sizes.
- The `mocked_celery` fixture now also patches `celery.chord` since the reindex canvas uses chords directly instead of `chain(group, ...)`.


feoh:
@copilot Please fix pre-commit errors.

feoh:
@rhysyngsun points out that this is too risky to be desirable before the weekend. I'll leave it for now.

PR #3646, opened 2026-07-19
Skip embedding tasks and Qdrant writes for unchanged content files

mbertrand:

Closes mitodl/hq#12454 (companion to mitodl/hq#12453 mechanics and mitodl/hq#12452 checksum).

When a course is re-ingested, skip re-embedding content files that are already up to date in Qdrant.

Every embedded file stores its content checksum in its Qdrant payload. Before doing any expensive work, `embed_run_content_files` now fetches those stored checksums in a few batched requests and compares them (plus metadata fields like title, description, and summary) against the database:

- **File missing from Qdrant, or checksum differs** → re-embed it.
- **Checksum matches but metadata changed** (edited title, new AI summary) → just rewrite its Qdrant payload, no re-embedding.
- **Everything matches** → skip the file. A fully unchanged course dispatches no embedding tasks — only the cheap pre-pass runs.

Because the check reads Qdrant rather than the database, it recovers from failures automatically: if an embedding task dies partway (deploy, autoscaling), the unfinished files still look stale on the next ingest and get retried, while completed files are skipped. Nothing is silently lost.

This replaces the earlier version of this branch, which detected changes by comparing database state before/after the load. As review pointed out, that couldn't recover from a failed embed: the database already said "done," so the file would never be retried.

Also in this PR:

- The per-file Qdrant lookups inside the embedding pipeline are batched (one request per batch instead of ~two per file). This covers backfills and other callers too.
- The `content_files_loaded` hook gains `removed_unpublished`, so the remove-unpublished-files task only runs when a file was actually unpublished.
- `embed_run_content_files` only embeds published files.

**Numbers** (1,770-file course, unchanged re-ingest): 31s of serialization and Qdrant writes on `main` → 0.09s and zero tasks on this branch. With 10% of files changed: 18 tasks instead of 177. A brand-new course does the same embedding work as `main` with fewer Qdrant calls. The branch never does more work than `main` in any scenario.

**Known limits:**

- Staleness is detected from each file's first chunk in Qdrant, so a file whose embedding died mid-file (first chunk written, rest lost) looks complete — same as `main`.
- Run-level payload fields (`run_title`, `course_number`, `offered_by`) refresh when a file changes or via the embedding backfill — roughly the same freshness as `main`, which only refreshed them when the course archive changed.
- The first ingest after deploy may rewrite payloads for older points missing newer fields (e.g. `summary`) — payload writes only, no re-embedding.
- Fields compared by the pre-pass must be exact serializer pass-throughs of `ContentFile` columns; a test fails if someone adds a transformed field (which would re-flag every file on every load).

### How can this be tested?

Use run `course-v1:MITxT+15.356.1x+3T2025` (300 content files, each with content and a checksum). If it's not in your local DB yet, ingest it first:

```bash
docker compose run --rm web python manage.py backpopulate_mitxonline_data
docker compose run --rm web python manage.py backpopulate_mitxonline_files --resource-ids <id of the course-v1:MITxT+15.356.1x course>
```

Queue the task for the run's pk and watch the pre-pass ratio in the celery logs:

```python
from vector_search.tasks import embed_run_content_files
embed_run_content_files.delay(<RUN_PK>)
```

```bash
docker compose logs celery --since 5m | grep 'files need embedding'
```

Expected sequence (verified on a 34-file run; the pre-pass on the 300-file run above logs `210 of 300` against a partially stale local Qdrant, matching step 1):

1. **First call** (run not yet fully in Qdrant): `N of N files need embedding` — stale/missing files embed. If the run was already fully embedded, you'll see `0 of N` immediately.
2. **Call again with no changes**: `0 of N` — no embedding tasks dispatched, nothing written.
3. **Metadata-only change** — retitle a file, re-queue:
   ```python
   from vector_search.tasks import embed_run_content_files
   ContentFile.objects.filter(run_id=<RUN_PK>, published=True).update(title='retitled')
   embed_run_content_files.delay(<RUN_PK>)
   ```
   Logs `N of N` (all titles changed); the files take the payload-only path — their Qdrant payload shows the new title, with no re-embedding. Re-queueing again logs `0 of N`.
4. **Content change** — edit a file's `content` and call `.save()` (which recomputes the checksum), re-queue: logs `1 of N` and re-embeds just that file.

Testing via a real course re-ingest works too, but note the embed task is only queued from the ETL hook when `QDRANT_ENABLE_INDEXING_PLUGIN_HOOKS=True` is set — with it off (the local default), a re-ingest logs nothing.

### DIFF
--- a/learning_resources/etl/loaders.py
+++ b/learning_resources/etl/loaders.py
@@ -24,7 +24,6 @@ from learning_resources.constants import (
     RunStatus,
 )
 from learning_resources.etl.constants import (
-    CONTENT_FILE_PAYLOAD_METADATA_FIELDS,
     CONTENT_TAG_CATEGORIES,
     READABLE_ID_FIELD,
     ContentTagCategory,
@@ -1029,17 +1028,6 @@ def load_content_files(
 
     """
     if course_run.learning_resource.resource_type == LearningResourceType.course.name:
-        # Snapshot existing files so we can embed only what actually changed this
-        # load, rather than re-embedding the whole run every time.
-        prior_files = {
-            key: (checksum, published, tuple(metadata))
-            for key, checksum, published, *metadata in ContentFile.objects.filter(
-                run=course_run
-            ).values_list(
-                "key", "checksum", "published", *CONTENT_FILE_PAYLOAD_METADATA_FIELDS
-            )
-        }
-
         content_files_ids = []
         content_tags = []
         for content_file in content_files_data:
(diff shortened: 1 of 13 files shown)


learning_resources/etl/loaders.py
```diff
@@ -1026,6 +1027,17 @@ def load_content_files(
 
     """
     if course_run.learning_resource.resource_type == LearningResourceType.course.name:
+        # Snapshot existing files so we can embed only what actually changed this
+        # load, rather than re-embedding the whole run every time.
+        prior_files = {
```
abeglova: This checks against the database for changes and not qdrant. The database etl job does not wait for the qdrant tasks to finish. The embedding is supposed to have 3 retries but if it does not go through for whatever reason (like celery becoming overwhelmed) qdrant will never be updated

abeglova: I think vector_search/utils.py#should_generate_content_embeddings already checks the checksome and does not do the expensive part of the embedding if there is no change

abeglova (changes requested):

mbertrand:
@abeglova ready for another look

learning_resources/etl/loaders.py
```diff
@@ -1050,7 +1050,7 @@ def load_content_files(
             .values_list("direct_learning_resource_id", flat=True)
             .distinct()
         )
-        stale_published_files.update(published=False)
+        removed_unpublished = stale_published_files.update(published=False) > 0
```
abeglova: So i think this means that if a file is unpublished but the qdrant unpublish action fails the orphaned content file in qdrant will never be removed. I think you can just leave the loader file unchanged? I don't think the delete action is costly if the point does not exist

abeglova (changes requested):

abeglova (approved):
This looks good but the pr title is no longer what the pr does - please change it

PR #3655, opened 2026-07-21
show all facets from url

abeglova:
closes https://github.com/mitodl/hq/issues/12383

Any facet that is in the url should be shown on the sidebar

Screenshot 2026-07-21 at 3 38 18 PM

### How can this be tested?
Run `docker-compose run web ./manage.py
backpopulate_ocw_data --course-name 14-12-economic-applications-of-game-theory-fall-2025`

Got to http://open.odl.local:8062/search?level=undergraduate
you should see the level facet in the sidebar after "department"

click "clear all"

You should no longer see the level facet

### DIFF
--- a/frontends/main/src/app-pages/SearchPage/SearchPage.test.tsx
+++ b/frontends/main/src/app-pages/SearchPage/SearchPage.test.tsx
@@ -12,7 +12,7 @@ import type {
   LearningResourcesSearchResponse,
   PaginatedLearningResourceOfferorDetailList,
 } from "api"
-import { ResourceTypeEnum } from "api"
+import { ResourceTypeEnum, LearningResourceRunLevelInnerCodeEnum } from "api"
 import invariant from "tiny-invariant"
 import { Permission } from "api/hooks/user"
 import {
@@ -160,7 +160,6 @@ describe("SearchPage", () => {
         "department",
         "free",
         "offered_by",
-        "professional",
         "resource_category",
         "resource_type",
         "resource_type_group",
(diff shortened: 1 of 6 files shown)


frontends/main/src/app-pages/SearchPage/searchRequests.ts
```diff
@@ -8,6 +11,46 @@ export const facetNames = [
   "topic",
   "offered_by",
   "free",
-  "professional",
```
abeglova, line 11: professional is removed because we don't actually need facets counts. The buttons don't have counts

frontends/main/src/app/(site)/search/page.tsx
```diff
@@ -31,7 +31,7 @@ const Page: React.FC<PageProps<"/search">> = async ({ searchParams }) => {
     // @ts-expect-error -- this will error until mitodl/mit-learn-api-axios is updated
     requestParams: validateRequestParams(search),
     constantSearchParams: {},
-    facetNames,
+    facetNames: defaultFacetNames,
```
mbertrand: The server prefetch here still uses `defaultFacetNames`, but the client now adds any extra facets from the URL to `aggregations`. So for exactly the URLs this feature targets (e.g. `/search?level=undergraduate`), the server prefetches without `level` in `aggregations` while the client queries with it — the React Query keys don't match, the dehydrated data is never used, and the client refetches on mount.

Repro (with the PR's test course ingested): a working prefetch produces **zero** client requests — load `http://open.odl.local:8062/search?topic=Economics` and the Network tab shows no `learning_resources_search` request at all (the browser query client's 30-min `staleTime` exists precisely so hydrated queries aren't refetched). Load `http://open.odl.local:8062/search?level=undergraduate` and the client makes one `learning_resources_search` request with `aggregations=level` included, because the dehydrated key (visible in view-source: `"level":["undergraduate"]` as a filter, but the default `aggregations` list without `level`) no longer matches. The server's own prefetch request only appears in server logs, not devtools.

Suggest computing the extras here too, e.g.:

```ts
const urlParams = new URLSearchParams(
  Object.entries(search).flatMap(([key, value]) =>
    Array.isArray(value) ? value.map((v) => [key, v]) : [[key, String(value)]],
  ),
)

const params = getSearchParams({
  // ...
  facetNames: [
    ...(defaultFacetNames ?? []),
    ...(getExtraFacetNames(urlParams) ?? []),
  ],
  page: Number(search.page ?? 1),
})
```

frontends/main/src/app-pages/SearchPage/SearchPage.test.tsx
```diff
@@ -172,6 +171,39 @@ describe("SearchPage", () => {
...
+    // ...and it renders as a facet after "Department" and before Admin Options.
+    const facetsContainer = screen.getByTestId("facets-container")
+    await within(facetsContainer).findByText("Level")
+    const facetTitles = within(facetsContainer)
+      .getAllByText(/Department|Level/)
+      .map((el) => el.textContent)
+    expect(facetTitles.indexOf("Level")).toBeGreaterThan(
+      facetTitles.indexOf("Department"),
+    )
+  })
```
mbertrand, line 205: This covers showing/aggregating the extra facet, but the PR's other behavior — "Clear all" removes the facet from the URL and the sidebar — isn't tested anywhere (the existing "Toggling facets" test only clears `topic`, a default facet). Since that's the path that depends on the new `searchParamFacets` wiring, it's the most likely to regress. This addition passes locally:

```tsx
const { location } = renderWithProviders(<SearchPage />, {
  url: "?level=graduate",
})
// ... existing assertions ...
const clearAll = await screen.findByRole("button", { name: /clear all/i })
await user.click(clearAll)
expect(location.current.search).toBe("")
await waitFor(() => {
  expect(within(facetsContainer).queryByText("Level")).toBeNull()
})
```

frontends/main/src/app-pages/SearchPage/SearchPage.tsx
```diff
@@ -217,7 +247,7 @@ const SearchPage: React.FC = () => {
           requestParams={params}
           setPage={setPage}
           facetManifest={facetManifest as FacetManifest}
-          facetNames={facetNames}
+          facetNames={augmentedFacetNames}
```
mbertrand, line 250: Passing the augmented list into the hybrid path surfaces extra facets in vector query search with misleading counts.

Repro (with the PR's test course ingested **and embedded** — `python manage.py generate_embeddings --resource-ids <course id> --skip-contentfiles`): `http://open.odl.local:8062/search?vector_search=true&q=game+theory&level=undergraduate&topic=Economics`

The course renders as the result — it passed the `level=undergraduate` filter server-side, since `level` isn't in `VECTOR_CLIENT_FILTER_FACETS` so `toUnfacetedVectorSearchParams` keeps it in the request. The client-computed counts are correct for the default facets (Topic "Economics 1", Free "1"), but the Level facet shows "Undergraduate 0" — `getResourceFacetValues` in `HybridSearchDisplay.tsx` has no cases for `level`/`platform`/`course_feature`, so they always count to zero even when every displayed result matches.

Admin-only experimental path, so probably fine to defer — but worth a follow-up to add those cases to `getResourceFacetValues`.

mbertrand (changes requested):
Looks good overall, noticed a few minor things

mbertrand (approved):
👍

PR #3662, opened 2026-07-22
Only re-embed a run's content files when it is actually republished

### LINKED ISSUE #3661: Unpublished course runs are re-embedded on every ETL sync
### Expected Behavior

A run's content files should only be re-embedded when the run is actually republished (goes from unpublished to published) or its content changes.

### Current Behavior

When an ETL sync loads a run that is stored as unpublished but still has content files, the loader treats it as "republished" and queues a full re-embedding of all its content files (`embed_run_content_files` with `overwrite=True`).

The check only looks at the run's *previous* published state, not its new one:

https://github.com/mitodl/mit-learn/blob/main/learning_resources/etl/loaders.py#L398-L402

So a run that stays unpublished trips this branch on every sync. Since some ETL pipelines run several times a day (e.g. mitxonline every 6 hours), the same unchanged content files are re-embedded again and again, wasting significant worker time and embedding API calls.

### Steps to Reproduce

1. Have a published resource with an unpublished run that has content files (common for retained sources, whose files stay in Qdrant).
2. Run the ETL sync for that source.
3. `embed_run_content_files` is queued for the run even though nothing changed.
4. Every subsequent sync does it again.

### Possible Solution

Only fire the republish actions when the run is now published:

```python
elif previously_published is False and learning_resource_run.published:
```

mbertrand:
Closes https://github.com/mitodl/mit-learn/issues/3661

Fixes a bug in `load_run` where the "run was republished" branch checked only the run's *previous* published state, not its new one. A run that stays unpublished (but has retained content files) tripped the branch on every ETL sync, queueing `embed_run_content_files` with `overwrite=True` and fully re-embedding unchanged content files several times a day.

The branch now also requires the run to currently be published, so the re-index/re-embed actions only fire on a real unpublished → published transition.

### How can this be tested?
1. Set `QDRANT_ENABLE_INDEXING_PLUGIN_HOOKS=True` in `env/backend.local.env` and start the stack (`docker compose up`).
2. In a second terminal, watch the celery logs for the embed task:
   `docker compose logs -f celery | grep embed_run_content_files`
3. Open a Django shell (`docker compose run --rm web python manage.py shell`) and set up an unpublished run that has content files on a published course:
   ```python
   from learning_resources.etl.loaders import load_run

   resource = LearningResource.objects.filter(
       published=True, etl_source="mitxonline",
       runs__published=False, runs__content_files__isnull=False,
   ).first()
   run = resource.runs.filter(published=False, content_files__isnull=False).first()
   # (if none exist, pick any run with content files and set run.published=False; run.save())
   ```
4. Simulate an ETL sync of the still-unpublished run:
   ```python
   load_run(resource, {"run_id": run.run_id, "published": False})
   ```
   **On `main`:** an `embed_run_content_files` task appears in the celery logs even though nothing changed. **On this branch:** no task is queued.
5. Now simulate the run actually being republished:
   ```python
   load_run(resource, {"run_id": run.run_id, "published": True})
   ```
   `embed_run_content_files` is queued once — same behavior as before the fix.

Automated coverage: `test_load_run_content_files_loaded_actions_only_on_republish` in `learning_resources/etl/loaders_test.py` covers both cases.

### DIFF
--- a/learning_resources/etl/loaders.py
+++ b/learning_resources/etl/loaders.py
@@ -395,10 +395,12 @@ def load_run(
                     )
 
                 transaction.on_commit(enqueue_content_tasks)
-            elif previously_published is False:
+            elif previously_published is False and learning_resource_run.published:
                 # Run was republished. Its content files are still present and
                 # published (retained sources keep them in Qdrant), just absent
                 # from OpenSearch. Re-index them without a full re-ingest.
+                # Runs that remain unpublished must not trip this on every
+                # sync — it re-embeds their files with overwrite=True.
                 content_files_loaded_actions(learning_resource_run)
     return learning_resource_run
 
--- a/learning_resources/etl/loaders_test.py
+++ b/learning_resources/etl/loaders_test.py
@@ -1346,18 +1346,9 @@ def test_load_run_content_files_loaded_actions_only_on_republish(mocker, new_pub
         published=True,
     )
     run = LearningResourceRunFactory.create(learning_resource=course, published=False)
-    ContentFileFactory.create_batch(2, run=run)
+    ContentFileFactory.create(run=run)
 
-    run_props = model_to_dict(run)
-    run_props["published"] = new_published
-    run_props["prices"] = []
-    run_props["instructors"] = []
-    run_props["image"] = None
-    del run_props["id"]
-    del run_props["learning_resource"]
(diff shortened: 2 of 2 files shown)


abeglova (approved):
LGTM

PR #3668, opened 2026-07-23
Add department 21T (Theater Arts)

zawan-ila:
Part of https://github.com/mitodl/hq/issues/5616

Introduces department **21T (Theater Arts)** in mit-learn:

- Adds `21T` → `Theater Arts` to the `DEPARTMENTS` constant (`learning_resources/constants.py`).
- Adds a `departments` fixture entry linked to the **School of Humanities, Arts, and Social Sciences** (`school_id` 3).
- Adds a data migration `0025_add_department_21T` (modeled on `0014_add_department_SP`) that creates the department and its channel.
- Regenerates the OpenAPI specs (`openapi/specs/v0.yaml`, `v1.yaml`) and the generated TypeScript client to include `21T` in the department enum.

### How can this be tested?
This and https://github.com/mitodl/course-search-utils/pull/212 should be tested together.

- Run docker compose up
- Verify that the department shows up in the django admin.
- Temporarily, point mitodl/course-search-utils in package.json to https://github.com/mitodl/course-search-utils/pull/212 i.e make the following change in the package.json file
  `"@mitodl/course-search-utils": "https://github.com/mitodl/course-search-utils#f75060db9010d3ab108bfb9968382bdca3bef514"`
- Now, ingest an ocw course with the 21T department. This is a bit awkward since we do not have any 21T courses in ocw. For testing, you need to run the minio container from https://github.com/mitodl/ocw-studio locally `docker compose up s3`. You may find https://github.com/mitodl/ocw-studio#local-s3-emulation-with-minio helpful for setting up minio. Now, visit the minio interface at localhost:9001. In minio, in some bucket, under some course path, `courses/14-01sc-principles-of-microeconomics-fall-2011/data.json`, add a data.json file that has the 21T department, and an appropriate course number `21T.789` (Just modify, for example, [this](https://ocw.mit.edu/courses/14-01sc-principles-of-microeconomics-fall-2011/data.json) data.json file to achieve the desired effect). Now, set up your learn instance to point to that minio instance. Set in your backend.local.env
```
OCW_LIVE_BUCKET=<name of minio bucket>
AWS_SECRET_ACCESS_KEY=minio_password
AWS_ACCESS_KEY_ID=minio_user
```
Also set `AWS_ENDPOINT_URL_S3="http://host.docker.internal:9000"` to allow learn to communicate with the minio container.

Now, run a backpopulate like `./manage.py backpopulate_ocw_data --course-name <s3 path segment identifying the course>`. Then, visit the search page and verify that the Theater Arts facet shows up under departments and clicking on it works as expected.

### Additional Context
- This branch and the companion "rename 21M → Music" branch each add a migration numbered `0025_*`. Whichever merges second will need renumbering (e.g. `0026_*`).


data_fixtures/migrations/0025_add_department_21T.py
```diff
@@ -0,0 +1,68 @@
...
+    Channel = apps.get_model("channels", "Channel")
+    ChannelDepartmentDetail = apps.get_model("channels", "ChannelDepartmentDetail")
+
+    school = LearningResourceSchool.objects.filter(name=SCHOOL_NAME).first()
+
+    department = LearningResourceDepartment.objects.create(
+        department_id="21T",
+        name="Theater Arts",
+        school=school,
+    )
```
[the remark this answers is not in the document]
zawan-ila: Gonna leave this to the reviewer. I see that this makes sense but I don't think we need this level of safety. The migration is pretty simple and extremely unlikely to fail.

data_fixtures/migrations/0025_add_department_21T.py
```diff
@@ -0,0 +1,68 @@
...
+    LearningResourceSchool = apps.get_model(
+        "learning_resources", "LearningResourceSchool"
+    )
+    Channel = apps.get_model("channels", "Channel")
+    ChannelDepartmentDetail = apps.get_model("channels", "ChannelDepartmentDetail")
+
+    school = LearningResourceSchool.objects.filter(name=SCHOOL_NAME).first()
+
+    department = LearningResourceDepartment.objects.create(
```
mbertrand: Change to `get_or_create` here.  On a fresh database, this will cause an error because the 21T department already gets created in the initial migration which loads data from the `departments.json` file, where 21T was also added in this PR.

```python
department = LearningResourceDepartment.objects.get_or_create(
    department_id="21T",
    defaults={
        "name": "Theater Arts",
        "school": school,
    }
)
```

data_fixtures/migrations/0025_add_department_21T.py
```diff
@@ -0,0 +1,68 @@
...
+    LearningResourceDepartment = apps.get_model(
+        "learning_resources", "LearningResourceDepartment"
+    )
+    LearningResourceSchool = apps.get_model(
+        "learning_resources", "LearningResourceSchool"
+    )
+    Channel = apps.get_model("channels", "Channel")
+    ChannelDepartmentDetail = apps.get_model("channels", "ChannelDepartmentDetail")
+
+    school = LearningResourceSchool.objects.filter(name=SCHOOL_NAME).first()
```
mbertrand: Maybe change to ` school = LearningResourceSchool.objects.get(name=SCHOOL_NAME)` - if this school does not exist, the department is created school-less, channels/plugins.py:99-102 deletes department channels with no school — you'd get a department with no channel and no error.  So best to error out here in that case. Though this shouldn't happen in practice because the school exists (created in initial migration from fixtures).

data_fixtures/migrations/0025_add_department_21T.py
```diff
@@ -0,0 +1,68 @@
...
+    channel = Channel.objects.create(
+        search_filter=urlencode({"department": department.department_id}),
+        channel_type=ChannelType.department.name,
+        name=slugify(department.name),
+        title=department.name,
+    )
+    ChannelDepartmentDetail.objects.create(
+        channel=channel,
+        department=department,
+    )
```
mbertrand: Same thing here - `get_or_create` for both:

```python
channel, _ = Channel.objects.get_or_create(
    search_filter=urlencode({"department": department.department_id}),
    channel_type=ChannelType.department.name,
    defaults={
        "name": slugify(department.name),
        "title": department.name,
    },
)
ChannelDepartmentDetail.objects.get_or_create(
    channel=channel,
    department=department,
)
```

learning_resources/constants.py
```diff
@@ -261,6 +261,7 @@ class LearningResourceRelationTypes(TextChoices):
     "21H": "History",
     "21L": "Literature",
     "21M": "Music and Theater Arts",
+    "21T": "Theater Arts",
```
mbertrand, line 268: These two should go out together with #3669 - until the 21M rename lands, we'd have both "Music and Theater Arts" and "Theater Arts" showing up in the department facet, each with its own channel. Not a problem with this PR itself, just worth merging them back to back (and #3669 will need the migration renumbered to 0026, as you called out in the description).

mbertrand, line 268: One deploy-time thing to be aware of: `extract_valid_department_from_id` only returns a department if it's a key in this dict, so any edX/MITx Online course already ingested with a `21T` prefix currently has no department attached. Those will pick it up on the next ETL run since `load_departments` re-`set()`s on every load, but the channel will look empty until then. If we want it immediate, `./manage.py update_course_number_departments` followed by a reindex does it - same approach as `learning_resources/migrations/0025_update_course_number_depts.py`.

OCW is the opposite case and this PR is a straight fix there: OCW departments come from `department_numbers` in the metadata without going through `DEPARTMENTS`, and `load_departments` does an unguarded `.get(department_id=...)`, so 21T courses were raising `DoesNotExist` before this.

mbertrand (changes requested):

data_fixtures/migrations/0025_add_department_21T.py
zawan-ila: Updated to incorporate suggestion

data_fixtures/migrations/0025_add_department_21T.py
zawan-ila: :thumbsup:

learning_resources/constants.py
zawan-ila: > These two should go out together

I [decided to send](https://github.com/mitodl/hq/issues/5616#issuecomment-5091729112) these out separately, though still as close to each other as possible.

The problem with sending them out together is that, for a while, until some of the existing 21M courses are moved to 21T (which is the plan) and the indexes updated, the "to be" Theater Arts courses would show up under the Music facet. The problem with sending them out separately, as you note, is that we have both the "Theater Arts" and "Music and Theater Arts" facets. To me, that is "more acceptable" than combining the changes in a single deployment (i.e "to be Theater Arts" courses showing up under "Music"). I also let Peter know about this (see item 1 below the divider [here](https://github.com/mitodl/hq/issues/5616#issuecomment-5091729112)) and he seemed fine with it.

I do think that this is not ideal, and would love any suggestions on approaches that eliminate the negative aspects of both combined and separate deploys noted above.

zawan-ila: I don't fully understand this, but I do plan on running `update_course_number_departments` and `update_index` post migration, so I think we should be good here.

mbertrand: Sounds good, maybe `./manage.py backpopulate_ocw_data` too

mbertrand (approved):
👍

PR #3669, opened 2026-07-23
Rename department 21M from Music and Theater Arts to Music

zawan-ila:
Part of https://github.com/mitodl/hq/issues/5616

Renames department **21M** from "Music and Theater Arts" to **Music** in mit-learn:

- Updates `21M` → `Music` in the `DEPARTMENTS` constant and the `departments` fixture.
- Adds a data migration `0025_rename_department_21M_to_music` that renames the existing department row and updates its channel title/slug.
- Regenerates the OpenAPI specs (`v0.yaml`, `v1.yaml`) and the generated TypeScript client for the renamed department display name.

### How can this be tested?
This needs to be tested together with https://github.com/mitodl/course-search-utils/pull/213. So, in your package.json file, make this change:-
`"@mitodl/course-search-utils": "https://github.com/mitodl/course-search-utils#95ff0bf8edf6ff8309984e0f5b81c1d6c2a57a70"`

- Ensure that you have existing 21M courses in your setup `./manage.py backpopulate_ocw_data --course-name 21m-051-fundamentals-of-music-spring-2007`
- `docker compose run --rm web python manage.py migrate` — the 21M department name is **Music** and its channel title/slug are updated.
-  Now, go to search. Verify that the facet "Music" shows up under departments, and clicking on it has the desired effect.

### Additional Context
- The migration also updates the department **channel** slug (`music-and-theater-arts` → `music`), which changes the department's channel URL.
- This branch and the companion "add 21T" branch each add a migration numbered `0025_*`; whichever merges second will need renumbering.


data_fixtures/migrations/0025_rename_department_21M_to_music.py
```diff
@@ -0,0 +1,47 @@
...
+    if not department:
+        return
+    department.name = name
+    department.save()
+
+    channel = Channel.objects.filter(department_detail__department=department).first()
+    if channel:
+        channel.title = name
+        channel.name = slugify(name)
+        channel.save()
```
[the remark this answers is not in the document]
zawan-ila: I don't think we associated multiple channels with the same Department so this is a non-issue.

data_fixtures/migrations/0025_rename_department_21M_to_music.py
mbertrand, line 1: Looks great overall!  Noticed a few things:

1. **Search facet still says "Music and Theater Arts."** The department filter label comes from `@mitodl/course-search-utils`, which has its own hardcoded department map. mitodl/course-search-utils#213 fixes it — this PR needs to pin `@mitodl/course-search-utils` to whatever version that gets released in.

2. **Old department name is cached in `Course.course_numbers`.** `etl/utils.py:1023` writes the department name into that JSON field, and it's served on the v1 API — so 21M courses will still report "Music and Theater Arts" there. The next ETL run overwrites it for courses still in the upstream feed, but archived ones keep the old name forever. Either update that JSON in the migration too, or say in the PR description that we're accepting the stale value on archived courses.

3. **`/c/department/music-and-theater-arts` will 404.** `next.config.js` already has legacy redirects; this is a few lines in the same array.

Also, I think the `update_index` command will need to be run after migrations to get the new label in opensearch.  Also maybe `generate_embeddings --courses`?

zawan-ila: 1. I don't really see a clear way to do this, and hence would appreciate some guidance. This PR needs to bump course-search-utils with the changes from https://github.com/mitodl/course-search-utils/pull/213 but https://github.com/mitodl/course-search-utils/pull/213 needs to bump its version of mit-learn-api-axios too, which won't be available until after this PR is on the release branch? So there is kind of a circular dependency here.
2. I am thinking it is easiest to just trigger a run of [update_course_number_departments](https://github.com/mitodl/mit-learn/blob/main/learning_resources/management/commands/update_course_number_departments.py) post migration? That should update the course number json field as appropriate.
3. Thanks for this. I have added the redirect.
4. I intend on running `recreate_index --all` post migration. Please let me know if you think that may be a bit too much for us.
5. I am not really familiar with generate_embeddings at all, but I'll make sure we run that post migration too. `./manage.py generate_embeddings --courses` Doesn't look like it can cause any harm. On this note, it appears that vector search is gated behind a flag and not currently active? Are there any other uses of embeddings/qdrant stuff?

data_fixtures/migrations/0025_rename_department_21M_to_music.py
```diff
@@ -0,0 +1,47 @@
+# Data migration: rename department 21M from "Music and Theater Arts" to "Music"
```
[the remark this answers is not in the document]
zawan-ila: Yup. This is a concern. I'll update the numbers before merge.

data_fixtures/migrations/0025_rename_department_21M_to_music.py
mbertrand: 1 - @ChristopherChudzicki might have some ideas on the best way to handle this.  Maybe it's okay (and even unavoidable) to have a slight mismatch temporarily until everything is merged & released across the repos.

2 - Sounds good, I forgot that command exists

4 - That might be overkill, `update_index` is generally much faster.  Probably worth trying it locally to verify if `update_index --courses` makes all necessary changes without the need for `recreate_index`

5 - I think vector search is used by default on the topics pages on production.  For example https://learn.mit.edu/c/topic/ai makes a call to https://api.learn.mit.edu/api/v0/vector_learning_resources_search/

mbertrand (approved):
Did Chris have any suggestions on the best way to handle the course-search-utils update?  Assuming a temporary mismatch in department names is okay, 👍 

The migration may still have to be renumbered if it's merged after the other PR with the same migration number.

zawan-ila:
> Did Chris have any suggestions on the best way to handle the course-search-utils update

No. I asked Anastasia and she pointed me to a [PR of Chris's](https://github.com/mitodl/course-search-utils/pull/193/changes) and I adopted the same approach :-)

PR #3663, opened 2026-07-22
Remove micromasters ETL

mbertrand:
### What are the relevant tickets?
N/A - all micromasters programs are now ingested via the mitxonline ETL

### Description (What does it do?)
Removes all code related to the MicroMasters ETL:

- The `micromasters` ETL module and its tests
- The `micromasters_etl` pipeline
- The `get_micromasters_data` celery task and its celerybeat schedule entry
- The `backpopulate_micromasters_data` management command
- The `MICROMASTERS_CATALOG_API_URL` setting and related env/devcontainer entries

Adds a data migration that unpublishes any remaining learning resources with `etl_source=micromasters` and deindexes them from OpenSearch/Qdrant via the standard `resource_unpublished_actions` plugin hooks.

Kept (not part of the ETL): `ETLSource.micromasters` enum (referenced by migration 0052), `CertificationType.micromasters` (used by the mitxonline ETL), and `MICROMASTERS_CMS_API_URL` (used for program letters).

### How can this be tested?
- Run the migration: `docker compose run --rm web python manage.py migrate learning_resources` — any published micromasters resources should be unpublished and deindex tasks dispatched.
- `docker compose run --rm web uv run pytest learning_resources/etl/pipelines_test.py learning_resources/tasks_test.py -n logical`

PR #3670, opened 2026-07-23
Delete micromasters resources and remove ETLSource constant

mbertrand:
N/A — follow-up to #3663

Follow-up cleanup to #3663, which removed the micromasters ETL pipeline and unpublished/deindexed all remaining `etl_source=micromasters` resources from OpenSearch and Qdrant.

- Deletes all `LearningResource` rows with `etl_source="micromasters"` via a new data migration (`0118`). No deindexing is needed — the resources were already removed from OpenSearch/Qdrant by the `0117` migration in #3663; related runs, content files, and relationship rows are removed by FK cascades.
- Removes the now-unused `ETLSource.micromasters` constant.
- Updates migration `0052` to use the `"micromasters"` string literal instead of `ETLSource.micromasters.name`, so historical migrations no longer depend on the removed enum member.

### How can this be tested?
- `docker compose run --rm web python manage.py migrate learning_resources` — `0118` should apply cleanly and delete any remaining micromasters resources.
- `docker compose run --rm web python manage.py makemigrations --check --dry-run` — no model changes detected (`etl_source` is a plain CharField, so removing the enum member causes no schema drift).
- `LearningResource.objects.filter(etl_source="micromasters").count()` should be 0 after migrating.

### DIFF
--- a/frontends/api/src/generated/v1/api.ts
+++ b/frontends/api/src/generated/v1/api.ts
@@ -8740,13 +8740,12 @@ export const SortbyEnum = {
 export type SortbyEnum = (typeof SortbyEnum)[keyof typeof SortbyEnum]
 
 /**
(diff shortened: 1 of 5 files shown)


shanbady (changes requested):
change looks good just needs an update to the openapi spec

mbertrand:
openapi spec is up to date, errors are due to this being considered a breaking change ( "micromasters" is no longer a valid value for the `source` parameter of the content_files webhook endpoint).

shanbady (approved):
LGTM

PR #3672, opened 2026-07-24
fix: Flaky JS test: ProgramAsCoursePage 'Enroll CTA posts program enrollment' times out on Loading button

ahtesham-quraish:
https://github.com/mitodl/hq/issues/12456

### Problem
`Enroll CTA posts program enrollment` intermittently failed in CI with
`Unable to find role="button" and name "Start Learning"` — the button was still
in its loading state (`aria-label="Loading"`). Re-running passed: a timing flake.

### Root cause
The "Start Learning" button is gated by `isStatusLoading`, which stays `true`
until the program-enrollment status query (`GET /api/v3/program_enrollments/`)
resolves — until then the button has no "Start Learning" name. `findAllByRole`
only retries for Testing Library's default ~1s, so under CI load the query
sometimes hadn't resolved in time.

### Fix
Added a shared `ENROLL_STATUS_TIMEOUT` constant and passed it as the `waitFor`
timeout on the three `findAllByRole("Start Learning")` call sites, so they retry
until the button leaves its loading state. No overhead on the fast path.

Also pinned `show_stay_updated: true` in the "Shows Stay Updated button" test —
the factory now emits a random value, making that assertion ~50% flaky (matches
the `CoursePage`/`ProgramPage` updates).

### Testing
Reproduced the flake locally (delayed status query) and confirmed the fix;
all `ProgramAsCoursePage.test.tsx` tests pass.

### DIFF
--- a/frontends/main/src/app-pages/ProductPages/ProgramAsCoursePage.test.tsx
+++ b/frontends/main/src/app-pages/ProductPages/ProgramAsCoursePage.test.tsx
@@ -44,6 +44,15 @@ const makeProgramAsCourse: typeof factories.programs.program = (
   })
 const makePage = factories.pages.programPageItem
 
+// The enroll button renders with aria-label="Loading" (and no "Start Learning"
+// accessible name) until the enrollment-status query resolves. Under CI load
+// that query can take longer than Testing Library's default 1s waitFor, leaving
+// the button in its loading state when findAllByRole gives up — a timing flake,
+// not a regression. Give these queries extra headroom so they retry until the
+// button leaves its loading state. See ProgramHeaderEnrollButton /
+// useProgramEnrollment.
+const ENROLL_STATUS_TIMEOUT = 5000
+
 const setupApis = ({
   program,
   page,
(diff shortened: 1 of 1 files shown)


frontends/main/src/app-pages/ProductPages/ProgramAsCoursePage.test.tsx
```diff
@@ -44,6 +44,15 @@ const makeProgramAsCourse: typeof factories.programs.program = (
   })
 const makePage = factories.pages.programPageItem
 
+// The enroll button renders with aria-label="Loading" (and no "Start Learning"
```
gumaerc: This comment could be made more concise, all that really needs to be said is that enrollment tests take more time and we need to wait for them.

gumaerc (approved):
👍 This looks good and solves the problem by adding extra timeout, which is fine. You may want to update the PR description before merging, it looks like there's some stale comments in there?

> Also pinned show_stay_updated: true in the "Shows Stay Updated button" test —
> the factory now emits a random value, making that assertion ~50% flaky (matches
> the CoursePage/ProgramPage updates).

PR #3626, opened 2026-07-16
dashboard card mobile design improvements

gumaerc:
### What are the relevant tickets?                                                                                                                                                                                                                   
   Closes https://github.com/mitodl/hq/issues/12305                                                                                                                                                                                                     
                                                                                                                                                                                                                                                        
   ### Description (What does it do?)                                                                                                                                                                                                                   
   Mobile-specific follow-up to the desktop dashboard card redesign, addressing design review feedback for course cards, program-as-course cards, program enrollment cards, and the B2B contract page header:                                           
                                                                                                                                                                                                                                                        
   - Fixed the date / "Certificate track" / upgrade-banner meta text so it wraps by whole segment rather than breaking mid-word or mid-number. Each segment (date, certificate track indicator, upgrade promo) stays intact; if a segment doesn't fit   
 next to the date, it wraps to its own line instead.                                                                                                                                                                                                    
   - Standardized the mobile/desktop breakpoint for dashboard cards to `sm` (600px) instead of `md` (900px), so tablet-width viewports use the desktop card layout (properly sized CTA buttons, inline metadata) instead of the stretched, stacked      
 mobile styling. True mobile layout now only applies below 600px.                                                                                                                                                                                       
   - Fixed the "Certificate track" indicator so earning a certificate replaces it in place with a "View Certificate" link, instead of showing both simultaneously.                                                                                      
   - Added the enrollment status (progress) indicator to the mobile module-row layout on program-as-course cards, matching the existing desktop behavior.                                                                                               
   - Fixed card/header content overflowing past the right edge of the viewport on mobile (a nested flexbox `min-width` issue introduced by the progress-indicator column).                                                                              
   - Adjusted spacing throughout: title-to-metadata gap, program card header/subheader gaps, module-row icon-to-content gap, and outer card border/shadow rendering on program-as-course cards.                                                         
   - Applied the same "flush to page background" mobile header treatment used for the My Learning section header to the B2B contract page header.                                                                                                       
   - Updated and added tests covering certificate/track-indicator behavior, enrollment status indicators, and the dual desktop/mobile render pattern.                                                                                                   
                                                                                                                                                                                                                                                        
   ### Screenshots (if appropriate):                                                                                                                                                                                                                    
<img width="768" height="1024" alt="image" src="https://github.com/user-attachments/assets/b7b7be00-1fba-4375-a62b-a19245a0e9be" />
[...]

PR #3674, opened 2026-07-24
fix: Remove course enrollment from cache on unenroll for instant UI update

daniellefrappier18:

 Closes https://github.com/mitodl/hq/issues/10104 

Unenrolling from a course on the dashboard was slow to update, the card stuck around until the enrollments list refetched. Program unenroll already handled this by dropping the enrollment from the cache in `onSuccess`, so I did the same thing for courses.

`useDestroyEnrollment` now filters the deleted enrollment out of the `courseRunEnrollments` cache in `onSuccess`, same as `useDestroyProgramEnrollment`. The card disappears as soon as the DELETE comes back. Left the `onSettled` invalidate in place so the list still reconciles with the server afterward.

The enroll-button spinner from the original ticket already went in with #3626, and I didn't touch the program flow.

- [ ] Desktop screenshots

BEFORE 

https://github.com/user-attachments/assets/bf117d45-a13c-4a65-835d-908385120927

AFTER

https://github.com/user-attachments/assets/9523a9a9-af85-46a4-9741-e260800cf961

### How can this be tested?

1. Log in as someone with an active course enrollment, go to My Learning.
2. On a course card, ⋯ → Unenroll → confirm.
3. Spinner runs until the DELETE finishes (same as before), then the card should drop right away instead of hanging around.
4. If the DELETE fails (throttle/offline), you should still get the error alert and the card should stay put.

### Additional Context

In the local k3d stack the course DELETE hangs ~6s on a dead `edx.odl.local` DNS lookup in mitxonline — separate fix in ol-infrastructure, nothing to do with this repo.

I added a hostAliases entry to the mitxonline-webapp pod spec in ol-infrastructure/local dev/apps/mitxonline/deployment.yaml, pinning the edX hostnames to the pod loopback: 

```
      hostAliases:
        - ip: "127.0.0.1"
          hostnames:
            - edx.odl.local
            - studio.edx.odl.local
```

### DIFF
--- a/frontends/api/src/mitxonline/hooks/enrollment/index.ts
+++ b/frontends/api/src/mitxonline/hooks/enrollment/index.ts
@@ -64,6 +64,12 @@ const useDestroyEnrollment = () => {
   return useMutation({
     mutationFn: (enrollmentId: number) =>
       courseRunEnrollmentsApi.enrollmentsDestroy({ id: enrollmentId }),
+    onSuccess: (_data, enrollmentId) => {
+      queryClient.setQueryData(
+        enrollmentQueries.courseRunEnrollmentsList().queryKey,
+        (data) => data?.filter((enrollment) => enrollment.id !== enrollmentId),
+      )
+    },
     onSettled: () => {
       queryClient.invalidateQueries({
         queryKey: enrollmentKeys.courseRunEnrollmentsList(),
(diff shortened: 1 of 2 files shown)


ahtesham-quraish (approved):
LGTM!

Minor consistency nit: this `onSuccess` references the cache key via `enrollmentQueries.courseRunEnrollmentsList().queryKey` (line 69), while the `onSettled` invalidate just below uses `enrollmentKeys.courseRunEnrollmentsList()` (line 75). Both resolve to the exact same key — `enrollmentQueries.courseRunEnrollmentsList().queryKey` is defined as `enrollmentKeys.courseRunEnrollmentsList()` — so there's no functional difference. But using two different accessors for the same list on adjacent lines is a little confusing to read. Consider using one consistently:

```ts
onSuccess: (_data, enrollmentId) => {
  queryClient.setQueryData(
    enrollmentKeys.courseRunEnrollmentsList(),
    (data) => data?.filter((enrollment) => enrollment.id !== enrollmentId),
  )
},
```

Not blocking — happy to keep the `enrollmentQueries` accessor if you prefer it for the typed queryKey.

PR #3675, opened 2026-07-24
Slim learning resource API payloads and cache rendered JSON

mbertrand:
Closes mitodl/hq#12534

1. **Stop returning content file text in API responses.** Search results, resource detail, and similar-resource responses no longer include `content`, `summary`, or `flashcards` inside nested `content_files`. The full text stays in the OpenSearch index, so searching within file content still works. Cuts the search payload roughly 20×.

2. **Cache rendered JSON instead of Python data.** The Redis view cache stored the un-rendered response data, so every cache hit still rebuilt the full JSON response. It now stores the rendered bytes and returns them directly for JSON requests. The browsable API still works (cache hits are re-rendered for it), and old cache entries are served correctly until they expire.
### How can this be tested?

**Setup** (same OCW sample data as #3018):
```
./manage.py backpopulate_ocw_data --course-name 15-s12-blockchain-and-money-fall-2018 --overwrite
./manage.py backpopulate_ocw_data --course-name 6-0001-introduction-to-computer-science-and-programming-in-python-fall-2016 --overwrite
./manage.py backpopulate_ocw_data --course-name res-6-011-the-art-of-insight-in-science-and-engineering-mastering-complexity-fall-2014 --overwrite
./manage.py recreate_index --all
```

1. **Slim search results:** open `http://localhost:8063/api/v1/learning_resources_search/?resource_type_group=learning_material&show_ocw_files=true`. You should see OCW learning materials (problem sets, exams, and "The Art of Insight in Science and Engineering" textbook). Each result's `content_files` entries should have metadata (`title`, `url`, `image_src`, etc.) but **no** `content`, `summary`, or `flashcards` keys.
2. **Searching file text still works:** search a phrase that only appears inside the textbook's text, not in any title or description: `http://localhost:8063/api/v1/learning_resources_search/?q=Merely+watching+workout+videos&show_ocw_files=true&min_score=0`. The Art of Insight textbook should come back — proving the full text is still indexed even though it's no longer in the response. (`min_score=0` is needed because the default relevance cutoff scores content-only matches below threshold on the small local dataset — pre-existing behavior, unrelated to this PR.)
3. **Detail and similar endpoints:** open `/api/v1/learning_resources/<id>/` for one of the materials from step 1 (the textbook is id 582 in my local db) — same slim `content_files` shape. If you have Qdrant embeddings locally, `/api/v1/learning_resources/<id>/vector_similar/` too.
4. **Frontend:** open `http://localhost:8062/search?resource_type_group=learning_material&show_ocw_files=true` — the Learning Materials tab should render the OCW materials normally. Then open `http://localhost:8062/search?resource_type_group=learning_material&show_ocw_files=true&resource=582` — the drawer should show the textbook with its description (which falls back to the content file's description) and a populated "Similar Learning Resources" carousel (fed by the slimmed `vector_similar` endpoint; needs local Qdrant embeddings). Video playback can't be checked with this dataset — the three OCW courses above ingest no video resources — so spot-check a video page on RC after deploy.
5. **Caching:** open a cached endpoint like `/api/v1/topics/` in a browser twice — the browsable API HTML should render both times (the second is a cache hit). Request it with `Accept: application/json` and confirm JSON comes back.
6. On RC, `curl -s '<search url>' | wc -c` — the default search page should be a few hundred KB, not several MB.

### Additional Context
After deploying (and before any rollback), run `main.utils.clear_views_cache()` in a web shell. After deploy it evicts cached responses that still contain the old multi-MB payloads; before a rollback it removes bytes-format entries the previous code can't read.

### Checklist:
- [ ] Run `clear_views_cache()` after deploy to RC/production

### DIFF
--- a/learning_resources/serializers.py
+++ b/learning_resources/serializers.py
@@ -1512,7 +1512,7 @@ class NestedContentFileSerializer(ContentFileSerializer):
         fields = [
             field
             for field in ContentFileSerializer.Meta.fields
-            if field not in ("content", "summary", "flashcards")
+            if field not in constants.CONTENT_FILE_LARGE_FIELDS
         ]
 
 
--- a/learning_resources/serializers.py
+++ b/learning_resources/serializers.py
@@ -1501,6 +1501,21 @@ class ContentFileSerializer(serializers.ModelSerializer):
         ]
 
 
+class NestedContentFileSerializer(ContentFileSerializer):
+    """
+    ContentFileSerializer without the large text fields (content, summary,
+    flashcards), for nesting inside learning resource API responses.
+    The search indexing path re-adds full content where needed.
+    """
+
+    class Meta(ContentFileSerializer.Meta):
+        fields = [
+            field
+            for field in ContentFileSerializer.Meta.fields
+            if field not in ("content", "summary", "flashcards")
+        ]
+
+
 class VideoResourceSerializer(LearningResourceBaseSerializer):
     """Serializer for video resources"""
 
@@ -1515,11 +1530,11 @@ class VideoResourceSerializer(LearningResourceBaseSerializer):
     content_files = serializers.SerializerMethodField()
     description = serializers.SerializerMethodField()
 
-    @extend_schema_field(ContentFileSerializer(many=True, allow_null=True))
+    @extend_schema_field(NestedContentFileSerializer(many=True, allow_null=True))
     def get_content_files(self, instance):
         """Serialize content files with prefetch."""
         content_files = instance.direct_content_files_for_serialization()
-        return ContentFileSerializer(
+        return NestedContentFileSerializer(
             content_files,
             many=True,
             read_only=True,
@@ -1547,11 +1562,11 @@ class DocumentResourceSerializer(LearningResourceBaseSerializer):
     content_files = serializers.SerializerMethodField()
     description = serializers.SerializerMethodField()
 
-    @extend_schema_field(ContentFileSerializer(many=True, allow_null=True))
+    @extend_schema_field(NestedContentFileSerializer(many=True, allow_null=True))
     def get_content_files(self, instance):
         """Serialize content files with prefetch."""
         content_files = instance.direct_content_files_for_serialization()
-        return ContentFileSerializer(
+        return NestedContentFileSerializer(
             content_files,
             many=True,
             read_only=True,
--- a/main/utils.py
+++ b/main/utils.py
@@ -42,9 +42,11 @@ def _sorted_query_string(query_dict):
 
 def _wants_html(request) -> bool:
     """Whether the request wants the browsable API (HTML) rather than JSON."""
-    return "text/html" in request.headers.get("Accept", "") or bool(
-        request.GET.get("format")
-    )
+    renderer = getattr(request, "accepted_renderer", None)
+    if renderer is not None:
+        # DRF negotiated the renderer in initial(), before the decorated handler
+        return renderer.format == "html"
+    return "text/html" in request.headers.get("Accept", "")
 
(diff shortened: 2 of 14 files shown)


learning_resources/serializers.py
```diff
@@ -1547,11 +1562,11 @@ class DocumentResourceSerializer(LearningResourceBaseSerializer):
     content_files = serializers.SerializerMethodField()
     description = serializers.SerializerMethodField()
 
-    @extend_schema_field(ContentFileSerializer(many=True, allow_null=True))
+    @extend_schema_field(NestedContentFileSerializer(many=True, allow_null=True))
     def get_content_files(self, instance):
         """Serialize content files with prefetch."""
         content_files = instance.direct_content_files_for_serialization()
-        return ContentFileSerializer(
+        return NestedContentFileSerializer(
```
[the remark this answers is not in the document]
mbertrand: added a test

openapi/specs/v1.yaml
```diff
@@ -11208,7 +11208,7 @@ components:
         content_files:
           type: array
           items:
-            $ref: '#/components/schemas/ContentFile'
+            $ref: '#/components/schemas/NestedContentFile'
           nullable: true
```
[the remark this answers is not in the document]
mbertrand: The load-testing client is intentionally out of scope here. It isn't covered by CI or any typecheck, and the k6 scripts only import `NewsEventsListParams` and `FeaturedListParams` from it — nothing that touches `content_files` or the renamed enum. It gets regenerated with `openapi-to-k6` whenever someone works on the load tests, so I'll leave syncing it to a separate housekeeping change rather than adding generated churn to this PR.

main/utils.py
```diff
@@ -99,12 +129,18 @@ async def inner_function(request, *args, **kwargs):
...
 
                 response = await func(request, *args, **kwargs)
 
                 if response.status_code == 200:  # noqa: PLR2004
-                    await cache_backend.aset(cache_key, response.data, cache_timeout)
+                    # Cache rendered JSON bytes so cache hits skip
+                    # DRF serialization entirely
+                    await cache_backend.aset(
+                        cache_key,
+                        JSONRenderer().render(response.data),
```
shanbady: on the performance side not sure how severe this is but if we have a cache miss we render the response twice (again at line [183](https://github.com/mitodl/mit-learn/pull/3675/changes#diff-2fda929525d41a6e403e5af8da689a7c030a73992de8fdcea1fb9772ad0f80fcR183) )

main/utils_test.py
```diff
@@ -412,6 +420,58 @@ def test_cache_returns_cached_response(mock_caches):
     assert response.data == cached_data
 
 
+@patch("main.utils.caches")
+def test_cache_stores_rendered_json_bytes(mock_caches):
+    """Responses are cached as rendered JSON bytes, not raw data."""
+    mock_cache = MagicMock()
+    mock_cache.get.return_value = None
+    mock_caches.__getitem__.return_value = mock_cache
```
shanbady: might be worth adding a test that sets REDIS_VIEW_CACHE_DURATION to non-zero (so it doesnt bypass the decorator)

shanbady (changes requested):
core functionality looks good. left some minor notes

main/utils.py
shanbady: note: if we do render this once and pass it along ensure the headers are carried through as well

mbertrand: done

main/utils_test.py
mbertrand: done

shanbady (approved):
👍 LGTM

PR #3679, opened 2026-07-24
feat(b2b): org analytics dashboard for the manager surface

blarghmatey:

N/A — no GitHub issue. This is the frontend piece of the B2B self-serve analytics work.

**Related PRs, and the required deploy order:**

1. [mitodl/mitxonline#3789](https://github.com/mitodl/mitxonline/pull/3789) — exposes `sso_organization_id` on `OrganizationPageSerializer` and adds the `?sso_organization_id=` filter to the manager org endpoint. **Must deploy first.**
2. [mitodl/ol-analytics-api#13](https://github.com/mitodl/ol-analytics-api/pull/13) — keys the org endpoints on the Keycloak org UUID. Deploying this before #3789 fails open: an older manager endpoint ignores the unknown filter param and returns *all* of the caller's managed orgs.
3. This PR — safe to merge at any point; the dashboard is behind a feature flag and reports itself unavailable until the API exists.

Identifier analysis: [mitodl/hq#11845](https://github.com/mitodl/hq/issues/11845).

Adds the **reporting** half of the B2B org-manager dashboard, alongside the seat-administration half that already exists in [`ContractAdminPage`](https://github.com/mitodl/mit-learn/blob/main/frontends/main/src/app-pages/ContractAdminPage/ContractAdminPage.tsx). Four sections at `/dashboard/organization/[orgSlug]/analytics`, each backed by one of `ol-analytics-api`'s org-scoped materialized views:

| Section | Endpoint | Backing view |
|---|---|---|
| Contract utilization (KPI cards) | `/contract-utilization` | `mv_b2b_contract_utilization` |
| Monthly engagement (line chart) | `/engagement-trend` | `mv_b2b_monthly_engagement_trend` |
| Course performance (table) | `/enrollment-funnel` | `mv_b2b_enrollment_completion_funnel` |
| Program funnel (bars + table) | `/program-funnel` | `mv_b2b_program_funnel` |

A query factory for the fifth endpoint (`/content-engagement`) is included; its panel is deliberately out of scope for this PR.

Reachable from a "View analytics" link added to the contract admin page header, behind the same flag.

#### Charting: `@mui/x-charts`, not `recharts`

The original scope called for `recharts`. After evaluating both against this codebase I went with `@mui/x-charts@^8`:

- **Theme integration.** The app is MUI v6 + emotion throughout, with design tokens in `theme.custom.colors` via `ol-components`. x-charts reads that theme directly, so axis, tooltip and legend styling comes from the same tokens as everything else. Recharts styles per-component through props, so every token would have to be threaded by hand.
- **Dependency weight.** Recharts 3 ships `@reduxjs/toolkit`, `react-redux`, `immer` and `es-toolkit` as *runtime* dependencies — a second state-management stack inside a chart library, in an app that uses TanStack Query. Roughly 18MB installed vs ~6.6MB for x-charts.
- **Compatibility.** `@mui/x-charts@8` peer-accepts `@mui/material ^6.0.0` (we're on 6.4.5) and React 19. The charts needed here (Line, Bar) are MIT/community; only the dedicated Funnel *widget* is Pro, and this uses ordinary bars.

Trade-off worth knowing: x-charts pulls `@mui/utils@^7` as a direct dependency alongside MUI v6, so that small utils package is duplicated in the tree.

#### Organizations are keyed on the Keycloak UUID, not the slug

Per ol-analytics-api#13, every org endpoint filters on `sso_organization_id` — the only identifier stable across the JWT, MITx Online and StarRocks. The route keeps `[orgSlug]` (human-facing, consistent with its sibling routes) and `AnalyticsContent` resolves slug → manager org → UUID.

Because mitxonline#3789 has not been cut into a release of `@mitodl/mitxonline-api-axios` (we're pinned to 2026.7.7), the generated `OrganizationPage` type does not declare the field yet. `Analytics/orgUuid.ts` reads it off the wire and returns `null` when absent — so pointed at an older MITx Online, the page shows "Analytics is not available for this organization yet" rather than issuing a request with `undefined` in the path. That shim is marked for deletion once the client is regenerated.

#### Suppressed values are never rendered as zero

The API nulls out any learner count below its k-anonymity floor, plus every rate and average derived from it. Those nulls are carried through as nulls:

- Table cells render a marked placeholder with the reason, exposed to assistive tech via `aria-label`, not hover alone.
- A suppressed month is a **gap** in the trend line, not a dip to zero.
- The program funnel is paired with a table of the same numbers — a bar chart can only omit a bar, which reads as zero.

#### Auth

Unchanged from the rest of the app: cookie-based, `withCredentials` + CSRF, no bearer token. The analytics API sits behind APISIX, which mints the JWT from the session cookie. Authorization is the API's job; this component only resolves *which* org is meant, via MITx Online's manager org list — the same source `ContractAdminPage` uses — so a non-manager never issues an analytics request.

#### Chart colors were validated, not chosen by eye

Hues come from the smoot-design token set, then checked against lightness band, chroma floor, CVD separation under simulated protanopia/deuteranopia, a normal-vision separation floor, and ≥3:1 contrast on the surface. Two slots are *stepped* versions of brand hues because the tokens themselves fall outside the readable band (`orange` `#FAB005` is far too light to carry a 2px line; `lightBlue` is too pale to distinguish from the card). Green and red are deliberately excluded from the series palette so they stay reserved for status. Rationale is documented in `Analytics/chartPalette.ts` — changing a value there means re-running the check, not swapping a hex.

Funnel stages take a single-hue light→dark ramp rather than three unrelated hues, because stage order carries meaning. The trend chart deliberately omits the view's `total_videos_watched` / `total_problems_attempted` / `total_chatbot_interactions` columns: those are counts of *events*, orders of magnitude above the learner counts, and putting them on the same plot would need a second y-axis.

#### Incidental changes

- **`components/B2BTable`** — the contract admin page's table primitives are extracted so both manager surfaces share them (including the sub-`md` reflow where rows become label/value stacks). `ContractAdminPage`'s 122 tests still pass unchanged.
- **`structuredClone` polyfill** in `frontends/jest-shared-setup.ts` — jsdom does not provide it and `@mui/x-charts` calls it, so without this *any* test rendering a chart dies with a `ReferenceError`. Implemented with the v8 serializer rather than a JSON round-trip so Dates/Maps/Sets survive.
- **`NEXT_PUBLIC_ANALYTICS_API_BASE_URL` is optional** in the yup schema. Environments without an analytics deployment still boot; `bootstrapApiClients()` simply skips configuring that client and the page reports itself unavailable. This avoids breaking startup for every environment that doesn't have the service yet.

### How can this be tested?

**Automated** — all green:

```
yarn workspace frontends run lint-check     # clean
yarn workspace api run global:typecheck     # clean
yarn workspace main run global:typecheck    # clean
cd frontends/main && NODE_ENV=test npx jest src/app-pages/DashboardPage src/app-pages/ContractAdminPage src/bootstrap
cd frontends/api  && NODE_ENV=test npx jest src/analytics
```

30 new tests covering: the five query factories (correct URL per endpoint, paging params, per-org key namespacing, org-id URL encoding); formatters; both charts rendering with the validated palette; and the page's access / availability / freshness / suppression behaviour.

Note on the full suite: `AssignSeatsSection`, `CoursePage`, `NewsEditor` and `SearchPage` fail intermittently under parallel load. All pass in isolation, and `AssignSeatsSection` and `NewsEditor` fail identically on an untouched `main` checkout — pre-existing flakes, not regressions from this branch.

**Manual**, once the dependencies are in place:

1. Deploy mitodl/mitxonline#3789, then mitodl/ol-analytics-api#13.
2. Add an APISIX route proxying to the analytics API with the JWT plugin (mirroring the existing `/ai/*` learn-ai route), and set `NEXT_PUBLIC_ANALYTICS_API_BASE_URL` to that route — **not** to the API directly; the API expects the JWT APISIX mints from the session cookie.
3. Create the `b2b-analytics-dashboard` PostHog flag and enable it for your user.
4. As an org manager, visit `/dashboard/organization/<org>/analytics`.

Worth exercising specifically:

- **Not a manager of that org** → "Access denied" (decided before any analytics request is issued).
- **An analytics 403** → propagates to the route error boundary. This is intentional and not handled in the component: [`makeBrowserQueryClient`](https://github.com/mitodl/mit-learn/blob/main/frontends/main/src/app/getQueryClient.ts) sets `throwOnError` for 400/401/403, so those never reach an `isError` branch. I wrote an in-component 403 branch first and removed it once a test proved it was unreachable.
- **Org without a Keycloak UUID** (or `NEXT_PUBLIC_ANALYTICS_API_BASE_URL` unset) → "not available" notice, and **no request is issued**.
- **An org whose data is below the anonymity floor** → placeholders reading "Withheld: too few learners…", never `0`.
- **A view that has never refreshed** (`as_of: null`) → "Data not yet refreshed" rather than an implied freshness.
- **Multiple contracts** → one KPI card group per contract (see below).

Locally without the API, set nothing and confirm the page renders the unavailable notice rather than erroring.

### Additional Context

**KPI cards are per-contract, not rolled up org-wide — deliberately.** `mv_b2b_contract_utilization`'s grain is org × contract, and none of the three headline figures can be honestly aggregated here: `active_learners` counts *distinct* learners per contract, so summing double-counts anyone on two contracts; and `seat_utilization_pct` / `completion_rate_pct` are rates, where averaging across contracts of different sizes is the wrong number and recomputing from raw counts is impossible whenever a count has been suppressed. Most orgs have one contract, so this reads as a single KPI row.

**A bug this caught:** `formatDate("2026-01-15")` was rendering as **Jan 14** — `new Date()` parses a date-only ISO string as UTC midnight, which `toLocaleDateString` then renders as the previous day for anyone west of Greenwich. A contract starting on the 15th must not display as starting on the 14th. Fixed for both contract dates and the `YYYY-MM` trend labels; regression tests cover both.

**Follow-up work already identified** (not in this PR):

- Add OpenAPI export + TypeScript client generation to `ol-analytics-api`, replacing the hand-written types in `frontends/api/src/analytics/types.ts`. Worth flagging for whoever picks it up: the b2b_dashboard tenant is a *mounted sub-app*, so it owns its own `/openapi.json` and the root schema does not include it; and its endpoints are registered in a loop with a runtime-parametrized generic, so the generated schema needs checking that each row model is named distinctly rather than collapsed into one `OrgAnalyticsResponse`.
- Remove the `sso_organization_id` shim once the regenerated mitxonline client lands.
- Build the content-engagement panel (factory and types are already here).
- Provision the APISIX route, the env var, and the PostHog flag.

**Screenshots** are omitted because no environment can render the dashboard until the analytics API is deployed. I'll attach desktop and mobile captures once it's up in CI/RC.

### DIFF
--- a/env/frontend.local.example.env
+++ b/env/frontend.local.example.env
@@ -12,8 +12,11 @@ NEXT_PUBLIC_STAY_UPDATED_HUBSPOT_FORM_ID=""
 # NEXT_PUBLIC_LEARN_AI_RECOMMENDATION_ENDPOINT=http://open.odl.local:8065/ai/http/recommendation_agent/
 # NEXT_PUBLIC_LEARN_AI_SYLLABUS_ENDPOINT=http://open.odl.local:8065/ai/http/syllabus_agent/
 
-# ol-analytics-api integration: powers the B2B org analytics dashboard at
-# /dashboard/organization/<org>/analytics. Leave unset to keep the dashboard
-# hidden. Point at the APISIX route rather than the API directly — the API
-# expects the JWT that APISIX mints from the session cookie.
+# ol-analytics-api integration: supplies the data for the B2B org analytics
+# dashboard at /dashboard/organization/<org>/analytics. Whether that route is
+# reachable at all is controlled by the b2b-analytics-dashboard PostHog flag,
+# not by this variable: with the flag on and this unset the page still renders,
+# reporting "Analytics is not available in this environment".
+# Point at the APISIX route rather than the API directly — the API expects the
+# JWT that APISIX mints from the session cookie.
 # NEXT_PUBLIC_ANALYTICS_API_BASE_URL=http://open.odl.local:8065/analytics/
--- a/env/frontend.local.example.env
+++ b/env/frontend.local.example.env
@@ -11,3 +11,9 @@ NEXT_PUBLIC_STAY_UPDATED_HUBSPOT_FORM_ID=""
 # in the README. Requires learn-ai running locally.
 # NEXT_PUBLIC_LEARN_AI_RECOMMENDATION_ENDPOINT=http://open.odl.local:8065/ai/http/recommendation_agent/
 # NEXT_PUBLIC_LEARN_AI_SYLLABUS_ENDPOINT=http://open.odl.local:8065/ai/http/syllabus_agent/
+
+# ol-analytics-api integration: powers the B2B org analytics dashboard at
+# /dashboard/organization/<org>/analytics. Leave unset to keep the dashboard
+# hidden. Point at the APISIX route rather than the API directly — the API
+# expects the JWT that APISIX mints from the session cookie.
+# NEXT_PUBLIC_ANALYTICS_API_BASE_URL=http://open.odl.local:8065/analytics/
--- a/frontends/api/src/analytics/types.ts
+++ b/frontends/api/src/analytics/types.ts
@@ -23,10 +23,16 @@
  * endpoint, so it is per-section rather than per-page: one lagging view cannot
  * make another section look fresher than it is. It is `null` until that view
  * has refreshed for the first time.
+ *
+ * `total_count` is how many rows the org has in that view across every page,
+ * already net of the anonymity floor. `data.length` alone cannot distinguish a
+ * complete result from one truncated at the page cap, so this is what lets the
+ * dashboard admit to showing a subset instead of quietly dropping the rest.
  */
 export type OrgAnalyticsResponse<RowT> = {
   organization_id: string
   as_of: string | null
+  total_count: number
   data: RowT[]
 }
 
--- a/frontends/api/src/analytics/types.ts
+++ b/frontends/api/src/analytics/types.ts
@@ -1,0 +1,125 @@
+/**
+ * Response types for the B2B dashboard tenant of the OL Analytics API.
+ *
+ * These are hand-written rather than generated: `ol-analytics-api` does not
+ * publish a TypeScript client the way MITx Online does (`@mitodl/mitxonline-api-axios`).
+ * They mirror `tenants/b2b_dashboard/models.py` — which in turn mirrors the
+ * StarRocks materialized views owned by dbt in `ol-data-platform` — column for
+ * column. When a view gains a column, update the matching type here.
+ *
+ * # Why so many nullable numbers
+ *
+ * The API applies a k-anonymity floor: any distinct-learner count below the
+ * floor, and every rate/average derived from it, comes back `null`. A `null`
+ * therefore means "suppressed to protect learner privacy", NOT zero and NOT
+ * missing — render it as such (see `SuppressibleValue` in the dashboard) and
+ * never coerce it to 0 in a chart or an average.
+ */
+
+/**
+ * Envelope shared by every org-scoped endpoint.
+ *
+ * `as_of` is the last refresh time of the single materialized view backing this
+ * endpoint, so it is per-section rather than per-page: one lagging view cannot
+ * make another section look fresher than it is. It is `null` until that view
+ * has refreshed for the first time.
+ */
+export type OrgAnalyticsResponse<RowT> = {
+  organization_id: string
+  as_of: string | null
+  data: RowT[]
+}
+
+/** `mv_b2b_contract_utilization` — grain: org x contract. */
+export type ContractUtilization = {
+  organization_key: string
+  organization_name: string
+  contract_pk: number
+  b2b_contract_name: string
+  b2b_contract_is_active: boolean
+  b2b_contract_start_date: string | null
+  b2b_contract_end_date: string | null
+  seat_limit: number | null
+  b2b_contract_membership_type: string | null
+  seats_consumed: number
+  active_learners: number | null
+  learners_certified: number | null
+  seat_utilization_pct: number | null
+  completion_rate_pct: number | null
+}
+
+/** `mv_b2b_enrollment_completion_funnel` — grain: org x contract x course run. */
+export type EnrollmentCompletionFunnel = {
+  organization_key: string
+  organization_name: string
+  contract_pk: number
+  b2b_contract_name: string
+  courserun_pk: number
+  courserun_readable_id: string
+  courserun_title: string
+  enrolled_learners: number
+  active_learners: number | null
+  passing_learners: number | null
+  certified_learners: number | null
+  active_rate_pct: number | null
+  completion_rate_pct: number | null
+}
+
+/**
+ * `mv_b2b_monthly_engagement_trend` — grain: org x year_month.
+ *
+ * `activity_year_and_month` is a `YYYY-MM` string, not a date.
+ */
+export type MonthlyEngagementTrend = {
+  organization_key: string
+  organization_name: string
+  activity_year_and_month: string
+  monthly_active_learners: number
+  new_enrollments: number | null
+  certificates_earned: number | null
+  total_videos_watched: number
+  total_problems_attempted: number
+  total_chatbot_interactions: number
+}
+
+/** `mv_b2b_program_funnel` — grain: org x contract x program. */
+export type ProgramFunnel = {
+  organization_key: string
+  organization_name: string
+  contract_pk: number
+  b2b_contract_name: string
+  program_pk: number
+  program_title: string
+  total_courses: number
+  enrolled_in_contract_courses: number
+  enrolled_via_program: number | null
+  program_course_completers: number | null
+}
+
+/** `mv_b2b_content_engagement_depth` — grain: org x course run, all-time. */
+export type ContentEngagementDepth = {
+  organization_key: string
+  organization_name: string
+  courserun_readable_id: string
+  courserun_title: string
+  total_enrolled_learners: number
+  engaged_learners: number | null
+  engagement_rate_pct: number | null
+  total_videos_watched: number | null
+  avg_videos_per_engaged_learner: number | null
+  total_problems_attempted: number | null
+  avg_problems_per_engaged_learner: number | null
+  total_chatbot_interactions: number | null
+  chatbot_users: number | null
+  chatbot_adoption_pct: number | null
+  certificates_earned: number | null
+}
+
+/**
+ * LIMIT/OFFSET paging, shared by every multi-row endpoint. The API caps `limit`
+ * at its own `max_page_size` and rejects anything larger with a 422.
+ */
+export type AnalyticsPageParams = {
+  limit?: number
+  offset?: number
+}
--- a/frontends/main/src/app-pages/ContractAdminPage/ContractAdminPage.tsx
+++ b/frontends/main/src/app-pages/ContractAdminPage/ContractAdminPage.tsx
@@ -12,6 +12,7 @@ import {
   alpha,
   Chip,
   Container,
+  Link,
   Pagination,
   SearchInput,
   Skeleton,
@@ -30,6 +31,18 @@ import {
 import { AssignSeatsSection } from "./AssignSeatsSection"
 import { RowActionMenu } from "./RowActionMenu"
 import {
+  EmptyTableMessage,
+  MobileLabel,
+  STUB,
+  TableCard,
+  TableCell,
+  TableFooter,
+  TableFootnote,
+  TableHeaderCell,
+  TableHeaderRow,
+  TableRow,
+} from "@/components/B2BTable/B2BTable"
+import {
   managerOrganizationQueries,
   type ManagerEnrollmentCode,
 } from "api/mitxonline-hooks/organizations"
@@ -38,6 +51,7 @@ import { matchOrganizationBySlug } from "@/common/utils"
 import { ForbiddenError } from "@/common/errors"
 import { FeatureFlags } from "@/common/feature_flags"
 import { useFeatureFlagsLoaded } from "@/common/useFeatureFlagsLoaded"
+import { organizationAnalyticsView } from "@/common/urls"
 import { ErrorContent } from "../ErrorPage/ErrorPageTemplate"
 import graduateLogo from "@/public/images/dashboard/graduate.png"
 
@@ -93,6 +107,12 @@ const ContractSubtitle = styled(Typography)(({ theme }) => ({
   ...theme.typography.subtitle1,
   color: theme.custom.colors.silverGrayDark,
 })) as typeof Typography
+
+const AnalyticsLink = styled(Link)(({ theme }) => ({
+  ...theme.typography.body2,
+  display: "inline-block",
+  paddingTop: "4px",
+}))
 
 const StatsSide = styled.div(({ theme }) => ({
   display: "flex",
@@ -185,89 +205,6 @@ const ControlsLeft = styled.div(({ theme }) => ({
   },
 }))
 
-const TableCard = styled.div(({ theme }) => ({
-  backgroundColor: theme.custom.colors.white,
-  border: `1px solid ${theme.custom.colors.lightGray2}`,
-  borderRadius: "8px",
-  padding: "24px",
-  [theme.breakpoints.down("md")]: {
-    padding: "16px",
-  },
-}))
-
-const TableHeaderRow = styled.div(({ theme }) => ({
-  display: "flex",
-  gap: "16px",
-  alignItems: "center",
-  paddingBottom: "16px",
-  borderBottom: `1px solid ${theme.custom.colors.silverGrayDark}`,
-  [theme.breakpoints.down("md")]: {
-    display: "none",
-  },
-}))
-
-const TableHeaderCell = styled("div", {
-  shouldForwardProp: (prop) => prop !== "$flex",
-})<{ $flex: number }>(({ $flex, theme }) => ({
-  flex: $flex,
-  minWidth: 0,
-  ...theme.typography.subtitle2,
-  color: theme.custom.colors.black,
-}))
-
-const TableRow = styled.div(({ theme }) => ({
-  display: "flex",
-  gap: "16px",
-  alignItems: "center",
-  padding: "14px 0",
-  borderBottom: `1px solid ${theme.custom.colors.silverGrayLight}`,
-  "&:last-child": {
-    borderBottom: "none",
-  },
-  [theme.breakpoints.down("md")]: {
-    position: "relative",
-    flexWrap: "wrap",
-    gap: "6px 0",
-    padding: "16px 40px 16px 0",
-  },
-}))
-
-const MobileLabel = styled.span(({ theme }) => ({
-  display: "none",
-  [theme.breakpoints.down("md")]: {
-    display: "inline",
-    ...theme.typography.subtitle2,
-    color: theme.custom.colors.darkGray2,
-    minWidth: "120px",
-    flexShrink: 0,
-  },
-}))
-
-const TableCell = styled("div", {
-  shouldForwardProp: (prop) => prop !== "$flex" && prop !== "$primary",
-})<{ $flex: number; $primary?: boolean }>(({ $flex, $primary, theme }) => ({
-  flex: $flex,
-  minWidth: 0,
-  ...theme.typography.body2,
-  color: theme.custom.colors.black,
-  overflow: "hidden",
-  textOverflow: "ellipsis",
-  whiteSpace: "nowrap",
-  [theme.breakpoints.down("md")]: {
-    flex: "none",
-    width: "100%",
-    display: "flex",
-    alignItems: "center",
-    gap: "8px",
-    overflow: "visible",
-    whiteSpace: "normal",
-    ...($primary && {
-      ...theme.typography.subtitle2,
-      marginBottom: "4px",
-    }),
-  },
-}))
-
 const StatusBadge = styled(Chip, {
   shouldForwardProp: (prop) => prop !== "$status",
 })<{ $status: "assigned" | "redeemed" }>(({ $status, theme }) => ({
@@ -299,27 +236,6 @@ const ActionCell = styled.div(({ theme }) => ({
     right: 0,
   },
 }))
-
-const TableFooter = styled.div({
-  display: "flex",
-  justifyContent: "space-between",
-  alignItems: "center",
-  paddingTop: "16px",
-})
-
-const TableFootnote = styled(Typography)(({ theme }) => ({
-  ...theme.typography.body2,
-  color: theme.custom.colors.silverGrayDark,
-})) as typeof Typography
-
-const EmptyTableMessage = styled(Typography)(({ theme }) => ({
-  ...theme.typography.body2,
-  color: theme.custom.colors.silverGrayDark,
-  padding: "32px 0",
-  textAlign: "center",
-})) as typeof Typography
-
-const STUB = "—"
 
 type StatusFilter = "all" | "pending" | "redeemed"
 
@@ -365,6 +281,9 @@ const ContractAdminPageInternal: React.FC<ContractAdminPageInternalProps> = ({
   contractSlug,
 }) => {
   const queryClient = useQueryClient()
+  const analyticsEnabled = useFeatureFlagEnabled(
+    FeatureFlags.B2BAnalyticsDashboard,
+  )
   const [statusFilter, setStatusFilter] = useState<StatusFilter>("all")
   const [searchQuery, setSearchQuery] = useState("")
   const [debouncedSearchQuery, setDebouncedSearchQuery] = useState("")
@@ -633,6 +552,19 @@ const ContractAdminPageInternal: React.FC<ContractAdminPageInternalProps> = ({
                   </>
                 ) : null}
               </ContractSubtitle>
+              {/* Analytics is the reporting half of this dashboard and is
+                  org-scoped, so it lives at its own URL rather than under this
+                  contract. Behind its own flag: the analytics API is not
+                  deployed everywhere this page is. */}
+              {analyticsEnabled ? (
+                <AnalyticsLink
+                  color="red"
+                  size="small"
+                  href={organizationAnalyticsView(orgSlug)}
+                >
+                  View analytics
+                </AnalyticsLink>
+              ) : null}
             </div>
           </OrgDetailsContainer>
           <StatsSide>
--- a/frontends/main/src/app-pages/DashboardPage/Analytics/EngagementTrendChart.tsx
+++ b/frontends/main/src/app-pages/DashboardPage/Analytics/EngagementTrendChart.tsx
@@ -1,0 +1,144 @@
+"use client"
+
+import React from "react"
+import { LineChart } from "@mui/x-charts/LineChart"
+import { Skeleton, styled, Typography } from "ol-components"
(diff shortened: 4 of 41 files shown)


env/frontend.local.example.env
```diff
@@ -11,3 +11,9 @@ NEXT_PUBLIC_STAY_UPDATED_HUBSPOT_FORM_ID=""
 # in the README. Requires learn-ai running locally.
 # NEXT_PUBLIC_LEARN_AI_RECOMMENDATION_ENDPOINT=http://open.odl.local:8065/ai/http/recommendation_agent/
 # NEXT_PUBLIC_LEARN_AI_SYLLABUS_ENDPOINT=http://open.odl.local:8065/ai/http/syllabus_agent/
+
+# ol-analytics-api integration: powers the B2B org analytics dashboard at
+# /dashboard/organization/<org>/analytics. Leave unset to keep the dashboard
+# hidden. Point at the APISIX route rather than the API directly — the API
+# expects the JWT that APISIX mints from the session cookie.
```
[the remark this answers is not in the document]
blarghmatey: Good catch — fixed in b2534db10.

You're right that the env var doesn't control visibility. Reachability is gated by the `b2b-analytics-dashboard` PostHog flag (`AnalyticsContent` throws `ForbiddenError` when it's off); with the flag on and the var unset, the route renders and shows the "Analytics is not available in this environment" notice, exactly as you describe.

Reworded both places — `env/frontend.local.example.env` said "keep the dashboard hidden" and `env/frontend.env` said "disable the org analytics dashboard", and both were wrong in the same way. They now say the variable decides whether there's an API to load data from, and name the PostHog flag as what gates the route.

frontends/main/src/app-pages/DashboardPage/AnalyticsContent.tsx
```diff
@@ -0,0 +1,325 @@
...
+            ? // The org record has no Keycloak organization UUID, which is
+              // what the analytics API keys on. Nothing the manager can fix.
+              "Analytics is not available for this organization yet. Please contact support if you expect to see data here."
+            : "Analytics is not available in this environment."}
+        </Notice>
+      </Stack>
+    )
+  }
+
+  const failed = [utilization, trend, courses, programs].some(
```
daniellefrappier18, line 305: Right now `isError` only shows the banner. Since the sections only receive `asOf`/`isLoading`/`rows` (all falsy on a failed query), a failed section ends up showing "Data not yet refreshed" and "No … recorded yet", which are both misleading. Mind passing isError into each section so they can render a proper error state?

frontends/main/src/app-pages/DashboardPage/Analytics/EngagementTrendChart.tsx
```diff
@@ -0,0 +1,144 @@
...
+  // The API orders by activity_year_and_month, but sorting here keeps the chart
+  // correct even if paging ever returns rows out of order.
+  const months = [...rows].sort((a, b) =>
+    a.activity_year_and_month.localeCompare(b.activity_year_and_month),
+  )
+  const labels = months.map((row) => row.activity_year_and_month)
+
+  return (
+    <ChartCard>
+      <LineChart
```
daniellefrappier18: `<ChartCard>` renders only the `<LineChart>`, with no text/table equivalent, so the monthly numbers aren't readable to a screen reader. `ProgramFunnelChart` already solves this with a paired `role="table"` of the same data. Can we mirror that here?

frontends/main/src/app-pages/DashboardPage/AnalyticsContent.tsx
```diff
@@ -0,0 +1,325 @@
...
+  borderRadius: "8px",
+  padding: "24px",
+})) as typeof Typography
+
+/**
+ * These views are small per org (contracts, programs, a couple of years of
+ * months), but the API caps every list endpoint, so ask for a page big enough
+ * that no org is silently truncated at the default.
+ */
+const PAGE_SIZE = 200
```
daniellefrappier18, line 117: The response is capped at 200 rows and the envelope carries no total count, so the UI can't show a "showing 200 of N" message or offer a way to load the rest. If an org ever has more than 200 rows in a section, the extras silently disappear with no indication.

frontends/main/src/app-pages/ContractAdminPage/ContractAdminPage.tsx
```diff
@@ -633,6 +552,19 @@ const ContractAdminPageInternal: React.FC<ContractAdminPageInternalProps> = ({
...
+                  org-scoped, so it lives at its own URL rather than under this
+                  contract. Behind its own flag: the analytics API is not
+                  deployed everywhere this page is. */}
+              {analyticsEnabled ? (
+                <AnalyticsLink
+                  color="red"
+                  size="small"
+                  href={organizationAnalyticsView(orgSlug)}
+                >
+                  View analytics
```
daniellefrappier18, line 565: This link is gated only on `analyticsEnabled`, but `useFeatureFlagEnabled` returns undefined until PostHog finishes loading flags. On a cold load that means the link is hidden until the flags come back over the network, then pops in; on a refresh it shows immediately from PostHog's cache. I hit this in testing (missing on first load, there after a refresh). Could we gate on `useFeatureFlagsLoaded()` too, the same way the page does at line 839, so the link's behavior is deterministic instead of "shows up once PostHog is warm"?

frontends/main/src/app-pages/DashboardPage/AnalyticsContent.tsx
blarghmatey: Fixed in 83bbba4. `isError` now goes to every section as well as the banner.

Each section renders a shared `SectionError` ("This data could not be loaded. Please try again later.") in place of its empty state, and `SectionHeader` drops the freshness slot entirely when `isError` — a failed request tells us nothing about when the view last refreshed, so "Data not yet refreshed" was a claim we had no grounds for, same as "No … recorded yet" was a claim about the org.

Covered by a new page test that 500s one endpoint and asserts the failed section shows the error copy while the other three keep their own "Data as of" stamps.

frontends/main/src/app-pages/DashboardPage/Analytics/EngagementTrendChart.tsx
blarghmatey: Fixed in 83bbba4 — mirrored `ProgramFunnelChart` exactly: same `role="table"` built from the B2BTable primitives, one row per month with Active learners / New enrollments / Certificates earned, and the same `SUPPRESSED_EXPLANATION` footnote when any month is withheld.

That last part turned out to matter more than the screen-reader case alone: a suppressed month is drawn as a gap in the line, which reads identically to "no data". Only the table can say it was withheld and why.

One pre-existing test needed scoping to the legend (`.MuiChartsLabel-root`) since the table now legitimately repeats each series name as a column header.

frontends/main/src/app-pages/DashboardPage/AnalyticsContent.tsx
blarghmatey: Good catch — this needed the API side first, so: mitodl/ol-analytics-api#15 adds `total_count` to both envelopes, and ef20f352 consumes it here.

Each section now compares its rendered rows against the total and renders "Showing 2 of 340." plus a "Show all 340" button that refetches that section alone at a larger limit. The ask is capped at the API's `max_page_size` (1000, it 422s above that) — past the cap the message stands alone rather than offering a button that cannot deliver the rest.

Worth flagging one thing the API PR had to get right: the count is not a plain `COUNT(*)`. Sub-floor rows are dropped by `suppress_small_cohorts` *after* the query returns, so a raw count would exceed anything paging can reach — and subtracting the rows you do receive would hand you the exact number of sub-floor cohorts in that org, which is what the k-anonymity floor exists to prevent. `build_count` applies the same primary-cohort gate in SQL, so the total is consistent with the data and discloses nothing the rows don't.

Note this PR now depends on ol-analytics-api#15 landing and deploying first.

frontends/main/src/app-pages/ContractAdminPage/ContractAdminPage.tsx
blarghmatey: Traced this one and I don't think the mechanism holds — leaving the thread open because you saw something real and I'd like to know what.

`AnalyticsLink` is inside `ContractAdminPageInternal`, and that component has exactly one render site (line 854), behind the outer `ContractAdminPage`'s own gate:

```
if (!flagsLoaded) return <Page><Skeleton /></Page>
```

So `ContractAdminPageInternal` never mounts until `flagsLoaded` is already true. And per `useFeatureFlagsLoaded`'s implementation, that returns true only once flags have come back from the PostHog *server* — it deliberately distinguishes that from bootstrapped-from-localStorage by watching `INTERNAL_BOOTSTRAPPING_FLAG` go `undefined`. Once the server payload is in, `useFeatureFlagEnabled(B2BAnalyticsDashboard)` is a real boolean, not `undefined`. Adding `useFeatureFlagsLoaded()` at line 284 would be reading a value that is already guaranteed true there.

It also doesn't match the direction of what you describe: bootstrapping makes things appear *sooner* on a repeat visit, and while bootstrapped `flagsLoaded` is false, so a refresh shows the skeleton first, not the link.

What I think you may have hit instead: a newly-created PostHog flag isn't in a given person's `/decide` payload until their flags refresh, so it reads `undefined` (falsy, link hidden) on one load and resolves on the next — which is PostHog-side propagation, and gating on `flagsLoaded` wouldn't change it.

Could you confirm whether it was the local dev PostHog or a deployed env, and whether the flag had just been created? If it reproduces somewhere the outer gate isn't in play, I've missed a path and will fix it — I just don't want to add a guard that's provably a no-op and leave the actual cause in place.

blarghmatey:
Review pass summary — 3 threads fixed and resolved, 1 replied to and left open.

**Fixed (83bbba4)**
- Failed sections no longer borrow the empty state. `isError` now reaches all four sections and `SectionHeader`, so a section whose query 500s says "This data could not be loaded" instead of "No … recorded yet" / "Data not yet refreshed" — neither of which was a claim we had grounds for.
- `EngagementTrendChart` is paired with a `role="table"` mirroring `ProgramFunnelChart`, including the suppression footnote. That mattered more than the screen-reader case alone: a withheld month is drawn as a gap in the line, which reads identically to "no data".

**Fixed (ef20f352, depends on mitodl/ol-analytics-api#15)**
- The page cap is no longer silent. The envelope gained `total_count`, so each section renders "Showing 2 of 340." with a "Show all" that refetches that section alone, capped at the API's `max_page_size`.
- Note the API side deliberately does *not* use a plain `COUNT(*)`: sub-floor rows are dropped after the query returns, so a raw count would exceed what paging can reach, and the difference would disclose the exact number of sub-floor cohorts in the org. `build_count` applies the same primary-cohort gate in SQL.

**Still open**
- The `ContractAdminPage` flag-gating thread. I couldn't reproduce the mechanism as described — `ContractAdminPageInternal` has a single render site already behind the outer `flagsLoaded` gate — so I replied with the trace instead of adding a guard that would be a no-op. Would rather find the real cause; details in the thread.

**Also**
- Merged `main` in. `openapi-diff` was failing on branch staleness (18 commits behind, `global_id` had landed on main), not on anything in this PR. All checks green now.

**Merge order:** ol-analytics-api#15 needs to land and deploy before this one, since the truncation UI reads `total_count`.

frontends/main/src/app-pages/DashboardPage/Analytics/chartPalette.ts
```diff
@@ -0,0 +1,61 @@
...
+  "#7FA4EA", // lightest — widest stage
+  "#1966FF",
+  "#002896", // darkest — narrowest stage
+] as const
+
+/**
+ * Non-data ink. Grid and axis lines stay recessive so the marks carry the
+ * chart; labels wear text tokens rather than the series color.
+ */
+const CHART_INK = {
```
daniellefrappier18: The `CHART_INK` values are hardcoded hex, but the trailing comments (// lightGray2, // silverGrayDark, etc.) give away that they're just copies of tokens we already have. That feels a little risky to me. If smoot-design ever retunes one of those tokens, the chart ink quietly falls out of sync with the rest of the UI and nothing here would catch it. Could we pull these from the theme instead of duplicating the hexes?

daniellefrappier18 (commented):
There is this strange truncation is a layout issue on the EngagementTrendChart 

Image

 The x-axis uses scaleType: "point" , which places the first and last categories flush against the left and right edges of the plot area. So the "Feb" tick is centered right at the far-right edge, and its label needs roughly half its width of overhang past that point to render fully. But the chart's right margin is only 8px, but 8px isn't enough for the label's overhang, so MUI x-charts truncates the tick with an ellipsis rather than let it spill outside the SVG.

Can you try:
`margin={{ left: 8, right: 32, top: 8, bottom: 0 }}`

32px is _should_ enough for a 3-letter month.

daniellefrappier18 (commented):
Heads up on mobile the hover/tap tooltip gets clipped off the left edge of the screen.

Image

A couple of things are combining to cause it:

- On small screens the chart lives inside `ChartWrapper`, which is an `overflowX: auto` scroll container with `minWidth: 560px`. So the plot itself is wider than the phone viewport and scrolls sideways.
- The x-charts tooltip is a portaled Popper anchored to the bar you tapped. Its header is the full y-axis label, which here is `program_title` plus the contract name (for example "Statistics and Data Science MicroMasters (FY26 Site License, Data Science Track)"). That string is long, and the default tooltip has no max width, so it renders wider than the screen. When the anchored bar sits toward the right of the 560px chart, the wide tooltip gets pushed left and its start runs off the edge.


Possible fixes, roughly in order of least invasive:

1. Constrain the tooltip and let it wrap. In x-charts v8 you can style the default tooltip through `slotProps={{ tooltip: { sx: { maxWidth: 280, "& .MuiChartsTooltip-...": { whiteSpace: "normal" } } } }}`, so a long title wraps to a couple of lines instead of overflowing. 
2. Pass Popper overflow modifiers (`preventOverflow` / `flip`) through the tooltip's popper slot so it shifts to stay inside the viewport rather than bleeding past the edge.
3. Trim what the tooltip shows. The header only needs to identify the row, and we already render the full program plus contract name in the accessible table right below the chart. We could shorten the tooltip header to just the program title (drop the contract suffix there) so it stops being the thing that blows out the width.

frontends/main/src/app-pages/DashboardPage/Analytics/chartPalette.ts
blarghmatey: Agreed, fixed in aef6458. `CHART_INK` is now a `chartInk(theme)` lookup reading `lightGray2` / `silverGrayLight` / `silverGrayDark` / `white` off `theme.custom.colors`.

I checked the four hexes against the current token set before switching — all four matched exactly, so this is a pure refactor with no rendering change today. `useTheme` is now re-exported from `ol-components` alongside `styled` and `alpha`, since smoot-design does not export its `colors` object publicly.

One deliberate asymmetry, now written into the module docstring: `CATEGORICAL` and `FUNNEL_STAGES` stay pinned as hexes. Those are validated data colors — lightness band, chroma floor, CVD separation under protanopia/deuteranopia, 3:1 against the surface — so they must not silently follow a token retune; changing one is supposed to mean re-running that check. Non-data ink is the opposite case and should track the design system, which is exactly your point.

Added a `chartInk` unit test asserting it resolves to the live tokens. Deliberately not asserting a hex literal there — that would just move the duplication into the test and would still pass if the module went back to hardcoding.

frontends/api/src/analytics/types.ts
```diff
@@ -0,0 +1,131 @@
...
+/**
+ * `mv_b2b_monthly_engagement_trend` — grain: org x year_month.
+ *
+ * `activity_year_and_month` is a `YYYY-MM` string, not a date.
+ */
+export type MonthlyEngagementTrend = {
+  organization_key: string
+  organization_name: string
+  activity_year_and_month: string
+  monthly_active_learners: number
```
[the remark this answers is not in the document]
blarghmatey: Verified against the API and this is a false positive — no change made.

`monthly_active_learners` is the **primary** cohort for this view, not a secondary count. The distinction decides the nullability:

```python
# tenants/b2b_dashboard/models.py
cohort_policy = CohortPolicy(
    primary="monthly_active_learners",
    secondary=("new_enrollments", "certificates_earned"),
)
```

A sub-floor **primary** does not null the field — it withholds the entire row:

```python
# core/anonymization.py, suppress_small_cohorts
if (row.get(policy.primary) or 0) < floor:
    continue          # row dropped, never emitted
```

Only `secondary` counts are nulled in place, which is why `new_enrollments` and `certificates_earned` are `number | null` here and `monthly_active_learners` is not. The API model types it `monthly_active_learners: int` (non-optional) — the TS mirrors the contract correctly.

So a month with 3 active learners never reaches the client at all; there is no response in which this field arrives as `null`. Widening the type would make the honest branches unreachable and force callers to handle a case the API cannot produce.

The premise "any distinct-learner count below the floor is returned as null" is right for secondary counts and wrong for the primary one, which gates the row instead.

frontends/main/src/app-pages/ContractAdminPage/ContractAdminPage.tsx
daniellefrappier18: This was a local dev PostHog and newly created and I can not reproduce the issue now so please ignore this one - it's a none issue.

frontends/main/src/app-pages/DashboardPage/Analytics/SectionTruncation.tsx
```diff
@@ -0,0 +1,76 @@
...
+  }
+
+  return (
+    <Root>
+      <Message>
+        Showing {formatCount(shown)} of {formatCount(total)}.
+      </Message>
+      {canShowAll ? (
+        <Button variant="tertiary" size="small" onClick={onShowAll}>
+          Show all {formatCount(total)}
```
daniellefrappier18: Clicking "Show all" raises the section's limit and refetches, but because of `keepPreviousData` the section gates its skeleton on isPending (false during this refetch) so the button stays enabled with no visible or announced change while the larger page loads. Could we mark the async state the same way the CSV export button does on ContractAdminPage (aria-busy + disabled while the expanded page is in flight, label → "Loading…"), and announce the result via a VisuallyHidden live region once it lands? Right now a screen reader gets no feedback that the click did anything.

frontends/main/src/app-pages/ContractAdminPage/ContractAdminPage.tsx
blarghmatey: Thanks for checking — local dev PostHog with a freshly created flag is exactly the propagation case I suspected, so that lines up. Resolving with no change.

frontends/main/src/app-pages/DashboardPage/Analytics/SectionTruncation.tsx
blarghmatey: Fixed in 57fbdfa. Good catch, and it turned out to be covering a second bug underneath it.

**The one you reported.** The control now reports its own state, following the CSV export button: `aria-busy` + `aria-disabled` while the expanded page is in flight, label → "Loading…", and the result announced through a polite live region. `aria-disabled` rather than `disabled` so a keyboard user does not lose focus mid-action, and the click handler no-ops while busy.

One wrinkle worth calling out: the live region sits *outside* the truncation message rather than inside it. When expanding completes a section, the message and the button both unmount — a live region rendered inside that branch would unmount in the same commit as the change it describes and announce nothing at all. The completion case is the one most worth announcing, so it has to outlive the control.

I also derived the announcement from render state rather than detecting a rising/falling edge in an effect. My first attempt did use an effect and the test caught it: an edge detector has to observe the in-flight render to arm itself, and React is free to batch a fast resolution into a single commit, so on a warm cache it would silently never fire.

**The bug underneath.** Your premise was that `keepPreviousData` keeps the rows while `isPending` stays false. It was not doing that at all — `placeholderData: keepPreviousData` is silently inert under `useQueries` here. I instrumented the transition and on a limit change the result went straight to `data: undefined`:

```
DBG courses hasData=true  placeholder=false limit=200
DBG courses hasData=false placeholder=false limit=340   <- blanked
DBG courses hasData=true  placeholder=false limit=340
```

So the section really was blanking back to its skeleton, which is the opposite of what the comment above it claimed. It behaves as documented on `useQuery` (as it already does on ContractAdminPage), so the four queries are now four `useQuery` calls — fixed count and order, so hook-safe. Same instrumentation after the change:

```
DBG courses hasData=true placeholder=true  limit=340   <- previous rows retained
DBG courses hasData=true placeholder=false limit=340
```

Covered by a new `SectionTruncation.test.tsx` for the branch logic (busy state, click-while-busy, both announcement branches, and the completion case where the control has gone) plus a page-level test for the wiring. I deliberately kept the page-level fixtures tiny — my first pass rendered 340 rows into jsdom and was flaky under parallel load.

frontends/main/src/app-pages/DashboardPage/Analytics/EngagementTrendChart.tsx
```diff
@@ -0,0 +1,249 @@
...
+    a.activity_year_and_month.localeCompare(b.activity_year_and_month),
+  )
+  const labels = months.map((row) => row.activity_year_and_month)
+  const hasSuppressed = months.some(
+    (row) => row.new_enrollments === null || row.certificates_earned === null,
+  )
+
+  return (
+    <ChartCard>
+      <LineChart
```
daniellefrappier18: Sorry I didn't see this before but wrap this chart and BarChart as well in an aria-hidden container. This will stop screen reader from traversing the chart's series/axis nodes. See attached recording 

https://github.com/user-attachments/assets/2d582b63-a04a-40a9-bd2d-0726aa87e1cd

daniellefrappier18 (approved):
Nicely done. I added another small a11y issue and these two small UX issues seem to be outstanding:
https://github.com/mitodl/mit-learn/pull/3679#pullrequestreview-4801387954
https://github.com/mitodl/mit-learn/pull/3679#pullrequestreview-4801436117

frontends/main/src/app-pages/DashboardPage/Analytics/EngagementTrendChart.tsx
blarghmatey: Fixed in 5ac42c5 — thanks, the recording made it obvious. Both charts are now wrapped in an `aria-hidden` container: the `LineChart` here and the `BarChart` in `ProgramFunnelChart`.

You're right that this is the natural consequence of the paired tables. Once the same numbers exist as text, the SVG isn't just redundant — a screen reader has to walk hundreds of series, axis and label nodes to reach data it's about to be handed properly a moment later.

One thing I checked before hiding rather than after: `aria-hidden` over a subtree containing focusable content is its own violation, since keyboard focus can land on a node assistive tech has been told doesn't exist. I probed both charts' rendered output and neither emits anything focusable (no `tabindex`, no buttons/links), so hiding them is safe as things stand. I added that as a test assertion alongside the hiding, so if a future `@mui/x-charts` starts emitting focusable nodes for tooltips or keyboard navigation, it fails loudly instead of quietly becoming a keyboard trap.

The tables stay exposed and keep carrying the accessible labels, so nothing is lost — the series names live on the table's column headers rather than the chart legend.

PR #3680, opened 2026-07-25
Set Posthog Cookie on site domain, not mit.edu

ChristopherChudzicki:
For 
- https://github.com/mitodl/hq/issues/11158

Set the posthog cookie on site domain (`open.odl.local`, or `learn.mit.edu`, or `rc.learn.mit.edu`) not MIT domain.

### How can this be tested?
1. With Posthog integrated locally, you should see a `ph_PROJECT_TOKEN_posthog` cookie set on your site domain exactly:
    - if site domain is `open.odl.local`, it should be set there, not on `odl.local`
    - if site domain is `learn.mit.dev`, it should be set there, not on `mit.dev`

### Additional Context
Setting the cookie on the site domain loses our ability to track **anonymous** (unauthenticated) users across sites within the same project, unless we do special setup like pass distinct_ids via query params in URLs, but that is deemed acceptable.

Companion PR, same change for the Keycloak login pages: https://github.com/mitodl/ol-keycloakify/pull/176

mitxonline already does this — it has passed `cross_subdomain_cookie: false` since 2023: https://github.com/mitodl/mitxonline/blob/80c2a512f1e3d2a93c06bb4270ba87976cf369ef/frontend/public/src/lib/util.js#L44

### DIFF
--- a/frontends/main/src/page-components/ConfiguredPostHogProvider/ConfiguredPostHogProvider.tsx
+++ b/frontends/main/src/page-components/ConfiguredPostHogProvider/ConfiguredPostHogProvider.tsx
@@ -68,6 +68,9 @@ const ConfiguredPostHogProvider: React.FC<{ children: React.ReactNode }> = ({
       posthog.init(POSTHOG_API_KEY, {
         api_host: POSTHOG_API_HOST,
         ui_host: POSTHOG_UI_HOST,
+        // Scope the posthog cookie to this site's exact domain. Posthog
+        // defaults this to true, which sets the cookie on the root domain
+        // (e.g. mit.edu), sharing it with every other site there.
         cross_subdomain_cookie: false,
         bootstrap: {
           featureFlags: featureFlags
--- a/frontends/main/src/page-components/ConfiguredPostHogProvider/ConfiguredPostHogProvider.tsx
+++ b/frontends/main/src/page-components/ConfiguredPostHogProvider/ConfiguredPostHogProvider.tsx
@@ -68,6 +68,7 @@ const ConfiguredPostHogProvider: React.FC<{ children: React.ReactNode }> = ({
       posthog.init(POSTHOG_API_KEY, {
         api_host: POSTHOG_API_HOST,
         ui_host: POSTHOG_UI_HOST,
+        cross_subdomain_cookie: false,
         bootstrap: {
           featureFlags: featureFlags
             ? {


frontends/main/src/page-components/ConfiguredPostHogProvider/ConfiguredPostHogProvider.tsx
```diff
@@ -68,6 +68,7 @@ const ConfiguredPostHogProvider: React.FC<{ children: React.ReactNode }> = ({
       posthog.init(POSTHOG_API_KEY, {
         api_host: POSTHOG_API_HOST,
         ui_host: POSTHOG_UI_HOST,
+        cross_subdomain_cookie: false,
         bootstrap: {
```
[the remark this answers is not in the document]
ChristopherChudzicki: I don't think this would provide much value.
- it would just check that two literals are the same—one in the config, one in the test.
- accidental deletion of the literal shows up plainly in a diff already

Analogy would be... This is more akin to env var values we set in ol-infra. We don't write tests in ol-infra that say `NEXT_PUBLIC_ORIGIN` has value `https://learn.mit.edu`.

I **do** think a playwright test asserting the cookie's origin would be useful.

daniellefrappier18 (approved):
LGTM

frontends/main/src/page-components/ConfiguredPostHogProvider/ConfiguredPostHogProvider.tsx
```diff
@@ -68,6 +68,10 @@ const ConfiguredPostHogProvider: React.FC<{ children: React.ReactNode }> = ({
       posthog.init(POSTHOG_API_KEY, {
         api_host: POSTHOG_API_HOST,
         ui_host: POSTHOG_UI_HOST,
+        // Scope the posthog cookie to this site's exact domain. Posthog
+        // defaults this to true, which sets the cookie on the root domain
+        // (e.g. mit.edu), sharing it with every other site there.
+        cross_subdomain_cookie: false,
         bootstrap: {
```
zawan-ila, line 75: We may also need this change in https://github.com/mitodl/ocw-hugo-themes/blob/f88bfc52184d19a7f366b65f225750602ccef6a5/base-theme/assets/js/posthog.ts#L18-L28, perhaps just scoped to the v3 theme? The v3 theme pages are served at urls like https://learn.mit.edu/courses/o/21h-151-dynastic-china-fall-2024/ Currently, they are "gated" behind the ocw-product-pages flag.

ChristopherChudzicki: - https://github.com/mitodl/ocw-hugo-themes/pull/1840

PR #3681, opened 2026-07-27
feat: soft delete feature for news and articles content

ahtesham-quraish:
https://github.com/mitodl/hq/issues/12513

As an article editor / super admin, I want to delete articles and news items directly from the UI, so I can remove outdated, incorrect, or unwanted content without needing Django admin access.

image
image

### How can this be tested?
For testing you need to go to following url `/news/draft`, '/article/draft' and `website_content/news/<draft_id>/edit` and see the delete button is functional


ChristopherChudzicki (changes requested):
This is working well, and for the model change a DateTime `deleted_on` field is definitely better than a boolean — `django-safedelete` uses the same shape (it calls the field `deleted`), so the change I'm suggesting below doesn't lose anything.

I do have some requested changes. The first is a fairly big change, but it simplifies the PR, and I think it's worth it.

I apologize for not suggesting `django-safedelete` when you first wrote the issue. (TBH, I didn't realize that's how we do it in Studio until now.)

### Let's use `django-safedelete`

Safe delete is well-trodden ground and Django has some standard libraries for this. The one we use in OCW Studio is `django-safedelete`, on a model with the same name as ours — see [`websites/models.py`](https://github.com/mitodl/ocw-studio/blob/8e0cd406da0030ffe8fbdce5e312e2b0057a40a0/websites/models.py#L332-L343) and [`websites/admin.py`](https://github.com/mitodl/ocw-studio/blob/8e0cd406da0030ffe8fbdce5e312e2b0057a40a0/websites/admin.py#L43-L67).

Using `django-safedelete` buys us fewer decisions of our own and a pattern to follow. That's a big win. But also:
1. `SafeDeleteModel` that makes `delete()` method soft by default (but optionally a hard delete) 
2. `SafeDeleteQuerySet` so our views etc naturally filter out deleted stuff.
3. `SafeDeleteAdmin` that shows the relevant stuff in Django Admin. *Normally this would take a bit of extra effort with the custom queryset*.
4. Cascade effects we'd need to handle ourselves. (*May not need these yet.*)

(1) + (2) removes essentially all the custom soft delete logic that is in the viewsets. The drafts-only rule stays in the view, either as `perform_destroy` logic or as a `CanEditWebsiteContent` object-level permission check.

### Minor Frontend Issues

- `CONTENT_TYPE_LABELS` is now duplicated two places.
    - Let's define it once, in `@/common/website_content`.
    - Don't use `Record<string, string>`. If we ever add a third content type, we want to update this mapping. And with `Record<string,string>`, nothing enforces that. Key it on the generated enum instead:[^labels]

      ```ts
      import { WebsiteContentContentTypeEnum } from "api/v1"

      export const CONTENT_TYPE_LABELS: Record<WebsiteContentContentTypeEnum, string> = {
        [WebsiteContentContentTypeEnum.News]: "News",
        [WebsiteContentContentTypeEnum.Article]: "Article",
      }
      ```
- **Opinion / Suggestion:** On the draft listing view, the card's delete button should be an icon-only `ActionButton`.
    - That's how we do card actions everywhere else
    - It looks kind of buggy on hover as-is... the background highlight touches bottom of card.
    - Since you added the confirmation dialog (Nice!) that's explicit enough. I don't think we need the "Delete" word also. Plus, trash can is very standard delete symbolism
- `WebsiteContentDraftListingPage.test.tsx` mocks `next/navigation` exports unnecessarily. The test file passes if you remove [these lines](https://github.com/mitodl/mit-learn/blob/b8ca2c9f8276d9825af229e7bc3be465f1d04e9d/frontends/main/src/app-pages/WebsiteContent/WebsiteContentDraftListingPage.test.tsx#L5-L10), since nothing triggers or asserts on router methods.
- Delete Button Placement: You could put `<Spacer />` between "Delete" and "Save As" to reduce misclicks. Misclicks aren't a big deal because of the dialog anyway, but Claude suggested this and I thought it was a reasonable addition.

[^labels]: There's a generated `WebsiteContentContentTypeEnumDescriptions` holding these exact strings already, but its values come from the `WebsiteContentType` labels in `website_content/constants.py`, so a backend rename would quietly change UI copy. Let's take the type from the generated code and keep the strings on the frontend. Note this isn't a pure find-and-replace: `WebsiteContentDraftListingPage`'s `contentType` prop is a plain `string`, so `CONTENT_TYPE_LABELS[type]` won't typecheck until that's narrowed to `WebsiteContentContentTypeEnum` as well.

### On Soft-deleting Published Content via Django Admin

@pdpinch commented on the issue:
> Can I suggest that ... and that we can use the django admin for (soft) deleting published articles when that becomes an issue?

With `django-safedelete`, the admin panel will allow soft-deleting (published or unpublished). If we use that, we should make sure the FeedItem gets deleted, too. Your original acceptance criteria included that:

> Deleting a CMS article/news WebsiteContent also soft-deletes (or suppresses) its derived FeedItem on /news.

I think that's probably pretty simple to add, so we might as well. Two notes:
- It can be a hard delete on FeedItem, no need for soft. The ETL can always reconstruct the FeedItem, and ETL hard-deletes non-existant items anyway at its next scheduled run.
- Should purge the listing page, too... Probably `purge_content_on_save` is what we want here.

frontends/main/src/app-pages/WebsiteContent/WebsiteContentDraftListingPage.tsx
```diff
@@ -112,30 +116,48 @@ const DraftItem: React.FC<{ contentItem: WebsiteContent; type: string }> = ({
...
+        reaches this list.
+      */}
+      {!contentItem.is_published && (
+        <Card.Actions>
+          <StyledActionButton
+            variant="text"
+            size="small"
+            aria-label={`Delete ${contentItem.title}`}
+            onClick={() => showDeleteWebsiteContentDialog(contentItem)}
+          >
```
ChristopherChudzicki: My 2c is to skip the StyledActionButton. Do it the way we do it everywhere: a raw ActionButton in Card.Actions slot, variant secondary

The current approach looks weird on hover:

Image

The approach we use elsewhere would look like this (similar to resource cards):

Image

website_content/admin.py
```diff
@@ -1,23 +1,38 @@
...
+    Soft-deleted rows stay visible here (struck through) so staff can undelete
+    them, unlike the API where the default manager hides them.
+    """
+
+    # highlight_deleted strikes through this field for soft-deleted rows.
+    field_to_highlight = "title"
 
     list_display = (
-        "title",
+        highlight_deleted,
```
ChristopherChudzicki: Apparently `highlight_deleted` from  `django-safedelete`:
1. uses `WebsiteContent.__str__` for to generate the column values
2. and labels the column "name"
3. isn't sortable

Since our `WebsiteContent.__str__` is the default implmentation, it looks bad (whereas before we got real titles):

Image

**Request:** Fix (1)... Implement `WebsiteContent.__str__` as returning the title. But I'd actually suggest doing slightly more, because LLMs make it easy.... Claude suggested using `highlight_deleted_field` instead of `highlight_deleted`. It's a method that allows customizing the column title and ordering. Implementation below, but then we get

Image

ChristopherChudzicki (changes requested):
Thanks for switching to `django-safedelete`! The implementation looks good in general. A few things I think we should improve before merge:

As I thought about what suggestions to make (and took screenshots) Claude was editing the code. Its edits look good to me, so I thought I'd push them up in case you want to use them.

**Suggestion:** if my changes look good to you, one option is to merge them in — the branch is `ahtesham/delete-feature-branch-cc-suggestions`, four commits on top of yours ([diff](https://github.com/mitodl/mit-learn/compare/ahtesham/delete-feature-branch...ahtesham/delete-feature-branch-cc-suggestions)). Take all of it, cherry-pick, or ignore it — whatever's easiest.

1. **Suggested: UI Tweak:** Card icon change, 
    - comment see https://github.com/mitodl/mit-learn/pull/3681/changes#r3691221229
    - Suggested code change [`a21392d`](https://github.com/mitodl/mit-learn/commit/a21392d4c7d2752aa1ef74a7cfc9401c0dc4be9b)
2. **Request: Admin Listing: Titles, sortable**: 
    - Comment https://github.com/mitodl/mit-learn/pull/3681/changes#r3691313268
    - Suggested code change [`ba930e6`](https://github.com/mitodl/mit-learn/commit/ba930e6a7fdc468165727c02626d933a44bfbf29)
3. **Strong Suggestion:** Let's make the unique field on slug ONLY apply to undeleted content.
    - Suggested code change [`4800f2e`](https://github.com/mitodl/mit-learn/commit/4800f2e0b9462f84103f8a14ae5a1ed524ee40ad). Worth taking before (3) if you split them up — once a deleted row stops reserving its slug, undelete has to re-derive it, which is what (3) does.
4. **Request: Make undelete work (or forbid it):** Saving an undeleted item in Django admin is currently a 500 error ... Two reasonable options would be (A) make it work (via django admin), or (B) forbid it entirely. 
    - The fundamental issue was `save()` using `objects` instead of `all_objects`
    - Suggested code change [`0d7fbbf`](https://github.com/mitodl/mit-learn/commit/0d7fbbfb3d3a9952fb9ab7df69ce1171e8d94f82), which does a bit of both. Undelete works, but not as a side effect of saving the change form — which is what a bare `save()` does today, with nothing on the form to say so. Published items are read-only there and have to go through "Undelete selected"; drafts stay editable, with a warning above Save.
    - :warning: This probably deserves a couple tests, which my commit did not implement

frontends/main/src/app-pages/WebsiteContent/WebsiteContentDraftListingPage.tsx
ahtesham-quraish: Addressed

frontends/main/src/app-pages/WebsiteContent/WebsiteContentDraftListingPage.tsx
```diff
@@ -38,19 +45,32 @@ const PageWrapper = styled.div`
...
+  // Card's internal wrapper around the footer and actions has no className of
+  // its own, so target it via the footer it always renders. Anchoring on the
+  // footer rather than the actions matters: Card omits the actions element
+  // entirely when it has no children, which is the case for published items.
+  // Keep the descendant form -- jsdom's selector engine rejects a leading
+  // combinator inside :has(), so `:has(> .MitCard-footer)` breaks the tests.
+  "> div:has(.MitCard-footer)": {
+    alignItems: "center",
+    paddingTop: "16px",
+  },
```
ChristopherChudzicki, line 68: Here and in `StyledActionButton`, I don't really think there's a good reason to deviate from how the rest of the site styles the action buttons on cards.

It's not a big deal, but I'd just go with regular secondary buttons (not black) and leave the padding the same as elsewhere.  It's less about what looks better (people could disagree) and more about consistency / use what's built.

**Suggestion:** Just match the rest of the site.
- No padding overrides
- No ActionButton overrides, just `variant="secondary"` `edge=circular`

ChristopherChudzicki (approved):
👍 Changes look good! One minor [comment](https://github.com/mitodl/mit-learn/pull/3681/changes#r3705462483) on style consistency, take it or leave it.

PR #3682, opened 2026-07-27
Fix topic page pagination

shanbady:

Closes https://github.com/mitodl/hq/issues/12397

This PR fixes an issue where the topic pages show more pages than can be paginated to. This is due to the use of exact=False for the final count returned by the api which ends up showing more results than there actually are.

### How can this be tested?
1. checkout main - make sure you have embeddings locally
2. go to a top-level topic page such as http://open.odl.local:8062/c/topic/engineering -> click on the very last page of results and it should come up empty.
3. checkout this branch and attempt the same - it should never be empty and the page count should be accurate

### Extra notes
the `exact=True` on the counts is only set on the learning resources collection since it will cause performance issues on the contentfiles collection which has north of 12m points on production

### DIFF
--- a/vector_search/views_test.py
+++ b/vector_search/views_test.py
@@ -1214,3 +1214,22 @@ def test_content_file_vector_search_partial_invalid_ids_searches_survivors(
     id_conditions = [c for c in must if getattr(c, "key", None) == "edx_module_id"]
     assert len(id_conditions) == 1
     assert list(id_conditions[0].match.any) == [valid_id]
+
+
+@pytest.mark.django_db(transaction=True)
+def test_vector_search_count_is_exact(client, mock_qdrant):
+    """The result total must be an exact count.
+
+    Qdrant's approximate count overestimates filtered collections, which made
+    the paginator advertise pages that returned no results.
+    """
+    client.get(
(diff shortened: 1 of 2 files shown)


vector_search/views_test.py
```diff
@@ -1214,3 +1214,24 @@ def test_content_file_vector_search_partial_invalid_ids_searches_survivors(
...
+        data={
+            "topic": "Art, Design & Architecture",
+            "resource_type_group": "learning_material",
+        },
+    )
+
+    assert response.status_code == 200
+
+    mock_qdrant.count.assert_awaited()
+    assert mock_qdrant.count.await_args.kwargs["exact"] is True
```
mbertrand, line 1237: Could we also add a test that the content-files endpoint still uses `exact=False`? Keeping exact counts off the contentfiles collection is the main performance safeguard in this PR (it has 12M+ points in prod), but nothing currently enforces it — if someone later switches this back to a plain `exact=True`, every test would still pass. A copy of this test that hits the content-files search endpoint and asserts `exact` is `False` would cover it.

mbertrand (approved):
LGTM, just one suggestion on an additional test

mbertrand (approved):
👍

PR #3683, opened 2026-07-27
Defer AI library imports to reduce process startup cost

blarghmatey:
N/A

`litellm`, `langchain_litellm`, `tiktoken`, the `opendataloader_pdf`/`cv2` (OpenCV) stack, and `langchain_text_splitters` were being imported at module scope in `learning_resources/tasks.py`, `vector_search/tasks.py`, `learning_resources/etl/utils.py`, and `vector_search/utils.py`.

Those modules sit on the import chain for ordinary web request handling:

- `learning_resources/tasks.py` is imported by `learning_resources/views.py`
- `vector_search/utils.py` is imported by `learning_resources_search/api.py`, `learning_resources_search/indexing_api.py`, and `vector_search/views.py`
- `learning_resources/etl/utils.py` is imported by `learning_resources/tasks.py` (and every ETL pipeline module)

Because Python imports are eager, every Django process (web/ASGI workers, every Celery worker, every management command) paid the full memory and startup-time cost of loading these libraries at process boot — even though they're only actually used by a handful of specific Celery tasks (content summarization, PDF-to-markdown OCR conversion, embedding-time text chunking) that the majority of code-paths never touch.

This PR moves each of those imports from module scope down to the single call site that uses them, so the import only happens when the relevant task/function actually runs. Log capture for these libraries is unaffected — Python's root logger configuration applies regardless of when a module (and its logger) is created, so litellm/langchain output is still captured once that code path is active.

N/A - no UI changes

### How can this be tested?
- `ruff check --select F401,F821` passes on all touched files (no unused imports, no unresolved names introduced by the deferred imports).
- Ran the full affected test suite via `docker compose run --rm web uv run pytest learning_resources/tasks_test.py learning_resources/content_summarizer_test.py learning_resources/converters/opendataloader_llm_converter_test.py vector_search/tasks_test.py vector_search/utils_test.py learning_resources/etl/utils_test.py` — all 545 tests pass.
- Two tests in `vector_search/utils_test.py` that previously patched `vector_search.utils.RecursiveCharacterTextSplitter` (a module attribute that no longer exists once the import is lazy) were updated to patch `langchain_text_splitters.RecursiveCharacterTextSplitter` directly, and one test in `learning_resources/tasks_test.py` was updated to patch `learning_resources.content_summarizer.ContentSummarizer` (the defining module) instead of `learning_resources.tasks.ContentSummarizer`.
- A reviewer can confirm the startup-cost reduction by checking that `sys.modules` does not contain `litellm`/`cv2`/`opendataloader_pdf` immediately after `django.setup()`, prior to any request being served.

### Additional Context
No functional behavior changes — this is purely an import-timing change. The libraries still fully load and function normally the first time their owning task/function runs; they're just no longer paid for by every process that never touches them.

### DIFF
--- a/learning_resources/tasks.py
+++ b/learning_resources/tasks.py
@@ -15,7 +15,6 @@ from django.db.models import Q
 from django.utils import timezone
 
 from learning_resources.constants import LearningResourceType
-from learning_resources.content_summarizer import ContentSummarizer
 from learning_resources.etl import ovs, pipelines, youtube
 from learning_resources.etl.canvas import (
     sync_canvas_archive,
@@ -549,6 +548,8 @@ def summarize_content_files_task(
     Returns:
         - None
     """
+    from learning_resources.content_summarizer import ContentSummarizer
+
     summarizer = ContentSummarizer()
     return summarizer.summarize_content_files_by_ids(content_file_ids, overwrite)
 
--- a/main/sentry.py
+++ b/main/sentry.py
@@ -4,8 +4,10 @@ import logging
 
 import sentry_sdk
 from celery.exceptions import WorkerLostError
+from sentry_sdk.integrations.boto3 import Boto3Integration
 from sentry_sdk.integrations.celery import CeleryIntegration
 from sentry_sdk.integrations.django import DjangoIntegration
+from sentry_sdk.integrations.httpx import HttpxIntegration
 from sentry_sdk.integrations.logging import LoggingIntegration
 from sentry_sdk.integrations.redis import RedisIntegration
 
@@ -74,17 +76,22 @@ def init_sentry(  # noqa: PLR0913
         before_send=before_send,
         traces_sample_rate=traces_sample_rate,
         profiles_sample_rate=profiles_sample_rate,
-        # Sentry's auto-enabling integrations (langchain, boto3, openai, etc.)
-        # import their target library at init time just to check whether it's
+        # Sentry's auto-enabling integrations (langchain, openai, etc.) import
+        # their target library at init time just to check whether it's
         # installed, regardless of whether it's ever used -- e.g. langchain is
         # installed as a litellm transitive dependency but never used directly,
         # and importing its Sentry integration alone pulls in the whole
         # langchain/tiktoken stack on every process. Opt in explicitly instead.
+        # Boto3/Redis/Httpx are included because boto3, redis, and httpx are
+        # already imported elsewhere at boot regardless of Sentry, so these
+        # add no import cost of their own.
         auto_enabling_integrations=False,
         integrations=[
             DjangoIntegration(),
             CeleryIntegration(),
             LoggingIntegration(level=log_level),
             RedisIntegration(),
+            Boto3Integration(),
+            HttpxIntegration(),
         ],
     )
--- a/main/sentry.py
+++ b/main/sentry.py
@@ -7,6 +7,7 @@ from celery.exceptions import WorkerLostError
 from sentry_sdk.integrations.celery import CeleryIntegration
 from sentry_sdk.integrations.django import DjangoIntegration
 from sentry_sdk.integrations.logging import LoggingIntegration
+from sentry_sdk.integrations.redis import RedisIntegration
 
 # these errors occur when a shutdown is happening (usually caused by a SIGTERM)
 SHUTDOWN_ERRORS = (WorkerLostError, SystemExit)
@@ -73,9 +74,17 @@ def init_sentry(  # noqa: PLR0913
         before_send=before_send,
         traces_sample_rate=traces_sample_rate,
         profiles_sample_rate=profiles_sample_rate,
+        # Sentry's auto-enabling integrations (langchain, boto3, openai, etc.)
+        # import their target library at init time just to check whether it's
+        # installed, regardless of whether it's ever used -- e.g. langchain is
+        # installed as a litellm transitive dependency but never used directly,
+        # and importing its Sentry integration alone pulls in the whole
+        # langchain/tiktoken stack on every process. Opt in explicitly instead.
+        auto_enabling_integrations=False,
         integrations=[
             DjangoIntegration(),
             CeleryIntegration(),
             LoggingIntegration(level=log_level),
+            RedisIntegration(),
         ],
     )
--- a/vector_search/tasks.py
+++ b/vector_search/tasks.py
@@ -10,7 +10,6 @@ from django.conf import settings
 from django.core.cache import caches
 from django.db.models import Q
 
-from learning_resources.content_summarizer import ContentSummarizer
 from learning_resources.models import (
     ContentFile,
     Course,
@@ -708,6 +707,8 @@ def _missing_summaries():
         # list the same as None (no restriction), so short-circuit here instead
         # of letting it scan every learning resource.
         return []
+
+    from learning_resources.content_summarizer import ContentSummarizer
 
     summarizer = ContentSummarizer()
     return summarizer.get_unprocessed_content_file_ids(
--- a/vector_search/utils.py
+++ b/vector_search/utils.py
@@ -8,16 +8,11 @@ from asgiref.sync import sync_to_async
 from django.conf import settings
 from django.db import close_old_connections
 from django.db.models import Prefetch, Q
-from langchain_text_splitters import (
-    MarkdownHeaderTextSplitter,
-    RecursiveCharacterTextSplitter,
-)
 from qdrant_client import AsyncQdrantClient, QdrantClient, models
 
 from learning_resources.constants import (
     PROGRAM_COURSE_CACHE_KEY_TEST_MODE,
 )
-from learning_resources.content_summarizer import ContentSummarizer
 from learning_resources.models import (
     ContentFile,
     LearningResource,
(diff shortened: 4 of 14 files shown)


vector_search/tasks.py
```diff
@@ -632,6 +631,8 @@ def embeddings_healthcheck():
 
 
 def _missing_summaries():
+    from learning_resources.content_summarizer import ContentSummarizer
+
     summarizer = ContentSummarizer()
     return summarizer.get_unprocessed_content_file_ids(
         LearningResource.objects.filter(require_summaries=True)
```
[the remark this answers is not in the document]
blarghmatey: Verified — this is a real, pre-existing bug (introduced in 3ba27bdc4e, 2025-11-07), unrelated to this PR's changes (which only defer the ContentSummarizer import a few lines above). Splitting it out and fixing it with regression tests in a separate PR: https://github.com/mitodl/mit-learn/pull/3686

blarghmatey:
Addressed review feedback: Sentry's bug-prediction finding in `_missing_summaries()` was verified as a real, pre-existing bug unrelated to this PR's import-timing changes — split out and fixed with regression tests in #3686. No other review comments outstanding. All checks (pytest, lint, CodeQL, GitGuardian, pre-commit.ci) are green.

learning_resources/tasks.py
```diff
@@ -549,6 +548,8 @@ def summarize_content_files_task(
     Returns:
         - None
     """
+    from learning_resources.content_summarizer import ContentSummarizer
```
mbertrand, line 551: There is a subtle side effect from this: `content_summarizer.py` sets `litellm.drop_params = True` at module scope. Before this change, that ran at boot in every process. Now it only runs once a summarizer is first used, so other litellm callers (the OCR converter's `batch_completion`, the embedding encoder) may run with `drop_params` unset, depending on which task happened to run first in that worker.

I checked both of those call sites and neither passes params that would be dropped, so nothing breaks today. .The lazy-safe fixes are to set `drop_params = True` in each module that imports litellm or to pass `drop_params=True` per call if it is needed.

vector_search/utils.py
mbertrand, line 1: `litellm` and `tiktoken` still load at boot in every web and celery process: this module imports `vector_search.encoders.utils` at the top, and that module imports `tiktoken` and `litellm.get_model_info` at module scope. Both are only used inside `truncate_to_model_limit`, so the same deferral pattern used in this PR applies — a two-line move, not a refactor (the encoder classes themselves are already lazy via `import_string`). `learning_resources/utils.py` also has a module-scope `import tiktoken` worth moving at the same time.

The `sys.modules` check suggested in the PR description doesn't catch this because it stops at `django.setup()`, but a web worker also imports the URLconf at boot — still before any request is served — and litellm loads there. Verified in the web container on this branch:

```
after django.setup():  ['tiktoken', 'langchain_text_splitters']
after URLconf import:  ['litellm', 'tiktoken', 'langchain_text_splitters']
```

This is the biggest remaining piece of the startup cost (litellm is roughly 4s and ~180MB per process).

vector_search/utils.py
```diff
@@ -393,6 +388,8 @@ def embed_topics():
 
 @cache
 def _get_text_splitter(**kwargs):
+    from langchain_text_splitters import RecursiveCharacterTextSplitter
```
mbertrand, line 391: Deferring these is correct, but it won't remove `langchain_text_splitters` from startup on its own: `learning_resources/serializers.py` still imports `RecursiveJsonSplitter` at module scope, and serializers load during `django.setup()` — the package shows up in `sys.modules` right after setup on this branch (see the probe output in the file-level comment). That import is only used in one method, so the same one-line move there would actually get the package (and the langchain_core/pydantic chain it pulls in) out of startup.

learning_resources/tasks.py
blarghmatey: Fixed in 674f78246: set `litellm.drop_params = True` in the other two modules that actually call litellm with params (`learning_resources/converters/opendataloader_llm_converter.py` for `batch_completion`, `vector_search/encoders/litellm.py` for `embedding`), mirroring `content_summarizer.py`. Since its just setting an attribute on the shared `litellm` module, doing it in each caller is idempotent and no longer depends on which lazy import happens to run first in a given worker. The `get_model_info` call sites (`vector_search/encoders/qdrant_cloud.py`, `vector_search/encoders/utils.py`) dont send provider params, so they dont need it.

vector_search/utils.py
blarghmatey: Fixed in 674f78246: moved the `tiktoken` and `litellm.get_model_info` imports in `vector_search/encoders/utils.py` into `truncate_to_model_limit`, and the module-scope `import tiktoken` in `learning_resources/utils.py` into `truncate_to_tokens`.

While verifying with the same sys.modules probe, found that `litellm` is now fully gone from boot (confirmed empty), but `tiktoken`/`langchain_text_splitters` were *still* loading at `django.setup()` -- turned out to be unrelated to any of our own import sites: sentry_sdk auto-detects langchain_core (a transitive dep of litellm/langchain-litellm, never used directly) and its Sentry integration module imports it unconditionally at module scope just to register the integration. Fixed in 5cade087a by disabling sentry_sdks auto-enabling integrations (`auto_enabling_integrations=False` in `main/sentry.py`) and re-adding `RedisIntegration()` explicitly, since redis is already loaded at boot regardless and its the only one of that auto-enabling list this app actually uses. Verified `tiktoken`/`litellm`/`langchain_core` are all absent from `sys.modules` after `django.setup()` and URLconf import now.

vector_search/utils.py
blarghmatey: Fixed in 674f78246: moved the `RecursiveJsonSplitter` import in `learning_resources/serializers.py` from module scope into `render_chunks`, the only method that uses it. Combined with the sentry_sdk fix in 5cade087a (see the reply on the sibling thread), `langchain_text_splitters` no longer loads at boot at all.

blarghmatey:
Addressed all three outstanding review threads from @mbertrand:
- `litellm.drop_params` ordering: set in each module that actually calls litellm with params, not just `content_summarizer.py` (674f78246)
- `vector_search/encoders/utils.py` (tiktoken/litellm.get_model_info) and `learning_resources/utils.py` (tiktoken) module-scope imports moved into the functions that use them (674f78246)
- `learning_resources/serializers.py`'s `RecursiveJsonSplitter` import moved into `render_chunks` (674f78246)

While verifying the last two with the same sys.modules probe from the review, found that tiktoken/langchain_text_splitters were still loading at django.setup() after those fixes -- via sentry_sdk's auto-enabling integrations, which import langchain_core unconditionally just to check whether the langchain integration applies. Confirmed this predates this PR and is unrelated to any of our own import sites. Fixed by disabling sentry_sdk's auto-enabling integrations and re-adding Redis explicitly, the only one of that list this app uses (5cade087a).

Verified via the web container: tiktoken, litellm, and langchain_core/langchain_text_splitters are now all absent from sys.modules after django.setup() and URLconf import.

main/sentry.py
```diff
@@ -73,9 +74,17 @@ def init_sentry(  # noqa: PLR0913
...
+        # installed, regardless of whether it's ever used -- e.g. langchain is
+        # installed as a litellm transitive dependency but never used directly,
+        # and importing its Sentry integration alone pulls in the whole
+        # langchain/tiktoken stack on every process. Opt in explicitly instead.
+        auto_enabling_integrations=False,
         integrations=[
             DjangoIntegration(),
             CeleryIntegration(),
             LoggingIntegration(level=log_level),
+            RedisIntegration(),
```
mbertrand, line 93: Maybe add `Boto3Integration ` and `HttpxIntegration` here too, if they don't add much memory usage?

Might want to double-check with @shanbady about whether or not sentry's `LangchainIntegration` ever proved useful here for debugging problems.

mbertrand (approved):
LGTM overall, just one question about which sentry integrations to include.

ChristopherChudzicki (changes requested):
@blarghmatey This is only a request. Feel free to reject it!

Before we merge this, I think we should have a test to prevent regressions. It seems to me very easy to fix the module-level imports now, and  have it re-introduced by re-organizing existing imports or adding a new file that does not lazy import. Indeed, I *think*  we have lazy-imported LLM related modules in the past to try and reduce memory / startup time. I might be wrong.

Could we add a test that:
1. runs in a new subprocess to avoid module caching
2. Bootstraps django
3. Serves a single, simple web request
4. Asserts that none of the offending packages have been imported

?

main/sentry.py
blarghmatey: Added both in 73d0d7772. Confirmed with the same sys.modules probe used to find the langchain issue: boto3 is already loaded during boot regardless of Sentry (via `learning_resources.urls` at URLconf-import time), and httpx is already loaded during `django.setup()` (via Django admin autodiscover) -- so `Boto3Integration()`/`HttpxIntegration()` add zero marginal import cost here.

Leaving this thread open on the `LangchainIntegration` usefulness question. @shanbady do you have any thoughts/input there?

blarghmatey:
@ChristopherChudzicki good call, added in 73d0d7772: `main/boot_imports_test.py`. It spawns a fresh subprocess (so nothing's cached from other tests), boots Django, serves a real request against `/health/liveness/` via `django.test.Client`, and asserts none of the packages this PR deferred (tiktoken, litellm, langchain_core/classic/text_splitters, opendataloader_pdf, cv2, pdf2image) show up in `sys.modules` afterward.

Verified it actually catches a regression, not just passes trivially: reverted the `auto_enabling_integrations=False` sentry fix and separately reverted the `RecursiveJsonSplitter` deferral in `learning_resources/serializers.py`, and confirmed the test fails on each before restoring both.

ChristopherChudzicki (approved):
👍

main/sentry.py
shanbady: no i dont think i've used that (or if it would be useful here)

PR #3684, opened 2026-07-27
Followup to #3646: skip contentless files in embed pre-pass, retry Qdrant blips, purge before embed

mbertrand:

Followup to #3646 (mitodl/hq#12454). Three enhancements to the change-detection pre-pass.

**1. Content-less files are excluded from the pre-pass.** A published `ContentFile`
with null/empty `content` never produces a Qdrant point, so the pre-pass counted it as
"missing" and re-flagged it on *every* load. Those files now never enter the candidate
list, so an unchanged run settles at a true `0 of N`.

**2. Transient Qdrant errors during the pre-pass retry instead of deferring the run.**
`embed_run_content_files` gets `max_retries=3`, and a `DEADLINE_EXCEEDED` / `UNAVAILABLE`
from the batched payload fetch retries with the existing `_retry_countdown` backoff.
Previously a single blip failed the task and pushed the whole run's embedding to the
next load.

**3. Purge runs before embed in the indexing chain.** `remove_unpublished_run_content_files`
is now queued ahead of `embed_run_content_files` in the `content_files_loaded` hook. The
two tasks touch disjoint sets (`published=False` vs `published=True`), so ordering is
free — but with purge first, unpublished files' points are removed even when the embed
step fails.

### How can this be tested?

Verified locally against OCW **15.S12 Blockchain and Money, Fall 2018**
(`15.S12+fall_2018`, run `eca120b8a78ad938e60920323e127f21`) — 58 published content
files, 3 of which are content-less. Get its run pk:

```python
run = LearningResourceRun.objects.get(run_id="eca120b8a78ad938e60920323e127f21")
print(run.id, run.content_files.filter(published=True).count())
```

If that run isn't in your DB, any run with a mix of content-bearing and content-less
published files works — find one with:

```python
from django.db.models import Q, Count
ContentFile.objects.filter(published=True).values("run_id").annotate(
    empty=Count("id", filter=Q(content__isnull=True) | Q(content="")),
    total=Count("id"),
).filter(empty__gt=0, total__gt=5).order_by("total")[:5]
```

**Restart the celery worker first** — prefork workers don't reload changed code, so a
worker started before checkout will silently run the old task and log nothing:

```bash
docker compose restart celery
```

**Content-less files excluded (fix 1)**

```python
from vector_search.tasks import embed_run_content_files
embed_run_content_files.delay(RUN_PK)
```

```bash
docker compose logs celery --since 5m | grep 'need embedding'
```

Queue it a second time once the first pass settles. Expect `0 of 55` — i.e. `N` equals
published-with-content, not total published:

```
embed_run_content_files run 4232: 0 of 55 files need embedding
```

On `mb/embed-change-detection` the same run reports `3 of 58` on every single load,
forever. The three files genuinely have no Qdrant point, which you can confirm directly:

```python
from django.db.models import Q
from learning_resources.models import ContentFile, LearningResourceRun
from vector_search.tasks import _stored_content_payloads
from vector_search.utils import vector_point_id, vector_point_key

run = LearningResourceRun.objects.select_related("learning_resource__platform").get(id=RUN_PK)
res = run.learning_resource

def pid(key):
    return vector_point_id(vector_point_key(
        {"platform": {"code": res.platform.code}, "resource_readable_id": res.readable_id,
         "run_readable_id": run.run_id, "key": key},
        chunk_number=0, document_type="content_file"))

empty = ContentFile.objects.filter(run=run, published=True).filter(
    Q(content__isnull=True) | Q(content="")).values_list("id", "key")
stored = _stored_content_payloads([pid(k) for _, k in empty], fields=("checksum",))
for cid, k in empty:
    print(cid, k, stored.get(pid(k)))   # all None
```

**Qdrant blip retries (fix 2)**

```bash
docker compose stop qdrant
```

Queue the task and watch the log — it should retry rather than fail terminally:

```
embed_run_content_files[...] retry: Retry in 64s: <_InactiveRpcError ...
        status = StatusCode.UNAVAILABLE
```

Bring Qdrant back inside that backoff window and the retry completes the pre-pass:

```bash
docker compose start qdrant
```

```
embed_run_content_files run 4232: 0 of 55 files need embedding
Task vector_search.tasks.embed_run_content_files[...] succeeded in 18.2s
```

Before this change the `UNAVAILABLE` was terminal and the run's embedding waited for
the next load.

**Purge before embed (fix 3)**

Unpublish one already-embedded file and fire the hook:

```python
from learning_resources.models import ContentFile, LearningResourceRun
from learning_resources.utils import content_files_loaded_actions

run = LearningResourceRun.objects.get(id=RUN_PK)
cf = ContentFile.objects.filter(run=run, published=True).exclude(
    content="").exclude(content__isnull=True).first()
print(cf.id, cf.key)
ContentFile.objects.filter(id=cf.id).update(published=False)
content_files_loaded_actions(run)
```

```bash
docker compose logs celery --since 5m | grep -E 'remove_unpublished|remove_embeddings|embed_run_content_files|need embedding'
```

Expected order — the purge fully completes before embedding starts:

```
remove_unpublished_run_content_files[...] received
remove_embeddings[...] received
remove_embeddings[...] succeeded
embed_run_content_files[...] received
embed_run_content_files run 4232: 0 of 54 files need embedding
```

Then confirm the unpublished file's point is gone (re-use the `pid()` helper above —
`_stored_content_payloads([pid(cf.key)], fields=("checksum",))` returns `{}`). It should
be absent regardless of how the embed step ends.

Republish it and fire the hook again to restore local state; the pre-pass picks it back
up as missing, which also shows the self-heal path:

```
embed_run_content_files run 4232: 1 of 55 files need embedding
```

### DIFF
--- a/vector_search/tasks.py
+++ b/vector_search/tasks.py
@@ -480,7 +480,8 @@ def embed_run_content_files(self, run_id):
     as missing/stale points, so they self-heal on the next load.
 
     Content-less files are excluded: they never produce Qdrant points, so they
-    would otherwise be re-flagged on every load. Transient Qdrant errors during
+    would otherwise be re-flagged on every load. Any point left over from when
+    such a file still had content is removed. Transient Qdrant errors during
     the pre-pass retry with backoff so a blip doesn't defer the run's embedding
     to the next load.
     """
@@ -509,17 +510,24 @@ def embed_run_content_files(self, run_id):
             )
         )
 
+    contentless = Q(content__isnull=True) | Q(content="")
     pid_rows = [
         (cf_id, chunk0_point_id(key), checksum, meta)
         for cf_id, key, checksum, *meta in ContentFile.objects.filter(
             run=run, published=True
         )
-        .exclude(Q(content__isnull=True) | Q(content=""))
+        .exclude(contentless)
         .values_list("id", "key", "checksum", *CONTENT_FILE_PREPASS_PAYLOAD_FIELDS)
+    ]
+    contentless_rows = [
+        (cf_id, chunk0_point_id(key))
+        for cf_id, key in ContentFile.objects.filter(
+            contentless, run=run, published=True
+        ).values_list("id", "key")
     ]
     try:
         stored = _stored_content_payloads(
-            [pid for _, pid, _, _ in pid_rows],
+            [pid for _, pid, _, _ in pid_rows] + [pid for _, pid in contentless_rows],
             fields=("checksum", *CONTENT_FILE_PREPASS_PAYLOAD_FIELDS),
         )
     except grpc.RpcError as err:
@@ -544,12 +552,20 @@ def embed_run_content_files(self, run_id):
         for cf_id, pid, checksum, meta in pid_rows
         if is_stale(pid, checksum, meta)
     ]
+    # A stored point for a now-contentless file is a leftover from when the
+    # file had content — remove it. Inline rather than a chained task: leftovers
+    # are rare and few, and a failed delete self-heals on the next load.
+    leftover_ids = [cf_id for cf_id, pid in contentless_rows if pid in stored]
     log.info(
-        "embed_run_content_files run %s: %d of %d files need embedding",
+        "embed_run_content_files run %s: %d of %d files need embedding, "
+        "%d leftover contentless points to remove",
         run_id,
         len(ids),
         len(pid_rows),
+        len(leftover_ids),
     )
+    if leftover_ids:
+        remove_qdrant_records(leftover_ids, CONTENT_FILE_TYPE)
     if not ids:
         return None
     return _replace_with_finalized_chain(self, ids, overwrite=True)
--- a/vector_search/tasks.py
+++ b/vector_search/tasks.py
@@ -465,7 +465,7 @@ def embed_new_content_files(self):
     )
 
 
-@app.task(bind=True)
+@app.task(bind=True, max_retries=3)
 def embed_run_content_files(self, run_id):
     """
     Embed the run's published content files whose Qdrant points are missing or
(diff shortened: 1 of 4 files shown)


vector_search/utils_test.py
```diff
@@ -1127,43 +1115,81 @@ def test_should_generate_for_changed_content_file(mocker):
...
-    mock_qdrant = mocker.MagicMock()
-    if stored_payload is None:
-        mock_qdrant.retrieve.return_value = []
-    else:
-        mock_point = mocker.MagicMock()
-        mock_point.payload = stored_payload
-        mock_qdrant.retrieve.return_value = [mock_point]
-    mocker.patch("vector_search.utils.qdrant_client", return_value=mock_qdrant)
+)  # stored_entry is the checksum in the stored payload dict
+def test_generate_content_points_checksum_gate(mocker, stored_entry, expect_regenerate):
```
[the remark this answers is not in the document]
mbertrand: Test passes, false alarm.

shanbady (approved):
Looks good. i'm unsure if its even worth resolving here since it probably doesnt happen very often if ever but there is one edge case:

A published file whose content goes from non-empty to empty keeps its old qdrant point indefinitely

mbertrand:
@shanbady I made another commit to deal with those contentless files.

vector_search/tasks.py
```diff
@@ -509,17 +510,24 @@ def chunk0_point_id(key):
...
         (cf_id, chunk0_point_id(key), checksum, meta)
         for cf_id, key, checksum, *meta in ContentFile.objects.filter(
             run=run, published=True
         )
-        .exclude(Q(content__isnull=True) | Q(content=""))
+        .exclude(contentless)
         .values_list("id", "key", "checksum", *CONTENT_FILE_PREPASS_PAYLOAD_FIELDS)
     ]
+    contentless_rows = [
+        (cf_id, chunk0_point_id(key))
```
shanbady: super tiny nit-pick with the name _chunk0_point_id_ (it feels a bit odd). could we rename to chunk_0_point_id or first_chunk_point_id?

vector_search/tasks.py
shanbady: Also this comment could be more clear. somehting along:

"returns the qdrant point id for the first chunk of the contentfile"

shanbady (changes requested):
left some minor nitpicks

shanbady (approved):
👍 LGTM

PR #3687, opened 2026-07-28
fix: Show error message when a podcast episode fails to play

ahtesham-quraish:
https://github.com/mitodl/hq/issues/12585

As a podcast listener, I want to be told when an episode won't play, so I can understand what went wrong instead of staring at a player that silently does nothing.

Screenshot 2026-07-28 at 2 02 21 PM
image

### How can this be tested?
- First test cast, for testing you can connect your local with PROD because we have some instances on PROD which are restricted to some countries like Pakistan, India and China, for example, `/podcast/14337/podcast_episode/14347/intro-women-in-leadership-series`, in this case you would see the error message and 
- Another test case, you can also load the episode page and before playing the audio you can make your browser offline from network tab and then try to play the audio, in this case you would see the message that you are offline

### DIFF
--- a/frontends/main/src/app-pages/PodcastPage/useAudioPlayer.ts
+++ b/frontends/main/src/app-pages/PodcastPage/useAudioPlayer.ts
@@ -21,6 +21,76 @@ export const formatClockTime = (seconds: number): string => {
   return h > 0 ? `${h}:${mm}:${ss}` : `${mm}:${ss}`
 }
 
+/**
+ * User-facing messages for the ways playback can fail. Kept deliberately short:
+ * the fixed player bar renders them in place of the progress bar, where there
+ * is room for about two lines.
+ */
+export const PLAYBACK_ERROR_MESSAGES = {
+  /** The episode has no audio URL at all — nothing to fetch. */
+  noSource: "Audio isn't available for this episode.",
+  /** The device has no connection at all. */
+  offline: "You appear to be offline. Check your connection and try again.",
+  /** Transport failed part-way through (connection dropped, timeout). */
+  network:
+    "We couldn't load this episode. Check your connection and try again.",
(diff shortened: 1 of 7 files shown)


daniellefrappier18 (commented):
In both `PodcastPlayer.tsx` and `PodcastEmbedPlayer.tsx`, the `error ? <PlaybackError> : <ProgressWrapper>` swap removes whichever control had focus with no focus management.

Suggested Fix: On the error→healthy transition after a retry, move focus to a stable element (the play/pause button is the natural landing spot). E.g. capture that a retry is in flight and, in an effect when error clears, call .focus() on the play button ref.

frontends/main/src/app-pages/PodcastPage/useAudioPlayer.ts
```diff
@@ -115,17 +201,37 @@ export const useAudioPlayer = (
...
+    void startPlayback()
+  }, [hasAudioSource, startPlayback])
+
   const togglePlay = useCallback(() => {
     if (!hasAudioSource) return
     const audio = audioRef.current
     if (!audio) return
+    // After a failure the play button acts as a second retry affordance, so
+    // pressing it does the same reload the "Try again" button does.
+    if (loadError) {
```
daniellefrappier18, line 224: When "Try again" is pressed, retry() clears the error and kicks off an async load() + play(). During that window the alert vanishes and nothing conveys "retrying" to a screen reader — no aria-busy, no status text.

Suggested Fix: Set aria-busy on the trigger while isBuffering/retry is in flight, or render a brief polite status ("Retrying…") so the operation's progress is perceivable non-visually.

daniellefrappier18 (approved):
This looks good and works as described. I found two small a11y issues that should be resolved before merging but otherwise looks good. 

Image

PR #3689, opened 2026-07-28
Fix flaky test:  deterministically sort  choices

shanbady:

Closes https://github.com/mitodl/hq/issues/12593

There PR solves some non-determinism in the update_hubspot_contact_property_choices command that caused tests to fail depending on the database row order.

I also moved the test file out of the commands directory (otherwise the test file shows up as an actual management command)

### How can this be tested?
Tests should pass.

If you want to see the failure case - checkout main and then drop in the new test (from this PR in) and run ```docker compose run --rm web uv run pytest ol_hubspot/update_hubspot_contact_property_choices_test.py -p no:randomly -q --no-cov```


abeglova (approved):
👍

PR #3691, opened 2026-07-28
Strip trailing slash from channel_url to avoid 308 redirects

### LINKED ISSUE #3688: channel_url has a trailing slash, causing a 308 redirect on every sitemap/API-driven request
### Expected Behavior
`Channel.channel_url` / `ChannelQuerySet.annotate_channel_url()` should return the canonical URL for a channel, matching the route the Next.js frontend actually serves, so that following a link from the sitemap or API never triggers a redirect.

### Current Behavior
`channel_url` is built with a trailing slash:

https://github.com/mitodl/mit-learn/blob/main/channels/models.py#L163-L167
```python
@cached_property
def channel_url(self) -> str | None:
    """Return the channel url"""
    if self.published:
        return frontend_absolute_url(f"/c/{self.channel_type}/{self.name}/")
    return None
```

and identically in the queryset annotation:

https://github.com/mitodl/mit-learn/blob/main/channels/models.py#L32-L48
```python
def annotate_channel_url(self):
    """Annotate the channel for serialization"""
    return self.annotate(
        channel_url=Case(
            When(
                published=True,
                then=Concat(
                    models.Value(frontend_absolute_url("/c/")),
                    "channel_type",
                    models.Value("/"),
                    "name",
                    models.Value("/"),
                ),
            ),
            default=None,
        )
    )
```

This value is used directly as the sitemap `<loc>`:

https://github.com/mitodl/mit-learn/blob/main/frontends/main/src/app/sitemaps/channels/sitemap.ts#L48
```ts
url: channel.channel_url,
```

But the Next.js frontend's own canonical route for this page has **no** trailing slash:

https://github.com/mitodl/mit-learn/blob/main/frontends/main/src/common/urls.ts#L79-L81
```ts
export const CHANNEL_VIEW = "/c/[channelType]/[name]" as const
export const makeChannelViewPath = (channelType: string, name: string) =>
  generatePath(CHANNEL_VIEW, { channelType, name })
```

`frontends/main/next.config.js` has no `trailingSlash` override, so Next.js uses its default (`trailingSlash: false`) — meaning any request arriving with a trailing slash gets a 308 redirect to the slash-less canonical form.

Net effect: every consumer that follows a `channel_url` from the sitemap or from an API response (search engine crawlers, cached links, RSC prefetches) hits `/c/{channel_type}/{name}/` and immediately gets 308-redirected to `/c/{channel_type}/{name}`, paying a wasted redirect round-trip on every single request.

### Steps to Reproduce
1. Fetch the channels sitemap or any API response containing a channel's `channel_url` (e.g. `/c/topic/<name>/`).
2. Request that URL directly.
3. Observe an HTTP 308 response redirecting to the same path without the trailing slash.

### Possible Solution
Strip the trailing slash from both `channel_url` and `annotate_channel_url` in `channels/models.py`, matching the frontend's existing no-trailing-slash convention in `common/urls.ts` (lower-risk than changing `next.config.js`'s `trailingSlash` default, since other URL builders already assume no trailing slash).

### Additional Details
Found while investigating why the `traefik-gateway-controller` HPA in `ol-infrastructure`'s `applications-production` EKS cluster was pegged at max replicas. Grafana Loki analysis of Traefik's access logs for `next.learn.mit.edu` over a 1-hour window showed:
- 663,395 × 200, **142,857 × 308** (~17.6% of all traffic), plus small 404/500/502/499 tails
- 308s were heavily concentrated on `/c/topic/<name>/?_rsc=<hash>` requests
- Some clients repeated the *exact same* stale URL (identical `_rsc` cache-buster hash) dozens of times within minutes, each time getting redirected again, meaning the wasted-redirect cost compounds under crawl/retry traffic at real production scale

blarghmatey:
Closes #3688

`Channel.channel_url` and `ChannelQuerySet.annotate_channel_url()` built the channel URL with a trailing slash (`/c/{channel_type}/{name}/`), but the Next.js frontend's canonical route for that page has no trailing slash (`CHANNEL_VIEW` in `frontends/main/src/common/urls.ts`) and Next.js defaults to `trailingSlash: false`. Every consumer that followed a `channel_url` from the sitemap or an API response (crawlers, cached links, RSC prefetches) hit a 308 redirect to the slash-less canonical path.

This was discovered while investigating the `traefik-gateway-controller` HPA being pegged at max replicas in production — Loki analysis of Traefik access logs for `next.learn.mit.edu` over a 1-hour window showed 142,857 / 806,252 requests (~17.6%) were 308s, heavily concentrated on `/c/topic/<name>/?_rsc=<hash>`.

This PR strips the trailing slash from both `channel_url` and `annotate_channel_url` in `channels/models.py`, matching the frontend's existing no-trailing-slash convention rather than changing `next.config.js`'s `trailingSlash` default (lower risk, since other URL builders already assume no trailing slash).

N/A — backend-only change, no UI impact.

### How can this be tested?
- `channels/models_test.py::test_channel_url_for_departments` asserts the channel URL path no longer has a trailing slash.
- `channels/serializers_test.py` and `learning_resources/serializers_test.py` assertions were updated to match the slash-less URL.
- Ran the full affected test suite: `docker compose run --rm web uv run pytest channels/ learning_resources/serializers_test.py -q` — all passing (103 tests across `channels/models_test.py`, `channels/serializers_test.py`, `learning_resources/serializers_test.py`).
- Manually verify: fetch a channel's `channel_url` from the API or sitemap and confirm requesting it directly returns 200 (no 308 redirect).

### Additional Context
N/A

### DIFF
--- a/channels/models.py
+++ b/channels/models.py
@@ -40,7 +40,6 @@ class ChannelQuerySet(TimestampedModelQuerySet):
                         "channel_type",
                         models.Value("/"),
                         "name",
-                        models.Value("/"),
                     ),
                 ),
                 default=None,
@@ -163,7 +162,7 @@ class Channel(TimestampedModel):
     def channel_url(self) -> str | None:
         """Return the channel url"""
         if self.published:
-            return frontend_absolute_url(f"/c/{self.channel_type}/{self.name}/")
+            return frontend_absolute_url(f"/c/{self.channel_type}/{self.name}")
         return None
 
     @property
--- a/channels/models_test.py
+++ b/channels/models_test.py
@@ -37,7 +37,7 @@ def test_channel_url_for_departments(published, channel_type, detail_factory):
     if published:
         assert (
             urlparse(channel.channel_url).path
-            == f"/c/{channel_type.name}/{channel.name}/"
+            == f"/c/{channel_type.name}/{channel.name}"
         )
     else:
         assert channel.channel_url is None
--- a/channels/serializers_test.py
+++ b/channels/serializers_test.py
@@ -105,7 +105,7 @@ def test_serialize_channel(  # pylint: disable=too-many-arguments
         "created_on": mocker.ANY,
         "id": channel.id,
         "channel_url": frontend_absolute_url(
-            f"/c/{channel.channel_type}/{channel.name}/"
+            f"/c/{channel.channel_type}/{channel.name}"
         ),
         "lists": [
             LearningPathPreviewSerializer(channel_list.channel_list).data
--- a/learning_resources/serializers_test.py
+++ b/learning_resources/serializers_test.py
@@ -395,7 +395,7 @@ def test_learning_resource_serializer(  # noqa: PLR0913
                 "department_id": dept.department_id,
                 "name": dept.name,
                 "channel_url": frontend_absolute_url(
-                    f"/c/department/{Channel.objects.get(department_detail__department=dept).name}/",
+                    f"/c/department/{Channel.objects.get(department_detail__department=dept).name}",
                 ),
                 "school": {
                     "id": dept.school.id,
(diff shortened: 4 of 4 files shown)


rhysyngsun (approved):
LGTM

PR #3693, opened 2026-07-28
Identify posthog persons by keycloak global_id, not django pk

ChristopherChudzicki:

- For https://github.com/mitodl/hq/issues/12601

Identifies posthog persons by the Keycloak `global_id` instead of the Django user pk, in both places this app names a person: the frontend `posthog.identify()` call and the server-side `account_created` event.

Users with no `global_id` are left unidentified rather than falling back to the pk.

### How can this be tested?

**Prerequesite:** Set `NEXT_PUBLIC_POSTHOG_API_KEY` to your **personal** (not mitodl) posthog project

**Posthog Query:** Run the following SQL throughout. You can run it at `https://us.posthog.com/project/YOUR_PROJECT_I_HERE/sql`

```sql
SELECT distinct_id,
       person_id,
       count() AS events,
       countIf(event = '$identify') AS identifies,
       min(timestamp) AS first_seen,
       max(timestamp) AS last_seen
FROM events
WHERE timestamp > now() - INTERVAL 500 SECOND
GROUP BY distinct_id, person_id
ORDER BY last_seen DESC
```

1. **On `main`**, start as an anonymous user. 
2. If you want to see how the data evolves, run the SQL above.
    - you should see recent data with an anonymous, posthog-generated  uuid-valued-`distinct_id` 
3. Log in. Re-run SQL. 
    - You should see a row with a primary key ID, same `person_id` as the events from (2).
    - **The PK-valued distinct_id row should have nonzero `identifies` events.**
4. Switch to this branch
5. Log out. Navigate a bit if you want. Re-run SQL You should see new row(s). New anonymous uuid-valued-distinct_id
6. Log in as the same user from (2). Navigate a bit. Re-run SQL.
    - you should **see a new row with nonzero `identifies` events**
    - The new row should have a `distinct_id` **that is your user's `user.global_id`**

The final SQL query results should look something like this:

| branch | distinct_id                          | events | identifies | last_seen                   | person_id                            |
| - | ------------------------------------ | ------ | ---------- | --------------------------- | ------------------------------------ |
| pr_3693 | user_gui | 21     | 3          | present | person_uuid_2 |
| pr_3693 | anon_uuid_4 | 2 | 0          | past | person_uuid_2 |
| pr_3693 | anon_uuid_3 | 24     | 0          | far_past | person_uuid_2 |
| main | user_primary_key | 63     | 6          | further_past | person_uuid_1 |
| main | anon_uuid_2 | 9      | 0          | furthest_past | person_uuid_1 |

⚠️ **NOTE:** The same human user has two `person_uuid` values. One human user, two Posthog persons.

This has been deemed acceptable. If we want to fix it, we can run a backend `$merge_dangerously`. (Though it's a little tricky, since xpro has been identifying posthog users in the same project with its primary key).

### DIFF
--- a/frontends/main/src/page-components/ConfiguredPostHogProvider/ConfiguredPostHogProvider.test.tsx
+++ b/frontends/main/src/page-components/ConfiguredPostHogProvider/ConfiguredPostHogProvider.test.tsx
@@ -1,8 +1,8 @@
 import React from "react"
 
-import { render, waitFor } from "@testing-library/react"
+import { waitFor } from "@testing-library/react"
 import { PosthogIdentifier } from "./ConfiguredPostHogProvider"
-import { QueryClientProvider, QueryClient } from "@tanstack/react-query"
+import { renderWithProviders } from "@/test-utils"
 
 // mock stuff
 import { setMockResponse, urls, factories } from "api/test-utils"
@@ -27,15 +27,15 @@ const mockPosthog = jest.mocked(posthog)
 
 describe("PosthogIdentifier", () => {
   const setup = (user: Partial<User>) => {
-    const queryClient = new QueryClient()
     const userData = factories.user.user(user)
 
     setMockResponse.get(urls.userMe.get(), userData)
-    render(
-      <QueryClientProvider client={queryClient}>
-        <PosthogIdentifier />
-      </QueryClientProvider>,
-    )
+    /**
+     * No `user` option: we want the user to arrive via the mocked request, so
+     * that the effect runs on a pending-then-resolved query as it does in the
+     * app, rather than on a pre-seeded cache.
+     */
+    renderWithProviders(<PosthogIdentifier />)
     return userData
   }
   test.each([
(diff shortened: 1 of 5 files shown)


zamanafzal (approved):
LGTM 

Image

ChristopherChudzicki:
@zamanafzal Mentioning: I don't think that screenshot actually shows the Learn GUID being used because the `identifies` event count is 0.

Posthog assigns a UUID `distinct_id` when you are **anonymous**.

When we call `identify(guid)`, your events now have a DIFFERENT globally unique UUID ... it should be the one that matches your keycloak user from `api/v0/users/me`.

So you should see something like 

| distinct_id | person_id | identifies |
|-------------|-----------|------------|
| uuid_3      | uuid_1    | > 1        |
| uuid_2      | uuid_1    | 0          |

where `uuid_1` and `uuid_2` are assigned by posthog, and `uuid_3` is your keycloak GUID.

zamanafzal:
@ChristopherChudzicki Good catch — re-tested from a clean anonymous state and it's working as expected:

| distinct_id                            | person_id      | events | identifies |
|----------------------------------------|----------------|--------|------------|
| `e8b61b06-…` (Keycloak GUID)           | `eef4feba-…`   | 50     | 1          |
| `019fadf9-…` (PostHog anon)            | `eef4feba-…`   | 5      | 0          |


Screenshot 2026-07-29 at 6 16 08 PM

PR #3700, opened 2026-07-29
Add course code to learning resource embedding context

shanbady:

Closes https://github.com/mitodl/hq/issues/12603

This PR adds the course readable id to the embedding context so that users who search by course code can find the right course (when using the vector endpoint). 

### How can this be tested?
1. checkout main
2. go to the search interface /search
3. login as an admin/superuser
4. on the search interface enable "vector search"
5. attempt to find a course by searching for its course code ex: MITx+15.481x is "Adaptive Markets: Financial Market Dynamics and Human Behavior" 
6. note that the right course does not get surfaced
7. checkout this branch
8. delete the resources collection from your local [qdrant collection](http://localhost:6333/dashboard#/collections/resource_embeddings.resources)
9. restart your celery container
10. run ```python manage.py generate_embeddings --courses --skip-contentfiles```
11. attempt the search again and see it finds the appropriate course

### Additional Context
There are additional changes that turn the context snippet into a proper markdown which aids with embedding context


vector_search/utils.py
```diff
@@ -476,9 +477,23 @@ def _learning_resource_embedding_context(document):
...
+            ],
+        )
+    )
+    context = dedent(
+        f"""\
+# {document.get("title")}
+
+{description}
+
+Course code: {document.get("readable_id")}
```
abeglova: To avoid introducing a new term for the same field this should be "course number". 
Also the course number is currently the same as the readable id for mitxonline but that is the bug. That should be fixed in the pr i currently have up https://github.com/mitodl/mit-learn/pull/3703/changes.  For example for https://api.learn.mit.edu/api/v1/learning_resources/2840/ the readable id is "course-v1:MITxT+10.50.CH09x". Currently it is the same but it will be "10.50.CH09x" once my pr is merged. For ocw they are different- the readable id has semester informartion.

vector_search/utils.py
```diff
@@ -476,9 +477,23 @@ def _learning_resource_embedding_context(document):
...
+"""
     )
```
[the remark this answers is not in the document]
abeglova: This is correct. You should use the actual course numbers. For example for videos the readable_id is the youtube id which is not readable

abeglova (changes requested):

vector_search/utils.py
shanbady: fixed

vector_search/utils.py
```diff
@@ -476,10 +477,21 @@ def _learning_resource_embedding_context(document):
...
+            [
+                document.get("description"),
+                document.get("full_description"),
+            ],
+        )
     )
+
+    parts = [f"# {document.get('title')}", description]
+    if document.get("resource_type_group") == "course" and document.get("readable_id"):
+        parts.append(f"Course number: {document.get('readable_id')}")
```
abeglova, line 507: the readable_id isn't the course number
the course numbers are in the  course_numbers json field in the Course model

abeglova (changes requested):

abeglova (approved):
lgtm

PR #3704, opened 2026-07-29
fix hidden cta on product pages when the window is too short

gumaerc:
Closes https://github.com/mitodl/hq/issues/12613

Fixes the sticky summary info box on course and program product pages so its call-to-action stays reachable when the box is taller than the viewport.

Previously the box pinned at a fixed offset below the header and stayed stuck there for the full length of the page, so the main content column scrolled first and the CTA at the bottom of the box only came into view once you had scrolled nearly to the end of the page.

Now the box uses `position: sticky` with a runtime-computed `top`:

`top = min(HEADER_HEIGHT + OFFSET_FROM_HEADER, viewportHeight − boxHeight)`

- **Shorter than the viewport:** the existing offset is used — behavior is unchanged.
- **Taller than the viewport:** a negative offset is used, so the box scrolls up with the page (revealing top → bottom) and locks once its bottom edge (the CTA) reaches the viewport bottom.

Because CSS can't read an element's own height into `top`, the offset is set by a small hook (`useStickyRevealTop`) that recomputes on resize and content-size change via `ResizeObserver`.

Changes:
- `ProductPageTemplate.tsx`: `SummaryCol` keeps a CSS `top` default for SSR/pre-hydration and the common case; a ref from the new hook overrides `top` at runtime.
- `useStickyRevealTop.ts` (new): resize-only hook that computes the sticky offset.

image

### How can this be tested?
1. Open a course or program product page with a tall info box (e.g. a program with certificate + free tracks).
2. Make the browser window short enough that the info box is taller than the viewport.
3. Scroll down from the top: the info box should scroll up with the page and then stick with its CTA flush at the bottom of the viewport — reachable without scrolling to the end of the page.
4. Regression checks:
   - Short info box / tall window: box stays pinned under the header with the original offset, fully visible (no change).
   - Below the `md` breakpoint (mobile): full-width, non-sticky layout is unaffected.
   - Resize the window while stuck: the offset recomputes and the CTA stays anchored to the viewport bottom.

Verified locally across short/tall viewports and mobile width; `typecheck`, `lint-check`, and the ProductPages test suite all pass.

### DIFF
--- a/frontends/main/src/app-pages/ProductPages/ProductPageTemplate.tsx
+++ b/frontends/main/src/app-pages/ProductPages/ProductPageTemplate.tsx
@@ -26,6 +26,7 @@ import { getStayUpdatedHubspotFormId } from "@/common/config"
 import { usePostHog } from "posthog-js/react"
 import { PostHogEvents } from "@/common/constants"
 import { PlatformEnum } from "api"
+import { useStickyRevealTop } from "./useStickyRevealTop"
 
 const LearningResourceDrawer = dynamic(
   () =>
@@ -306,6 +307,7 @@ const ProductPageTemplate: React.FC<ProductPageTemplateProps> = ({
   resource,
 }) => {
   const posthog = usePostHog()
+  const summaryColRef = useStickyRevealTop(HEADER_HEIGHT + OFFSET_FROM_HEADER)
   const stayUpdatedFormId = getStayUpdatedHubspotFormId()
   const shouldShowStayUpdatedButton = Boolean(
     stayUpdatedFormId && showStayUpdated,
@@ -414,7 +416,7 @@ const ProductPageTemplate: React.FC<ProductPageTemplateProps> = ({
         </TopContainer>
       </GradientBanner>
       <BottomContainer>
-        <SummaryCol>{infoBox}</SummaryCol>
+        <SummaryCol ref={summaryColRef}>{infoBox}</SummaryCol>
         <MainCol>
           <SectionsWrapper>{children}</SectionsWrapper>
         </MainCol>
--- a/frontends/main/src/app-pages/ProductPages/useStickyRevealTop.ts
+++ b/frontends/main/src/app-pages/ProductPages/useStickyRevealTop.ts
@@ -1,0 +1,43 @@
+import { useEffect, useRef } from "react"
+
+/**
+ * Positions a `position: sticky` element so that, when it's taller than the
+ * viewport, it scrolls up with the page until its bottom edge reaches the
+ * viewport bottom and then sticks there - keeping its lower content (e.g. a
+ * CTA) reachable without an internal scrollbar or a scroll listener.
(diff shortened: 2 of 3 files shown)


frontends/main/src/app-pages/ProductPages/ProductPageTemplate.tsx
```diff
@@ -306,6 +307,7 @@ const ProductPageTemplate: React.FC<ProductPageTemplateProps> = ({
   resource,
 }) => {
   const posthog = usePostHog()
+  const summaryColRef = useStickyRevealTop(HEADER_HEIGHT + OFFSET_FROM_HEADER)
   const stayUpdatedFormId = getStayUpdatedHubspotFormId()
```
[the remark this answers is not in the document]
ahtesham-quraish: @gumaerc can you look into this please?

frontends/main/src/app-pages/ProductPages/useStickyRevealTop.ts
```diff
@@ -0,0 +1,43 @@
...
+    const el = ref.current
+    if (!el) return
+
+    const update = () => {
+      const top = Math.min(defaultOffset, window.innerHeight - el.offsetHeight)
+      el.style.top = `${top}px`
+    }
+
+    update()
+    const resizeObserver = new ResizeObserver(update)
```
ahtesham-quraish, line 39: ### Question: 

`update()` reads `window.innerHeight` but only recomputes on the element `ResizeObserver` and the `window` `"resize"` event. On mobile, when the browser URL/toolbar shows or hides, the visible viewport height changes without a reliable `window` `resize` event — that change comes through `window.visualViewport`'s `resize` event. Since this hook exists specifically to keep the CTA reachable on short viewports, that case is worth covering. Could we also listen to `visualViewport` (and use its height in the calc)?

```ts
const update = () => {
  const viewportHeight = window.visualViewport?.height ?? window.innerHeight
  const top = Math.min(defaultOffset, viewportHeight - el.offsetHeight)
  el.style.top = `${top}px`
}

update()
const resizeObserver = new ResizeObserver(update)
resizeObserver.observe(el)
window.addEventListener("resize", update)
window.visualViewport?.addEventListener("resize", update)

return () => {
  resizeObserver.disconnect()
  window.removeEventListener("resize", update)
  window.visualViewport?.removeEventListener("resize", update)
}
```

gumaerc: The sticky positioning is only applied on `[theme.breakpoints.up("md")]` so this would only ever apply to desktop. On mobile, the info box follows normal scrolling. I did implement this however, as using the viewport dimensions is just generally a better idea.

gumaerc:
@ahtesham-quraish Thanks, this is ready for another look.

ahtesham-quraish (approved):
LGTM!

PR #3707, opened 2026-07-30
Add trackFilterCourseCatalog GTM event

cp-at-mit:
NA

Adds a new `trackFilterCourseCatalog` function that fires a `filter-course-catalog` GTM event when a user applies a filter in the course catalog. Replaces the previous `trackCatalogFilter` call in SearchDisplay with this new dedicated event.

### How can this be tested?
unpause the filter-course-catalog GTM tag here: https://tagmanager.google.com/?authuser=1#/container/accounts/6351372910/containers/250259136/workspaces/14/tags.  Then preview and test the changes and verify that the tag triggers when filtering the course catalog.

### DIFF
--- a/frontends/main/src/common/analytics/gtm.test.ts
+++ b/frontends/main/src/common/analytics/gtm.test.ts
@@ -260,7 +260,10 @@ describe("trackCatalogFilter", () => {
 
 describe("trackFilterCourseCatalog", () => {
   it("pushes a filter-course-catalog event with filter name and value", () => {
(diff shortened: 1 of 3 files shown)


pdpinch:
Curious what this is even for? Is it part of a search journey of some kind?

cp-at-mit:
> Curious what this is even for? Is it part of a search journey of some kind?

It was on the spreadsheet

dsubak (approved):

PR #3709, opened 2026-07-31
anonymous checkout

gumaerc:
https://github.com/mitodl/hq/issues/12530

Lets a logged-out learner start a paid purchase without creating an account first. Behind the `anonymous-checkout` feature flag, clicking a paid enroll CTA while unauthenticated now hands off directly to the MITx Online basket/checkout instead of showing the signup popover; account creation happens later in the MITx Online flow (Review → Account → Verify → Payment).

- **Feature flag** `AnonymousCheckout` (`anonymous-checkout`), **off by default → no behavior change**. With the flag off, an unauthenticated paid (or free) click still routes to `onRequireSignup`, exactly as today.
- **Enroll hooks** (`useCourseEnrollment`, `useProgramEnrollment`): the auth guard is now `!isAuthenticated && (kind === "free" || !anonymousCheckoutEnabled) → onRequireSignup`. With the flag on, an unauthenticated **paid** click proceeds to the basket handoff; **free** enrollment still requires an account (there is no anonymous audit/free enrollment).
- **`useReplaceBasketItem`**: threads the created basket's `anonymous_id` into the `/cart/` redirect as an `anonymous_basket_id` query param. Django's session cookie for the anonymous basket is host-only and can't cross from Learn's proxied API host to MITx Online's own domain, so the id is handed off explicitly in the URL (MITx Online's `AnonymousBasketHandoffMiddleware` adopts it into the session and strips it).
- **Basket mutation hooks** (`useAddToBasket` / `useClearBasket`): skip `basketState` query invalidation when unauthenticated. `basketState` hits the checkout-payload endpoint, which creates a real order against the basket's purchaser; an anonymous basket has no purchaser, so invalidation must wait until after the anonymous→authenticated conversion.
- Adds `enroll` / `begin_checkout` analytics on the paid path.

### How can this be tested?
- Make sure you have mitxonline running and connected to Learn as described in the readme
- Make sure your mitxonline data includes a purchasable program or course
- Make sure you have the Cybersource credentials from RC loaded into your local, which should be all keys prefixed with `MITOL_PAYMENT_GATEWAY_CYBERSOURCE_` as well as `ECOMMERCE_DEFAULT_PAYMENT_GATEWAY="Cybersource"`
- Make sure you have Posthog connected and have the `anonymous-checkout` feature flag enabled
- Install a temporary branch client of the `@mitodl/mitxonline-api-axios` package with `docker compose exec yarn up @mitodl/mitxonline-api-axios@https://github.com/mitodl/mitxonline-api-clients/raw/954216380a8d7f09f9b7bfa00c09d09b28bd4cfe/src/typescript/mitxonline-api-axios/package.tgz`
- Open an incognito window and browse to the product page for the purchasable program / course you set up
- Click the Enroll in Program / Course
- Verify you are sent to the cart page
- Click Place your order
- You should be sent to Keycloak login, where you should create an account or log in to an existing one (best to test both paths)
- Once you have verified your email address, you should be sent to Cybersource for payment info
- Enter fake / test payment info and checkout
- Verify that you are sent to onboarding in Learn
- After going through onboarding, verify that you are sent to the dashboard and you see the Program / Course that you purchased

### Additional Context
- **Temporary API client pin.** `@mitodl/mitxonline-api-axios` is pinned to a branch build of the pending MITx Online release (for `anonymous_id` and the new basket types). This must be swapped for the official published version once that release ships.
- **Bundled type change.** That pending release also carries the b2b `sso_organization_id` field (MITx Online #3807), which is required on `OrganizationPage`. Adopting the client therefore also requires a small test-fixture update for the org factory/tests — coupled to the client bump, not to anonymous-checkout behavior. (Runtime already tolerates it via `orgUuid.ts`, added earlier in #3679.)
- The feature flag makes this safe to merge ahead of a coordinated launch — the reordered flow only activates when the flag is turned on.

### Checklist:
- [ ] Swap the temporary `@mitodl/mitxonline-api-axios` pin for the official release build before/at release
- [ ] Confirm the `anonymous-checkout` PostHog flag exists in each environment

### DIFF
--- a/frontends/api/src/mitxonline/hooks/baskets/index.ts
+++ b/frontends/api/src/mitxonline/hooks/baskets/index.ts
@@ -1,6 +1,7 @@
 import { basketQueries } from "./queries"
 import { useMutation, useQueryClient } from "@tanstack/react-query"
 import { basketsApi } from "../../clients"
+import { useUserIsAuthenticated } from "../../../hooks/user"
 import type { BasketWithProduct } from "@mitodl/mitxonline-api-axios/v2"
 
 /**
@@ -9,6 +10,7 @@ import type { BasketWithProduct } from "@mitodl/mitxonline-api-axios/v2"
  */
 const useAddToBasket = () => {
   const queryClient = useQueryClient()
+  const isAuthenticated = useUserIsAuthenticated()
   return useMutation({
     mutationFn: async (productId: number): Promise<BasketWithProduct> => {
       const response = await basketsApi.basketsCreateFromProductCreate({
@@ -17,10 +19,16 @@ const useAddToBasket = () => {
       return response.data
     },
     onSuccess: async () => {
-      // Invalidate checkout query to ensure fresh data
-      queryClient.invalidateQueries({
-        queryKey: basketQueries.basketState().queryKey,
-      })
+      // Invalidate checkout query to ensure fresh data - but only once
+      // authenticated. basketState hits the checkout-payload endpoint, which
+      // creates a real order against the basket's purchaser; an anonymous
+      // basket has no purchaser, so this has to wait until after the
+      // anonymous-checkout conversion step, not fire on every cart add.
+      if (isAuthenticated) {
+        queryClient.invalidateQueries({
+          queryKey: basketQueries.basketState().queryKey,
+        })
+      }
     },
   })
 }
@@ -30,14 +38,17 @@ const useAddToBasket = () => {
  */
 const useClearBasket = () => {
   const queryClient = useQueryClient()
+  const isAuthenticated = useUserIsAuthenticated()
   return useMutation({
     mutationFn: async (): Promise<void> => {
       await basketsApi.basketsClearDestroy()
     },
     onSuccess: () => {
-      queryClient.invalidateQueries({
-        queryKey: basketQueries.basketState().queryKey,
-      })
+      if (isAuthenticated) {
+        queryClient.invalidateQueries({
+          queryKey: basketQueries.basketState().queryKey,
+        })
+      }
     },
   })
 }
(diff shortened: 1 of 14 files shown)


ChristopherChudzicki (changes requested):
👍 Working well, I am going to make a few changes, primarily (1) and (2) below.

1. **Revert the auth check on the basket invalidations** — restore the unconditional `invalidateQueries` in `useAddToBasket` and `useClearBasket`.[^1]
    - Nothing should be subscribe without authentication: UI either doesn't load that, or it would be gated by `enabled`.
    - If something actually is subscribed, it should be invalidated.
    - *Aside:* Nothing actually uses that query right now at all.
    - Plus the gating as-is is on Learn session; if it's gated at all, it should be gated on MITxOnline session
3. **Test what the PR actually does** — add a test that feeds a basket carrying an `anonymous_id` through the real `useReplaceBasketItem` and asserts the redirect URL has `?anonymous_basket_id=`. Nothing covers that seam today.[^3]
3. **Share pending state across basket mutation instances** — `CoursePage.tsx:52` and `CourseEnrollArea.tsx:22` each own a separate `useReplaceBasketItem` and each disables only on its own pending state, so clicking one CTA then the other starts two clear→add chains.[^2] `CourseEnrollmentDialog.tsx:271` needs the same treatment; it's `disabled={!enabled}` today, so a plain double-click does it.
4. **Add `isError` to `useEnrollmentHandler`** — it returns `isPending` but not `isError` ([lines 175–181](https://github.com/mitodl/mit-learn/blob/ad56f7a0131a0dd4ce8766ccc212370dae962ea2/frontends/main/src/app-pages/DashboardPage/CoursewareDisplay/useEnrollmentHandler.ts#L175-L181)), and `UnenrolledCourseCard.tsx:51` consumes only `isPending`, so a failed dashboard upgrade shows the user nothing at all.

PR #3712, opened 2026-08-03
Harden GH Actions supply chain: add zizmor static analysis + 7-day dependency cool-down

blarghmatey:
N/A

Hardens GitHub Actions supply-chain security and reduces exposure to newly-introduced malicious/vulnerable dependency releases via three changes:

1. **Add a zizmor GitHub Actions static-analysis workflow** (`.github/workflows/actions-static-analysis.yml`). It runs on push/PR whenever files under `.github/workflows/**` change, checking out the repo with `persist-credentials: false` and running [zizmorcore/zizmor-action](https://github.com/zizmorcore/zizmor-action) (pinned to a specific commit SHA, tagged v0.6.2) with `min-severity: high` / `min-confidence: medium`. This is the same pattern already in use in the sibling repo mitodl/mitxonline. zizmor detects common GitHub Actions security issues such as unpinned third-party actions, overly broad `permissions:`, and injectable template expressions in `run:` steps.

2. **Add zizmor as a local pre-commit hook** (`.pre-commit-config.yaml`), using `zizmorcore/zizmor-pre-commit` at `v1.29.0`, placed alongside the other lint-style hooks (next to `shellcheck`). This surfaces the same class of workflow issues at commit time instead of only in CI.

3. **Delay dependency version resolution by 7 days** (`pyproject.toml`). Added `exclude-newer = "7d"` to the existing `[tool.uv]` table. uv 0.11+ treats a relative duration string like `"7d"` as a rolling window rather than a fixed date — it's stored in `uv.lock` as a self-updating `exclude-newer-span = "P7D"` (confirmed by inspecting the regenerated lockfile), not a one-time cutoff. This means `uv lock` will never resolve a package version that was published less than 7 days ago, giving the community time to catch and yank newly-published malicious or broken releases before this repo picks them up.

N/A — no UI changes.

### How can this be tested?
- **Workflow YAML validity**: `python3 -c "import yaml; yaml.safe_load(open('.github/workflows/actions-static-analysis.yml'))"` parses without error, as does `.pre-commit-config.yaml`.
- **zizmor pre-commit hook**: ran locally with `pre-commit run zizmor --files .github/workflows/actions-static-analysis.yml` — passed cleanly (network-enabled sandbox, so the hook environment was actually fetched and executed, not just statically reviewed).
- **uv lock regeneration**: ran `uv lock` in the repo root after adding `exclude-newer = "7d"`. It re-resolved (`Resolving despite existing lockfile due to addition of global exclude newer ...`) and succeeded — `Resolved 385 packages`. The resulting `uv.lock` now contains:
  ```
  exclude-newer = "0001-01-01T00:00:00Z" # This has no effect and is included for backwards compatibility when using relative exclude-newer values.
  exclude-newer-span = "P7D"
  ```
  confirming the rolling-window (not fixed-date) behavior. No package versions changed materially — the only lockfile diff besides the new `exclude-newer`/`exclude-newer-span` lines was a platform-marker change on the `ansicon` package (`sys_platform == 'win32'` marker dropped), which is unrelated resolver churn, not a version bump.
- Reviewers should sanity-check the new workflow file renders correctly on GitHub's Actions tab once merged, and that the zizmor CI job fires only on `.github/workflows/**` changes as scoped by the `paths:` filter.

### Additional Context
The zizmor workflow and pre-commit hook pin their action/hook versions by SHA/tag (`actions/checkout@3d3c42e...` = v7.0.1, `zizmorcore/zizmor-action@3dc1ecc...` = v0.6.2, `zizmorcore/zizmor-pre-commit` = v1.29.0) rather than a floating major-version tag, consistent with supply-chain hardening best practice for GitHub Actions.


blarghmatey:
Follow-up: added \`[tool.uv.exclude-newer-package]\` to exempt MIT ODL's own in-house packages (ol-concourse, django-aqueduct, the open-edx-plugins, and ol-django packages) from the new 7-day \`exclude-newer\` cool-down.

The 7-day delay is meant to guard against newly-published third-party malicious/vulnerable packages — it shouldn't slow down consuming our own releases, which we control and iterate on rapidly. Each exempted package gets \`"0d"\` (always resolve to latest), while the global \`exclude-newer = "7d"\` still applies to everything else.

Ran \`uv lock\` to regenerate the lockfile — it succeeded, and the only diff is the new \`[options.exclude-newer-package]\` metadata block recording the per-package overrides; no dependency versions changed.

pyproject.toml
```diff
@@ -162,6 +162,7 @@ dev = [
 package = false
 default-groups = "all"
 override-dependencies = ["setuptools<80"]
+exclude-newer = "7d"
```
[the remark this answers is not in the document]
blarghmatey: Fixed in da49128f2: added required-version = ">=0.9.17" to [tool.uv] — the exact uv release that introduced relative exclude-newer duration support (https://github.com/astral-sh/uv/releases/tag/0.9.17). An older uv now fails loudly with a clear version-mismatch error instead of silently mishandling the setting.

uv.lock
```diff
@@ -1848,7 +1899,7 @@ name = "jinxed"
 version = "1.3.0"
 source = { registry = "https://pypi.org/simple" }
 dependencies = [
-    { name = "ansicon", marker = "sys_platform == 'win32'" },
+    { name = "ansicon" },
 ]
```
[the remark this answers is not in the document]
blarghmatey: Verified — the marker drop is real (confirmed by diffing against main's uv.lock), but the CRITICAL/"will break builds" framing is incorrect: ansicon (https://pypi.org/project/ansicon/) is a pure-Python universal wheel (py2.py3-none-any) with no OS-specific code or extra requirements — it installs fine as an unused package on Linux/macOS, it will not break the build. This looks like a shift in uv's marker-propagation output during relock rather than a bug in this PR's config; tracked in follow-up issue https://github.com/mitodl/mit-learn/issues/3715 for a closer look rather than blocking this PR.

pyproject.toml
```diff
@@ -162,6 +162,55 @@ dev = [
 package = false
 default-groups = "all"
```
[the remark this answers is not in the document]
blarghmatey: Verified — this is incorrect. Not just from the changelog (uv 0.9.17's release notes list 'Add support for relative durations in exclude-newer (#16814)' as new in that release), but empirically: I installed uv 0.9.17 itself and ran `uv lock` against this exact exclude-newer = "7d" config. It resolves cleanly with no error and produces the correct exclude-newer-span = "P7D" in the lockfile. required-version = ">=0.9.17" is the correct, precise minimum — no change needed.

pyproject.toml
```diff
@@ -162,6 +162,55 @@ dev = [
...
 override-dependencies = ["setuptools<80"]
+exclude-newer = "7d"
+required-version = ">=0.9.17"
+
+[tool.uv.exclude-newer-package]
```
ChristopherChudzicki, line 168: @blarghmatey Most of these aren't used in Learn? Do you just want a global whitelist? Probably would get out of sync eventually anyway.

ChristopherChudzicki (approved):
👍

.github/workflows/actions-static-analysis.yml
```diff
@@ -0,0 +1,32 @@
+name: GitHub Actions Static Analysis
+
+on:
+  push:
+    paths:
+      - ".github/workflows/**"
+  pull_request:
+    paths:
+      - ".github/workflows/**"
```
ChristopherChudzicki, line 11: `push` + `pull` is redundant, right?

**Suggestion:** Just do `push`.

ChristopherChudzicki (approved):
Still approving, but maybe remove the `pull_request` trigger?

.github/workflows/actions-static-analysis.yml
```diff
@@ -0,0 +1,32 @@
...
+      - name: Checkout repository
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+        with:
+          persist-credentials: false
+
+      - name: Run zizmor 🌈
+        uses: zizmorcore/zizmor-action@3dc1ecc9bcb9e94e9b2c709687979e1298497054 # v0.6.2
+        with:
+          inputs: ".github/workflows/"
+          min-severity: high
```
ChristopherChudzicki: Should we lower this? (Future work, per repo, probably)

ChristopherChudzicki (approved):

blarghmatey:
Dropped the `pull_request` trigger — good catch, it was firing zizmor twice per PR commit (once from `push`, once from `pull_request`) since both were scoped to the same paths with no branch filter. `push` alone still covers PR branch commits.

pyproject.toml
blarghmatey: Trimmed in 089d5a4 — kept only the 4 in-house packages mit-learn actually locks (mitol-django-common, mitol-django-observability, mitol-django-scim, mitol-drf-lint), dropped the other ~40 (mostly Open edX plugins this repo doesn't depend on). Confirmed by intersecting the allowlist against uv.lock's locked package set, applied per-repo across all 8 repos that had this allowlist.

.github/workflows/actions-static-analysis.yml
blarghmatey: Done org-wide, not just here — lowered `min-severity` from `high` to `medium` (kept `min-confidence: medium`) across all 33 repos in this rollout, fixed the 89 findings that surfaced (mostly missing `permissions:` blocks, plus a few dynamic-secret-lookup and one archived-action case), and verified every repo comes back clean at the new gate. mit-learn itself had 10 findings (all `excessive-permissions`, all resolved with workflow-level `contents: read` in ci.yml/publish-pages.yml + `pull-requests: write` on the openapi-diff job, which posts a PR comment). `min-confidence` is still `medium` — going to `low` only turned up 2 more org-wide (both a low-confidence `cache-poisoning` finding on `actions/setup-node@v7`'s default caching, which isn't cleanly fixable via config alone) so didn't seem worth the noise. Happy to revisit if you want confidence lowered too.

PR #3713, opened 2026-08-03
serializer performance - cache/re-use child serializer

shanbady:

Closes https://github.com/mitodl/hq/issues/12610

This PR caches the concrete child serializers that LearningResourceSerializer returns which on a
100-resource page, Serializer.get_fields accounts for ~80% of serialization time
(0.442s of 0.549s cumulative, 1352 calls).

more details in the [ticket](https://github.com/mitodl/hq/issues/12610)

### How can this be tested?
1. download [this](https://github.com/user-attachments/files/30668223/bench_serializer.py) serializer benchmark script
2. checkout main and run it via `docker compose run --rm -T web python manage.py shell < bench_serializer.py`
3. note the stats - in particular the p50 ms/resource
4. checkout this branch and run the same. note the difference

### Additional Context
There are no functional changes so all tests should pass etc - this is purely a performance related tweak

### DIFF
--- a/learning_resources/serializers.py
+++ b/learning_resources/serializers.py
@@ -1582,11 +1582,29 @@ class LearningResourceSerializer(serializers.Serializer):
         )
     }
 
+    def _child_serializer(self, resource_type):
+        """
+        Return a reusable serializer instance for a resource_type.
+
+        Constructing a DRF serializer deep-copies its whole field tree, so
+        building one per object makes bulk serialization (search hydration,
+        indexing) dominated by setup rather than output. Under many=True the
+        ListSerializer reuses a single child, so this cache spans the list --
+        which is how a normal many=True serializer already behaves.
+        """
+        if not hasattr(self, "_child_serializer_cache"):
+            self._child_serializer_cache = {}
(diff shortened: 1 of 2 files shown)


learning_resources/serializers.py
```diff
@@ -1582,11 +1582,29 @@ class LearningResourceSerializer(serializers.Serializer):
         )
     }
 
+    def _child_serializer(self, resource_type):
+        """
+        Return a reusable serializer instance for a resource_type.
+
+        Constructing a DRF serializer deep-copies its whole field tree, so
+        building one per object makes bulk serialization (search hydration,
+        indexing) dominated by setup rather than output. Under many=True the
```
mbertrand: I think that indexing doesn't actually benefit from this cache — it constructs a fresh `LearningResourceSerializer` (and therefore a fresh, empty cache) for every resource.  So could either drop "indexing" from the docstring, or hoist one shared `LearningResourceSerializer()` above the loop and pass it down so bulk indexing gets the win too.

mbertrand (approved):
Works great, just one nitpick.

PR #3716, opened 2026-08-03
recreate index refactor

abeglova:
https://github.com/mitodl/hq/issues/11265

recreate_index rebuilds the OpenSearch indexes by fanning out thousands of Celery subtasks. Previously start_recreate_index replaced itself with a chord — a group of indexing tasks with finish_recreate_index as the callback that swaps the aliases. The management command blocked on task.get() until the whole thing finished.

On production Kubernetes, autoscaling routinely culls the pod running the coordinating task. When that happens the chord never fires, the aliases never switch, and the reindex silently never completes.

This PR moves all orchestration state out of Celery/Redis and into the database, so no process's lifetime has to span the reindex. Every actor — the start task, each batch, the finish step, the safety-net sweep — does one short, idempotent step against durable state and exits. Any of them can be killed at any point and the system converges.

It does this by creating TaskJob and TaskBatch models in the database that store the in progress tasks. These can be used for other jobs that fan out subtasks and must wait for completion.

### How can this be tested?
`docker compose run --rm web ./manage.py recreate_index --all ` 
this should start the job

run
`docker compose run --rm web ./manage.py recreate_index --status  `

Poll for status a few times and you should see progras

The job should finish. After it finishes 
`docker compose run --rm web ./manage.py recreate_index --status  `

should show success

Search should work normally

```
from learning_resources_search.connection import  get_conn
conn = get_conn()
conn.indices.get_alias(index="*")
```

should show that the indexes from the status output are set as default

Go to
http://api.open.odl.local:8065/admin/main/taskbatch
and
http://api.open.odl.local:8065/admin/main/taskjob

You should see a record of the job


learning_resources_search/management/commands/recreate_index.py
```diff
@@ -61,32 +111,55 @@ def handle(self, *args, **options):  # noqa: ARG002
...
+                TaskJob.objects.filter(
+                    id=active_job.id, status__in=TaskJob.ACTIVE_STATUSES
+                ).update(
+                    status=TaskJob.Status.FAILED,
+                    error=f"superseded by reindex job {job.id}",
+                )
 
-        total_seconds = (now_in_utc() - start).total_seconds()
-        self.stdout.write(f"Recreate index finished, took {total_seconds} seconds")
+        start_recreate_index.delay(job.id)
```
shanbady, line 158: in start_recreate_index the job is set to RUNNING before the run_reindex_batch.delay() loop. worker may die mid-loop , the redelivered task hits if job.status != QUEUED: return and the never-enqueued batches sit QUEUED indefinitely

shanbady, line 158: run_reindex_batch autoretries RetryError (default max_retries=3). When retries exhaust, the task fails without touching the batch row, so the batch stays RUNNING and the job never reaches FINISHING. we could consider adding an on_failure handler on the task that marks the batch FAILED and nudges _maybe_finish_reindex_job, or a periodic task to sweep for stale RUNNING batches. 

Same applies to finish_reindex_job exhausting retries while the job is FINISHING.

shanbady (changes requested):
Left a few notes.

Also two other items:
- running this locally just 1 or 2 times accumulated 16k taskbatch objects in the database. I think we need some scheduled task to purge items older than n number of days. Every content-file batch stores up to 100 ids in its params JSON, so a full reindex creates roughly total_content_files / 100 rows (plus resource and dispatch batches), each carrying ~0.5–1KB of JSON. With content files in the hundreds of thousands to millions, that's tens of thousands of rows / tens of MB per run - can easily grow to  several gb of data over a year
- I noticed the initial ticket was created to address youtube but i dont see any youtube related changes here - is that still something we need to look at?
- given how frequently our etls run and how many resources and tasks we have - there is potential for this to generate an unbounded number of rows and flood our postgres instance

learning_resources_search/management/commands/recreate_index.py
abeglova: Fixed! Now the start_recreate_index job can restart if it is stopped halfway and re-enqueued

abeglova: I added a  on_failure handler

shanbady (approved):
LGTM

PR #3717, opened 2026-08-03
receipt page

gumaerc:
Closes https://github.com/mitodl/hq/issues/12783

This PR adds the initial implementation of the receipt page in Learn, both mobile and desktop.

desktop-1440
mobile-375

### How can this be tested?
- Ensure you have set up and configured a linked instance of mitxonline as described in the readme here or by setting up the local-dev setup in ol-infrastructure
- Ensure that your user has at least one purchased course or program with a receipt
- Browse to the dashboard logged in as said user
- Find the card for the course / program you have purchased
- Click the three dots to open the context menu and click on Receipt to view the receipt page
- Ensure that everything looks as it should from the designs including desktop and mobile views
- Confirm an audit enrollment (or a verified one with no purchase behind it) shows no Receipt item in the context menu. The link is only rendered once an order resolves, so it can no longer lead to a dead page

### Additional Context
The Figma includes Tax, Total Before Tax, HSN, GSTIN and Company Name rows. The MITx Online receipt payload has no fields behind any of them, so they are omitted rather than rendered blank — the existing MITx Online receipt does not show them either, so this is not a regression. CEUs is wired up but MITx Online currently returns null for it, so that row does not render either.

### Checklist
- [ ] After https://github.com/mitodl/mitxonline/pull/3823 is deployed to production, update this branch with the latest version of `@mitodl/mitxonline-api-axios`


gumaerc:
Some upstream changes we should make in MITx Online before continuing work on the frontend:

1. **`Order.purchaser`: drop `many=True` and add a name field.** It's annotated as a list but returns an object, so the generated type is wrong; and `ExtendedLegalAddress` is only `country`/`state`/`email`. Learn ignores the field entirely and calls `users/me` just to get a name. Fix both and the receipt renders from one response.

2. **Normalize money field precision.** One receipt payload returns `total_price_paid: "149.00000"`, `total_paid: "149.00"`, `price: "149.00"`, `discount: "0.0"`, and `refunds[].amount` as a *number*. Four formats and two types for currency in a single object. Consistent 2-decimal strings (or consistent numbers) removes per-field parsing.

3. **Make `transactions` and `street_address` consistently null-or-populated.** They currently come back as `None`, or as an object with every field null, depending on whether a transaction row exists — so consumers handle three shapes for one field.

4. **Always paginate `orders/history`.** With no `PAGE_SIZE` configured, omitting `limit` bypasses pagination and returns a bare array instead of the `{count, results}` envelope the schema declares. A generated client that omits `limit` breaks.

5. **Populate `CEUs` or remove it.** Declared in the schema and hardcoded to `None` in `TransactionLineSerializer`, so it's permanently blank.

6. **Add a discriminator to `Product.purchasable_object`.** Untagged union whose three variants all expose a bare `id` from different tables; Learn currently identifies them by which optional fields are present.

jkachel:
Looked at this a bit and have no comments - seems fine to me. We discussed (Linda, Steve, me) in a meeting this morning about this too and I was using the screenshots for reference during that. For the elements that we don't have in the API result currently, omitting them from the display is a good choice. (The fields you have listed in the description are all ones we don't have _yet_. Tax support is coming at some point, and so is the migration of xPRO transactions into MITx Online. I think the other big one that was in Figma but not here was payment info; we don't expose this in a "nice" manner and it's not guaranteed to be there now, and Stripe doesn't include it by default in its checkout payloads, so unless we decide to add payment info to the receipt, it's fine to omit.)

The CEU issue is _probably_ an oversight but it's going to be empty most of the time, I think.. MITx Online does store this data but it's in the certificate, and most of them set it to nothng. Just noting for completeness, fixing that is a separate issue.

frontends/main/src/app/(site)/receipt/[orderId]/page.tsx
```diff
@@ -0,0 +1,26 @@
...
+  title: "Receipt",
+  robots: { index: false },
+})
+
+/**
+ * Fetched client-side: MITx Online's session cookie is not forwarded on
+ * server-side requests, so the order cannot be prefetched here.
+ */
+const Page: React.FC<PageProps<"/receipt/[orderId]">> = async ({ params }) => {
+  const { orderId } = await params
```
Comment (daniellefrappier18) [inline frontends/main/src/app/(site)/receipt/[orderId]/page.tsx:17]: One thing I don't think is covered yet: none of the three /receipt/* routes have an auth gate on them. I know that sounds like an edge case but I imagine that these are bookmark-prone, people save them for expense reports and reimbursement and come back days or weeks later and there's a good chance the cookie's already stale. Same if you leave a dashboard tab open overnight and hit Receipt in the morning, or log out in another tab.

The fix is small and there's already a pattern for it in `dashboard/layout.tsx ` 

A new frontends/main/src/app/(site)/receipt/layout.tsx covers all three routes, and none of the page files change.

```
import React from "react"
import RestrictedRoute from "@/components/RestrictedRoute/RestrictedRoute"
import { Permission } from "api/hooks/user"

const Layout: React.FC<{ children: React.ReactNode }> = ({ children }) => {
  return (
    <RestrictedRoute requires={Permission.Authenticated}>
      {children}
    </RestrictedRoute>
  )
}

export default Layout
```

That gets us redirectAuthToCurrent(), so the user goes through Keycloak and comes back to the receipt they originally asked for rather than dead-ending on an error page.

frontends/main/src/common/mitxonline/useOrderIdForResource.ts
```diff
@@ -0,0 +1,85 @@
...
+  })
+
+  if (resourceId === null) {
+    return { isPending: false, orderId: null }
+  }
+  if (history.isPending) {
+    return { isPending: true, orderId: null }
+  }
+  // Reported as "not found": callers render the same 404 either way.
+  if (history.isError || !history.data) {
```
daniellefrappier18, line 71: The Receipt menu item used to be a plain link built from the run id, so it was always there for verified enrollments. Now it needs an order id, which nothing on the enrollment carries, so each card fetches the last 100 orders and searches them for a matching line.

The problem is that `useOrderIdForResource` returns `orderId: null` in three pretty different cases: still loading, the request failed, and loaded fine but found nothing. But `receiptMenuItem.ts` treats all three the same and renders nothing. So if orders/history 500s, Receipt quietly disappears from every card on the dashboard with no error and no retry, and it looks identical to "you never paid for this."

I think the fix is already in this PR though. `/receipt/by-run/[runId]` does this same lookup on its own page with a real skeleton and a real 404, and receiptByRunView/receiptByProgramView currently have zero callers. If `getReceiptMenuItem` took an href instead of an order id, the cards could link straight there:

```

const receiptMenuItem = getReceiptMenuItem(
  enrollment?.enrollment_mode,
  receiptByRunView(enrollment.run.id),
)
```
That drops the hook from both cards and loses the extra dashboard request also the redirect pages stop being dead code.

One thing that's your call: the comment on `receiptMenuItem.ts `mentions verified enrollments that genuinely have no order. Today those users never see the item; this way they'd click and hit a 404. If we'd rather keep hiding it, the smaller fix is adding `isError` to `OrderIdResolution` and only hiding when the lookup actually succeeded and came back empty.

frontends/main/src/app-pages/ReceiptPage/ReceiptOrderSummary.tsx
```diff
@@ -0,0 +1,121 @@
...
+            <RowLabel>Discount</RowLabel>
+            <DiscountValue>{`- ${formatMoney(totalDiscount)}`}</DiscountValue>
+          </Row>
+        ) : null}
+        <Row>
+          <RowLabel>Quantity</RowLabel>
+          <RowValue>{`x ${totalQuantity}`}</RowValue>
+        </Row>
+      </Rows>
+      <TotalRow>
```
daniellefrappier18, line 157: This doesn't look at `order.refunds` or `order.state`, so a refunded order renders as a normal receipt with the full "Total Paid" and nothing saying the money came back. `total_price_paid` gets stamped at fulfillment and isn't reduced by a refund, so that number stays at the original amount.

The serializer returns `refunds` as `[{amount, date}]` built from the refund transactions also it returns  `state` (`refunded `/ `partially_refunded)` and `refund_eligible`, all on the order we're already fetching. And OrderReceiptView.get_queryset() only filters by purchaser with no state filter, so refunded orders definitely land on this page.

Simplest fix is probably a row per refund in this card when `order.refunds.length > 0`, showing the amount and date.

frontends/main/src/app-pages/ReceiptPage/ReceiptPage.tsx
```diff
@@ -0,0 +1,331 @@
...
+              variant="tertiary"
+              startIcon={<RiPrinterLine />}
+              onClick={() => window.print()}
+            >
+              Print
+            </Button>
+          </Actions>
+        </TitleRow>
+
+        {orderQuery.isPending ? (
```
daniellefrappier18, line 285: The `<h1>Receipt</h1>` and the Back/Print buttons render immediately, then this swaps between a skeleton and either the receipt or an error, and none of that gets announced. MUI's Skeleton sets no ARIA of its own, so a screen reader user hears "Receipt", then silence while it loads, then nothing when it either lands or fails.

Two small additions would cover it: `role="status"` with an accessible name on the skeleton wrapper (or `aria-busy` on the content container), and `role="alert"` on `ErrorState` so the failure actually gets read out. Same applies to the skeleton in ReceiptRedirect, which sits there and then navigates with no announcement either.

frontends/main/next.config.js
```diff
@@ -107,6 +107,7 @@ const nextConfig = {
     // Turbopack filesystem caching is enabled by default in Next.js 16.1+
     // Explicitly enable it for clarity (optional - already default)
     turbopackFileSystemCacheForDev: true,
+    turbopackMemoryLimit: 2 * 1024 ** 3, // bytes; triggers Turbopack GC above ~2GB
```
daniellefrappier18: Is this an intentional change?

frontends/main/src/app-pages/ReceiptPage/ReceiptPage.tsx
```diff
@@ -0,0 +1,331 @@
...
+  return (
+    <Background>
+      <PageContainer>
+        <TitleRow>
+          <PageHeading>Receipt</PageHeading>
+          <Actions>
+            <Button
+              variant="tertiary"
+              startIcon={<RiArrowLeftLine />}
+              onClick={() => router.back()}
```
daniellefrappier18: The Back button calls `router.back()`, which is just `history.back()` underneath. If the receipt is the first page in the tab, there's nothing to go back to, so the button does nothing at all. Opening bookmarked link in a fresh tab is probably the most common way someone revisits a receipt.

Suggested Fix: `ButtonLink` to `urls.DASHBOARD_HOME`, or gate on `window.history.length > 1`.

daniellefrappier18 (commented):
Overall this looks good. I worked through the test steps in the description. I logged in as a user with a purchased course and a purchased program, opened each card's context menu and clicked through to the receipt, and checked the page against the designs on desktop and mobile.

Also confirmed that audit enrollments and verified enrollments with no order behind them show no Receipt item.

I left a few comments inline, nothing that changes the overall approach.

The two I'd most want a look at are the missing auth gate on the /receipt/* routes and the way a failed orders/history request silently drops the Receipt item.

Image

Image

Image

frontends/main/next.config.js
gumaerc: Good catch, that was accidental.

daniellefrappier18 (approved):
Looks good!

PR #3719, opened 2026-08-04
added a fallback to populate run readable ids for contentfiles withou…

shanbady:
Closes https://github.com/mitodl/hq/issues/12691

This PR resolves an issue I noticed where at some point the vector endpoints stopped returning data from the marketing sites we scrape for programs and courses. 
For example: https://api.learn.mit.edu/api/v0/vector_content_files_search/?resource_readable_id=program-v1%3AMITx%2BDEDP

The program does have a ContentFile with the site content but because it had None for a run_readable_id it gets skipped in the embedding process

### How can this be tested?
1. checkout main
2. run `./manage.py scrape_marketing_pages`
3. run the following script in shell which gets a marketing page , embeds it and then checks for existence in the api:
```python
import time

from learning_resources.etl.constants import MARKETING_PAGE_FILE_TYPE
from learning_resources.tasks import scrape_marketing_pages
from vector_search.constants import CONTENT_FILES_COLLECTION_NAME
from vector_search.utils import (
    _content_file_vector_hits,
    best_run_ids_for_resources,
    embed_learning_resources,
    remove_points_matching_params,
    qdrant_client,
    qdrant_query_conditions,
)

# 1. Grab any scraped marketing page (they are all run-less contentfiles)
cf = (
    ContentFile.objects.filter(
        file_type=MARKETING_PAGE_FILE_TYPE, published=True, run=None
    )
    .exclude(content="")
    .select_related("learning_resource")
    .first()
)
assert cf, "No scraped marketing pages in DB - run scrape_marketing_pages.apply() first"
rid = cf.learning_resource.readable_id
print(f"Testing contentfile {cf.id} for resource {rid!r}")

client = qdrant_client()

# 2. Clean slate: delete any existing points for this marketing page so the
#    script can be re-run (stale points from a previous run on either branch
#    would otherwise muddy the before/after)
remove_points_matching_params(
    {"key": cf.key, "resource_readable_id": rid},
    collection_name=CONTENT_FILES_COLLECTION_NAME,
)
time.sleep(2)  # deletes use wait=False

# 2. Embed it (synchronously, bypassing celery)
embed_learning_resources([cf.id], "content_file", overwrite=True)
time.sleep(3)  # upserts use wait=False

# 3. Rebuild EXACTLY the filter ContentFilesVectorSearchView uses for
#    ?resource_readable_id=<rid> (it swaps that param for a run filter)
client = qdrant_client()
params = {"run_readable_id": best_run_ids_for_resources([rid]) + [rid]}
flt = qdrant_query_conditions(params, collection_name=CONTENT_FILES_COLLECTION_NAME)
n = client.count(
    collection_name=CONTENT_FILES_COLLECTION_NAME, count_filter=flt, exact=True
).count
print(f"Points matching the API filter: {n}")

# 4. Check hit hydration for a run-less file
pts, _ = client.scroll(
    collection_name=CONTENT_FILES_COLLECTION_NAME, scroll_filter=flt,
    limit=10, with_payload=True,
)
marketing = [p for p in pts if p.payload.get("file_type") == "marketing_page"]
hit = _content_file_vector_hits(marketing)[0] if marketing else None
print("Hydrated from DB:", bool(hit and hit.get("id") == cf.id))
print("Payload run_readable_id fallback:", marketing[0].payload.get("run_readable_id") == rid if marketing else None)
```
4. note that there are 0 results
5. checkout this branch and run the same script and see we have marketing page content

### DIFF
--- a/vector_search/utils.py
+++ b/vector_search/utils.py
@@ -502,6 +502,27 @@ def _content_file_embedding_context(document):
     return document.get("content", "")
 
 
+def _with_run_readable_id_fallback(serialized_document):
+    """
+    Return the document with run_readable_id defaulted to the resource
+    readable_id for run-less content files (e.g. scraped marketing pages).
+
+    The content-file search API rewrites resource_readable_id filters into
+    run_readable_id filters (see ContentFilesVectorSearchView), so every point
+    payload must carry a run_readable_id to stay reachable -- course-metadata
+    points already follow this convention. Only the Qdrant payload gets the
+    fallback; point ids still derive from the raw serialized document.
+    """
+    if serialized_document.get("run_readable_id") or not serialized_document.get(
+        "resource_readable_id"
+    ):
+        return serialized_document
+    return {
+        **serialized_document,
+        "run_readable_id": serialized_document["resource_readable_id"],
+    }
+
+
 def _process_resource_embeddings(serialized_resources):
     docs = []
     metadata = []
@@ -559,7 +580,7 @@ def update_content_file_payload(serialized_document):
 
     _set_payload(
         points,
-        serialized_document,
+        _with_run_readable_id_fallback(serialized_document),
         param_map=QDRANT_CONTENT_FILE_PARAM_MAP,
         collection_name=CONTENT_FILES_COLLECTION_NAME,
     )
@@ -847,16 +868,18 @@ def _generate_content_file_points(serialized_content, stored_payloads):
                     break
                 chunk_id, split_doc = valid_chunks[relative_index]
 
-                metadata = {
-                    "resource_point_id": str(resource_vector_point_id),
-                    "chunk_number": chunk_id,
-                    "chunk_content": split_doc.page_content,
-                    **{
-                        key: split_doc.metadata[key]
-                        for key in QDRANT_CONTENT_FILE_PARAM_MAP
-                        if key in split_doc.metadata
-                    },
-                }
+                metadata = _with_run_readable_id_fallback(
+                    {
+                        "resource_point_id": str(resource_vector_point_id),
+                        "chunk_number": chunk_id,
+                        "chunk_content": split_doc.page_content,
+                        **{
+                            key: split_doc.metadata[key]
+                            for key in QDRANT_CONTENT_FILE_PARAM_MAP
+                            if key in split_doc.metadata
+                        },
+                    }
+                )
 
                 point_id = vector_point_id(
                     vector_point_key(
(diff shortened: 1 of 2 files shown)


vector_search/utils.py
```diff
@@ -1184,15 +1207,25 @@ def _content_file_vector_hits(search_result):
...
+    # run_readable_id; their Qdrant payloads carry the resource readable_id
+    # in that field instead, so key them the same way here.
     [
-        contentfiles_dict.update({(cf["run_readable_id"], cf["key"]): cf})
+        contentfiles_dict.update(
+            {
+                (
+                    cf.get("run_readable_id") or cf.get("resource_readable_id"),
+                    cf["key"],
+                ): cf
```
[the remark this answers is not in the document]
mbertrand: worth making this change

vector_search/utils.py
mbertrand, line 1: This fixes new points, but nothing will fix the marketing-page points already in Qdrant, so I don't think the DEDP example in the description actually starts working when this merges.

`daily_embed_new_content_files` runs with `overwrite=False` and only looks at recently-created files, and with `overwrite=False` `process_batch` drops any doc whose point already exists. `scrape_marketing_pages` only re-embeds resources that don't already have a marketing page. So the existing points keep their missing `run_readable_id` and stay unmatched by the resource->run filter rewrite.

Could fix those by running something like this in a shell after deploy:

```python
from itertools import islice
from learning_resources.etl.constants import MARKETING_PAGE_FILE_TYPE
from vector_search.utils import embed_learning_resources

ids = list(
    ContentFile.objects.filter(
        file_type=MARKETING_PAGE_FILE_TYPE, published=True, run=None
    ).exclude(content="").values_list("id", flat=True)
)
it = iter(ids)
while batch := list(islice(it, 200)):
    embed_learning_resources(batch, "content_file", overwrite=True)
```


Semi-related: the description says run-less content files "get skipped in the embedding process." — `_queue_program_content_file_embedding_tasks` has been queuing them since #3143, so the points exist, they were just unreachable (no `run_readable_id` for the filter) and unhydratable (`run__run_id__in` can't match a NULL run).

vector_search/utils.py
```diff
@@ -1184,15 +1207,25 @@ def _content_file_vector_hits(search_result):
     keys = [hit.payload.get("key") for hit in search_result]
 
     serialized_content_files = ContentFileSerializer(
-        ContentFile.objects.for_serialization().filter(
-            run__run_id__in=run_readable_ids, key__in=keys
-        ),
+        ContentFile.objects.for_serialization()
+        .filter(key__in=keys)
+        .filter(Q(run__run_id__in=run_readable_ids) | Q(run__isnull=True)),
```
mbertrand: `Q(run__isnull=True)` isn't scoped to the resources we actually got hits for — it matches any run-less content file that happens to share a key, and those extra rows all get pushed through `ContentFileSerializer` (which resolves topics/departments/offered_by per row).

In practice marketing-page keys are full URLs so it's probably close to a no-op today, but since the payload values here *are* resource readable_ids for run-less hits, we can just say so:

```python
.filter(Q(run__run_id__in=run_readable_ids) | Q(run__isnull=True, learning_resource__readable_id__in=run_readable_ids))
```

That also closes a small correctness hole: two run-less files with the same `(resource_readable_id, key)` currently collide in `contentfiles_dict` and the last one wins.

vector_search/utils_test.py
```diff
@@ -1270,6 +1271,132 @@ def test_update_payload_no_points(mocker):
     mock_qdrant.set_payload.assert_not_called()
 
 
+def test_generate_content_points_runless_run_readable_id_fallback(mocker):
```
mbertrand, line 1274: These cover the payload write, the payload refresh, and hydration, which is good — but nothing asserts that the payload we write is actually matched by the filter the view builds.

The whole fix rests on `ContentFilesVectorSearchView` appending raw resource ids to the rewritten run filter (views.py around line 722). If someone tidies that line up later, every test here still passes and the feature quietly breaks again. A small test asserting `qdrant_query_conditions({"run_readable_id": [resource.readable_id]}, CONTENT_FILES_COLLECTION_NAME)` matches a generated marketing-page payload would tie the two halves together.

mbertrand (changes requested):

mbertrand (approved):

vector_search/utils.py
shanbady: I was planning on manually deleting those points post-deploy

PR #3726, opened 2026-08-05
feat(settings): change email and password via Keycloak

ahtesham-quraish:
### Ticket:
https://github.com/mitodl/hq/issues/10815

Part of https://github.com/mitodl/hq/issues/10815

Adds an **Email & Password** section to the dashboard Settings page, with
**Change Email** and **Change Password** buttons, behind the PostHog flag
`account-management`.

Keycloak owns these credentials, so Learn doesn't collect them. Both buttons hand
the user off to Keycloak using its [application initiated
actions](https://www.keycloak.org/docs/latest/server_admin/#con-aia_server_administration_guide),
then return the user to Settings with an alert describing what happened. This is
the same approach MITx Online uses, in `authentication/api_gateway/views.py`.

Users who authenticate through an external identity provider are excluded — those
credentials belong to their institution, not to Learn.

Scope is the UI plus the endpoints and serializer fields behind it. **Getting a
changed address from Keycloak into Learn is not part of this PR** — see
_Propagation_ below.

#### The handoff

Two new endpoints, in `authentication/views.py`:

| Endpoint | Does |
|---|---|
| `account/action/start/<action>/` | Redirects to Keycloak's authorization endpoint with `kc_action=UPDATE_EMAIL` or `UPDATE_PASSWORD`, and a callback URL pointing at the endpoint below. |
| `account/action/complete` | Reads the `kc_action_status` Keycloak appends, and redirects to Settings with the outcome as query params. `AccountActionAlert` renders them once and strips them from the URL. |

Both legs build the callback URL through one helper,
`build_account_action_callback_url()`, so the value Keycloak sees when starting
the action matches the one it redirects back to.

`KEYCLOAK_CLIENT_ID` is a new setting with **no default**. The callback URL has to
be a registered redirect URI on that client, so a guessed default would fail inside
Keycloak with an opaque error. When it's unset the flow returns an error alert
rather than raising.

#### Propagation — deliberately not handled here

Learn keeps its own copy of the email, so a change in Keycloak has to reach the
Django `User`. That's **SCIM's job**, not this PR's:

- #3747 makes `ApisixUserMiddleware`'s userinfo updates togglable, so deployed
  environments stop overwriting SCIM-delivered values with a stale gateway header
- Keycloak pushes the change to Learn over SCIM, which Learn already accepts
  (`django_scim` + `mitol.scim` are wired up in `main/urls.py`)

`docs/keycloak-scim-request.md` is the request for that push configuration, which
lives outside this repo. Worth flagging: **it isn't set up for Learn yet**, and
once #3747 disables middleware updates, SCIM is the only path left — so it needs to
be in place before the flag goes on.

Because propagation is asynchronous and out of band, the callback can't tell an
applied change from one still waiting on its confirmation link. So the update-email
success alert points at the inbox rather than claiming the address already changed:

| Situation | Message |
|---|---|
| Email change accepted by Keycloak | "Check your inbox for a confirmation link to finish updating your email address." |
| Password changed | "Your password has been updated." |
| User cancels at Keycloak | no alert — cancelling is deliberate |
| Error, or unknown action | error alert |
| SSO user | "…managed by your organization's single sign-on provider." |

This is accurate for every deployed realm, which all have `verify_email` on. It's
mildly pessimistic where verification is off (the change *is* immediate then) —
happy to split the copy by realm behaviour if you'd rather, but that needs the
authorization-code exchange back to detect which case applies.

#### Excluding SSO users

The Settings section is hidden for them, and `AccountActionStartView` re-checks
independently so the URL can't simply be visited. `is_sso_user()` reads a user's
federated identities from the Keycloak Admin API and caches the result per user for
an hour.

This needs `MITOL_KEYCLOAK_ADMIN_CLIENT_*`, which **no deployed environment sets
yet** — those come from ol-infrastructure#5252 and #5292. Until they're applied the
check fails open, the buttons show for everyone, and Keycloak itself remains the
only real gate. It also activates
`profiles.api.sync_email_optin_to_keycloak`, which is gated on the same setting and
has therefore never run anywhere.

#### API change

`GET /api/v0/users/me/` now returns `email` and `is_sso_user`, via a new
`CurrentUserSerializer`. `UserSerializer.email` stays `write_only`, so the staff
`/api/v0/users/` list doesn't start exposing addresses.

#### Local development

`UPDATE_EMAIL` is a Keycloak **preview** feature, so two things are needed for the
flow to run at all locally, and both are committed:

- the `update-email` feature is added to the `keycloak` service's `--features` flag
- the `UPDATE_EMAIL` required action is registered in the realm export

Without them Keycloak fails with a generic error page whose only clue is
`NullPointerException … "requiredActionProvider" is null` in its logs.
`--import-realm` **skips realms that already exist**, so an existing local setup
needs the action registered once by hand — step 4 below does exactly that.

`UPDATE_PASSWORD` is built in and needs neither.

- [ ] Desktop screenshots

### How can this be tested?

All commands below were run against a working local setup. Two things genuinely
can't be checked locally and are listed at the end so they aren't mistaken for
breakage.

#### 1. Start the containers

Needs `COMPOSE_PROFILES=backend,frontend,keycloak,apisix` in `.env`.

```bash
docker compose up -d
```

#### 2. Get an admin token (reused throughout)

```bash
TOKEN=$(curl -s -X POST "http://localhost:8066/realms/master/protocol/openid-connect/token" \
  -d "client_id=admin-cli" -d "username=admin" -d "password=admin" -d "grant_type=password" \
  | python3 -c "import sys,json;print(json.load(sys.stdin)['access_token'])")
```

It expires in a few minutes — re-run it if you start getting `401`s.

#### 3. Check the realm has the required action

`--import-realm` **skips realms that already exist**, so if your Keycloak predates
this branch the realm change won't have applied:

```bash
curl -s -H "Authorization: Bearer $TOKEN" \
  "http://localhost:8066/admin/realms/ol-local/authentication/required-actions" \
  | python3 -c "import sys,json;print([r['alias'] for r in json.load(sys.stdin) if r['alias'].startswith('UPDATE')])"
```

Want to see `UPDATE_EMAIL` in the list:

```
['UPDATE_PASSWORD', 'UPDATE_PROFILE', 'UPDATE_EMAIL']
```

#### 4. If it's missing, register it once

```bash
curl -s -X POST -H "Authorization: Bearer $TOKEN" -H "Content-Type: application/json" \
  -d '{"providerId":"UPDATE_EMAIL","name":"Update Email"}' \
  "http://localhost:8066/admin/realms/ol-local/authentication/register-required-action"
```

Re-run step 3 to confirm, then `docker compose restart keycloak`.

#### 5. Confirm Django has the Keycloak client id

Already in `env/backend.env`:

```bash
docker compose exec web python manage.py shell -c "
from django.conf import settings as s; print('client_id:', s.KEYCLOAK_CLIENT_ID)"
```

Use `manage.py shell -c`, not `python -c` — a bare `python -c` raises
`ImproperlyConfigured` because Django isn't set up.

#### 6. Enable the feature flag

Add to `env/frontend.local.env`, then restart the frontend:

```
NEXT_PUBLIC_FEATURE_account_management=True
NEXT_PUBLIC_POSTHOG_API_KEY=any-non-empty-value
```

Three ways this silently fails:

- the part after the prefix becomes the flag name with `_`→`-` and is
  **case-sensitive** — must be lowercase, or you get `ACCOUNT-MANAGEMENT`
- the value must be exactly `True`
- `NEXT_PUBLIC_POSTHOG_API_KEY` must be non-empty or `ConfiguredPostHogProvider`
  never mounts and bootstrap flags are ignored entirely. A dummy value is fine

---

### Change email

Log in at http://open.odl.local:8062 as **`student@odl.local` / `student`**, then go
to **Settings**.

1. **Change Email** → new address → submit
2. Back on Settings, expect the alert **"Check your inbox for a confirmation link to
   finish updating your email address."**
3. Confirm Keycloak took the change:
   ```bash
   curl -s -H "Authorization: Bearer $TOKEN" \
     "http://localhost:8066/admin/realms/ol-local/users?search=student" \
     | python3 -c "import sys,json;print([u.get('email') for u in json.load(sys.stdin)])"
   ```

The local realm has `verifyEmail` off, so Keycloak applies the change immediately
and there's no confirmation email to open. Deployed realms have it on, where
Keycloak emails a link first — which is why the alert is worded the way it is.

**Learn's own copy of the address will not change locally**, and that's expected
rather than a bug: propagation is SCIM's job (see _Propagation_ above) and there's
no SCIM push against a local Keycloak. Check Keycloak's value, per step 3.

### Change password

**Change Password** → Keycloak's form → submit → back to Settings with *"Your
password has been updated."* Log out and in with the new password. Keycloak applies
password changes immediately, so nothing async here.

### Cancel and error paths

- Start either action, click **Cancel** in Keycloak → back to Settings with **no
  alert**. Cancelling is deliberate, not an error.
- An unknown action returns an error alert rather than a 500:
  ```bash
  curl -s -o /dev/null -D - "http://localhost:8063/account/action/start/delete-account/" \
    | grep -i '^location'
  # → …/dashboard/settings?account_action=delete-account&account_action_status=error
  ```

### Automated tests

```bash
docker compose exec web uv run pytest authentication/ profiles/ -q

yarn test frontends/main/src/app-pages/DashboardPage/SettingsContent.test.tsx \
          frontends/main/src/app-pages/DashboardPage/AccountActionAlert.test.tsx \
          frontends/main/src/common/urls.test.ts
```

### Not testable locally

- **Hiding the section from SSO users.** Needs `MITOL_KEYCLOAK_ADMIN_CLIENT_*`,
  which no environment sets yet (ol-infrastructure#5252, #5292). Locally
  `is_sso_user()` short-circuits to `False`, so **the section always shows** — this
  is expected, not a bug. Review via `authentication/api_test.py::test_is_sso_user*`,
  or force it with
  `mocker.patch("authentication.views.is_sso_user", return_value=True)`.
- **The SCIM push carrying the new address into Learn.** No SCIM code in this PR.
  The push needs a deployed environment — the Keycloak plugin is
  enterprise-licensed and absent from the upstream image used locally.

### Additional Context

**This PR is safe to merge on its own** with the `account-management` flag **off**:
the endpoints exist but nothing links to them, and `is_sso_user()` fails open.

Before turning the flag on, in order:

1. **This PR** — merge with the flag off
2. **#3747** — so deployed environments stop overwriting SCIM
3. **Keycloak SCIM push configured for Learn** — `docs/keycloak-scim-request.md`.
   Without this, step 2 leaves nothing writing the new address at all
4. **ol-infrastructure#5252 → #5292** — the admin client, so SSO users stop seeing
   the buttons
5. Enable `account-management`, RC first

**No new email template is needed.** The ticket asks. Keycloak sends the
confirmation itself from its own theme, so Learn sends no mail in this flow.

**Reviewer question worth a look:** the update-email success copy now points at the
inbox in all cases, since without the authorization-code exchange the callback can't
distinguish an applied change from a pending one. Accurate for deployed realms, a
little pessimistic where verification is off. Say the word if you'd rather have the
exchange back to tell them apart.


authentication/views.py
```diff
@@ -106,3 +119,208 @@ def get(
...
+        next_url = get_account_action_redirect_url(self.request)
+        raw_action = self.request.GET.get(ACCOUNT_ACTION_PARAM)
+        keycloak_status = self.request.GET.get(KEYCLOAK_ACTION_STATUS_PARAM)
+
+        action = parse_account_action(raw_action)
+        status = KEYCLOAK_ACTION_STATUSES.get(keycloak_status)
+
+        if action is None or status is None:
+            # Not something we can report on, so redirect without an alert
+            # rather than guess at an outcome.
```
ChristopherChudzicki, line 267: If this code triggers, that means user tried to initiate a change, but we weren't able to complete it. Something went wrong. (E.g., keycloak not configured to allow updating email).

Am I reading that correctly? If yes, I think we should show an opaque "something went wrong" alert in the redirect param.

I think you'd nee a corresponding change [here](https://github.com/mitodl/mit-learn/pull/3726/changes#diff-679881a110c4fc6c50018aa96d726339be27a973c5dcf4b7db66bd52acd74f57R44)

ChristopherChudzicki (commented):
I didn't get a chance to try the actual email change flows locally. I did look over the code some. Overall, I think I'd like someone with more experience on our Keycloak integrations (@rhysyngsun or @cp-at-mit, perhaps) to take a look. 

But here are my impressions:

1. **Rewrite the description against `main`**: The PR description is pretty long, and references this branch's own history, rather than what has changed relative to `main`. For example, 
    - "What this PR now does:...", and I think the entire blockquote at the beginning, is in reference to an old version of the PR
    - "... Claiming "updated" here was the bug this
PR fixes..." I think the "bug" referenced here was an earlier commit in this PR, never on `main`.

    This is a fairly large PR. Don't describe its history, describe what it is intended to do and how it does that.
2. **I think drop `provision_scim_client.py`:** We already have a SCIM credential in production — a token on a service account. I don't think we need to add it here?
    - Potentially useful to codify it as a management command, I guess? But the default username should match prod if you want to keep it.
3. **Minor, but timeout could go in ol-django** Rather than additing it here, add the timeout to [get_admin_client](https://github.com/mitodl/ol-django/blob/a5b72402953c6832fec9c6aff5599c270bcb7b99/src/keycloak/mitol/keycloak/api.py#L9)
    - ❓ Is there more that should move to ol-django?
4. **Consider dropping token exchange:** As far as I know, MITxOnline does not use Keycloak authorization codes at all for this sync. I think we will always use SCIM in production. I don't think having the token exchange just for local dev is worth it. (Though I'd defer to @rhysyngsun / @cp-at-mit).
5. **Email Syncing:** Currently, this PR syncs keycloak fields [except email](https://github.com/mitodl/mit-learn/pull/3726/changes#diff-bd8d2bd1180347c167769273604116125f9d89ddd7cdbf73af1d8a8429fb1867R49) on requests. As written, it skips email *but not username* so they get out of sync. (We set username = email.)

    A better approach *might* be: Only sync fields from header if the field is unset, otherwise only SCIM can change them. (I.e., header = create, SCIM = update).
6. **Say something when the callback params don't parse:** see [here](https://github.com/mitodl/mit-learn/pull/3726/changes#r3732567440)

authentication/api.py
```diff
@@ -39,6 +49,131 @@ def create_user(username, email, profile_data=None, user_extra=None):
...
+
+    cache = caches["redis"]
+    cache_key = f"{SSO_USER_CACHE_PREFIX}{global_id}"
+    cached = cache.get(cache_key)
+    if cached is not None:
+        return cached
+
+    try:
+        admin_client = main_keycloak.get_admin_client()
+        is_sso = bool(admin_client.get_user_social_logins(global_id))
```
rhysyngsun, line 81: I think we should actually cache this on the `User` model instead of putting it in the cache - it's a value that rarely changes and it's possible that how we determine if a user is sso could change in the future. Additionally, storing it as a field allows us to enable a user to change their email simply by toggling a flag, in case in the future it's valid for a user to take their account with them when they leave an org - for example if someone with an MIT email wanted to change their account email to a personal one so they don't lose access to courses associated with that account. You can default the column to null and if the value is null and a `GET` is performed on the `user/me` API, we do a first-time fetch of this and store the value then.

rhysyngsun (changes requested):
I think generally this PR is doing too much - can you reduce the scope so that it's only the frontend changes and the backend changes needed to support the logic there (e.g. the `is_sso_user()` function.

I opened the draft PR https://github.com/mitodl/mit-learn/pull/3747 to address the SCIM updates getting overridden by the middleware - for deployed environments we'll just depend on that mechanism to propagate the changes. If we need to do more to make this more robust we can do so separately in another PR but I think the SCIM updates should be enough for now.

authentication/api.py
ahtesham-quraish: Addressed

authentication/views.py
```diff
@@ -106,3 +118,170 @@ def get(
                 profile.save()
 
         return redirect(redirect_url)
+
+
+ACCOUNT_SETTINGS_PATH = "/dashboard/settings"
+KEYCLOAK_ACTION_STATUS_PARAM = "kc_action_status"
```
rhysyngsun: These should be in `constants.py` for consistency.

docs/keycloak-scim-request.md
```diff
@@ -0,0 +1,196 @@
+# Request: push user changes from Keycloak to MIT Learn over SCIM
```
rhysyngsun: Can you remove this from your PR? we already have SCIM setup in deployed environments and it's not relevant to the work here.

rhysyngsun (commented):
Partial pass on this, I might have more but wanted to get this over to you before the end of my day.

docs/keycloak-scim-request.md
ahtesham-quraish: Done

authentication/views.py
ahtesham-quraish: Done

profiles/serializers.py
```diff
@@ -395,6 +395,33 @@ class Meta:
...
+    read-only: users change their email through Keycloak, not through us.
+    """
+
+    email = serializers.SerializerMethodField()
+    is_sso_user = serializers.SerializerMethodField()
+
+    def get_email(self, instance) -> str:
+        """Get the user's email address (blank for anonymous users)"""
+        return getattr(instance, "email", "") or ""
+
```
rhysyngsun: Should be able to accomplish this by changing the definition on `UserSerializer` to `email = serializers.CharField(read_only=True)`

docs/keycloak-scim-request.md
rhysyngsun: It looks like this file is still here

rhysyngsun (changes requested):
A couple of small things, I wasn't able to verify the email change end-to-end because keycloak is kicking me back to the application, I had to end my day before I could investigate that so will need to circle back on that Monday.

rhysyngsun (approved):
LGTM

PR #3730, opened 2026-08-06
remove GITHUB_ACCESS_TOKEN

abeglova:
None

The GITHUB_ACCESS_TOKEN is only used by the podcast etl to get our podcast repo. That repo is public. The etl job works without GITHUB_ACCESS_TOKEN and only runs twice a day so we are unlikely to be rate limited. With GITHUB_ACCESS_TOKEN it throws an error due to the new mitodl organization token restrictions 
https://celery-monitoring.odl.mit.edu/task/?app=mitlearn&env=production&uuid=e255344d-4996-4c97-b002-edba5c606323

### How can this be tested?
run `docker-compose run web ./manage.py backpopulate_podcast_data`
it should finish

### DIFF
--- a/learning_resources/etl/podcast.py
+++ b/learning_resources/etl/podcast.py
@@ -40,10 +40,7 @@ def github_podcast_config_files():
         A list of pyGithub contentFile objects
     """  # noqa: D401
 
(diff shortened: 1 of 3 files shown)


shanbady (approved):
LGTM 👍

PR #3734, opened 2026-08-06
Update @mitodl/mitxonline-api-axios to 2026.8.6

ChristopherChudzicki:
None, but doing this as a prereq to #3709 

Bumps the mitxonline api client; adds fields to factories, etc.

### How can this be tested?
Tests should pass


frontends/main/src/app-pages/DashboardPage/AnalyticsContent.tsx
```diff
@@ -154,7 +153,12 @@ const AnalyticsContentInternal: React.FC<AnalyticsContentInternalProps> = ({
   } = useQuery(managerOrganizationQueries.managerOrganizationsList())
 
   const org = managerOrgs?.find(matchOrganizationBySlug(orgSlug))
-  const orgUuid = getOrgUuid(org)
```
ChristopherChudzicki, line 157: Previously, this did 
```ts
  const value = (org as { sso_organization_id?: unknown } | undefined)
    ?.sso_organization_id
  return typeof value === "string" && value.length > 0 ? value : null
```

I.e., it was a type-safe way of getting `sso_organization_id` before it was included in the API client.

(I guess it also omitted empty string `""` uuids. The new version, direct access, does not do that. But ~if the uuid is empty, then we have bigger problems...~ we should trust the API)

daniellefrappier18 (approved):
LGTM

PR #3737, opened 2026-08-06
fix broken images in subscription emails

shanbady:

Closes https://github.com/mitodl/hq/issues/12708

This PR ensures that images in emails properly fallback instead of appearing broken. There does not appear to be a single solution that works client-side and is supported by all mail clients. The proper fix in this case is to do a HEAD request - set the default (on the backend) if it broken.

### How can this be tested?
1. checkout main
2. run the following script which generates a resource with an invalid image url and then sends a subscriptoion email with it:
```python
import datetime

from django.db import transaction

from learning_resources.models import LearningResource, LearningResourceImage
from learning_resources_search.api import percolate_matches_for_document
from learning_resources_search.models import PercolateQuery
from learning_resources_search.tasks import send_subscription_emails
from main.celery import app
from profiles.models import Profile

User = get_user_model()
DEAD_IMAGE_URL = "https://ocw.mit.edu/images/this-image-was-removed-404.jpg"

# run the celery task inline, and print emails instead of sending them
app.conf.task_always_eager = True
settings.NOTIFICATION_EMAIL_BACKEND = "django.core.mail.backends.console.EmailBackend"


class Rollback(Exception):
    """Raised to roll the transaction back once the demo has run."""


try:
    with transaction.atomic():
        since = now_in_utc() - datetime.timedelta(days=1)
        recent = LearningResource.objects.filter(published=True, created_on__gt=since)

        # Two resources matching the SAME query, so both land in one digest
        chosen, query = [], None
        for resource in recent.select_related("image"):
            matches = percolate_matches_for_document(resource.id).filter(
                source_type=PercolateQuery.CHANNEL_SUBSCRIPTION_TYPE
            )
            if not resource.image:
                continue
            if query is None:
                query = matches.filter(users__isnull=False).first()
                if query:
                    chosen.append(resource)
            elif matches.filter(id=query.id).exists():
                chosen.append(resource)
            if len(chosen) == 2:
                break

        if len(chosen) < 2:
            raise SystemExit(
                "Could not find two recently-created resources that percolate-match "
                "a channel subscription query. Load fresh data and reindex, then retry."
            )

        keep, break_image_on = chosen
        print(f"\nUsing subscription query {query.id}: {query.original_query}\n")
        print(f"  intact image : {keep.title[:60]}")
        print(f"                 {keep.image.url[:90]}")
        print(f"  broken image : {break_image_on.title[:60]}")
        print(f"                 {DEAD_IMAGE_URL}\n")

        # Point one resource at a dead image URL
        break_image_on.image = LearningResourceImage.objects.create(
            url=DEAD_IMAGE_URL, alt=break_image_on.title
        )
        break_image_on.save()

        # Keep the digest to just these two resources
        recent.exclude(id__in=[keep.id, break_image_on.id]).update(
            created_on=since - datetime.timedelta(days=30)
        )

        # Send to a throwaway subscriber. EVERY query matching either resource
        # must be reassigned, not just the one we picked, or the digest goes to
        # real subscribers and prints their addresses.
        subscriber = User.objects.create(
            username="digest-preview-user", email="reviewer@example.com"
        )
        Profile.objects.create(user=subscriber, email_optin=True)
        matching_query_ids = {
            match.id
            for resource in (keep, break_image_on)
            for match in percolate_matches_for_document(resource.id)
        }
        for match in PercolateQuery.objects.filter(id__in=matching_query_ids):
            match.users.set([subscriber])

        print("=" * 78)
        print("Running send_subscription_emails - email printed below")
        print("=" * 78)

        send_subscription_emails.apply(("channel_subscription_type",), {"period": "daily"})

        raise Rollback
except Rollback:
    print("\n" + "=" * 78)
    print("Transaction rolled back - no data was changed, no email was sent.")
    print("=" * 78)
```
3. copy and paste the generated email content (copy from opening <html> to closing not the text version) and save it as email.html and view it in the browser
4. note that the image is broken
5. checkout this branch and note it resolves

### Additional Context
There is a codeql failure in this PR that is being addressed separately as part of https://github.com/mitodl/hq/issues/12773


learning_resources/tasks.py
```diff
@@ -847,3 +852,35 @@ def cleanup_deleted_content_files():
...
+    instead of rendering a broken image.
+    """
+    image_ids = (
+        LearningResource.objects.filter(published=True, image__isnull=False)
+        .values_list("image_id", flat=True)
+        .distinct()
+    )
+    removed = 0
+    for image in LearningResourceImage.objects.filter(id__in=image_ids).iterator():
+        if image_url_is_reachable(image.url):
```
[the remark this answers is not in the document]
shanbady: fixed

learning_resources/utils.py
```diff
@@ -227,6 +230,40 @@ def load_course_blocklist():
...
+            if response.status_code in (
+                HTTPStatus.METHOD_NOT_ALLOWED,
+                HTTPStatus.NOT_IMPLEMENTED,
+            ):
+                # Some servers reject HEAD requests; retry without the body
+                response = requests.get(
+                    url, timeout=settings.REQUESTS_TIMEOUT, stream=True
+                )
+                response.close()
+            reachable = response.ok
```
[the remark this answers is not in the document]
shanbady: fixed

learning_resources/tasks.py
```diff
@@ -847,3 +852,60 @@ def cleanup_deleted_content_files():
...
+            continue
+        resource_ids = list(image.learningresource_set.values_list("id", flat=True))
+        log.info(
+            "Pruning unreachable image %s from resources %s", image.url, resource_ids
+        )
+        # FK is SET_NULL, so deleting the record clears it on referencing resources
+        image.delete()
```
mbertrand: I think this will delete images on transient failures. `image_url_is_reachable` returns False for timeouts, connection errors, 429s and 5xx as well as real 404s, and caches that False for 24h in the same cache the email path writes to — so a CDN blip during Saturday's digest send gets read back here on Sunday and the image is deleted without ever being re-checked. This task is also a burst of python-requests HEADs from datacenter IPs, which is what CDN bot protection tends to answer with 403/429, so one throttling episode could wipe a lot of images in a single run.

For the email fallback a false negative is harmless, but before deleting I'd want a definitive status (404/410) rather than any failure, and probably skip the cached value on this path.

mbertrand (changes requested):

learning_resources/tasks.py
shanbady: good point. what do you think about removing the task altogether (is it overkill for a pretty rare occurrance?) and just sticking to that validate/fallback check when sending the actual email.

mbertrand: That seems like it would be fine

mbertrand (approved):

PR #3705, opened 2026-07-30
Rate-limit content_feedback submissions per user

zamanafzal:
### What are the relevant tickets?

mitodl/hq#12355

### Description (What does it do?)

Rate-limits `POST /api/v0/content_feedback/` to **10 submissions/min per authenticated user**, returning HTTP 429 when exceeded. Adds a reusable `RedisScopedRateThrottle` (redis-backed DRF `ScopedRateThrottle`) and opts the view in via `throttle_scope`. The rate is env-configurable via `CONTENT_FEEDBACK_THROTTLE_RATE`.

### Screenshots (if appropriate):

N/A — backend only.

### How can this be tested?

**1. Automated:**

```
docker compose run --rm web uv run pytest main/throttles_test.py content_feedback/ -v
```

`test_submit_rate_limited` forces a low rate and asserts `201, 201, 429` for one user, while a different user still gets `201` (per-user keying).

**2. Live endpoint through APISIX (real auth path, default 10/min):**

Same approach as the base PR #3593 — log in, then run the snippet from the devtools console on the `open.odl.local:8065` origin. It submits **11 times in the same minute**: the first **10 return 201**, the **11th returns 429**.

1. `docker compose up -d`, then log in at `http://open.odl.local:8065/login` (complete Keycloak). Sessions lapse; if you see `NotAuthenticated`, log in again.
2. On the `open.odl.local:8065` origin, open the console and run:

```js
(async () => {
  const API = "http://open.odl.local:8065";
  const me = await fetch(`${API}/api/v0/users/me/`, { credentials: "include" });
  if (me.status !== 200) { console.warn("Not logged in (", me.status, ") → visit /login first"); return; }
  console.log("authed as:", (await me.json()).username);
  const csrf = document.cookie.match(/(?:^|; )csrftoken-local=([^;]+)/)?.[1];
  const body = JSON.stringify({
    course_id: "course-v1:MITx+6.00+2T2026",
    course_name: "Introduction to Computer Science",
    block_usage_key: "block-v1:MITx+6.00+2T2026+type@video+block@abc123",
    block_type: "video",
    block_display_name: "Lecture 3: Recursion",
    unit_title: "Recursion and Dictionaries",
    url: "https://apps.mitxonline.mit.edu/learn/course/x/y/abc123",
    sentiment: "idea",
    comment: "throttle test",
  });
  for (let i = 1; i <= 11; i++) {
    const res = await fetch(`${API}/api/v0/content_feedback/`, {
      method: "POST",
[...]

PR #3738, opened 2026-08-07
feat(content_feedback): allow anonymous submissions

zamanafzal:

Part of content feedback (mitodl/hq#11629). Addresses the courseware-learner `403`: courseware-only learners have no mit-learn/APISIX session, so the `IsAuthenticated` gate rejected nearly all of them with a `NotAuthenticated` 403 (before CSRF was ever checked).

Switches `POST /api/v0/content_feedback/` from `IsAuthenticated` to `AllowAny`, and attributes the row to the request user only when authenticated (else stores `user=null`). This mirrors learn-ai/AskTIM's anonymous-tolerant posture.

- The `ContentFeedback.user` FK is already `null=True`, so **no migration** is required.
- With no authenticated user, DRF's `SessionAuthentication` returns `None`, so CSRF is not enforced on anonymous submissions.
- Session-bearing (authenticated) callers are unchanged: CSRF is still enforced for them — a session POST without an `X-CSRFToken` is rejected (`403`), covered by `test_authenticated_without_csrf_token_rejected`.

Screenshot 2026-08-07 at 2 52 21 PM
Screenshot 2026-08-07 at 2 55 05 PM

### How can this be tested?

**1. Automated**

```
docker compose run --rm web uv run pytest content_feedback/views_test.py -v
```

`test_submit_allows_anonymous` asserts an anonymous POST returns `201` and stores `user=None`; the authenticated tests confirm the row is still attributed to the request user.

**2. Anonymous — no session at all (the behavior this PR adds)**

The whole point of the change: a caller with **no auth and no CSRF** is now accepted. Post straight to the web container:

```bash
curl -s -w '\n--> HTTP %{http_code}\n' -X POST http://localhost:8061/api/v0/content_feedback/ \
  -H 'Content-Type: application/json' \
  -d '{"course_id":"course-v1:Test+Anon+2026","block_usage_key":"block-v1:demo+type@video+block@anon-'"$(date +%s)"'","sentiment":"positive","comment":"anon test"}'
```

Returns `--> HTTP 201`. Verify the stored row has no user:

```bash
docker compose run --rm web python manage.py shell -c \
  "from content_feedback.models import ContentFeedback as F; r=F.objects.filter(course_id='course-v1:Test+Anon+2026').latest('created_on'); print('user=', r.user, '| sentiment=', r.sentiment)"
# -> user= None | sentiment= positive
```

Before this PR the identical request returned `403 NotAuthenticated`.

**3. Authenticated submissions are still attributed (live through APISIX)**

Same as #3705 §2 — the real auth path. `docker compose up -d`, log in at `http://open.odl.local:8065/login` (complete Keycloak), then from the console on the `open.odl.local:8065` origin submit once:

```js
(async () => {
  const API = "http://open.odl.local:8065";
  const me = await fetch(`${API}/api/v0/users/me/`, { credentials: "include" });
  const authed = me.status === 200;
  console.log(authed ? "authed as: " + (await me.json()).username : "logged out (" + me.status + ")");
  const csrf = document.cookie.match(/(?:^|; )csrftoken-local=([^;]+)/)?.[1];
  const res = await fetch(`${API}/api/v0/content_feedback/`, {
    method: "POST",
    credentials: "include",
    headers: { "Content-Type": "application/json", ...(csrf ? { "X-CSRFToken": csrf } : {}) },
    body: JSON.stringify({
      course_id: "course-v1:MITx+6.00+2T2026",
      block_usage_key: "block-v1:MITx+type@video+block@abc123",
      sentiment: "idea",
      comment: "live test",
    }),
  });
  console.log("submit →", res.status); // 201
})();
```

Logged in → `201`, attributed to your user. Logged out → **also `201`** now (previously `403`), stored with `user=null`.

### DIFF
--- a/content_feedback/views.py
+++ b/content_feedback/views.py
@@ -2,7 +2,7 @@
 
 from drf_spectacular.utils import OpenApiResponse, extend_schema
 from rest_framework.generics import CreateAPIView
-from rest_framework.permissions import IsAuthenticated
+from rest_framework.permissions import AllowAny
 
 from content_feedback.models import ContentFeedback
 from content_feedback.serializers import ContentFeedbackSerializer
@@ -18,14 +18,21 @@ from main.throttles import RedisScopedRateThrottle
     },
 )
 class ContentFeedbackView(CreateAPIView):
-    """Accept per-block content feedback submissions (append-only)."""
+    """Accept per-block content feedback submissions (append-only).
+
+    Anonymous submissions are allowed: courseware-only learners have no
+    mit-learn/APISIX session, so requiring authentication would reject nearly
+    all of them (403). This mirrors learn-ai/AskTIM, which tolerates anonymous
+    callers and stores no user. Attributed submissions still record the user.
+    """
 
     queryset = ContentFeedback.objects.all()
     serializer_class = ContentFeedbackSerializer
-    permission_classes = (IsAuthenticated,)
+    permission_classes = (AllowAny,)
     throttle_classes = (RedisScopedRateThrottle,)
     throttle_scope = "content_feedback"
 
     def perform_create(self, serializer):
-        """Attribute the feedback to the authenticated user (server-side)."""
-        serializer.save(user=self.request.user)
+        """Attribute to the request user when authenticated, else store null."""
+        user = self.request.user if self.request.user.is_authenticated else None
+        serializer.save(user=user)
--- a/content_feedback/views.py
+++ b/content_feedback/views.py
@@ -20,10 +20,9 @@ from main.throttles import RedisScopedRateThrottle
 class ContentFeedbackView(CreateAPIView):
     """Accept per-block content feedback submissions (append-only).
 
-    Anonymous submissions are allowed: courseware-only learners have no
-    mit-learn/APISIX session, so requiring authentication would reject nearly
-    all of them (403). This mirrors learn-ai/AskTIM, which tolerates anonymous
-    callers and stores no user. Attributed submissions still record the user.
+    Uses AllowAny: courseware-only learners have no mit-learn/APISIX session,
+    so requiring auth would 403 nearly all of them. Authenticated rows record
+    the user; anonymous rows store null (mirrors learn-ai/AskTIM).
     """
 
     queryset = ContentFeedback.objects.all()
--- a/content_feedback/views.py
+++ b/content_feedback/views.py
@@ -18,15 +18,13 @@ from main.throttles import RedisScopedRateThrottle
     },
 )
 class ContentFeedbackView(CreateAPIView):
-    """Accept per-block content feedback submissions (append-only).
-
-    Uses AllowAny: courseware-only learners have no mit-learn/APISIX session,
-    so requiring auth would 403 nearly all of them. Authenticated rows record
-    the user; anonymous rows store null (mirrors learn-ai/AskTIM).
-    """
+    """Accept per-block content feedback submissions (append-only)."""
 
     queryset = ContentFeedback.objects.all()
     serializer_class = ContentFeedbackSerializer
+    # AllowAny: courseware-only learners have no mit-learn/APISIX session, so
+    # requiring auth would 403 nearly all of them. Authenticated rows record the
+    # user; anonymous rows store null (mirrors learn-ai/AskTIM).
     permission_classes = (AllowAny,)
     throttle_classes = (RedisScopedRateThrottle,)
     throttle_scope = "content_feedback"
--- a/content_feedback/views_test.py
+++ b/content_feedback/views_test.py
@@ -26,11 +26,12 @@ def _payload(**overrides):
     return payload
 
 
-def test_submit_requires_authentication(client):
-    """Anonymous users cannot submit feedback."""
+def test_submit_allows_anonymous(client):
+    """Anonymous users can submit feedback; the record has no user."""
     response = client.post(reverse("content_feedback:v0:content_feedback"), _payload())
-    assert response.status_code in (401, 403)
-    assert ContentFeedback.objects.count() == 0
+    assert response.status_code == 201
+    assert ContentFeedback.objects.count() == 1
+    assert ContentFeedback.objects.get().user is None
 
 
 def test_submit_creates_record_for_user(user_client, user):
--- a/content_feedback/views_test.py
+++ b/content_feedback/views_test.py
@@ -2,6 +2,7 @@
 
 import pytest
 from django.urls import reverse
+from rest_framework.test import APIClient
 
 from content_feedback.factories import ContentFeedbackFactory
 from content_feedback.models import ContentFeedback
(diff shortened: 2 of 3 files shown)


content_feedback/views.py
```diff
@@ -18,14 +18,20 @@
...
+
+    Uses AllowAny: courseware-only learners have no mit-learn/APISIX session,
+    so requiring auth would 403 nearly all of them. Authenticated rows record
+    the user; anonymous rows store null (mirrors learn-ai/AskTIM).
+    """
 
     queryset = ContentFeedback.objects.all()
     serializer_class = ContentFeedbackSerializer
-    permission_classes = (IsAuthenticated,)
+    permission_classes = (AllowAny,)
```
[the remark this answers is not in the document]
zamanafzal: Correct — anonymous requests key by IP, so learners behind one egress IP (campus NAT) share the `10/min` bucket. Accepting this for now: it's a small, reversible change still gated behind a product decision (draft), usage is async/self-paced so 10 submissions from one IP in 60s is an edge case, and the rate is env-tunable (`CONTENT_FEEDBACK_THROTTLE_RATE`). A stable anonymous identity isn't available here by design. If anonymous ships, we'll revisit the throttle (real client IP via `NUM_PROXIES` + a separate anon rate) as a rollout follow-up.

content_feedback/views.py
```diff
@@ -18,14 +18,20 @@
     },
 )
 class ContentFeedbackView(CreateAPIView):
-    """Accept per-block content feedback submissions (append-only)."""
+    """Accept per-block content feedback submissions (append-only).
+
+    Uses AllowAny: courseware-only learners have no mit-learn/APISIX session,
+    so requiring auth would 403 nearly all of them. Authenticated rows record
+    the user; anonymous rows store null (mirrors learn-ai/AskTIM).
+    """
```
mbertrand: nit: drf-spectacular publishes the *whole* docstring, not just the first line (`AutoSchema.get_description()` calls `get_doc(view_class)`), so this rationale becomes the public `description` in `openapi/specs/v0.yaml` and the JSDoc on the generated TS client. I regenerated locally to check and that's what comes out.

I'd keep the docstring to the one-line summary and move the why to a plain `#` comment above `permission_classes` — same value to anyone reading this file, without shipping "mirrors learn-ai/AskTIM" to API consumers.

content_feedback/views.py
mbertrand, line 28: One gap worth deciding on deliberately: CSRF is still enforced for callers who happen to carry a mit-learn session. Auth runs before permissions, so `AllowAny` doesn't stop `SessionAuthentication` from checking CSRF — anonymous posts skip it (as your description says), but a learner whose browser still has a session cookie gets 403 unless the request sends `X-CSRFToken`. Measured with `APIClient(enforce_csrf_checks=True)`: anonymous, no token → 201; `force_login`, no token → 403; same caller with the token → 201.

The drawer does handle this — it primes the CSRF cookie via `FEEDBACK_CSRF_PRIME_URL` and echoes it into the header — so in practice it only bites if there is a cookie-name or `CSRF_COOKIE_DOMAIN` change in any of the four repos involved (smoot-design, lehrer, ol-infrastructure, here). When it does, the same learner gets an intermittent 403.  Not sure if this is something that might happen often enough to handle here, but just thought it's worth mentioning.

content_feedback/views.py
```diff
@@ -18,14 +18,20 @@
...
 
     queryset = ContentFeedback.objects.all()
     serializer_class = ContentFeedbackSerializer
-    permission_classes = (IsAuthenticated,)
+    permission_classes = (AllowAny,)
     throttle_classes = (RedisScopedRateThrottle,)
     throttle_scope = "content_feedback"
```
mbertrand, line 30: Not something this PR broke, but it becomes load-bearing here: with anonymous as the primary path, this throttle is the only thing between the endpoint and unauthenticated write abuse, and I don't think the anon key holds up.

`ScopedRateThrottle.get_cache_key` falls back to `get_ident()` when there's no user, and `get_ident` reads `NUM_PROXIES` — which isn't set anywhere in the repo (grepped, zero hits), so it defaults to `None` and the key becomes the entire `X-Forwarded-For` chain. `config/nginx.conf.erb:53` uses `$proxy_add_x_forwarded_for`, which appends rather than replaces, so a client-supplied value survives into the key.

Measured with the rate dropped to `2/min`:

- 6 posts rotating `X-Forwarded-For: 203.0.113.{0..5}` → `[201, 201, 201, 201, 201, 201]`, all 6 rows written. No throttling at all.
- same 3 posts with a *fixed* XFF → `[201, 201, 429]`, so the throttle itself is working fine, it's purely the key.
- the key it builds, printed directly: `throttle_content_feedback_198.51.100.7,10.0.0.1`

Flip side of the same problem: 3 separate clients sending *no* XFF gave me `[201, 201, 429]` — they all key on the peer IP and share one 10/min bucket, so one noisy learner could throttle everyone.

Fix is probably `"NUM_PROXIES": <hop count>` in `REST_FRAMEWORK`, or override `get_ident` on `RedisScopedRateThrottle` to use ipware (already a dep, we use it in `authentication/middleware.py`) with a configured proxy count. Plain ipware defaults to leftmost-XFF and is spoofable too, so the hop count is the load-bearing part either way. Do you know how many proxy hops we actually have in front of this in prod? Happy for this to be a follow-up issue rather than blocking here, since it predates the PR.

content_feedback/views_test.py
```diff
@@ -26,11 +26,12 @@ def _payload(**overrides):
...
-def test_submit_requires_authentication(client):
-    """Anonymous users cannot submit feedback."""
+def test_submit_allows_anonymous(client):
+    """Anonymous users can submit feedback; the record has no user."""
     response = client.post(reverse("content_feedback:v0:content_feedback"), _payload())
-    assert response.status_code in (401, 403)
-    assert ContentFeedback.objects.count() == 0
+    assert response.status_code == 201
+    assert ContentFeedback.objects.count() == 1
+    assert ContentFeedback.objects.get().user is None
```
mbertrand, line 39: The `user is None` assertion is the right one to make here. But this doesn't pin down the CSRF half of the claim in the description: the `client` fixture is a bare `APIClient()` (`fixtures/users.py:64`), which defaults to `enforce_csrf_checks=False`, so this test passes whether or not CSRF is enforced for anonymous callers.

`APIClient(enforce_csrf_checks=True)` makes it a real assertion. I tried it and it does still return 201, so the behavior is right — it's just not currently being tested.

Two gaps I'd add while you're in here: a test that an authenticated session POST with no `X-CSRFToken` still 403s (that's the regression this change could most plausibly have caused, and it's the one in my `permission_classes` comment), and an anonymous throttle case with two different `X-Forwarded-For` values, which would surface the keying problem directly.

Also `test_submit_rate_limited`'s comment further down says the limit is "keyed per authenticated user" — still true of what that test covers, but now only half the picture.

mbertrand (changes requested):
Looks great overall, OpenAPI specs need to be regenerated.

content_feedback/views.py
zamanafzal: I think it's probably fine to leave the session-caller case as-is for now — it's the behavior we want, and the drawer already primes the cookie for the real path. The intermittent 403 would only surface if the cookie name or CSRF_COOKIE_DOMAIN drifted across the four repos, which isn't really something the view can guard against. Happy to revisit if you feel it's worth handling here.

mbertrand (approved):
👍

content_feedback/views.py
zamanafzal: Thank you for sharing this. I've created an issue to handle this https://github.com/mitodl/hq/issues/12775

zamanafzal:
It's now working as expected, and in case user mit-learn session is not available, his feedback will be stored anonymously. 
Steps followed:
I opened the [url](https://courses.rc.learn.mit.edu/learn/course/course-v1:MITxT+3.012Sx+3T2024/block-v1:MITxT+3.012Sx+3T2024+type@sequentia
l+block@40af5fa3fe714e07a872b2bfd004de0e/block-v1:MITxT+3.012Sx+3T2024+type@vertical+block@7fe893afc4044b18abef3ee484118f30) in an incognito window.

Click on the Sign in button, sign in, and submit the feedback. Feedback is submitted successfully without a user.

Screenshot 2026-08-12 at 2 34 55 PM
Screenshot 2026-08-12 at 2 37 08 PM

cc: @pdpinch

PR #3744, opened 2026-08-10
fix: add navbar for games

ahtesham-quraish:
https://github.com/mitodl/hq/issues/12717

- We are implemented cleaner design for hacksnack game for better user experience  

image
image

### How can this be tested?
For testing you need to install hacksnack dependency in learn and then run `yarn run dev` for running the game locally and then run the learn app using the following command `yarn run workspace dev` and then visit the following url `http://open.odl.local:8062/games/hacksnack#/demo?date=5-20-2026`

### DIFF
--- a/frontends/main/src/components/GameSubNav/GameSubNav.tsx
+++ b/frontends/main/src/components/GameSubNav/GameSubNav.tsx
@@ -1,0 +1,96 @@
+"use client"
+
+import React from "react"
+// next-nprogress-bar wraps next/navigation's router so navigation drives the
+// site's progress bar; it is what the rest of the app uses.
+import { useRouter } from "next-nprogress-bar"
+import { styled, Typography } from "ol-components"
+import type { TypographyProps } from "ol-components"
+import { RiArrowLeftSLine } from "@remixicon/react"
+
+const Bar = styled.div(({ theme }) => ({
+  display: "flex",
+  alignItems: "center",
+  justifyContent: "center",
+  padding: "16px 200px",
+  backgroundColor: theme.custom.colors.white,
+  boxShadow: "2px 2px 15px 0 rgba(0, 0, 0, 0.05)",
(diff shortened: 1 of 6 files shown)


frontends/main/src/components/GameSubNav/GameSubNav.tsx
```diff
@@ -0,0 +1,96 @@
...
+ */
+const GameSubNav: React.FC<GameSubNavProps> = ({ title }) => {
+  const router = useRouter()
+  return (
+    <Bar>
+      <BarInner>
+        <BackButton
+          type="button"
+          aria-label="Go back"
+          onClick={() => router.back()}
```
abeglova: should this go back? Or go to the main learn page/eventual game page. Just going to router.back() is weird if the user goes to the game directly through the url

package.json
```diff
@@ -31,7 +31,8 @@
   "resolutions": {
     "@types/react": "^19.0.0",
     "@types/react-dom": "^19.0.0",
-    "video.js": "^8.23.7"
+    "video.js": "^8.23.7",
+    "@mitodl/hacksnack": "https://github.com/mitodl/hacksnack.git#aq/board-design"
```
abeglova, line 35: Obviously this needs to not link to github before you merge. Also this duplicates the addition of @mitodl/hacksnack in frontends/main/package.json. hacksnack should only be added in one of those places

abeglova (changes requested):
Looks good with two minor comments

package.json
ahtesham-quraish: I will do it

frontends/main/src/components/GameSubNav/GameSubNav.tsx
ahtesham-quraish: Addressed

abeglova (approved):
LGTM but make sure to move hacksnack out of package.json before merging

PR #3747, opened 2026-08-10
Make apisix userinfo updates togglable

rhysyngsun:

Part of https://github.com/mitodl/hq/issues/10815

This makes the create/update userinfo settings configurable via environment variable. It's unlikely we'll ever set the create flag to False, but I added it just in case. Disabling updates, however, are critical in deployed environments where we update user data in Learn via SCIM because the APISIX userinfo is always stale unless the user _just_ logged in. So we need to disable updates for deployed environments so the middleware doesn't overwrite the SCIM update, because this makes it impossibly for the user to ever see the correct value from keycloak since the middleware updates the user info in learn at the start of the request.

Note: mit-learn doesn't use `mitol-django-apigateway` (we need to switch to this, but it's very involved because mit-learn's implementation has drifted so much), so this parallels work in https://github.com/mitodl/ol-django/pull/543.

### How can this be tested?

- In the main branch of Learn:
  - Login as a user in Learn
  - In a private window, login as an admin and change the first user's email in django admin
  - Refresh a frontend page in Learn as the first user and observe that the API responses reflect the old email
  - As the admin user, you should observe that the user's email reverted
- In this branch, this should be fixed:
  - Login as a user in Learn
  - In a private window, login as an admin and change the first user's email in django admin
  - Refresh a frontend page in Learn as the first user and observe that the API responses still reflect the changed email
  - As the admin user, you should observe that the user's email did not revert


README-keycloak.md
```diff
@@ -51,6 +51,23 @@ follow these steps:
 2. Add `DISABLE_APISIX_USER_MIDDLEWARE=True` to your `backend.local.env` file
 3. Set `COMPOSE_PROFILES=backend,frontend` in your .env file
 
+### Controlling user provisioning from APISIX headers
+
+By default, `ApisixUserMiddleware` both creates users it hasn't seen before and keeps
+existing users and their profiles in sync with the APISIX userinfo headers. Two settings
+control that, in `backend.local.env` (both default to `True`):
```
gumaerc: `MITOL_APIGATEWAY_USERINFO_UPDATE` defaults to `False` now.

env/backend.local.example.env
```diff
@@ -24,6 +24,11 @@ MAILGUN_SENDER_DOMAIN=open.odl.local
 MAILGUN_KEY=fake
 
 # APISIX/Keycloak settings
+# Set to False to stop provisioning users from the APISIX userinfo headers
+# MITOL_APIGATEWAY_USERINFO_CREATE=False
+# Set to True to enable syncing known users/profiles from the APISIX userinfo
+# headers (requires SCIM to keep them in sync instead)
```
gumaerc: I think maybe here it might be better to say "(requires SCIM to keep them in sync otherwise)?" The current language is a little confusing imo. At a glance this reads like SCIM is required if it's set to true.

gumaerc (approved):
Just a few minor comment nitpicks:

PR #3749, opened 2026-08-10
passing canvas_course_ids to scheduled celery task

shanbady:

Closes https://github.com/mitodl/hq/issues/12793

This PR resolves an issue where a signature mismatch with learning_resources.sync_canvas_courses caused the weekly scheduled task to fail

### How can this be tested?
1. checkout main
2. in settings_celery.py find the "sync_canvas_courses-every-1-weeks" entry and set it to the following (so it tries running every minute):
```
"sync_canvas_courses-every-1-weeks": {
            "task": "learning_resources.tasks.sync_canvas_courses",
            "schedule": crontab(minute="*/1"),
            "kwargs": {"overwrite": False},
        },
```
3. set CELERY_BEAT_DISABLED = False (in the same file)
4. restart celery and note that the task continuously fails with `celery.beat.SchedulingError: Couldn't apply scheduled task sync_canvas_courses-every-1-weeks: sync_canvas_courses() missing 1 required positional argument: 'canvas_course_ids'`
5. checkout this branch and set the above settings once again, restart celery and see the task executes.

### DIFF
--- a/learning_resources/tasks.py
+++ b/learning_resources/tasks.py
@@ -610,11 +610,13 @@ def ingest_edx_run_archive(
 
 
 @app.task(acks_late=True)
-def sync_canvas_courses(canvas_course_ids, overwrite):
+def sync_canvas_courses(canvas_course_ids=None, overwrite=False):  # noqa: FBT002
     """
     Sync all canvas course files
 
     Args:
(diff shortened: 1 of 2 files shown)


abeglova (approved):
👍

PR #3758, opened 2026-08-11
Sitemap fix, 1 of 2: Add canonical_parent_ids and resource_type to the summary endpoint

ChristopherChudzicki:
- For https://github.com/mitodl/hq/issues/12814

Adds `canonical_parent_ids` and `resource_type` to the `api/v1/learning_resources/summary/` endpoint for use in podcast episode + video pages

### How can this be tested?
1. Verify that https://api.learn.mit.dev/api/v1/learning_resources/summary/?resource_type=podcast_episode, https://api.learn.mit.dev/api/v1/learning_resources/summary/?resource_type=video include `canonical_parent_ids`

### DIFF
--- a/learning_resources/views.py
+++ b/learning_resources/views.py
@@ -202,7 +202,7 @@ class SummaryPagination(LargePagination):
     """
 
     def get_count(self, queryset):
-        """Count distinct pks, which the annotations play no part in"""
+        """Count distinct pks; .values() drops the annotation, .only() would not"""
         return queryset.values(*self.count_fields).distinct().count()
 
 
--- a/learning_resources/views.py
+++ b/learning_resources/views.py
@@ -192,6 +192,20 @@ class BaseLearningResourceViewSet(viewsets.ReadOnlyModelViewSet):
         return super().list(request, *args, **kwargs)
 
 
+class SummaryPagination(LargePagination):
+    """
+    LargePagination that keeps annotations out of the count query.
+
+    Django keeps annotations in the count of a distinct queryset, so
+    canonical_parent_ids would be evaluated once per row counted. Counting
+    distinct pks is the same number for a fraction of the work.
+    """
+
+    def get_count(self, queryset):
+        """Count distinct pks, which the annotations play no part in"""
+        return queryset.values(*self.count_fields).distinct().count()
+
+
 @extend_schema_view(
     list=extend_schema(
         description="Get a paginated list of learning resources.",
@@ -337,7 +351,15 @@ class LearningResourceViewSet(
         detail=False,
         methods=["GET"],
         name="Get learning resources summary",
-        pagination_class=LargePagination,
+        pagination_class=SummaryPagination,
+    )
+    @method_decorator(
+        # no timeout: REDIS_VIEW_CACHE_DURATION is then read per request, so
+        # setting it to 0 disables this cache
+        cache_page_for_anonymous_users(
+            cache="redis",
+            key_prefix="learning_resources_summary",
+        )
     )
     def summary(self, request, **kwargs):  # noqa: ARG002
         """
@@ -351,7 +373,9 @@ class LearningResourceViewSet(
             # we don't use `self.get_queryset()` here because there are incomplatible
             # `select_related()` invocations and we don't need related data anyway
             LearningResource.objects.filter(published=True)
-            .only("id", "last_modified", "url", "title")
+            # a deferred field the serializer reads costs a query per row
+            .only("id", "last_modified", "url", "title", "resource_type")
+            .with_canonical_parent_ids()
             .distinct()
             # Deterministic order so offset pagination has stable page
             # boundaries.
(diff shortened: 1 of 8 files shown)


learning_resources/views.py
```diff
@@ -337,7 +351,15 @@ def vector_similar(self, request, *_, **kwargs):
...
-        pagination_class=LargePagination,
+        pagination_class=SummaryPagination,
+    )
+    @method_decorator(
+        # no timeout: REDIS_VIEW_CACHE_DURATION is then read per request, so
+        # setting it to 0 disables this cache
+        cache_page_for_anonymous_users(
+            cache="redis",
+            key_prefix="learning_resources_summary",
+        )
```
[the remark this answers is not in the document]
ChristopherChudzicki: This came up in local review. It's also true afaik of all our cached endpoints. @abeglova or @shanbady can you advise if this is worth worrying about?

learning_resources/views.py
```diff
@@ -192,6 +192,20 @@ def list(self, request, *args, **kwargs):
...
+    LargePagination that keeps annotations out of the count query.
+
+    Django keeps annotations in the count of a distinct queryset, so
+    canonical_parent_ids would be evaluated once per row counted. Counting
+    distinct pks is the same number for a fraction of the work.
+    """
+
+    def get_count(self, queryset):
+        """Count distinct pks, which the annotations play no part in"""
+        return queryset.values(*self.count_fields).distinct().count()
```
ahtesham-quraish, line 206: Consider adding a one-line note on why this uses .values() rather than the base class's .only().

The base DefaultPagination.get_count uses queryset.only(*self.count_fields).count(), but .only() only defers concrete model columns — it does not strip the canonical_parent_ids annotation, so left alone the subquery would still be evaluated per counted row. Switching to .values(*self.count_fields).distinct() is what actually removes the annotation from the SELECT list (Django keeps only the fields named in .values()), so COUNT(DISTINCT ...) runs over just pk.

That behavior is the entire reason this class exists. It's already covered by test_summary_count_omits_the_parent_ids_annotation (which asserts "canonical_parent_ids" not in count_sql), so a revert to .only() would fail CI rather than break silently — but a short inline comment like # .values("pk") — not .only() — is required to drop the annotation from the count query would make that intent explicit at the call site without having to go find the test. Non-blocking.

learning_resources/views.py
```diff
@@ -337,7 +351,15 @@ def vector_similar(self, request, *_, **kwargs):
         detail=False,
         methods=["GET"],
         name="Get learning resources summary",
-        pagination_class=LargePagination,
+        pagination_class=SummaryPagination,
+    )
+    @method_decorator(
+        # no timeout: REDIS_VIEW_CACHE_DURATION is then read per request, so
+        # setting it to 0 disables this cache
+        cache_page_for_anonymous_users(
```
ahtesham-quraish: ### Question:

This action is now cached under key_prefix="learning_resources_summary", and the cache is invalidated by clear_views_cache() (called with no prefix, so it flushes views.*). That works for the ETL tasks that call it — get_podcast_data, get_ovs_data, the edX/MITx/OLL/xPro pipelines, etc.

The gap is the video/YouTube path: get_youtube_channel_data and get_youtube_playlist_data (learning_resources/tasks.py) load playlists and their video memberships via loaders.load_playlist, but none of them calls clear_views_cache(). Note that load_playlist does call update_index(...) after writing the memberships, so the search index stays fresh — the gap is specifically the Redis view cache, which nothing on the YouTube path touches. test_cache_is_cleared_after_task_run also explicitly documents that get_youtube_data is excluded because "it only queues the fan-out."

Since a video's canonical_parent_ids is exactly its playlist membership, a YouTube ETL run that adds/reorders a video's playlists won't invalidate this newly-cached summary page — it will keep serving stale parent ids until an unrelated ETL task that does call `clear_views_cache()` next runs.".

Given this PR exists specifically so the sitemaps can rely on canonical_parent_ids, can you confirm the staleness window is acceptable, or add a clear_views_cache() at the point the playlist loader finishes writing? load_playlist (learning_resources/etl/loaders.py) is the natural spot since it already does post-write index maintenance there.

ahtesham-quraish (changes requested):
Left few comments

learning_resources/views.py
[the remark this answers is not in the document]
ChristopherChudzicki: moot; decided to remove caching changes from this PR.

learning_resources/views.py
ChristopherChudzicki: Clarified the comment.

learning_resources/views.py
ChristopherChudzicki: Decided to remove caching from this PR to keep it simpler. That said, the issue you point out affects regular ETL tasks, and maybe is worth addressing at some point. 

 (Not sure about putting the cache clear in etl/loaders, though; that would be a first for the repo).

ahtesham-quraish (approved):
Approved! Please rebase it

PR #3765, opened 2026-08-12
Order topics deterministically in serialization queryset

mbertrand:
Closes https://github.com/mitodl/hq/issues/12833

`LearningResourceTopic` had no `Meta.ordering`, so topics came back in whatever order Postgres chose. `test_serialize_bulk_learning_resources` builds its expected values from one query and the serializer builds its results from another, so the two disagreed intermittently. Adding `ordering = ["name"]` to the model makes topic order deterministic in the test and in API responses — same thing `LearningResourceDepartment` already does.

### How can this be tested?

```bash
docker compose run --rm web uv run pytest \
  learning_resources_search/serializers_test.py::test_serialize_bulk_learning_resources \
  --count=50 -q --no-cov
```

Should pass every time


zamanafzal (approved):
LGTM 👍🏻

PR #3766, opened 2026-08-12
feat: Collect required compliance fields before any enrollment or checkout

daniellefrappier18:

Closes https://github.com/mitodl/hq/issues/12713

MITx Online  now reports which export-compliance (OFAC) fields a profile is missing, via `compliance_missing_fields` on `users/me`. This drives the just-in-time dialog from that field instead of the hardcoded country/year-of-birth check, and applies it to **every** path that enrolls or adds to the basket rather than only B2B enrollment.

**Questions - Open for Discussion**

**Enrollment-code redemption is ungated. Is that the behavior we want?**
Redeeming an enrollment code is ungated. It only attaches the user to the org/contract, with no compliance check. But that attachment alone has no course/program consequence because actual enrollment happens separately via useEnrollmentHandler's isB2B branch, which does run the compliance gate. So there's no path today where a user ends up in a course or program without hitting the gate.

Is that sequencing fine as-is, or do we want to require compliance before the code can even be redeemed, rather than deferring it to enrollment time?

**The dialog**

- Renders the whole compliance set **prefilled** from the existing profile
- Email is displayed **read-only** 
- State and postal code become required reactively from the **selected** country
- Postal code is hidden entirely — not just optional — unless the selected country is US or Canada, and its label switches between "Zip Code" (US) and "Postal Code" (Canada)
- Yup schema is built from the same `isFieldRequired` predicate the inputs use for their `required` attribute

**Year of birth stays separate.** It backs MIT's minimum-age rule rather than export compliance, and is deliberately absent from `compliance_missing_fields` (that list is CyberSource bill-to shaped). It's read from `user_profile`, checked Learn-side, and still written back on submit.

**Where the gate lives**

- Paid and checkout paths are gated centrally inside `useReplaceBasketItem`, so a new checkout entry point can't skip the check.
- Free and verified-program enrollments call the gate directly

**Forward-compatible with [mitxonline#3850](https://github.com/mitodl/mitxonline/pull/3850)** (not yet in production). That PR makes `compliance_missing_fields` flag `postal_code`/`state` as missing whenever they're empty, for *every* country — not just US/CA, unlike the CyberSource check itself, which stays US/CA-only. Since the dialog only ever collects those two fields for US/CA, gating on them everywhere would trap every non-US/CA user in a dialog they can never satisfy. The gate now disregards `postal_code`/`state` entries in `compliance_missing_fields` unless the stored country actually requires them, so nothing further is needed here once #3850 ships — see the Additional Context checklist below for the one manual check worth doing once it's live.

**Also in here**

- Gates the dashboard **"Upgrade for certificate"** action — a fifth entry point not listed on the ticket, which reaches enrollment without touching any of the paths that were.
- Bumps `@mitodl/mitxonline-api-axios` to `2026.8.12` for the new field. That bump also made `Order.refund_eligible` required

- [ ] Desktop screenshots

Blank dialog (compliance-empty user)
Screenshot 2026-08-14 at 8 43 08 AM

Prefilled dialog (compliance-partial user) showing prefilled state select
State/Postal appearing after selecting United States or Canada 
No Country = No Postal Code Field Shown 
Screenshot 2026-08-14 at 8 44 47 AM

Canada = Postal Code Label 
Screenshot 2026-08-14 at 8 44 35 AM

United States = Zip Code Label 
Screenshot 2026-08-14 at 8 44 18 AM

B2B with no profile on file 
Screenshot 2026-08-14 at 8 53 46 AM

Validation errors on empty submit
Screenshot 2026-08-14 at 8 59 49 AM

### How can this be tested?

These instructions assume you are using the **Tilt / k3d stack in ol-infrastructure** 

#### 1. Seed the test data

The seed script isn't committed — grab it from this gist: 
https://gist.githubusercontent.com/daniellefrappier18/55a665a8ced15b0e133aa41beaf320cc/raw/c26ae56554723fc6b88fb53b945b9a9d6431591b/gistfile1.txt

Drop it into your local mitxonline checkout (Tilt live-syncs it into the pod) and run it:

```bash
cd ~/path/to/mitxonline
curl -sL  https://gist.githubusercontent.com/daniellefrappier18/55a665a8ced15b0e133aa41beaf320cc/raw/c26ae56554723fc6b88fb53b945b9a9d6431591b/gistfile1.txt\
  -o compliance/management/commands/seed_compliance_scenarios.py

kubectl exec -it -n mitxonline deploy/mitxonline-webapp -c app -- \
  python manage.py seed_compliance_scenarios
```

It will show up as untracked in mitxonline — don't commit it.

It creates four users (in **Keycloak and Django**, linked — login matches on `global_id`, so a Django-only user can't sign in), a paid+free course, a program with two required courses, products for each, CMS pages, an audit enrollment for the upgrade scenario, and a dedicated B2B org/contract/course/program with the fourth user pre-attached to it. Re-running resets every profile, which is the quickest way back to a blocked state after completing the dialog once.

| User | Password | Profile on file | Reported missing |
| --- | --- | --- | --- |
| `compliance-empty@odl.local` | `localdev123` | nothing, no year of birth | `city, country, first_name, last_name, street_address_1` |
| `compliance-partial@odl.local` | `localdev123` | Ada Lovelace · US · MA · Cambridge · 1988 | `postal_code, street_address_1` |
| `compliance-upgrade@odl.local` | `localdev123` | nothing, audit-enrolled in the upgrade course | `city, country, first_name, last_name, street_address_1` |
| `compliance-b2b@odl.local` | `localdev123` | nothing, pre-attached to the B2B contract | `city, country, first_name, last_name, street_address_1` |

Seeded courseware:

- `/courses/course-v1:MITxT+JIT-CHECKOUT` — $100, audit + verified
- `/programs/program-v1:MITxT+JIT-PROGRAM` — $250, two required courses
- `course-v1:MITxT+JIT-UPGRADE` — the audit-enrolled course, for the dashboard upgrade link
- `/dashboard/organization/jit-compliance-b2b-org/contract/jit-compliance-b2b-contract` — dedicated B2B contract with one course, for the `isB2B` enrollment branch

#### 2. One rule that shapes the testing

**The gate is per user, not per path.** Completing the dialog once makes that user compliant everywhere, so they stop triggering it. One user therefore verifies one **submit** before needing a reseed. **Cancelling costs nothing**, so all the appearance / prefill / validation checks can be repeated freely.

Note: Checkout will redirect to CyberSource's test acceptance page and fail with "You are not authorized to view this page" locally. That is to be expected without real CyberSource sandbox credentials. The compliance gate this PR adds runs entirely before that redirect, so it's fully testable up to that point.

#### 3. Free to repeat (cancel at the end of each)

| # | User | Do this | Expect |
| --- | --- | --- | --- |
| 1 | `empty` | Course page → paid CTA | dialog instead of the cart |
| 2 | `empty` | Course page → "Start Learning" | dialog instead of enrolling |
| 3 | `empty` | Program page → "Earn Certificate" | dialog instead of the cart |
| 4 | `empty` | Program page → "Start Learning" | dialog instead of enrolling |
| 5 | `upgrade` | Dashboard → "Upgrade for certificate" | dialog instead of upgrading |
| 6 | `upgrade` | Cancel the dialog from the upgrade link | **no** "There was a problem upgrading your enrollment" — a cancel isn't a failure |
| 7 | `empty` | Cancel from a paid CTA, then check the cart and network tab | nothing added, no redirect, no `PATCH users/me` |
| 8 | `empty` | Inspect the blank form | First/Last Name, Email, Country, Address, Address Line 2 (optional), City, Year of Birth. No State, and no Zip/Postal Code field at all, until a country that needs them is selected |
| 9 | `empty` | Try to edit Email | read-only, not disabled — still focusable and announced |
| 10 | `empty` | Submit everything empty | required errors on First name, Last name, Country, Address, City, Year of birth; no `PATCH` |
| 11 | `empty` | Type three spaces in First Name, submit | still rejected — whitespace isn't a value |
| 12 | `empty` | Submit with errors | focus lands on the first invalid field |
| 13 | `partial` | Open the dialog | Ada / Lovelace / United States / **Massachusetts** / Cambridge / 1988 prefilled; Address and Zip Code empty (labeled "Zip Code", not "Postal Code" — country is US) |
| 14 | `partial` | Change a prefilled value (e.g. City → Boston) | editable, not locked — wrong data on file can be corrected |
| 15 | `partial` | Clear a prefilled required field, submit | rejected like any empty required field |
| 16 | `empty` | Select United Kingdom | no State field; no Zip/Postal Code field either — hidden entirely, not just optional |
| 17 | `empty` | Select United States | State appears; a required field labeled **Zip Code** appears too |
| 18 | `empty` | Select Canada | State becomes the province list; a required field labeled **Postal Code** appears (not "Zip Code") |
| 19 | `empty` | US → pick Massachusetts, type a Zip Code → UK → back to US | State reset to "Please Select" and Zip Code cleared, not stale from before |
| 20 | `empty` | With a screen reader, change country to US | "State/Province and Zip Code are required for United States." announced |
| 21 | `empty` | Keyboard only: tab through, operate both selects, submit, Escape | all reachable, focus visible, Escape closes |
| 22 | `b2b` | Contract dashboard (seeded URL above) → click enroll on the course card | dialog instead of enrolling |
| 23 | `b2b` | Cancel the dialog from the contract dashboard | back on the dashboard, no enrollment, no `PATCH users/me` |

#### 4. One submit per user, reseed between rounds

The recurring assertion: after saving, **the action you originally clicked carries on by itself** — no clicking Enroll twice.

**Round 1**

| User | Do this | Expect |
| --- | --- | --- |
| `empty` | Course page → paid CTA → fill → Submit | dialog closes and you land in the cart holding the $100 product |
| `partial` | Dashboard → "Upgrade for certificate" → add Address + Zip Code → Submit | upgrade proceeds, no error banner |
| `upgrade` | Course page → "Start Learning" → fill → Submit | enrollment created, dashboard success state |
| `b2b` | Contract dashboard → click enroll on the course card → fill → Submit | isB2B enrollment created; you land on the course's courseware URL, not a "problem enrolling" error |

Reseed.

**Round 2**

| User | Do this | Expect |
| --- | --- | --- |
| `empty` | Program page → "Earn Certificate" → fill → Submit | cart holds the $250 program product |
| `partial` | Program page → "Start Learning" → fill → Submit | program enrollment created |
| `upgrade` | Dashboard → "Upgrade for certificate" → fill the **blank** form → Submit | same result as the prefilled run in round 1 |

### Additional Context

- **After [mitxonline#3850](https://github.com/mitodl/mitxonline/pull/3850) reaches production** — do one manual check: a non-US/CA account with no postal code should *not* get re-gated on every enroll attempt (confirms the `compliance_missing_fields` filtering works against real backend data, not just a local deployment that predates #3850), and a US/CA account should behave exactly as it does today.


frontends/main/src/common/mitxonline/useComplianceGate.ts
```diff
@@ -0,0 +1,79 @@
...
+ * `useMutation` contract has no way to express "cancelled" that does not either
+ * reject (which reads as a failure to the caller) or resolve as if the
+ * enrollment had happened.
+ */
+const useComplianceGate = () => {
+  const queryClient = useQueryClient()
+  // Collapses concurrent callers (e.g. a double-clicked enroll button) onto a
+  // single check, so a second click can't slip through and start its own
+  // mutation while the first is still awaiting the profile fetch or dialog.
+  const inFlight = React.useRef<Promise<boolean> | null>(null)
```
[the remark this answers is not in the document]
daniellefrappier18: Moving to shared module/provider state to fix that would add more complexity. The scope is intentionally per-instance

pdpinch:
Why are we presenting email address here? It's not like the user can change it from this screen.

pdpinch:
Does Cybersource require the address be split across lines?

gumaerc:
> Why are we presenting email address here? It's not like the user can change it from this screen.

This was going to be my first question in the review, but because you mentioned it I will boost it now. I'm not sure why we'd display this here either. Firstly, it plays our hand so to speak in that we are collecting information for a compliance check. If a bad actor were actually trying to create an account this would tip them off. Realistically a form asking for all this information has kind of already done that, but no need to make it extra obvious. Besides, the email isn't required for the profile update API endpoint. MITx Online just pulls that from your session. You need an email to create an account (and it's resynced from Keycloak anyway) so it can never be in `compliance_missing_fields`. I'd say just remove anything in the form to do with email. This is what AI had to say about what should be cleaned up surrounding it:

```
 JustInTimeDialog.tsx                                                                                                                                                                                                                                                   
 - 156–167 — the spec.kind === "readonly" branch. This is the rendered field and the help text.                                                                                                                                                                         
 - 107 — the field !== "email" && guard and as EditableJitField cast in the focus effect                                                                                                                                                                                
 - 169 — const name = field as EditableJitField cast                                                                                                                                                                                                                    
 - 33 — the EditableJitField import                                                                                                                                                                                                                                     
                                                                                                                                                                                                                                                                        
 Both casts exist only because email had to be excluded from the editable set; once it's gone, JitField and the editable set are the same type and the casts are unnecessary.                                                                                           
                                                                                                                                                                                                                                                                        
 complianceFields.ts                                                                                                                                                                                                                                                    
 - 43 — "email" in JIT_FIELDS (the entry that drives rendering)                                                                                                                                                                                                         
 - 79 — the FIELD_SPECS.email entry                                                                                                                                                                                                                                     
 - 71 — the "readonly" member of the kind union, plus the doc line at 69 describing it — email was its only user                                                                                                                                                        
 - 56–61 — the doc comment and EditableJitField = Exclude<JitField, "email">. This collapses to an identity alias, so it's cleanest to drop it and use JitField at its six references (63, 221, 227, 236, 250, 341)                                                     
 - 283 — the {@link EditableJitField} reference in the jitPatchPayload docs                                                                                                                                                                                             
                                                                                                                                                                                                                                                                        
 Keep line 19 — "email" in COMPLIANCE_FIELDS. That mirrors mitxonline's BILL_TO_FIELD_TO_PROFILE_FIELD value set and stays accurate regardless of what the dialog draws. 
```

Your call on whether or not to remove the "read only field" machinery. I'm not sure if we have a need for it now though.

gumaerc:
@daniellefrappier18 I've been doing manual testing on this today and things are working well so far aside from what I noted above. I have some other things I want to investigate that I will get back to first thing Monday.

daniellefrappier18:
@gumaerc @pdpinch COMPLIANCE_FIELDS mirrors the value set of mitxonline's BILL_TO_FIELD_TO_PROFILE_FIELD, and email is one of those eight bill-to fields. The dialog was built to render the whole compliance set, prefilled, rather than only the reported-missing subset (so wrong-but-present data can be corrected). Email came along as a member of that set. That being said I'm not wedded to keeping it so I'm happy to remove it, since like you said the user can not update it.

daniellefrappier18:
> Does Cybersource require the address be split across lines?

@pdpinch No, CyberSource doesn't require the split. Only `address1` is required for the export compliance check. `address2 `is optional additional info, and on the payment side CyberSource concatenates the two anyway. The second line comes from mitxonline's `LegalAddress`, which stores `street_address_1` and s`treet_address_2 `as separate columns, so the form is mirroring our storage rather than anything CyberSource asks for.

Which means we can drop it. Removing the field and omitting it from the PATCH leaves any stored value untouched, however there is the one caveat is that anyone who already has an "Apt 4B" on file keeps it but can no longer see or edit it in this dialog. That seems like the right trade for a just-in-time form whose only job is unblocking enrollment; the full profile page is the place to change it.

Happy to collapse it to a single Address field if you'd prefer.

rhysyngsun:
FYI - we have the multiple address fields because cybersource caps them at 60 characters each, so the extra one is in case 60 chars isn't enough.

frontends/main/src/page-components/EnrollmentDialogs/complianceFields.ts
```diff
@@ -0,0 +1,345 @@
...
+    // meaningless under a country that has no subdivisions, so a
+    // previously-stored value is dropped rather than carried over.
+    state: requiresSubdivision(values.country) ? values.state : "",
+    postal_code: requiresSubdivision(values.country)
+      ? values.postal_code.trim()
+      : "",
+  }
+  const yearOfBirth = Number.parseInt(values.year_of_birth, 10)
+  const payload: PatchedUserRequest = { legal_address: legalAddress }
+  if (yearOfBirth !== currentYearOfBirth) {
```
gumaerc, line 292: I did some research into `addl_field_flag` on the MITx Online side, and it seems that it you are correct in that the value gets clobbered. However, it seems the value is completely unused. AI flagged this and I did a bit of manual research and it seems to be correct. There could be some kind of edge case here but it seems to be a case of the comments about it being outdated in MITx Online, therefore AI hallucinated that this was a situation worth handling.

 The only application read is EditProfileForm.js:28, which loads it as a Formik initial value and submits it straight back — and EditProfilePage.js:49 overwrites it to true on submit regardless. No Django admin, no HubSpot sync, no exports.

 It used to be read, and two stale comments still claim it is. DashboardPage.js:158-176 and ProgramProductDetailEnroll.js:94-110 both open with:

 ```js
   /*
   If we've got addl_field_flag, then display the extra info modal. Otherwise,
   send the learner directly to the page.
   */
 ```

 That was accurate when written. Before commit 01e296b09 (https://github.com/mitodl/mitxonline/pull/2611, 2025-05-05, "add country and year of birth to the dashboard modal") the condition was literally the flag:

 ```js
   // BEFORE #2611
   if (currentUser.user_profile && currentUser.user_profile.addl_field_flag) {
     return
   }
 ```

https://github.com/mitodl/mitxonline/pull/2611 replaced it with a direct data check in both files and left the comments untouched:

 ```js
   // AFTER #2611 — current main
   if (
     currentUser &&
     currentUser.legal_address &&
     currentUser.legal_address.country !== "" &&
     currentUser.legal_address.country !== null &&
     currentUser.user_profile &&
     currentUser.user_profile.year_of_birth !== "" &&
     currentUser.user_profile.year_of_birth !== null
   ) {
     return
   }
 ```

frontends/main/src/common/mitxonline/useComplianceGate.ts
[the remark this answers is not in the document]
gumaerc: Claude flagged this for me as well, but for a different reason. It claimed that it was possible on the product page to click the audit enrollment button, then quickly click the verified enrollment button before the popup shows. I had to enable "slow 4G" throttling to reproduce it, but I was able to. What happened was I got an audit enrollment, and also had the verified track added to a cart. So, when I filled out the missing profile fields dialog, I was redirected to the cart in MITx Online with the verified track in it. We haven't actually created a verified enrollment at that point, but it could be potentially confusing for a user. Conceivably, someone could accidentally tap both on their phone. It's not likely to happen though in any normal situation though, as the dialog will come up quickly enough that it doesn't matter. It would be a similar situation with what Copilot describes here; most of the time everything is going to happen quickly enough that a user isn't going to be able to trigger the dialog through two cards before it pops up. I think the fix is pretty straightforward though, and because the dialog initialization was changed as part of this PR, I think it makes sense to scope the change here. I asked Claude to summarize:

 On "straightforward" — one caveat worth flagging first. Copilot suggests "shared module/provider state," and the module version is a trap. I tried it: it does fix the race, but it breaks three tests in UnenrolledCourseCard.test.tsx, and it's not test flakiness. Those tests open the dialog and never close it, so NiceModal.show()'s promise never settles and the module-level value stays pinned. The same thing happens in the browser if someone opens the dialog and navigates away instead of closing it —
 from then on every enroll click silently does nothing until a reload. That's worse than the race we're fixing.

 The provider version avoids that, because the state dies with the tree. I prototyped it and it's small:

 ```diff
   --- a/frontends/main/src/common/mitxonline/useComplianceGate.ts
   +++ b/frontends/main/src/common/mitxonline/useComplianceGate.ts

   +/**
   + * Holds the single in-flight compliance check for the whole tree.
   + *
   + * It has to be shared rather than per-hook: a product page has two gate
   + * instances (the free-enrollment gate, plus the one inside
   + * `useReplaceBasketItem`) and the dashboard has one per card. `NiceModal.show`
   + * returns the *same* promise when the dialog is already open, so without
   + * sharing, each instance awaits one dialog and each resumes its own mutation
   + * from a single submit.
   + *
   + * A provider rather than module-level state, so the ref cannot outlive the
   + * tree. A dialog left unresolved (user navigates away instead of closing it)
   + * leaves the promise pending forever; module state would pin it for the rest
   + * of the page session and silently make every later enroll click a no-op.
   + */
   +const ComplianceGateContext =
   +  React.createContext<React.MutableRefObject<Promise<boolean> | null> | null>(
   +    null,
   +  )
   +
   +const ComplianceGateProvider: React.FC<{ children: React.ReactNode }> = ({
   +  children,
   +}) => {
   +  const inFlight = React.useRef<Promise<boolean> | null>(null)
   +  return React.createElement(
   +    ComplianceGateContext.Provider,
   +    { value: inFlight },
   +    children,
   +  )
   +}
   +
    const useComplianceGate = () => {
      const queryClient = useQueryClient()
   -  // Collapses concurrent callers (e.g. a double-clicked enroll button) onto a
   -  // single check, so a second click can't slip through and start its own
   -  // mutation while the first is still awaiting the profile fetch or dialog.
   -  const inFlight = React.useRef<Promise<boolean> | null>(null)
   +  // Falls back to a per-instance ref when no provider is mounted, so a gate
   +  // still works in isolation -- it just can't collapse calls from sibling
   +  // instances. Mount ComplianceGateProvider alongside NiceModal's provider.
   +  const shared = React.useContext(ComplianceGateContext)
   +  const local = React.useRef<Promise<boolean> | null>(null)
   +  const inFlight = shared ?? local

   @@
        inFlight.current = promise
        return promise
   -  }, [queryClient])
   +  }, [queryClient, inFlight])

   -export { useComplianceGate }
   +export { useComplianceGate, ComplianceGateProvider }
 ```

 Then mount it next to NiceModalProvider in the two places that have one — app/providers.tsx and TestProviders in test-utils/index.tsx:

 ```diff
        <ThemeProvider>
   -      <NiceModalProvider>{children}</NiceModalProvider>
   +      <NiceModalProvider>
   +        <ComplianceGateProvider>{children}</ComplianceGateProvider>
   +      </NiceModalProvider>
        </ThemeProvider>
 ```

 With that in place, two dashboard cards clicked before the dialog appears produce one enrollment instead of two, all 999 tests in the affected suites pass, and typecheck and lint are clean.

 A few things I left rough, since this was just to prove it out:

 - React.createElement rather than JSX because the file is .ts — might be nicer to rename it .tsx.
 - The context fallback is lenient: no provider means gates quietly degrade to per-instance. Keeps isolated renders working, but a future refactor could drop the provider and reintroduce this silently. Throwing instead would be defensible.
 - Worth adding a regression test for it — two cards, both clicked before the dialog shows, assert a single enrollment. None of the current tests cover that, since they all exercise one gate instance.

frontends/main/src/app-pages/DashboardPage/CoursewareDisplay/hooks/useEnrollmentHandler.ts
gumaerc: Claude pointed out that this is now dead code. The query and where it returns below can be removed.

frontends/main/src/page-components/EnrollmentDialogs/complianceFields.ts
```diff
@@ -0,0 +1,345 @@
...
+} from "@mitodl/mitxonline-api-axios/v2"
+
+/**
+ * Profile field names MITx Online can report in `User.compliance_missing_fields`.
+ *
+ * This is the value set of `BILL_TO_FIELD_TO_PROFILE_FIELD` in mitxonline's
+ * `compliance/api.py` — CyberSource bill-to fields mapped back to profile field
+ * names so a consumer can render them.
+ */
+const COMPLIANCE_FIELDS = [
```
gumaerc: This and `ComplianceField` below don't seem to be referenced anywhere else and can be deleted.

gumaerc (commented):
I was able to successfully test all of this manually. I'll approve it once the email field is removed as discussed above, but I noted a few more things below. One has to do with some comments surrounding `addl_field_flag` and a gate built around not clobbering it, as well as an expansion on what copilot flagged with the inFlight check. None of the stuff I highlighted below is blocking, but would be good to address.

frontends/main/src/page-components/EnrollmentDialogs/complianceFields.ts
daniellefrappier18: removed

frontends/main/src/app-pages/DashboardPage/CoursewareDisplay/hooks/useEnrollmentHandler.ts
daniellefrappier18: Yup, this is dead - I removed it

frontends/main/src/common/mitxonline/useComplianceGate.ts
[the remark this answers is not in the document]
daniellefrappier18: Moved to a ComplianceGateProvider shared ref like in your prototype and added a regression test in useComplianceGate.test.tsx covering two cards clicked before the dialog appears.

frontends/main/src/page-components/EnrollmentDialogs/complianceFields.ts
daniellefrappier18: Yup, you are correct `addl_field_flag` has no live consumer, but the guard costs nothing and stays correct if mitxonline ever reads it again, so I'm going to leave this since it's not causing any issues.

daniellefrappier18:
@steven-hatch tagging you if you wanted to take a look at the screenshots provided. I implement the zip/postal code logic we talked about last week.

frontends/main/src/page-components/EnrollmentDialogs/complianceFields.ts
```diff
@@ -0,0 +1,313 @@
...
+ * Two entries are not export-compliance fields:
+ *  - `street_address_2` is never required, but belongs to an address form.
+ *  - `year_of_birth` backs MIT's minimum-age requirement, which Learn enforces
+ *    itself. It is deliberately absent from `compliance_missing_fields`: that
+ *    list is CyberSource-shaped and read-only, and year of birth is not a
+ *    bill-to field. We read it straight off `user_profile` instead.
+ *
+ * Email is deliberately not collected here: it comes from the user's SSO
+ * identity, so editing it would silently diverge the MITx Online account from
+ * Keycloak, and the dialog can't change it anyway.
```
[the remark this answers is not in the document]
gumaerc: Not Copilot completely ignoring the detailed discussion above about why it was removed...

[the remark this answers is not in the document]
daniellefrappier18: Haha. Read the room copilot. In fairness though I'll update the PR description

gumaerc:
@daniellefrappier18 It looks like this branch needs a rebase

daniellefrappier18:
> @daniellefrappier18 It looks like this branch needs a rebase

@gumaerc all set now

gumaerc (approved):
👍 LGTM!

pdpinch:
Late comment acknowledging that I reconsidered and decided that two address lines are fine.

PR #3773, opened 2026-08-14
feat(b2b): render the analytics dashboard scoped to a contract

blarghmatey:

N/A — no GitHub issue. Tracked in the work graph as `tk-point-the-mit-learn-analytics-dashboard-at-the-c-d9d6e1`. Part of the B2B self-serve analytics effort ([mitodl/hq discussion #11845](https://github.com/mitodl/hq/discussions/11845)).

**This is the last of three PRs and cannot work alone. Draft until the other two land.** All three must ship, in this order:

| # | PR | What it adds | Why it blocks the next |
|---|----|--------------|------------------------|
| 1 | [mitodl/ol-data-platform#2559](https://github.com/mitodl/ol-data-platform/pull/2559) | `contract_id` on four MVs; two new contract-grained MVs | Without it the API's `SELECT` names a column that does not exist |
| 2 | [mitodl/ol-analytics-api#34](https://github.com/mitodl/ol-analytics-api/pull/34) | `/organizations/{org}/contracts/{contract}/…` | Without it every request from this PR 404s |
| 3 | **this PR** | the dashboard that calls those routes | — |

Between 1 and 2 there is a **deploy step, not just a merge**: the next `b2b_analytics_starrocks_job` run has to rebuild the materialized views. That is automatic (ol-data-platform#2554's drift detection, verified in production 2026-08-14) but it is not instant — merging 1 and deploying 2 in the same window will fail.

The dashboard was org-scoped at every layer while [MITx Online's manager dashboard](https://github.com/mitodl/mitxonline/blob/main/b2b/views/v0/urls.py) it mirrors is contract-scoped, so a manager clicking "View analytics" from a contract page landed on org-wide numbers.

- New route `dashboard/organization/[orgSlug]/contract/[contractSlug]/analytics`, alongside the existing org-level one.
- `analyticsContractsApi` + `analyticsContractQueries`, alongside the org versions. The org endpoints are staying, so nothing currently deployed changes shape.
- `AnalyticsContent` takes an optional `contractSlug` and picks the scope once. Everything below that choice is untouched — both scopes return the identical envelope and row shapes.
- The "View analytics" button now links to the contract-scoped view, and the comment above it explaining why analytics was org-scoped is updated rather than left contradicting the code.

**The slug in the route is not the id the API filters on.** Three identifiers are in play: the route's contract **slug**, MITx Online's **`contract_id`** (`ContractPage.page_ptr_id`, what the API filters on), and the warehouse's **`contract_pk`** (an md5 surrogate that must never reach a URL). The slug resolves through the org's `contracts` list — the same lookup `ContractAdminPage` already does — and a slug matching no contract of the org reads as "unavailable" rather than issuing a request with `undefined` in the path, which the API would answer 403, indistinguishable from a real denial.

### How can this be tested?

`yarn install && npx tsc --noEmit -p main/tsconfig.json` — clean.

Tests: `npx jest src/analytics` in `frontends/api` (15 passed) and `npx jest src/app-pages/DashboardPage` in `frontends/main` (644 passed across 29 suites); `src/app-pages/ContractAdminPage` also passes (120) since the link changed. New cases:

- five `analyticsContractQueries` cases asserting each hits the contract-nested path;
- contract query keys nest under the org's (so invalidating an org drops its contracts) and differ per contract, and never collide with an org-scoped key;
- the contract id is URL-encoded, not spliced raw;
- `AnalyticsContent` requests the contract-nested endpoints with the resolved id — **only the contract URLs are mocked, so an org-scoped fallback would fail the render**, which is the point of the test;
- a contract slug not in the org issues no `/contracts/` request at all.

`prettier --check` and `eslint` clean on every changed file; pre-commit hooks passed on commit.

End-to-end against a real API is not possible until the two upstream PRs deploy.

### Additional Context

**Three type corrections fall out of touching these files.** All are pre-existing drift against the API, not consequences of this change — I fixed them here rather than leaving `types.ts` half-right, since the file is hand-written and this is the second time it has needed chasing:

- `contract_pk`, `courserun_pk` and `program_pk` were typed `number` but have been `str` in the API since [ol-analytics-api#29](https://github.com/mitodl/ol-analytics-api/pull/29) — they are md5 surrogates, not integers. `CoursePerformanceTable`'s grouping `Map<number, …>` follows.
- `MonthlyEngagementTrend`'s three activity totals became nullable in [ol-analytics-api#33](https://github.com/mitodl/ol-analytics-api/pull/33) and were still typed non-null; its five cohort columns, and `ContentEngagementDepth`'s two, were missing entirely.

This closes the separately-tracked `tk-update-mit-learn-s-hand-written-analytics-types--5f2d5b`, and is a standing argument for generating this file from the API's OpenAPI schema instead.

**One deliberate cast, in `AnalyticsContent`.** The org and contract query factories build keys of different lengths and `queryOptions` is invariant in the key type, so the inferred union leaves `useQuery` unable to pick an overload. The scope selector erases the key while keeping the row type — the same trade this repo's own query tests already make (`erase` in `hooks/organizations/queries.test.ts`). The key still comes from the query factories; nothing in the component reads it.

**Privacy note carried down from the API PR.** With both grains published, a suppressed contract is recoverable as `org_total − (other contracts)`. 4 of 58 orgs already hold more than one contract, so this is reachable on release rather than hypothetical — tracked as `tk-k-anonymity-suppress-complement-disclosure-near--8818e9`.

https://claude.ai/code/session_01QC8Vb9vZiGMPeHd97Fy4Gj

### DIFF
--- a/frontends/main/src/app-pages/DashboardPage/AnalyticsContent.test.tsx
+++ b/frontends/main/src/app-pages/DashboardPage/AnalyticsContent.test.tsx
@@ -1,6 +1,6 @@
 import React from "react"
 import { renderWithProviders, screen, TestingErrorBoundary } from "@/test-utils"
-import { setMockResponse } from "api/test-utils"
+import { setMockResponse, makeRequest } from "api/test-utils"
 import { factories, urls } from "api/mitxonline-test-utils"
 import {
   factories as analyticsFactories,
@@ -431,7 +431,7 @@ describe("AnalyticsContent", () => {
     const courseRows = (count: number) =>
       Array.from({ length: count }, (_, index) =>
         analyticsFactories.enrollmentCompletionFunnel({
-          courserun_pk: index + 1,
+          courserun_pk: String(index + 1),
           courserun_title: `Course ${index + 1}`,
         }),
       )
@@ -606,6 +606,97 @@ describe("AnalyticsContent", () => {
       expect(
         screen.queryByRole("button", { name: /Show all/ }),
       ).not.toBeInTheDocument()
+    })
+  })
+})
+
+describe("AnalyticsContent, contract-scoped", () => {
+  beforeEach(() => {
+    mockedUseFeatureFlagsLoaded.mockReturnValue(true)
+    mockedUseFeatureFlagEnabled.mockReturnValue(true)
+    setMockResponse.get(
+      urls.userMe.get(),
+      factories.user.user({ email: "manager@test.com" }),
+    )
+  })
+
+  test("requests the contract-nested endpoints, resolving the slug to a contract id", async () => {
+    const contract = factories.contracts.contract()
+    const org = orgWithUuid({ contracts: [contract] })
+    setManagerOrgs([org])
+
+    const contractId = String(contract.id)
+    const page = { limit: 200 }
+    // Only the contract-nested URLs are mocked. An org-scoped request would
+    // find no mock and fail the render, which is the assertion: this route must
+    // not silently fall back to org-wide numbers for a contract-scoped page.
+    setMockResponse.get(
+      analyticsUrls.contracts.contractUtilization(ORG_UUID, contractId, page),
+      analyticsFactories.envelope([analyticsFactories.contractUtilization()], {
+        as_of: AS_OF,
+      }),
+    )
+    setMockResponse.get(
+      analyticsUrls.contracts.engagementTrend(ORG_UUID, contractId, page),
+      analyticsFactories.envelope(
+        [analyticsFactories.contractMonthlyEngagementTrend()],
+        { as_of: AS_OF },
+      ),
+    )
+    setMockResponse.get(
+      analyticsUrls.contracts.enrollmentFunnel(ORG_UUID, contractId, page),
+      analyticsFactories.envelope(
+        [analyticsFactories.enrollmentCompletionFunnel()],
+        { as_of: AS_OF },
+      ),
+    )
+    setMockResponse.get(
+      analyticsUrls.contracts.programFunnel(ORG_UUID, contractId, page),
+      analyticsFactories.envelope([analyticsFactories.programFunnel()], {
+        as_of: AS_OF,
+      }),
+    )
+
+    renderWithProviders(
+      <AnalyticsContent
+        orgSlug={org.slug.replace(/^org-/, "")}
+        contractSlug={contract.slug}
+      />,
+    )
+
+    await waitFor(() => {
+      expect(makeRequest).toHaveBeenCalledWith(
+        expect.objectContaining({
+          url: analyticsUrls.contracts.contractUtilization(
+            ORG_UUID,
+            contractId,
+            page,
+          ),
+        }),
+      )
+    })
+  })
+
+  test("a contract slug that is not in this org requests nothing at all", async () => {
+    const org = orgWithUuid()
+    setManagerOrgs([org])
+
+    renderWithProviders(
+      <AnalyticsContent
+        orgSlug={org.slug.replace(/^org-/, "")}
+        contractSlug="not-a-contract-of-this-org"
+      />,
+    )
+
+    // The slug resolves to no contract, so there is no id to put in the path.
+    // Better an empty state than a request with `undefined` in the URL, which
+    // the API would answer 403 — indistinguishable from a real denial.
+    await waitFor(() => {
+      expect(makeRequest).not.toHaveBeenCalledWith(
+        expect.objectContaining({
+          url: expect.stringContaining("/contracts/"),
+        }),
+      )
     })
   })
 })
--- a/frontends/main/src/app-pages/DashboardPage/AnalyticsContent.test.tsx
+++ b/frontends/main/src/app-pages/DashboardPage/AnalyticsContent.test.tsx
@@ -13,6 +13,7 @@ import type { OrganizationPage } from "@mitodl/mitxonline-api-axios/v2"
 import { useFeatureFlagEnabled } from "posthog-js/react"
 import { allowConsoleErrors } from "ol-test-utilities"
 import { ForbiddenError } from "@/common/errors"
+import { contractAdminView } from "@/common/urls"
 import { useFeatureFlagsLoaded } from "@/common/useFeatureFlagsLoaded"
 import AnalyticsContent from "./AnalyticsContent"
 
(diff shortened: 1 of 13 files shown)


frontends/main/src/app-pages/DashboardPage/AnalyticsContent.tsx
```diff
@@ -159,7 +195,19 @@ const AnalyticsContentInternal: React.FC<AnalyticsContentInternalProps> = ({
...
   // unavailable" rather than a request with `undefined` in the path.
   const orgUuid = org?.sso_organization_id ?? null
-  const analyticsAvailable = isAnalyticsConfigured() && !!orgUuid
+  // MITx Online's contract id, resolved from the slug the route carries -- the
+  // same lookup ContractAdminPage does. The analytics API filters on this id,
+  // never on the slug, and never on the warehouse's `contract_pk` surrogate.
+  const contract = contractSlug
+    ? org?.contracts.find((c) => c.slug === contractSlug)
+    : undefined
+  const contractId = contract ? String(contract.id) : null
```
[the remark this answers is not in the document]
blarghmatey: Real bug — fixed in ec5585b4. It now uses `contract?.slug ?? org.contracts[0]?.slug`, so the fallback only applies on the org-wide page where there is no current contract. Added a regression test with a two-contract org viewing the second, which fails against the old `contracts[0]` version.

frontends/main/src/app-pages/DashboardPage/AnalyticsContent.test.tsx
```diff
@@ -609,3 +609,94 @@ describe("AnalyticsContent", () => {
...
+    // The slug resolves to no contract, so there is no id to put in the path.
+    // Better an empty state than a request with `undefined` in the URL, which
+    // the API would answer 403 — indistinguishable from a real denial.
+    await waitFor(() => {
+      expect(makeRequest).not.toHaveBeenCalledWith(
+        expect.objectContaining({
+          url: expect.stringContaining("/contracts/"),
+        }),
+      )
+    })
```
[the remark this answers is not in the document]
blarghmatey: Correct, the assertion was vacuous — fixed in ec5585b4. It now awaits the settled unavailable state (`findByText(/Analytics is not available for this organization/)`) and then asserts, rather than wrapping the negative in `waitFor`.

frontends/main/src/app-pages/DashboardPage/AnalyticsContent.tsx
```diff
@@ -180,31 +228,64 @@ const AnalyticsContentInternal: React.FC<AnalyticsContentInternalProps> = ({
...
+  const scoped = React.useMemo(() => {
+    const orgId = orgUuid ?? ""
+    return contractId
+      ? {
+          utilization: (page: { limit: number }) =>
+            analyticsContractQueries.contractUtilization(
+              orgId,
+              contractId,
+              page,
+            ),
```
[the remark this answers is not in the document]
blarghmatey: Accurate, and deliberate — same decision recorded on the two upstream PRs (ol-data-platform#2559, ol-analytics-api#34). Publishing both grains was chosen over the alternative that structurally removes the attack (serving only one grain), with this exposure accepted rather than missed.

Not gating this PR on it. The mitigation is tracked as tk-k-anonymity-suppress-complement-disclosure-near--8818e9 in ol-analytics-api, raised p2 -> p1 today with the concrete recovery path, the 4-of-58 multi-contract figure, and the constraint that any cohort-minus-count rule must exclude the event-count columns. Worth noting the fix has to live in the API rather than here: the frontend cannot suppress what the endpoints already returned.

frontends/main/src/app-pages/DashboardPage/AnalyticsContent.tsx
```diff
@@ -180,31 +228,64 @@ const AnalyticsContentInternal: React.FC<AnalyticsContentInternalProps> = ({
...
+          utilization: (page: { limit: number }) =>
+            analyticsOrganizationQueries.contractUtilization(orgId, page),
+          trend: (page: { limit: number }) =>
+            analyticsOrganizationQueries.engagementTrend(orgId, page),
+          courses: (page: { limit: number }) =>
+            analyticsOrganizationQueries.enrollmentFunnel(orgId, page),
+          programs: (page: { limit: number }) =>
+            analyticsOrganizationQueries.programFunnel(orgId, page),
+        }
+  }, [orgUuid, contractId]) as unknown as SectionQueries
```
ahtesham-quraish, line 270: The `as unknown as SectionQueries` here erases the whole object's type, not just the query-key. That's broader than the `erase` helper this mirrors in `hooks/organizations/queries.test.ts` — that one keeps a typed `UseQueryOptions<...>` shell and only widens the four key type-params to `unknown`. Going through bare `as unknown` also drops the row types, so a future mismatch (e.g. wiring `courses` to a factory that returns `ProgramFunnel` rows) would compile silently and only fail at render.

Could we narrow the cast to just the query-key type so the row types stay checked against `SectionQueries`? Something like a small per-factory helper that widens only the key param — same trade as the test's `AnyQueryOptions`, but keeping `ContractUtilization` / `MonthlyEngagementTrend` / etc. intact:

```ts
const eraseKey = <RowT>(
  factory: (page: { limit: number }) => UseQueryOptions<OrgAnalyticsResponse<RowT>, Error, OrgAnalyticsResponse<RowT>, readonly unknown[]>,
): SectionQuery<RowT> => factory
```

Then wrap each branch's factories with `eraseKey<...>()` and drop the `as unknown as SectionQueries` on the memo. Not blocking — the current code is correct and tested; this just keeps the row-type check from being silently disabled.

frontends/main/src/app-pages/DashboardPage/AnalyticsContent.tsx
```diff
@@ -159,7 +195,19 @@ const AnalyticsContentInternal: React.FC<AnalyticsContentInternalProps> = ({
...
+  // never on the slug, and never on the warehouse's `contract_pk` surrogate.
+  const contract = contractSlug
+    ? org?.contracts.find((c) => c.slug === contractSlug)
+    : undefined
+  const contractId = contract ? String(contract.id) : null
+  // A contract route whose slug matches nothing in this org is as unavailable
+  // as a missing org UUID: better an empty state than a request with
+  // `undefined` in the path, which the API would answer with a 403 for a
+  // contract that simply does not exist.
+  const analyticsAvailable =
```
ahtesham-quraish, line 209: ### Suggestion.

A contract-scoped route whose slug doesn't resolve lands in the same `!analyticsAvailable` branch as a missing org UUID, so a manager on a perfectly valid org sees "Analytics is not available for this organization yet. Please contact support…". But the org is fine here — only the contract slug is unknown (usually a stale or mistyped URL), and there's nothing support can do about that.

The request-suppression is correct (`enabled: analyticsAvailable` keeps `undefined` out of the path), but the messaging conflates two different failures. Could we split them out — e.g. a distinct "contract not found" notice, or a 404 for the contract route — so this case reads as "this contract doesn't exist / check the link" rather than "your organization has no analytics"?

Roughly:

```tsx
const contractUnresolved = !!contractSlug && !!orgUuid && !contractId
```

and branch the `Notice` on `contractUnresolved` before the org-level copy, pointing the manager back to the org analytics page or telling them the contract link is stale.

ahtesham-quraish (commented):
LGTM! Left few suggestions

ahtesham-quraish (approved):
Left few suggestions otherwise looks good

daniellefrappier18 (approved):
I also reviewed this and don't have any other comments outside of what @ahtesham-quraish has already suggested.

frontends/main/src/app-pages/DashboardPage/AnalyticsContent.tsx
blarghmatey: Fixed in fec3a1bb4: `eraseKey` now widens only the query-key type param, so each section stays checked against its row type. Also added `eraseContractTrendRow` -- the contract-scoped engagement-trend factory returns `ContractMonthlyEngagementTrend` (a superset of `MonthlyEngagementTrend`), and `UseQueryOptions` uses the row type contravariantly in `select`, so that one section still needs an explicit cast; commented inline with why.

frontends/main/src/app-pages/DashboardPage/AnalyticsContent.tsx
blarghmatey: Fixed in fec3a1bb4: split out a `contractUnresolved` case that shows a distinct notice linking to organization-wide analytics, instead of folding it into the org-not-configured "contact support" message.

blarghmatey:
Addressed both review threads (fec3a1bb4): narrowed the query-key type erasure so section row types stay checked, and split the contract-not-found notice out from the org-not-configured message. Merged main (eb61d2495) to clear stale openapi-diff drift unrelated to this PR. All checks green.

frontends/main/src/app-pages/DashboardPage/AnalyticsContent.tsx
```diff
@@ -266,7 +402,10 @@ const AnalyticsContentInternal: React.FC<AnalyticsContentInternalProps> = ({
     return <ErrorContent title="Access denied" timSays="403" />
   }
 
-  const firstContractSlug = org.contracts[0]?.slug
+  // On a contract-scoped page this must be the contract being viewed, not the
+  // org's first one, or "Manage seats" silently sends the manager to a
+  // different contract's admin page.
+  const manageSeatsSlug = contract?.slug ?? org.contracts[0]?.slug
```
daniellefrappier18: The ?? fallback is correct on the org-wide route, but the unresolved-contract branch also renders header, so a manager on a stale or mistyped contract link gets "This contract link could not be found" with a "Manage seats" button directly above it 
```suggestion
  const manageSeatsSlug = contractSlug  ? contract?.slug  : org.contracts[0]?.slug
```

daniellefrappier18 (commented):
One more minor thing, nothing on the contract-scoped page says which contract you are looking at.

Image

frontends/main/src/app-pages/DashboardPage/AnalyticsContent.tsx
blarghmatey: Fixed in 97de0e967: manageSeatsSlug only falls back to org.contracts[0] on the org-wide route now (no contractSlug); on a contract-scoped route it's contract?.slug, so it's undefined when the contract fails to resolve and the button no longer shows on the "could not be found" notice.

blarghmatey:
Addressed the outstanding review feedback in 97de0e967:
- `manageSeatsSlug` no longer falls back to `org.contracts[0]` on the unresolved-contract branch (thread resolved).
- The contract-scoped page's header now shows the contract's name (@daniellefrappier18's comment about not knowing which contract you're looking at).

blarghmatey:
Rebased onto main (328cbfc1d) to resolve the merge conflict blocking the mergeable check. That pulled in #3806's content-engagement panel, whose query was hardcoded to the org-wide endpoint with no contract-scoped branch — so the contract-scoped page here would have shown the whole org's content-engagement numbers instead of the contract's, unlike the other four sections. Wired it to `analyticsContractQueries.contentEngagement` (already present in the client layer, just unused) the same way the other sections are scoped.

blarghmatey:
Rebase also picked up a new lint rule (hq#12925, AppPageProps over PageProps) that this route's page.tsx predated — fixed in 8e159d642, matching the sibling org-wide analytics page's existing pattern. All checks green, no merge conflicts.

